PR #37, opened 2023-07-27
Fix superlinter on the master branch and ensure it runs on PRs

jeancochrane:
Our superlinter checks [are failing on the master branch](https://github.com/ccao-data/data-architecture/actions/runs/5684699106/job/15407982040) but they [passed on the most recent PR](https://github.com/ccao-data/data-architecture/actions/runs/5684620829/job/15407722505) (#23) because the config points to the wrong name for our main branch:

https://github.com/ccao-data/data-architecture/blob/72bf150a8eb4117191f4a84053527c60ae5f8818/.github/workflows/lint.yaml#L30

``` 
2023-07-27 19:23:43 [INFO]   Gathering user validation information...
2023-07-27 19:23:43 [INFO]   - Only validating [new], or [edited] files in code base...
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

fatal: ambiguous argument 'main...06569ff79d05062bf89c03297192d23ee331442c': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
2023-07-27 19:23:[49](https://github.com/ccao-data/data-architecture/actions/runs/5684620829/job/15407722505#step:4:50) [WARN]   No files were found in the GITHUB_WORKSPACE:[/github/workspace] to lint!
```

This PR points superlinter to the correct main branch name (`master`) and tweaks some files/configs to ensure it will pass on the next merge.

You can confirm that superlinter is in fact running the tests by checking out the [latest run for this PR](https://github.com/ccao-data/data-architecture/actions/runs/5685167543/job/15409465635).

PR #39, opened 2023-07-28
20 decompose modelling views into reusable components

wrridgeway:
There are 3 issues to consider before merging this work:

1. There are 6 new columns in the condo input view (nearest neighbor columns) that can only be extricated by shifting a significant amount of code back into the condo and res views
2. vw_pin_shared_input isn't a good name for the precursor view
3. We lose the ability to order columns for the input views with this schema

All of this being noted, we do significantly diminish the amount of code in each of the modelling views by shifting shared code into the precursor view.


.github/workflows/lint.yaml
```diff
@@ -29,6 +29,7 @@ jobs:
           VALIDATE_ALL_CODEBASE: false
           DEFAULT_BRANCH: main
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VALIDATE_PYTHON: true
           VALIDATE_R: true
           VALIDATE_SQLFLUFF: true
-          SQLFLUFF_CONFIG_FILE: /.sqlfluff
+          SQLFLUFF_CONFIG_FILE: ./.sqlfluff
```
jeancochrane, line 35: **[Question, non-blocking]** Does this change work? In my branch to [fix superlinter](https://github.com/ccao-data/data-architecture/pull/37) I had to set `LINTER_RULES_PATH` to `/` because it's [set to `.github/linters` by default](https://github.com/super-linter/super-linter#environment-variables), so I would assume that `./.sqlfluff` would expand out to `.github/linters/.sqlfluff` without the change to `LINTER_RULES_PATH`. Maybe pulling in the latest change from `master` will resolve the SQLFluff problems you were experiencing?

.github/workflows/lint.yaml
```diff
@@ -29,6 +29,7 @@ jobs:
           VALIDATE_ALL_CODEBASE: false
           DEFAULT_BRANCH: main
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VALIDATE_PYTHON: true
```
jeancochrane, line 32: **[Question, non-blocking]** According to the [environment variable docs](https://github.com/super-linter/super-linter#environment-variables), `VALIDATE_PYTHON` should be set to `true` by default; are you experiencing unexpected behavior with the default config that this change is intended to resolve?

aws-athena/views/model-vw_card_res_input.sql
```diff
@@ -16,58 +16,15 @@ extracts for modeling
...
-    FROM iasworld.pardat AS par
-    LEFT JOIN iasworld.legdat AS leg
-        ON par.parid = leg.parid
-        AND par.taxyr = leg.taxyr
-    LEFT JOIN spatial.parcel AS sp
-        ON SUBSTR(par.parid, 1, 10) = sp.pin10
-        AND par.taxyr = sp.year
-    LEFT JOIN spatial.township AS twn
-        ON leg.user1 = CAST(twn.township_code AS VARCHAR)
+    SELECT * FROM model.vw_pin_shared_input
```
jeancochrane, line 19: **[Praise]** Nice cleanup!

jeancochrane (commented):
I'm not yet enough of an expert to have a useful opinion on the SQL refactor yet, so I would wait for Dan's approval before merging, but at a high level this looks great to me! A couple superlinter questions below that may be resolved by merging in the latest from `master`.

.github/workflows/lint.yaml
wrridgeway: No, you're completely right. We could eliminate it, the only reason I had it in here was an attempt to be about explicit which languages we're looking to lint, but it's really rather moot now that we're probably going to have more in this repo than we used to.

PR #41, opened 2023-07-28
Symlink dbt models to Athena views to reduce duplication

### LINKED ISSUE #36: [Data catalog] Factor out dbt models to share logic with Athena views during development
Inspired by [this thread](https://github.com/ccao-data/data-architecture/pull/23#discussion_r1275265607). Briefly: We want to avoid duplication between `dbt/models/` and `aws-athena/views/` as much as possible during development of the [data catalog](https://github.com/ccao-data/data-architecture/milestone/1); in order to do that, we need to have these two versions of our views share as much logic as possible.

To do that, we plan to create a long-lived branch named `data-catalog` that we will use for development of the data catalog. We will merge all of our data catalog work into this branch. On the `data-catalog` branch, dbt models will be symlinked to the view definitions in `aws-athena/views/`, and tweaked minimally so as to make them work with dbt (most importantly, removing `CREATE TABLE AS` statements and using dbt references for other views in the DAG). Then, as changes are made to the views on the `master` branch, we can merge them into `data-catalog` and resolve merge conflicts; when we're ready to go live with the data catalog, we can merge `data-catalog` into `master`.

jeancochrane:
This PR changes up the way that we store our dbt models to reduce duplication during development of our dbt flow. Briefly, we symlink model definitions to their counterparts in `aws-athena/views/` and make as few edits as possible in order to support dbt's model definition syntax. This should allow us to work on models in the long-lived `data-catalog` branch (the base branch for this PR) without introducing duplication or causing too many conflicts.

The PR also folds in a small related change to support the long-lived `data-catalog` branch: Updating the GitHub Actions config to ensure CI treats this branch like a main branch when running tests.

Closes #36.


dfsnow:
**question (non-blocking):** I take it the symlink issue is fixed then? Or does it not apply in this case?

**question (non-blocking):** Is there any reason to keep test in the filename now that these are just symlinks?

dfsnow (approved):
Looks good! Nice cleanup and sorry for my dumb Qs. Merge at your leisure

Issue #28, opened 2023-07-26
[Data catalog] Add production profile to the dbt configuration


jeancochrane:
The dbt profile created in https://github.com/ccao-data/data-architecture/pull/23 only supports a development Athena database. We would also like to add a profile that supports building views and tables into the production Athena databases.

This might be slightly tricky, since we override the database defined in `profiles.yml` via a folder-level config in the `models` attribute of `dbt_project.yml`. We do this so that we can build models into separate databases depending on their surrounding subdirectory in the `models/` directory, e.g. `default/` or `location/`. This means we can't just create a new `prod` entry in `profiles.yml`, which is the recommended way of supporting [multiple environments](https://docs.getdbt.com/docs/core/dbt-core-environments).

We may want to investigate using a separate [Athena Data Catalog](https://docs.aws.amazon.com/athena/latest/ug/understanding-tables-databases-and-the-data-catalog.html) for dev and production, but that would depend on dev being able to read data from the production catalog, since our source data is built directly into the production Data Catalog.

jeancochrane:
Closed by https://github.com/ccao-data/data-architecture/pull/44.

Issue #31, opened 2023-07-26
[Data catalog] Define GitHub Actions workflows for building the dbt DAG and running tests


jeancochrane:
## Overview

We want to use GitHub Actions to run the `dbt run` and `dbt test` commands that build and test our DAG, respectively. To do this, we'll need to add GitHub Actions workflow definitions to this repository.

## Workflow definitions

Our workflow definition should support two different types of flows:

1. Rebuild all models that have changed since the last cached run, and run their tests
2. Run all the tests, regardless of the result of the last cached run

These two flows will be used in two different ways:

1. will provide continuous integration (CI) for dbt models in this repository, building and testing models when we make code changes to them; while
2. will provide a test interface that we can call from the GitHub Actions workflow API to run data integrity checks after pulling fresh source data from the system of record each night.

Caching is important in CI to help speed up development cycles, but it's unnecessary in the context of our nightly data integrity checks, where we want to validate all of the data on each run.

## Cache behavior

The CI workflow (1 above) should exhibit the following cache behavior:

* On every PR:
  * Run the build and tests for models that have changed since last commit to `master` OR since the last successful workflow run for this PR
    * In other words: The first workflow run for any PR should use the cache from the `master` branch, and subsequent runs should use  the cache from the first successful run on the PR branch
    * These builds and tests should run in a separate development environment, ideally one that is created exclusively for the PR and not shared by other PRs; we should use the same environment scheme set up in https://github.com/ccao-data/data-architecture/issues/28
    * The `master` branch cache should never be updated by this flow
* On commits to `master`:
  * Run builds and tests for models that have changed since the last commit to `master`
    * These builds and tests should run against the `prod` Athena environment
    * The `master` branch cache should be updated when this flow succeeds

## AWS credentials

In order to run dbt commands against Athena from the context of a GitHub Action workflow, we'll need to inject valid AWS credentials into the workflow. Credentials should be stored as [encrypted secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) and should have their permissions restricted as much as possible to reduce the attack surface of the credentials. See the [`dbt-athena` docs](https://dbt-athena.github.io/docs/getting-started/prerequisites/iam-permissions) for a list of the required permissions for the adapter.

## Incremental testing

Since our CI tests will only be useful if they can distinguish between expected and unexpected data integrity issues, this issue depends on https://github.com/ccao-data/data-architecture/issues/32.

jeancochrane:
[...]

PR #44, opened 2023-07-31
Adjust dbt profiles and schema macro to support prod/CI targets

jeancochrane:
This PR adjusts our dbt profiles and our custom `generate_schema_name` macro to support production and CI targets for the data catalog.

The way this configuration works is:

* We can choose which target environment we want to build or test using the `--target` flag that is accepted by all dbt commands. If no flag is provided, the default environment is `dev`.
* When building a target in a specific environment, dbt will build the target into an Athena database with an automatic prefix prepended to the database name in order to namespace the environment. We choose to use prefix namespacing to distinguish between environments instead of building databases in isolated data catalogs because we need to leverage prebuilt tables from the default catalog in our dev and CI environments in order to save time when running running build and test commands.
    * The `dev` environment adds a prefix to the database name matching to the caller's Unix username, e.g. the output database will be called `dev-jecochr-default` for targets built into the database named `default` when `dbt run` is run on my machine.
    * The `ci` environment adds a prefix matching to the name of the branch, e.g. `ci-feature-branch-1-default`, with special characters removed and all letters forcibly lowercased to match [`dbt-athena`'s requirements](https://dbt-athena.github.io/docs/known-issues).
        * When we set up CI in https://github.com/ccao-data/data-architecture/issues/31, we will pull this value from the builtin `$GITHUB_BASE_REF` env variable ([docs](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)). We will also detect when the pull request has been closed or merged ([docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges)) and use that as a trigger to cleanup any resources that were created for the PR.
    * The `prod` environment does not add any prefix, e.g. `default`.

A quick test confirming this behavior on my machine:

```console
(venv) jecochr@21CCAO-LAPTOP54:~/code/data-architecture/dbt$ dbt run --select vw_pin10_location_test
16:49:17  Running with dbt=1.5.4
16:49:18  Registered adapter: athena=1.5.1
16:49:18  Unable to do partial parsing because change detected to override macro. Starting full parse.
16:49:18  Found 10 models, 19 tests, 0 snapshots, 0 analyses, 456 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics, 0 groups
16:49:18
16:49:22  Concurrency: 5 threads (target='dev')
16:49:22
16:49:22  1 of 1 START sql view model jecochr-location.vw_pin10_location_test ............ [RUN]
16:49:25  1 of 1 OK created sql view model jecochr-location.vw_pin10_location_test ....... [OK -1 in 2.71s]
16:49:25
16:49:25  Finished running 1 view model in 0 hours 0 minutes and 6.22 seconds (6.22s).
16:49:25
16:49:25  Completed successfully
16:49:25
16:49:25  Done. PASS=1 WARN=0 ERROR=0 SKIP=0 TOTAL=1
```

This PR also adds a new framework for unit testing macros. Macro tests can now be run via the following command:

```console
$ dbt run-operation test_all
```

Sample output:

```console
(venv) jecochr@21CCAO-LAPTOP54:~/code/data-architecture/dbt$ dbt run-operation test_all
20:51:23  Running with dbt=1.5.4
20:51:23  Registered adapter: athena=1.5.1
20:51:24  Found 10 models, 19 tests, 0 snapshots, 0 analyses, 475 macros, 0 operations, 0 seed files, 0 sources, 0 exposures, 0 metrics, 0 groups
20:51:24  test_kebab_slugify_lowercases_strings - PASS
20:51:24  test_kebab_slugify_replaces_spaces - PASS
20:51:24  test_kebab_slugify_replaces_slashes - PASS
20:51:24  test_kebab_slugify_replaces_underscores - PASS
20:51:24  test_kebab_slugify_removes_special_characters - PASS
20:51:24  test_kebab_slugify_handles_leading_numbers - PASS
20:51:24  test_generate_schema_name_handles_dev_env - PASS
20:51:24  test_generate_schema_name_handles_ci_env - PASS
20:51:24  test_generate_schema_name_handles_prod_env - PASS
20:51:24  test_generate_schema_name_raises_for_default_schema_name - PASS
```

Closes https://github.com/ccao-data/data-architecture/issues/28.


dbt/dbt_project.yml
```diff
@@ -28,6 +28,6 @@ models:
   athena:
     +materialized: view
     default:
-      +schema: dbt-test-default
```
jeancochrane, line 31: We no longer need to manually specify this prefix, because the `generate_schema_name` macro is doing it for us.

dbt/profiles.yml
```diff
@@ -11,3 +11,22 @@ athena:
...
+    ci:
+      type: athena
+      s3_staging_dir: s3://ccao-dbt-athena-ci-us-east-1/results/
+      s3_data_dir: s3://ccao-dbt-athena-ci-us-east-1/data/
+      region_name: us-east-1
+      schema: dbt-test
+      database: awsdatacatalog
+      # Prefix all generated data by schema, so that we can delete it when the
+      # PR is merged
+      s3_data_naming: schema_table
```
jeancochrane, line 23: I am actually not totally sure if this works yet, since we're not yet using dbt to manage our CTAs, but the idea here is that eventually when dbt builds CTAs and stores them in S3 the `schema_table` config will tell it to store those files in the configured `s3_data_dir` bucket with the path `{s3_data_dir}/{schema}/{table}/` ([docs](https://dbt-athena.github.io/docs/configuration/table-configuration#table-data-location)). That way when we're ready to have CD clean up the resources generated for the PR, we can use the schema name defined by the `generate_schema_name` to delete everything in `{s3_data_dir}/{schema}/` and leave resources that are being used by other PRs intact.

dbt/profiles.yml
jeancochrane, line 1: In case you need a refresher on the `dbt-athena` configuration options we're adding here, see [the docs](https://dbt-athena.github.io/docs/getting-started/installation#configuring-dbt-athena).

jeancochrane:
Requesting both @dfsnow and @wrridgeway since this PR makes some major architectural choices about how we plan to structure dbt environments, but I only need a detailed review from one of you!

dbt/README.md
```diff
@@ -48,14 +48,24 @@ Build the models to create views in our Athena warehouse:
 dbt run
 ```
 
+By default, all `dbt` commands will run against the `dev` environment, which
+namespaces the resources it creates by prefixing target database names with
+your Unix `$USER` name (e.g. `jecochr-default` for the `default` database when
```
dfsnow, line 53: How does this work when the name of a branch includes an unexpected character (i.e. the `/` in `jeancochrane/28-data-catalog-add-production-profile-to-the-dbt-configuration`)?

dbt/macros/generate_schema_name.sql
```diff
@@ -1,15 +1,41 @@
...
+    {#
+        According to the dbt docs linked above, this is required to be set by
+        the built-in macro that we are overriding, but we don't actually use it
+    #}
     {%- set default_schema = target.schema -%}
+
+    {%- if target.name == "dev" -%}
+        {%- set schema_prefix = env_var("USER") -%}
+    {%- elif target.name == "ci" -%}
+        {%- set schema_prefix = env_var("GITHUB_BASE_REF") -%}
```
dfsnow, line 15: **suggestion**: Won't `GITHUB_BASE_REF` give the target of the PR i.e. `main`? I think this should be `GITHUB_HEAD_REF` instead.

dfsnow:
Overall, I think this looks great. My one note is that it might be worthwhile to prefix the prefix with something that identifies a dev table so that we can easily grep them using the CLI (I suggest `dev-`).

So `jecochr-location.vw_pin10_location_test` would become `dev-jecochr-location.vw_pin10_location_test`.

dbt/macros/generate_schema_name.sql
jeancochrane: Great catch, I got these mixed up after reading the docs! Fixed in 62dc2b34473ddf9a4f63d2a1fa6a8e39368c80f8 👍🏻

dbt/README.md
jeancochrane: Good question! I think we can replace special characters (and uppercase letters, which are [not allowed in `dbt-athena`](https://dbt-athena.github.io/docs/known-issues)) using Jinja template filters. I'll take a stab at that now.

jeancochrane: @dfsnow This spiraled a little bit, but I was able to make it work in https://github.com/ccao-data/data-architecture/pull/44/commits/181e516d577c82add3b37a50c5e5e64c7a03d6d6! It required a new custom macro, but that led to me realize that we were starting to build a lot of logic into a collection of untested macros, so I took a minute to design a quick unit testing framework for our macros. Curious what you think!

dfsnow: This is great! Nice work and definitely needed since this is getting more complex. I'm a dumb dumb and commented on a commit instead of the PR. Linking here: https://github.com/ccao-data/data-architecture/commit/181e516d577c82add3b37a50c5e5e64c7a03d6d6#r123228596

dfsnow (approved):

Issue #32, opened 2023-07-26
[Data catalog] Add script to support incremental testing of dbt models


jeancochrane:
I anticipate we will be much faster at defining data integrity checks for our models than we are at actually _fixing_ the data problems revealed by those checks. If that's true, we'll need to wrap `dbt test` in a script that can distinguish between expected and unexpected test failures, or else our CI test runs will fail constantly.

I'm not totally sure what the best way to do this is, so this issue will require some design work. We'll probably want some sort of version controlled config file that we can use to map tests to their expected number of failures; perhaps we can even include this in the dbt `schema.yml` files via a plugin API.

jeancochrane:
Closed by https://github.com/ccao-data/data-architecture/pull/45.

Issue #27, opened 2023-07-26
[Data catalog] Add remaining tests to DAG models


jeancochrane:
Once we've added models and documentation for the remaining Athena views that we want to build with dbt (https://github.com/ccao-data/data-architecture/issues/26), we'll want to add tests for them. We'll also want to finish up the tests for all of the existing models that are marked with TODO (indicating that Jean didn't understand the test and required clarification). Billy will be a helpful resource for defining the remaining tests.

See [this doc](https://cookcounty-my.sharepoint.com/:x:/r/personal/william_ridgeway_cookcountyil_gov/_layouts/15/Doc.aspx?sourcedoc=%7BA8C09072-E8B8-41E9-961E-8F96A7C5BB41%7D&file=sql_guide_q_and_a.xlsx&wdLOR=c19D97A6F-0CBA-4889-86F7-AEC0EE2B452F&action=default&mobileredirect=true) for explanations of the remaining tests.

This is a separate story since I'm anticipating that the work to define and write these tests will generate enough code to warrant a dedicated pull request, but if that's not true, feel free to roll it into https://github.com/ccao-data/data-architecture/issues/26.

Depends on https://github.com/ccao-data/data-architecture/issues/26 and #61.

jeancochrane:
Superseded by #113,

PR #45, opened 2023-08-01
Support incremental testing of dbt models through failure thresholds

jeancochrane:
We know there are some existing integrity issues with our data that the data catalog is designed to help us address, but we don't want our dbt tests to unnecessarily alert us about these known issues while we're developing the catalog. To handle this problem, this PR adds some simple thresholds to the tests to ensure that they only fail when the number of test failures exceeds the number of expected failures.

The numbers used in these thresholds may not accurately reflect underlying integrity issues, and instead may simply represent problems with the tests themselves; we haven't thoroughly audited our tests yet so we can't rule that out. But I'm considering that effort out of scope for now, and will instead be a part of https://github.com/ccao-data/data-architecture/issues/27.

Adding these failure thresholds was significantly easier than I thought it would be (#32) since it turns out test severity thresholds [have native support in dbt](https://docs.getdbt.com/reference/resource-configs/severity). Nice! In case we want to support running these tests with the option to ignore the thresholds, we can do so with the following incantation:

```
$ dbt --warn-error test
```

The PR also makes two unrelated changes that helped me debug failing tests:

* Fix a bug that was preventing the build of the `default.vw_pin_sale` model due to a missing column in a subquery
* Add simple model namespacing to tests to work around the fact that dbt does not 

Closes https://github.com/ccao-data/data-architecture/issues/32.

### DIFF
--- a/aws-athena/views/default-vw_pin_sale.sql
+++ b/aws-athena/views/default-vw_pin_sale.sql
@@ -185,7 +185,8 @@ mydec_sales AS (
             ) > 0 AS sale_filter_ptax_flag,
             COUNT() OVER (
                 PARTITION BY line_1_primary_pin, line_4_instrument_date
-            ) AS num_cards_sale
+            ) AS num_cards_sale,
+            year_of_sale
         FROM sale.mydec
         WHERE is_earliest_within_doc_no
     )
--- a/dbt/models/default/schema.yml
+++ b/dbt/models/default/schema.yml
@@ -85,11 +85,7 @@ models:
               case when char_renovation = '1' then true else false end
             )
           config:
-            # This error threshold represents an upper bound, since the
-            # underlying query that generates `char_recent_renovation`
-            # is currently not deterministic. When we fix the underlying
-            # view to make that column deterministic, we should lower this.
-            error_if: ">74500"
+            error_if: ">73925"
       # TODO: Characteristics columns should adhere to pre-determined criteria
   - name: vw_pin_address_test
     description: '{{ doc("vw_pin_address_test") }}'
--- a/dbt/models/default/schema.yml
+++ b/dbt/models/default/schema.yml
@@ -7,7 +7,7 @@ models:
     tests:
       # Unique by 14-digit PIN and year
       - unique_combination_of_columns:
-          name: unique_by_14_digit_pin_and_year
+          name: vw_pin_history_unique_by_14_digit_pin_and_year
           combination_of_columns:
             - pin
             - year
@@ -16,6 +16,7 @@ models:
     tests:
       # Unique by 14-digit PIN and year
       - unique_combination_of_columns:
+          name: vw_pin_value_unique_by_14_digit_pin_and_year
           combination_of_columns:
             - pin
             - year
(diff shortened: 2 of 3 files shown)


aws-athena/views/default-vw_pin_sale.sql
```diff
@@ -185,7 +185,8 @@ mydec_sales AS (
             ) > 0 AS sale_filter_ptax_flag,
             COUNT() OVER (
                 PARTITION BY line_1_primary_pin, line_4_instrument_date
-            ) AS num_cards_sale
+            ) AS num_cards_sale,
+            year_of_sale
```
jeancochrane, line 189: This missing column was causing `dbt run` to fail to build this model:

```
18:53:31  Runtime Error in model vw_pin_sale_test (models/default/vw_pin_sale_test.sql)
18:53:31    line 212:18: Column 'mydec_sales.year_of_sale' cannot be resolved
```

I had forgotten that I had made this adjustment prior to factoring out shared logic between the Athena and dbt views in https://github.com/ccao-data/data-architecture/pull/41.

Perhaps this change should also be pulled into `master`? It seems like the Athena version of the view correctly selects this column, so it must just be a discrepancy between the production version of the view and the version stored in this file. (Another nice category of problem that dbt will eventually resolve for us.)

dbt/models/default/schema.yml
```diff
@@ -56,36 +63,46 @@ models:
...
           # char_renovation can be null, which we count here as falsey
           expression: >-
             char_recent_renovation = (
               case when char_renovation = '1' then true else false end
             )
+          config:
+            # This error threshold represents an upper bound, since the
+            # underlying query that generates `char_recent_renovation`
+            # is currently not deterministic. When we fix the underlying
+            # view to make that column deterministic, we should lower this.
```
jeancochrane, line 91: Very strange! I think this has to do with the ordered window functions that we use to generate `vw_card_res_char.char_recent_renovation`:

https://github.com/ccao-data/data-architecture/blob/73cdfec06aa2730d41a06558bcf469b4a3f9c124/dbt/models/default/vw_card_res_char_test.sql#L90-L124

Billy's looking into it!

dbt/models/default/schema.yml
```diff
@@ -7,7 +7,7 @@ models:
     tests:
       # Unique by 14-digit PIN and year
       - unique_combination_of_columns:
-          name: unique_by_14_digit_pin_and_year
+          name: vw_pin_history_unique_by_14_digit_pin_and_year
```
jeancochrane, line 10: An example of what test output looks like before adding this namespacing, to motivate the change:

```
19:53:13  1 of 19 START test unique_by_14_digit_pin_and_year ................... [RUN]
```

There are a bunch of tests named `unique_by_14_digit_pin_and_year` in particular.

jeancochrane:
Requesting both @dfsnow and @wrridgeway on this for the sake of keeping everyone up to date on how we're handling failure thresholds, but I only really need one thorough review!

aws-athena/views/default-vw_pin_sale.sql
dfsnow: Can you cherry-pick this commit and open a PR to master? This seems like a mistake

dbt/models/default/schema.yml
dfsnow: Seemingly fixed by #47.

jeancochrane: Yup, that fixes it -- resolving this now!

aws-athena/views/default-vw_pin_sale.sql
jeancochrane: Can do! [PR up here](https://github.com/ccao-data/data-architecture/pull/48), I'll request your review once tests pass.

dbt/models/default/schema.yml
```diff
@@ -97,17 +114,21 @@ models:
             - mail_address_state
             - mail_address_zipcode_1
             - mail_address_zipcode_2
+          config:
+            error_if: ">879261"
```
dfsnow, line 114: **issue (non-blocking)**: This high error rate is a result of using `CONCAT_WS` to construct `prop_address_full` and `mail_address_full`. Any address missing a unit number or description will have trailing whitespace. See https://github.com/ccao-data/data-architecture/blob/36a42a7d304577fd77440ff1d807905d5bae1a75/aws-athena/views/default-vw_pin_address.sql#L20C12-L20C21.

@wrridgeway Can you make a new PR to master wrapping `CONCAT_WS` in `TRIM()` for these address fields.

@jeancochrane We can keep this PR moving but we'll need to remember to update this threshold once Billy's PR is merged.

dbt/models/default/schema.yml
dfsnow: Love this, good thinking!

**question (non-blocking)**: Is there no way to define the namespace in the test definition? If not, this totally works.

dfsnow (approved):

aws-athena/views/default-vw_pin_sale.sql
jeancochrane: Merged to `master` 👍🏻

PR #47, opened 2023-08-01
Add card to partition for recent renovation

wrridgeway:


jeancochrane (approved):
Good catch, thanks! Doesn't have to be blocking, but I'm curious why the missing partition would lead to inconsistent ordering.

wrridgeway:
The issue was that dweldat isn't unique by PIN, it's unique PIN and CARD. So I was generating inconsistent ordering within thechar_recent_renovation column (by ignoring CARD), which depends on ordering for the LAG operation (it indicates if there's been a switch from 0 to 1 in char_renovation within the last three years).

PR #48, opened 2023-08-01
Add missing year_of_sale column to mydec_sales subquery in vw_pin_sale

jeancochrane:
It seems like `vw_pin_sale` is missing a column in the `mydec_sales` subquery, causing dbt to raise this error when attempting to build the models:

```
18:53:31  Runtime Error in model vw_pin_sale_test (models/default/vw_pin_sale_test.sql)
18:53:31    line 212:18: Column 'mydec_sales.year_of_sale' cannot be resolved
```

This isn't a problem in the production Athena view, so this probably represents a discrepancy caused by drift between the prod view definition and the view definition stored in this repo. This PR updates the view definition in this repo to correctly select `year_of_sale` when generating the `mydec_sales` subquery.


dfsnow (approved):
Looks good, thanks for the quick turnaround.

PR #49, opened 2023-08-01
Add sqlfmt pre-commit hook for autoformatting dbt files

### LINKED ISSUE #30: [Data catalog] Improve Jinja templating indentation in dbt files
It seems like SQLFluff doesn't quite understand how to handle `for` loops in our dbt Jinja templates, and it keeps wanting to pull the `{% endfor %}` tag onto its own line. For example:

https://github.com/ccao-data/data-architecture/blob/3f31ca090d1fa198585480924e4c2e252c93f499/dbt/tests/generic/test_column_length.sql#L16-L20

It's also not totally clear if SQLFluff wants the body of `{% test %}` and `{% macro %}` blocks to be indented. I would prefer if their bodies weren't indented, since the tags act more like file metadata than proper blocks given dbt's organization schema for tests and macros (where each test/macro gets its own file), but either way we should get SQLFluff to enforce the rule automatically.

Update our SQLFluff indentation rules so that we can fix and enforce both of these policies.

jeancochrane:
This PR sets up a pre-commit hook for [sqlfmt](https://docs.sqlfmt.com/) so that we can autoformat our dbt files and check for consistency on CI. It also runs sqlfmt autoformatting on all of our existing dbt files except our models, to which we are currently attempting to minimize edits as we build out the data catalog on the `data-catalog` branch.

Closes #30.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,3 +5,17 @@ repos:
     hooks:
     -   id: sqlfluff-lint
     -   id: sqlfluff-fix
+  - repo: https://github.com/tconbeer/sqlfmt
+    rev: v0.19.2
+    hooks:
+      - id: sqlfmt
+        entry: >-
+          sqlfmt -v dbt
+            --exclude dbt/venv/**/*
+            --exclude dbt/models/**/*
+            --exclude dbt/target/**/*
+            --exclude dbt/dbt_packages/**/*
+            --exclude dbt/dbt_modules/**/*
+        files: ^dbt/
+        language_version: python
+        additional_dependencies: ['.[jinjafmt]']
--- a/dbt/.sqlfluff
+++ b/dbt/.sqlfluff
@@ -1,0 +1,29 @@
+# These rules are required to support sqlfmt. See:
+# https://docs.sqlfmt.com/integrations/sqlfluff
+[sqlfluff]
+exclude_rules = layout.indent, layout.cte_bracket, layout.select_targets, ambiguous.column_count, structure.column_order, RF04, ST05
+max_line_length = 88
+
+[sqlfluff:rules]
+capitalisation_policy = lower
+extended_capitalisation_policy = lower
+
+[sqlfluff:rules:convention.terminator]
+multiline_newline = True
+
+# These rules are required to override the sqlfluff config at the root of
+# the project repo, since those rules interfere with sqlfmt standards.
+[sqlfluff:rules:capitalisation.keywords]
+capitalisation_policy = lower
+
+[sqlfluff:rules:capitalisation.identifiers]
+capitalisation_policy = lower
+
+[sqlfluff:rules:capitalisation.functions]
+extended_capitalisation_policy = lower
+
+[sqlfluff:rules:capitalisation.literals]
+capitalisation_policy = lower
+
+[sqlfluff:rules:capitalisation.types]
+extended_capitalisation_policy = lower
--- a/dbt/macros/generate_schema_name.sql
+++ b/dbt/macros/generate_schema_name.sql
@@ -2,13 +2,17 @@
 -- and replace it with our own namespacing on dev and CI.
 -- See: https://docs.getdbt.com/docs/build/custom-schemas
 {% macro generate_schema_name(custom_schema_name, node) -%}
-    {{ return(_generate_schema_name(
-        custom_schema_name,
-        node,
-        target,
-        env_var,
-        exceptions.raise_compiler_error
-    )) }}
+    {{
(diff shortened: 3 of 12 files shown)


.pre-commit-config.yaml
jeancochrane, line 1: The changes to this file and the addition of the `dbt/.sqlfluff` file are the only substantive changes on this branch; everything else is just the diff generated by running `sqlfmt` against our non-model dbt files for the first time.

.pre-commit-config.yaml
```diff
@@ -4,4 +4,18 @@ repos:
...
+    rev: v0.19.2
+    hooks:
+      - id: sqlfmt
+        entry: >-
+          sqlfmt -v dbt
+            --exclude dbt/venv/**/*
+            --exclude dbt/models/**/*
+            --exclude dbt/target/**/*
+            --exclude dbt/dbt_packages/**/*
+            --exclude dbt/dbt_modules/**/*
```
jeancochrane, line 18: Annoyingly, these `--exclude` options can't be combined via the CLI, so each pattern requires its own `--exclude` flag ([docs](https://docs.sqlfmt.com/getting-started/configuring-sqlfmt#configuration-reference)). We could combine them if we used a `pyproject.toml` file to configure sqlfmt, but that seemed like overkill for one config value; still, I could be convinced to do that if this command seems unwieldy.

.pre-commit-config.yaml
```diff
@@ -4,4 +4,18 @@ repos:
...
+        files: ^dbt/
```
jeancochrane, line 19: The `files` attribute is required here, since otherwise passing the `--all-files` flag to `pre-commit` (as we do in CI) will run sqlfmt on _everything_ in this repo.

https://github.com/ccao-data/data-architecture/blob/5b8279a83a1bf7a127209311e34e3bc56373da82/.github/workflows/pre-commit.yaml#L36-L38

As an aside, I'm surprised that this is how pre-commit behaves, given that it claims [the `entry` command options will not be overriden](https://pre-commit.com/#hooks-entry). I'm guessing there's a special case for `file`, which must do some globbing of its own behind the scenes.

.pre-commit-config.yaml
dfsnow: I think this is totally fine, no need for additional files.

.pre-commit-config.yaml
dfsnow: **question (blocking):** How will we reconcile our existing SQLfluff config with the dbt-specific one? Will we eventually need to reformat the model views using `sqlfmt` instead of `sqlfluff-fix`? Or will the model files use `sqlfluff-fix` while the dbt/jinja files use `sqlfmt`?

dfsnow (commented):
Looks great besides the blocking questions above!

.pre-commit-config.yaml
jeancochrane: Good question @dfsnow! I was expecting that eventually when we pull `data-catalog` into `master` we would want to start formatting the model files with sqlfmt instead of sqlfluff-fix for the sake of consistency, but there's nothing that says we _have_ to do that, particularly given that I anticipate our models will continue to make minimal use of Jinja templating beyond resolving references. Do you have opinions on the two options? I'd be happy to chat out loud about this if you have bigger-picture concerns, too.

dbt/.sqlfluff
dfsnow, line 1: **question (non-blocking):** Given the question above about merging sqlfmt and sqlfluff, will we also need to merge `.sqlfluff` files, or will there be separate `dbt` and `model` `.sqlfluff` files?

jeancochrane:
> question (non-blocking): Also @jeancochrane, does this deprecate https://github.com/ccao-data/data-architecture/pull/35?

Yes, thanks for catching that @dfsnow! I just closed it with a note linking to this PR.

.pre-commit-config.yaml
dfsnow: Hmmm, let's just solve that problem when we reach it. I think I'd need to see the output of `sqlfmt` vs `sqlfluff-fix` on some of the more complex views. Either one is _probably_ fine, but ideally we choose the one with clearer output (which can really matter for some of the complex queries).

For now, go ahead with the merge!

dbt/.sqlfluff
jeancochrane: @dfsnow I think that depends on how we decide to format the models. If we want to use some but not all of the dbt sqlfmt rules, we'll indeed need two separate `.sqlfluff` files. But if we decide to just continue using sqlfluff for formatting models going forward, we can continue to exclude the models from the sqlfmt search path and skip the extra `.sqlfluff` file. Does that make sense?

PR #50, opened 2023-08-03
Define GitHub actions workflows for building and testing dbt

jeancochrane:
This PR sets up GitHub Actions workflows for two particular cases that we want to support when building and testing dbt:

1. A fast build-and-test workflow that runs on pull requests and pushes to main branches (`build_and_test_dbt`)
    * This workflow will provide CI/CD when developing against this repo
    * Tests will be run against isolated databases that get torn down when the PR is merged
3. A workflow that runs the full test suite on manual dispatch (`test_dbt_models`)
    * This workflow will provide an interface to allow extraction jobs like [`service-sqoop-iasworld`](https://github.com/ccao-data/service-sqoop-iasworld/) to run tests after pulling in new data (see e.g.: https://github.com/ccao-data/service-sqoop-iasworld/issues/1)
    * Tests will run against the production environment

Closes https://github.com/ccao-data/data-architecture/issues/31.


jeancochrane:
One more time for testing the cleanup step 🤞🏻

jeancochrane:
Testing cleanup again!

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Cache dbt manifest
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.MANIFEST_DIR }}
+          key: ${{ env.CACHE_NAME }}-${{ env.CACHE_KEY }}
+          restore-keys: |
+            ${{ env.CACHE_NAME }}-data-catalog
+            ${{ env.CACHE_NAME }}-master
```
jeancochrane, line 72: The `restore-keys` setting here means that we will fall back to the `data-catalog` or `master` branch caches if the PR branch cache does not exist (which we only ever expect to happen on the first pushes to a PR, before the PR has had a successful run).

.github/workflows/test_dbt_models.yaml
jeancochrane, line 1: This workflow represents the tests that we will want to trigger from our nightly import process (https://github.com/ccao-data/service-sqoop-iasworld/issues/1). It is not actually tested yet as part of this project, since manual workflow dispatch is not possible until the workflow has been pulled into the main branch ([source](https://stackoverflow.com/a/71423764)).

Note that I sort of tested this workflow by manually editing the `on:` key below to run the workflow on pushes to this branch; test output [here](https://github.com/ccao-data/data-architecture/actions/runs/5788187928/job/15686641383).

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -0,0 +1,122 @@
...
+    branches: [master, data-catalog]
+
+jobs:
+  build-and-test-dbt:
+    runs-on: ubuntu-latest
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint
+    # so that we can authenticate with AWS
+    permissions:
+      id-token: write
+      contents: read
```
jeancochrane, line 23: More details on these permissions settings [here](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings).

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -0,0 +1,122 @@
...
+on:
+  pull_request:
+    branches: [master, data-catalog]
+    # Specifying event types manually allows us to run this flow when the
+    # PR is closed so that we can clean up staging dbt resources
+    types:
+      - opened
+      - synchronize
+      - closed
+      - reopened
```
jeancochrane, line 12: Note that these event types represent the default event types plus `closed`. [Docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request):

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -0,0 +1,122 @@
+name: Build and test dbt
+
+on:
+  pull_request:
+    branches: [master, data-catalog]
```
jeancochrane, line 5: There are a number of references to `data-catalog` here that cover for the fact that we don't expect this workflow to be pulled into `master` immediately, and that we want to treat `data-catalog` as a CI environment until we're ready to merge `data-catalog` into `master` (https://github.com/ccao-data/data-architecture/issues/55). Part of that merge issue will be stripping out the references to `data-catalog` in these workflows and ensuring that they are ready to be run against `master`.

dbt/models/default/schema.yml
```diff
@@ -31,15 +31,15 @@ models:
             - pin
             - year
           config:
-            error_if: ">280655"
+            error_if: ">280659"
```
jeancochrane, line 34: I have confirmed by running these tests locally that these changes represent additional malformed data that has been added to our source data since this PR opened. The fact that this happened twice during the ~week that this PR was open is an indication to me that I may want to pivot soon to focusing on resolving these data problems, since they're going to become annoying quite quickly once we're running these tests on every PR.

.github/workflows/test_dbt_models.yaml
jeancochrane, line 1: This workflow represents the tests that we will want to trigger from our nightly import process (https://github.com/ccao-data/service-sqoop-iasworld/issues/1). It is not fully tested yet as part of this project, since manual workflow dispatch is not possible until the workflow has been pulled into the main branch ([source](https://stackoverflow.com/a/71423764)). Instead, I tested the workflow by manually editing the `on:` key below to run the workflow on pushes to this pull request; test output [here](https://github.com/ccao-data/data-architecture/actions/runs/5789370346/job/15690283366).

.github/actions/load_environment_variables/action.yaml
jeancochrane, line 1: Similarly, both workflows share a common set of environment variables that are loaded using this action.

jeancochrane:
This is finally ready for a review! Probably best if both @dfsnow and @wrridgeway take a look, since the workflows introduced in this PR will operate as the runners of our tests.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -0,0 +1,122 @@
...
+            echo "Running tests on modified resources only"
+            dbt test --target "$TARGET" -s state:modified --state target/
+          else
+            echo "Running tests on all resources"
+            dbt test --target "$TARGET"
+          fi
+        working-directory: ${{ env.PROJECT_DIR }}
+        shell: bash
+
+      - if: ${{ github.event_name == 'pull_request' && github.event.action == 'closed' }}
```
dfsnow, line 113: **suggestion (blocking):** Maybe factor these out into a separate workflow if possible. Currently closing/merging a PR would re-run the entire build and test process, which will happen anyways in the prod env when merged to master.

There's also a risk that someone does something silly and gets their branch into a non-zero exit state in the builds/tests, see https://github.com/ccao-data/data-architecture/pull/52. If they then wanted to abandon/close that branch without merging and delete the ci env, they would need to reconcile their branch until everything runs in order to reach this step.

.github/workflows/test_dbt_models.yaml
```diff
@@ -0,0 +1,32 @@
...
+        uses: ./.github/actions/load_environment_variables
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+          aws-region: us-east-1
+
+      - name: Test models
+        run: dbt test --target ci
```
dfsnow, line 30: **question (non-blocking):** This switches to prod once `data-catalog` is merged, correct?

dbt/models/default/schema.yml
dfsnow: **suggestion (non-blocking):** Some of these are likely just under-specified tests, but I agree it's going to be a problem. Grab @wrridgeway this week for a review/debugging session of the current tests. He knows more than anyone but Mirella about the horrors of our changing data.

dfsnow (commented):
Overall amazing work @jeancochrane! This couldn't have been easy to debug. Just a couple outstanding blockers and we're good to go.

.github/actions/load_environment_variables/action.yaml
```diff
@@ -0,0 +1,15 @@
...
+    description: |
+      File path to variable file or directory.
+      Defaults to ./.github/variables/* if none specified
+      and runs against each file in that directory.
+    required: false
+    default: ./.github/variables/*
+runs:
+  using: composite
+  steps:
+    - run: sed "" ${{ inputs.env_var_file_path }} >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
jeancochrane: Embarrassing honesty: I just copy/pasted this from a super simple published action without thinking too hard about the choice of command! I did a quick [code search](https://github.com/search?q=Defaults+to+.%2F.github%2Fvariables%2F*+if+none+specified&type=code) to try to remember where I got the code from (based on the results it seems like it's common to copy/paste this action 😅) and I noticed [this variant](https://github.com/tw3lveparsecs/github-actions-set-variables/blob/23da06e2c4d5804d2082ecf4d79b83a2338fb419/action.yml#L16) that uses `sed` to remove comment lines. I thought that was a nice idea so I updated our action to do the same in e51c677bc22593f33a71131d7f7662c3cf348989, but let me know if you have a strong preference for `cat`!

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -0,0 +1,122 @@
+name: Build and test dbt
```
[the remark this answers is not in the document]
jeancochrane: Makes sense, done in f9b5d86bc32c8f345fb2d07d136eae1ac23e29f9 👍🏻

.github/workflows/build_and_test_dbt.yaml
jeancochrane: Smart, done in 704ed13d79918028c3bc0e66daec37ac759be9fc!

jeancochrane:
Closing this temporarily to test the new `cleanup-dbt-resources` workflow!

.github/workflows/build_and_test_dbt.yaml
jeancochrane: Here's a test run demonstrating that the factored-out workflow still works: https://github.com/ccao-data/data-architecture/actions/runs/5799412187/job/15719353357

dbt/models/default/schema.yml
jeancochrane: I scheduled some time tomorrow afternoon! In the meantime, I had to bump the thresholds yet again in 672a7d47628fcd1f868dbc07cd8ae8e8c560efff ☹️

jeancochrane:
This is ready for another look @dfsnow! I took just about all of your advice; the only thread with an open question is https://github.com/ccao-data/data-architecture/pull/50#discussion_r1287299584.

jeancochrane:
Actually, hold off on that review @dfsnow -- my testing on the docs branch (#57) is revealing that there may be a problem with the way we're setting things up for the `data-catalog` branch here.

jeancochrane:
Alright @dfsnow, this is ready again! I made one change that wasn't part of our PR discussion after realizing that we can't override the `GITHUB_ENV_VAR` environment variable in order to run the `ci` target on the `data-catalog` branch. More context in the commit message for https://github.com/ccao-data/data-architecture/pull/50/commits/60c2bb10b4731869ef51e25c15ef71ffcf5087a4:

> It turns out that we can't override the GITHUB_HEAD_REF variable in a GitHub Actions workflow,
which means we can't set it in order to treat the data-catalog branch environment as a CI target.
>
> To get around this, this commit changes up the generate_schema_name macro to read from a new
HEAD_REF env var instead of GITHUB_HEAD_REF, and then factors out a configure_dbt_environment
composite action that our dbt workflows can use to set this environment variable appropriately.

dfsnow (approved):
Looks great @jeancochrane, thanks for making those changes! Merge at your leisure

PR #52, opened 2023-08-07
Add stage indicators and deduplicate conditionally rather than through aggregation

wrridgeway:


dfsnow:
@wrridgeway I don't know enough about the underlying schema to say whether this will cause a regression / breaking change. This would really be a perfect case for dbt tests once they're ready.

**question (blocking):** How did you test that this doesn't cause a regression?

wrridgeway:
I've removed anything that could cause regression. All that's been added is two new columns indicating stage name and number.

dfsnow (approved):
This looks good to me. Nice work @wrridgeway

Issue #34, opened 2023-07-26
[Data catalog] Define GitHub Actions workflow for rebuilding and deploying docs site


jeancochrane:
Every commit to the `master` branch should run a workflow that rebuilds and deploys our documentation site (https://github.com/ccao-data/data-architecture/issues/33).

Bonus points if you can get [deploy previews](https://docs.netlify.com/site-deploys/deploy-previews/) set up, but not required.

Depends on #33.

Issue #55, opened 2023-08-07
[Data catalog] Merge `data-catalog` branch into `master`


jeancochrane:
Once we're feeling ready to put our data catalog work into production, merge the `data-catalog` branch into the `master` branch of this repo.

This will require updating our GitHub Actions workflows (#31) to remove references to `data-catalog` and point any operations that were running against that environment to run against the `prod` environment. Note that since the `data-catalog` branch runs dbt in the `ci` profile, we'll also want to double-check references to `--target ci` in the `data-catalog` branch context and switch them to `--target prod` where appropriate.

PR #57, opened 2023-08-08
Deploy dbt docs to GitHub Pages

jeancochrane:
This PR adds a GitHub Actions workflow for deploying our dbt docs to GitHub Pages. It implements the flow outlined in the GitHub documentation on [creating a custom GitHub Actions workflow to publish your site](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#creating-a-custom-github-actions-workflow-to-publish-your-site); this feels like a more appropriate choice than [publishing from a branch](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-from-a-branch) because we don't want to keep our generated docs files under version control if we can avoid it.

For a sample of the deployed site, see: https://ccao-data.github.io/data-architecture

For an example of a successful run that built the site, see: https://github.com/ccao-data/data-architecture/actions/runs/5800847533/job/15723822771

Closes https://github.com/ccao-data/data-architecture/issues/34.

### DIFF
--- a/.github/workflows/deploy_dbt_docs.yaml
+++ b/.github/workflows/deploy_dbt_docs.yaml
@@ -1,12 +1,12 @@
-name: Deploy dbt docs
+name: deploy-dbt-docs
 
 on:
   push:
-    branches: [master, data-catalog]
-  # TODO: Remove this after testing
-  pull_request:
     branches:
-      - 'jeancochrane/31-data-catalog-define-github-actions-workflows-for-building-the-dbt-dag-and-running-tests'
+      - master
+      - data-catalog
+      # TODO: Remove this after testing
+      - 34-data-catalog-define-github-actions-workflow-for-rebuilding-and-deploying-docs-site
 
 jobs:
   deploy-dbt-docs:
@@ -36,6 +36,8 @@ jobs:
         run: dbt docs generate --target ci
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
+        env:
+          GITHUB_HEAD_REF: 34-data-catalog-define-github-actions-workflow-for-rebuilding-and-deploying-docs-site
 
       - name: Upload docs directory artifact
         uses: actions/upload-pages-artifact@v2
--- a/.github/workflows/deploy_dbt_docs.yaml
+++ b/.github/workflows/deploy_dbt_docs.yaml
@@ -14,7 +14,6 @@ jobs:
     permissions:
       pages: write      # To deploy to Pages
       id-token: write   # To verify the deployment comes from an valid source
-    if: github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success'
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/deploy_dbt_docs.yaml
+++ b/.github/workflows/deploy_dbt_docs.yaml
@@ -33,7 +33,7 @@ jobs:
           aws-region: us-east-1
 
       - name: Generate docs
-        run: dbt --debug docs generate --target ci
+        run: GITHUB_HEAD_REF=$GITHUB_HEAD_REF dbt docs generate --target ci
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
         env:
--- a/.github/workflows/deploy_dbt_docs.yaml
+++ b/.github/workflows/deploy_dbt_docs.yaml
@@ -3,7 +3,7 @@ name: deploy-dbt-docs
 on:
   workflow_run:
     workflows: [build-and-test-dbt]
-    branches: [master, data-catalog, 34-data-catalog-define-github-actions-workflow-for-rebuilding-and-deploying-docs-site]
+    branches: [master, data-catalog]
     types: [completed]
   workflow_dispatch:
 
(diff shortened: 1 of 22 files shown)


.github/workflows/deploy_dbt_docs.yaml
```diff
@@ -0,0 +1,71 @@
+name: deploy-dbt-docs
+
+on:
+  workflow_run:
+    workflows: [build-and-test-dbt]
+    branches: [master, data-catalog]
+    types: [completed]
```
jeancochrane, line 7: The idea here is that we only want this workflow to run after the `build-and-test-dbt` workflow has completed on the `master` or `data-catalog` branch. Unfortunately, there's no `type` that corresponds to a _successful_ build, so the best we can do is check for the `completed` type and gate each job step behind the `github.event.workflow_run.conclusion == 'success'` expression (see [this SO thread](https://stackoverflow.com/a/64733705) and the [GitHub Actions feature request](https://github.com/actions/runner/issues/662) that it links to).

.github/workflows/deploy_dbt_docs.yaml
```diff
@@ -0,0 +1,71 @@
...
+        env:
+          HEAD_REF: data-catalog
+
+      - if: github.event.workflow_run.conclusion == 'success'
+        name: Package doc files for upload
+        run: |
+          mkdir _site
+          for file in index.html catalog.json manifest.json; do
+            cp "target/$file" "_site/$file"
+          done
```
jeancochrane, line 58: The `target` directory includes a bunch of metadata about the run, stored in files that could be accessed via GitHub Pages if someone knew what to look for. I scanned these files and couldn't find anything that looked obviously sensitive, but my confidence in that threat assessment is low (the files can be big), so I'm playing it safe and only explicitly packaging up files that are necessary to serve the docs site.

.github/workflows/deploy_dbt_docs.yaml
dfsnow: **suggestion (blocking):** I believe the entire `deploy-dbt-docs` job can be gated with a single if statement. See example here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow. I would refactor for simplicity if possible, but I might be missing some caveat here.

.github/workflows/deploy_dbt_docs.yaml
```diff
@@ -0,0 +1,71 @@
...
+    # These permissions are required to make a GitHub Pages deployment
+    permissions:
+      pages: write      # To deploy to Pages
+      id-token: write   # To verify the deployment comes from an valid source
+
+    steps:
+      # There is currently no way to early-exit a run with a success status
+      # in GitHub actions, so we have to mark each of these steps with the
+      # same `if` condition to restrict deploys to successful builds :(
+      # https://github.com/actions/runner/issues/662
```
dfsnow, line 21: I _think_ just gating the job with the if statement will prevent the steps from running at all, negating the need for the early exit.

.github/workflows/deploy_dbt_docs.yaml
```diff
@@ -0,0 +1,71 @@
+name: deploy-dbt-docs
+
+on:
+  workflow_run:
```
dfsnow, line 4: **suggestion (non-blocking):** I would add a [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) option on the off chance that we need to manually rebuild the pages from a run.

dfsnow (changes requested):

.github/workflows/deploy_dbt_docs.yaml
jeancochrane: I think that's correct as well, but I'm testing in 4a46220341271be54b98015106a13337761aba2f!

.github/workflows/deploy_dbt_docs.yaml
jeancochrane: Good call, done in ee4a7602b27bc7142e7f6bcc3e778187946dc3c6 👍🏻

.github/workflows/deploy_dbt_docs.yaml
jeancochrane: It turns out we can't test this behavior because both [workflow_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) and [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) will only run if the workflow file exists on the default branch 🙃 

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

Are we comfortable merging this without testing and testing on prod instead? Note another downside of this limitation is we won't start building docs until we merge `data-catalog` into `master` (#55), but maybe we can bump the timeline on that issue?

dfsnow: I think that's fine, nothing will break if it fails in prod except for the docs. Let's see how the checks meeting goes re: bumping #55.

dfsnow (approved):

Issue #26, opened 2023-07-26
[Data catalog] Document remaining tables and views


jeancochrane:
The initial dbt project we added in https://github.com/ccao-data/data-architecture/pull/23 contains documentation for many of the Athena views that we build on top of our source data, but not all of them. It also does not contain documentation for source tables yet. Add documentation for these tables and views.

jeancochrane:
Views have been added in https://github.com/ccao-data/data-architecture/pull/64, but we're blocked on documentation by #61.

jeancochrane:
Superseded by #113.

PR #64, opened 2023-08-10
Add stub dbt models for the remaining Athena views and tables

### LINKED ISSUE #24: [Data catalog] Add stub models for tables/views defined outside the DAG
The models defined in our dbt project currently only represent the Athena views that we build on top of our source data, but we would also like to include models for our source data tables so that we can add them to our data catalog documentation. To accomplish this, we'll need to add "stub" models that don't actually build new data, they just allow dbt to recognize the model as a valid entity that can be documented.

For guidance, see this dbt community post: https://discourse.getdbt.com/t/representing-non-sql-models-in-a-dbt-dag/2083

If this issue gets picked up before we have a proper dev/prod split set up (https://github.com/ccao-data/data-architecture/issues/28), it might be wise to continue following the pattern of suffixing models with `_test` until we have environments up and we can safely remove the suffix (https://github.com/ccao-data/data-architecture/issues/29).

jeancochrane:
This PR updates our dbt config to add a large number of stubs for the remaining models that are not yet being managed by dbt. By "stubs", I mean that these models currently contain no documentation, tests, or substantial queries; instead, the configuration is intentionally kept as simple as possible so that we can start tracking and fleshing out these models incrementally.

In other words: Don't be alarmed by the 192 changed files, the diff here is in fact quite simple (<500 lines of code)!

All views are added to the dbt config by symlinking the SQL file in `aws-athena/views` to an entry in the corresponding schema directory in `dbt/models`, following the pattern defined in #36. Tables are added as [ephemeral models](https://docs.getdbt.com/docs/build/materializations#ephemeral) whose definition is simply `select * from <original_table>`, which allows us to track those models in the DAG and test/document them without having to actually recreate them when we run `dbt run`.

Closes #24 and connects to #26.


dbt/models/model/schema.yml
```diff
@@ -0,0 +1,21 @@
...
+  - name: feature_importance
+  - name: final_model
+  - name: metadata
+  - name: parameter_final
+  - name: parameter_range
+  - name: parameter_search
+  - name: performance
+  - name: performance_quantile
+  - name: shap
+  - name: test_card
```
jeancochrane, line 16: Do we actually want to keep `test_card` in the DAG, or was this a temporary testing table?

dbt/models/rpie/schema.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+
+
+models:
+  - name: pin_codes
+  - name: pin_codes_dummy
```
jeancochrane, line 6: Same here: Is this "dummy" table necessary to track in the DAG?

dfsnow: Would need to confirm with @wrridgeway, but I'm pretty confident this is in fact an important table.

.github/scripts/cleanup_dbt_resources.sh
```diff
@@ -29,7 +29,7 @@ echo "Deleting the following schemas from Athena:"
 echo
 echo "$schemas"
 
-echo "$schemas" | xargs -i bash -c 'aws glue delete-database --name {} || exit 255'
+echo "$schemas" | xargs -i bash -c 'aws glue delete-database --name {}'
```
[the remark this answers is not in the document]
dfsnow: **question (non-blocking):** So this only runs to cleanup PR branches and therefore has no chance of deleting the prod ephemeral tables (which _do_ exist), is that correct?

If we swaps our definitions such that ephemeral must be manually defined and views are the default, would `--select config.materialized:view` then work? Is there any benefit to this besides the additional guarantee of returning no prod tables?

dfsnow (approved):
@jeancochrane Looks great! Just a few non-blocking questions. Merge at your leisure!

.github/scripts/cleanup_dbt_resources.sh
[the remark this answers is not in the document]
jeancochrane: > If we swaps our definitions such that ephemeral must be manually defined and views are the default, would --select config.materialized:view then work?

Good point @dfsnow! I played around with making `materialized: view` the default in 19520879d140433c08ac48b4a25e358000db7a28 and managed to get it filtering properly with `--exclude config.materialized:ephemeral` (which will be slightly more robust if/when we introduce `table` or `incremental` models). I'm worried that the config in `dbt_project.yml` is more confusing as a result, but I left a long comment that will hopefully help document the pattern.

> So this only runs to cleanup PR branches and therefore has no chance of deleting the prod ephemeral tables (which do exist), is that correct?

Yup, that's right! For an additional layer of protection, the `cleanup_dbt_resources.sh` script will error if you try to run it on prod:

https://github.com/ccao-data/data-architecture/blob/dda75c833b8636da3080157fbae6bc3e571b1d33/.github/scripts/cleanup_dbt_resources.sh#L17-L20

>  Is there any benefit to this besides the additional guarantee of returning no prod tables?

By "is there any benefit to this" I assume you're asking what the benefit of selecting only non-ephemeral models in `dbt list` is, right? If so, the main benefit I see is that by filtering for only resources we expect to exist, we can operate with the assumption that if `aws glue delete-database` ever completes with a non-zero status code, it's an indication of a real problem worth investigating and we can fail the CI job accordingly. Does that make sense?

dbt/models/model/schema.yml
jeancochrane: Gotcha, I'll leave this in then!

dbt/models/rpie/schema.yml
jeancochrane: Cool, I'll leave it in unless/until Billy tells me to remove it.

jeancochrane:
Thanks for the thoughts and questions @dfsnow! Mind taking a quick look at my response to your question about materialization configs (and the corresponding commit, if you'd like) to make sure my updated approach makes sense?

.github/scripts/cleanup_dbt_resources.sh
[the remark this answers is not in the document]
dfsnow: > By "is there any benefit to this" I assume you're asking what the benefit of selecting only non-ephemeral models in dbt list is, right? 

Correct!

> Does that make sense?

Yes! I think this is a good change, as now there's no chance of even listing the ephemeral models and we get extra failure indicators in the CI. Good work!

dfsnow:
> Thanks for the thoughts and questions @dfsnow! Mind taking a quick look at my response to your question about materialization configs (and the corresponding commit, if you'd like) to make sure my updated approach makes sense?

This update looks sensible to me @jeancochrane. I think it will be more flexible in the future despite being slightly confusing. Your comment in `dbt_project.yml` is good, but we should probably add this somewhere to the dbt docs or to a README file in this repo.

jeancochrane:
Good call @dfsnow, I added an issue to the data catalog milestone and the Architecture Upgrades project to track needs around doc updates (https://github.com/ccao-data/data-architecture/issues/67) 👍🏻

Issue #68, opened 2023-08-14
[Data catalog] dbt CI is not properly selecting for new and modified models


jeancochrane:
I noticed that the [most recent CI run](https://github.com/ccao-data/data-architecture/actions/runs/5857518680/job/15879535306) for the `data-catalog` branch did not (re)build any of the models we added or changed in https://github.com/ccao-data/data-architecture/pull/64. Testing things out locally, I noticed two problems:

1. We're missing the `state:new` selector in our workflow configuration, which is intended to select for new models
2. In my local environment, neither `state:new` nor `state:modified` properly picks up additions or changes to the model definition

dfsnow:
Closing manually since for some reason #69 did not close this on merge.

PR #69, opened 2023-08-14
Separate dbt target and state directories to properly support slim CI

jeancochrane:
One underdocumented subtlety of dbt's [state selection options](https://docs.getdbt.com/reference/node-selection/methods#the-state-method) is that when selecting resources based on state attributes (e.g. selecting only resources that are new or modified since last run, for "slim CI") you have to point to a _separate_ state directory containing a manifest file from the previous run, you can't just point to the default state directory that dbt uses (`target/`). I had assumed it was best practice to use one state directory, so I designed the `build-and-test-dbt` workflow to save/restore the state cache from the default `target/` state directory; however, it appears that dbt updates its own manifest file before comparing it to the state manifest file to see which resources have changed, so if those manifest files are the same file then dbt will never think that any resources have changed.

This PR fixes our `build-and-test-dbt` workflow such that we compare dbt state to the cached manifest file using a new `state/` directory that is separate from `target/`.

Closes https://github.com/ccao-data/data-architecture/issues/68.

### DIFF
--- a/.github/variables/dbt.env
+++ b/.github/variables/dbt.env
@@ -1,3 +1,3 @@
 CACHE_NAME=dbt-cache
-MANIFEST_DIR=dbt/target
+MANIFEST_DIR=dbt/state
 PROJECT_DIR=dbt
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -61,11 +61,11 @@ jobs:
       - name: Build models
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
-            echo "Running build on modified resources only"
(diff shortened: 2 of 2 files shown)


dbt/models/default/schema.yml
```diff
@@ -31,15 +31,15 @@ models:
             - pin
             - year
           config:
-            error_if: ">280662"
+            error_if: ">280679"
```
jeancochrane, line 34: As per usual these days, I also had to bump these thresholds to get tests to pass. Billy is still investigating but his hunch is that the assumptions behind these two tests are not actually correct.

dfsnow (approved):
This looks great! Nice debugging @jeancochrane. Merge at your leisure.

PR #74, opened 2023-08-16
Remove deactivated rows from views

wrridgeway:


wrridgeway:
this is done as far as i can tell. next is to fix aggregated land in every view once i'm sure we're calculating it correctly.

dfsnow (approved):

PR #77, opened 2023-08-17
Replace all aggregate land calculations with pulls from agg land view

wrridgeway:
Aggregating land square footage needs to incorporate the `influ` column in `iasworld.land` and some land line heuristics, rather than just summing by parcel.

We need to make sure every view that aggregates land sf uses this new view for consistency and accuracy.


dfsnow (changes requested):
@wrridgeway I'm seeing some differences between the original `aggregate_land` CTE total row count and the new `default.vw_pin_land` total row count. Here's an example:

```sql
SELECT COUNT(*)
FROM "default"."vw_pin_land"
WHERE year = '2022';
```

Row count: 1,864,776

```sql
SELECT COUNT(*)
FROM (
    SELECT parid
    FROM "iasworld"."land"
    WHERE taxyr = '2022'
        AND land.cur = 'Y'
        AND land.deactivat IS NULL
    GROUP BY parid, taxyr
)
```

Row count: 1,864,932

Can you look into the difference between these two?

wrridgeway:
Good catch. Should be sorted now, can't use lline = 1 since that row can be deactivated in some cases.

dfsnow (approved):
Great! Looks good now @wrridgeway. I'll go ahead and merge this.

PR #78, opened 2023-08-17
Merge data catalog branch into master

jeancochrane:
Now that our dbt work is relatively stable (https://github.com/ccao-data/data-architecture/milestone/1), we're ready to pull it in and start deploying it to production. This PR merges the long-lived `data-catalog` branch that we've been using for MVP dbt development (containing 12 PRs!) into `master`.

The diff here is very large, but only a few commits are new to this PR, while the rest represent PRs that have already been approved and merged into `data-catalog`. See [this filtered file view](https://github.com/ccao-data/data-architecture/pull/78/files/7a3d0e0704ec0d34fa9ec7138e1c3c040aacc2ce..0efef26f1718f43741694b9443a1fca973589e7e) to restrict your review to only unreviewed commits.

A few changes to expect once this PR lands:

* dbt will start deploying our models to the production views in Athena on every merge to `master`
* dbt will start running builds and tests for every PR opened against `master`
* The data catalog docs will start deploying to GitHub Pages on every merge to `master`
* The `test-dbt-models` workflow will become available to run via workflow dispatch, unblocking https://github.com/ccao-data/service-sqoop-iasworld/issues/1

Note that I'm planning to merge these commits in without squashing, so that we can preserve the PR history embedded in this long-lived branch.

Closes https://github.com/ccao-data/data-architecture/issues/55.


jeancochrane:
Requesting both @dfsnow and @wrridgeway since this is a pretty big change that will affect both of you (and will cause merge conflicts with #74 and #77 if it lands first)! But I'm expecting that Dan will do the review.

.github/workflows/deploy_dbt_docs.yaml
```diff
@@ -32,14 +32,9 @@ jobs:
...
       - name: Generate docs
-        # Target is currently set to CI because we expect this action to be
-        # run against the long-lived data-catalog branch, but we should change
-        # this to prod when we merge that branch into master
-        run: dbt docs generate --target ci
+        run: dbt docs generate --target "$TARGET"
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
-        env:
-          HEAD_REF: data-catalog
```
dfsnow: **question (dumb):** This does _not_ need to be set for master right? It was a specific setup for the long-lived catalog branch?

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -0,0 +1,87 @@
...
+            dbt run -t "$TARGET"
+          fi
+        working-directory: ${{ env.PROJECT_DIR }}
+        shell: bash
+
+      - name: Test models
+        run: |
+          if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
+            echo "Running tests on modified/new resources only"
+            dbt test -t "$TARGET" -s state:modified state:new --state "$STATE_DIR"
```
dfsnow, line 76: **question (non-blocking):** So dbt knows to run tests for any part of the DAG where the models are modified/new. Does it also know to run tests for anywhere that the _tests themselves_ are modified/new? I.e. if we start creating PRs that don't modify a view but add many tests, will dbt know to run them?

dfsnow (approved):
Amazing work @jeancochrane. I'm excited to see this merged and browse around the live docs for the first time. This will be a boon to the whole department and should save us from _so_ many headaches in the long run.

🎉 🎉

.github/workflows/deploy_dbt_docs.yaml
jeancochrane: That's right! We only need `HEAD_REF` to be set on CI in order to properly generate the schema; dev and prod follow their own pattern that doesn't rely on this env var.

https://github.com/ccao-data/data-architecture/blob/0efef26f1718f43741694b9443a1fca973589e7e/dbt/macros/generate_schema_name.sql#L29-L34

Normally we pull `HEAD_REF` automatically from the `GITHUB_HEAD_REF` variable that is [included by default](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) in `pull_request` worfklow events, but we needed to set `HEAD_REF` here manually because we wanted to pretend that `data-catalog` was a CI branch for the sake of testing. (Unfortunately this didn't end up mattering, because GitHub Pages can only deploy from workflows that are on the main branch 😅 )

https://github.com/ccao-data/data-architecture/blob/0efef26f1718f43741694b9443a1fca973589e7e/.github/actions/configure_dbt_environment/action.yaml#L10-L16

.github/workflows/build_and_test_dbt.yaml
jeancochrane: > Does it also know to run tests for anywhere that the tests themselves are modified/new?

That's my understanding, yes! Tests are a type of node, just like models, and my read of [the docs](https://docs.getdbt.com/reference/node-selection/methods#the-state-method) is that state selection applies to all node types. But I haven't confirmed this locally yet, I'll do so now!

PR #81, opened 2023-08-18
Update data catalog design doc to explain how non-SQL scripting fits into the catalog

jeancochrane:
This PR updates the data catalog design doc with my thoughts on a proposed design for integrating our Glue jobs and our R extraction/transformation scripts into our dbt catalog. While Glue and R do not have native integrations into dbt that would allow us to manage our entire data pipeline with one, I think it will only take a relatively small amount of engineering to set up dev workflows where these three tools can interact well and understand how to use one another's output. 

Closes https://github.com/ccao-data/data-architecture/issues/83.


documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,68 @@ and validating our data using dbt:
...
+* We will keep our raw data extraction scripts separate from the dbt DAG, per
+  [dbt's recommendation](https://docs.getdbt.com/terms/data-extraction).
+* Raw data will be referenced in dbt transformations using the [`source()`
+  function](https://docs.getdbt.com/reference/dbt-jinja-functions/source).
+* [Source freshness
+  checks](https://docs.getdbt.com/docs/deploy/source-freshness) will be used to
+  ensure that raw data is updated appropriately prior to transformation.
+* Where possible, the intermediate transformations defined in the
+  `aws-s3/scripts-ccao-data-warehouse-us-east-1` subdirectory will be rewritten
+  in SQL and moved into the dbt DAG. During the transition period, while some
```
jeancochrane, line 368: Transitioning our R transformation scripts to SQL is a big new step that we haven't talked about yet. Based on my read of `aws-s3/scripts-ccao-data-warehouse-us-east-1`, it seems like everything can either be transitioned to SQL or comfortably considered a "source" instead of a data transformation; does that match your understanding? It will be a lift to move all this stuff but I think our future selves will thank us if we can move as much stuff as possible into the DAG.

This also reminds me that we don't really have an issue yet to track the work to transition CTAs to the DAG; perhaps I'll go ahead and open that in the Architecture Upgrades milestone and we can prioritize it later?

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,68 @@ and validating our data using dbt:
...
+  services.
+* Glue jobs will be kept under version control and deployed to AWS using
+  [Terraform](https://www.terraform.io/) run in GitHub Actions on
+  commits to their repo's main branch. We will write a reusable [composite
+  action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action)
+  that performs the following operations:
+    1. Runs `terraform apply` to recreate the Glue job definition in AWS
+      1. In doing so, inserts the current Git SHA as a command argument in the
+         Glue job definition so that the job script can read the SHA and use it
+         for versioning.
```
jeancochrane, line 390: Does this make sense? Do others agree that this will allow us to perform any versioning we need to do (e.g. store the SHA of the commit that generated the job script in a metadata table) while also allowing us to maintain a single job definition file in S3?

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,68 @@ and validating our data using dbt:
...
+      dbt), we will separate the dbt config into two targets, use the second
+      bullet approach (dbt -> Glue) to trigger the Glue job once the first
+      target has completed, and update the dbt wrapper script to initiate
+      the second dbt target build once the Glue job has completed.
+        * This wrapper script should also provide the caller with the option
+          to skip running the Glue job if the AWS CLI can determine that
+          the output of the Glue job already exists.
+        * The opposite circular dependency (Glue -> dbt -> Glue) should not
+          require a special solution since it is just a combination of the
+          first and second bullets above.
```
jeancochrane, line 417: I think this is correct, but it's a bit confusing. Am I wrong here?

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+      dbt), we will document the Glue job as an [ephemeral
+      model](https://docs.getdbt.com/docs/build/materializations#ephemeral) in
+      dbt so that we can specify its dependencies using [the `depends_on`
+      attribute](https://docs.getdbt.com/reference/dbt-jinja-functions/ref#forcing-dependencies).
+      If we would like to be able to build the entire DAG from scratch,
+      including running the Glue jobs and transforming their output using
+      dbt, we can separate the dbt config into two targets, use the second
+      bullet approach above (dbt -> Glue) to trigger the Glue job once the first
+      target has completed, and update the dbt wrapper script to initiate
+      the second dbt target build once the Glue job has completed.
```
jeancochrane, line 421: I think we should only do this in cases where we absolutely need to be able to build the full DAG from scratch, since it would introduce a lot of complexity into the build process. But I wanted to sketch it out here to make it clear that we should be able to tie together basically any Glue job with our dbt catalog given enough engineering effort.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+  function](https://docs.getdbt.com/reference/dbt-jinja-functions/source).
+* [Source freshness
+  checks](https://docs.getdbt.com/docs/deploy/source-freshness) will be used to
+  ensure that raw data is updated appropriately prior to transformation.
+* Where possible, the intermediate transformations defined in the
+  `aws-s3/scripts-ccao-data-warehouse-us-east-1` subdirectory will be rewritten
+  in SQL and moved into the dbt DAG. During the transition period, while some
+  transformations are still written in R, we will treat their output as if it
+  were raw data and reference it using `source()`. Any transformations that
+  can't be easily rewritten in SQL will continue to be defined this way in the
```
dfsnow, line 371: It might be useful long-term to rewrite some of them as Python, especially if we expect dbt to add better Python support in the near future. Certainly we can write any _new_ transform scripts that don't work as SQL using Python.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+  long term.
+* Intermediate or final transformations that require CPU- or memory-intensive
+  operations like running machine learning models will be defined in Python,
+  run as AWS Glue jobs, and defined as [ephemeral
+  models](https://docs.getdbt.com/docs/deploy/source-freshness) in the dbt DAG.
```
dfsnow, line 376: **nitpick (blocking):** Incorrect link, should be: https://docs.getdbt.com/docs/build/materializations#ephemeral.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+  action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action)
+  that performs the following operations:
+    1. Runs `terraform apply` to recreate the Glue job definition in AWS
+      1. In doing so, inserts the current Git SHA as a command argument in the
+         Glue job definition so that the job script can read the SHA and use it
+         for versioning.
+      2. Supports creating staging jobs that we can use for testing during CI.
+    2. Uploads the newest version of the script to the proper bucket in S3
+* There will be three expected ways in which we handle dependencies between
+  dbt and glue, depending on the direction of the dependency graph:
```
dfsnow, line 394: **nitpick:** glue --> Glue, for consistency.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+    * In cases where dbt depends on the output of a Glue job (Glue -> dbt), we
+      will treat the Glue job output as an  or
```
dfsnow, line 396: **nitpick:** Missing a noun here.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+          instead of the Glue job scheduler and configure GitHub Actions to
+          rerun dbt in case of a successful Glue job run.
+    * In cases where a Glue job depends on the output of dbt (dbt -> Glue),
+      we will document the Glue job as an
+      [exposure](https://docs.getdbt.com/docs/build/exposures) in the DAG.
+      If we would like to ensure that we run the Glue job every time the
+      dbt source data updates, we can write a wrapper script around `dbt run`
+      that uses the Glue `StartJobRun` API
+      ([docs](https://docs.aws.amazon.com/glue/latest/webapi/API_StartJobRun.html))
+      to trigger a job run once the dbt build completes successfully.
```
dfsnow, line 410: **suggestion:** I don't love this. Maybe we can instead integrate it with Actions and query the dbt state to trigger the job? Unsure what's possible here or how often this will really come up.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+      will treat the Glue job output as an  or
+      a [source](https://docs.getdbt.com/docs/build/sources) in the DAG and
+      schedule the job as necessary to maintain freshness.
+        * If we would like to rebuild the dbt models every time the Glue
+          source data updates, we can schedule the job via GitHub Actions
+          instead of the Glue job scheduler and configure GitHub Actions to
+          rerun dbt in case of a successful Glue job run.
+    * In cases where a Glue job depends on the output of dbt (dbt -> Glue),
+      we will document the Glue job as an
+      [exposure](https://docs.getdbt.com/docs/build/exposures) in the DAG.
```
dfsnow, line 405: **suggestion:** We should also take advantage of the `depends_on` attribute for exposures here.

dfsnow (changes requested):
Great work @jeancochrane. I generally agree with everything outlined here and will continue to think about any clever solutions we can use to simplify things a bit.

See my blocking comments for changes.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
   failures](https://docs.getdbt.com/reference/resource-configs/store_failures)
```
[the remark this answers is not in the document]
jeancochrane: Yeah, I think this will be a useful feature when we start engineering a pipeline to report test output to different stakeholders! I've played around with it a bit; at a high level, each test gets its own view with a set of rows that failed for the test. I think it would take some engineering effort to transform the raw set of rows into a format that highlights the error at hand and helps a person make a decision about how to fix it, but it's nice to know we have this option as we start to think about the notification stream for our tests.

documentation/design-docs/data-catalog.md
jeancochrane: Definitely aligned with this! Should I add it as a note to the doc, or are we fine with keeping this as tacit knowledge?

documentation/design-docs/data-catalog.md
jeancochrane: Oops, good catch! Fixed in b92c3ebe281ccd9b71f0069654727c1f4c587afe.

documentation/design-docs/data-catalog.md
jeancochrane: That's reasonable; should we strike this section of the doc until we make a decision on that, then?

documentation/design-docs/data-catalog.md
jeancochrane: Good catch, done in a04e4f5fd42ee0af8de59b1afc6a1293442a957a!

documentation/design-docs/data-catalog.md
jeancochrane: Fixed in a04e4f5fd42ee0af8de59b1afc6a1293442a957a.

documentation/design-docs/data-catalog.md
jeancochrane: True, I added a brief note about this in a04e4f5fd42ee0af8de59b1afc6a1293442a957a.

documentation/design-docs/data-catalog.md
jeancochrane: I think that makes sense! It seems like it just depends on whether we want dbt to be coupled to the Glue job, or the Glue job to be coupled to dbt; given the dependency order here I agree that it probably makes sense to couple Glue to dbt, since dbt doesn't need to care about how its downstream consumers consume it. I adjusted this note in da2d0acf35495fff221b5c8675eb5ab398b1f429, but I'm up for tweaking it further if we come up with a better solution.

documentation/design-docs/data-catalog.md
```diff
@@ -339,6 +354,79 @@ and validating our data using dbt:
...
+      we will document the Glue job as an
+      [exposure](https://docs.getdbt.com/docs/build/exposures) in the DAG.
+      If we would like to ensure that we run the Glue job every time the
+      dbt source data updates, we can write a wrapper script around `dbt run`
+      that uses the Glue `StartJobRun` API
+      ([docs](https://docs.aws.amazon.com/glue/latest/webapi/API_StartJobRun.html))
+      to trigger a job run once the dbt build completes successfully.
+    * In case of a circular dependency between dbt and Glue (dbt -> Glue ->
+      dbt), we will document the Glue job as an [ephemeral
+      model](https://docs.getdbt.com/docs/build/materializations#ephemeral) in
```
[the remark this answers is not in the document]
jeancochrane: You actually can query from ephemeral models in the context of a view! As an example, I tested with two dummy models, one of which is ephemeral and one of which is a view:

```sql
-- models/iasworld/valclass_ephemeral.sql
{{ config(materialized='ephemeral') }}

select * from {{ source('iasworld', 'valclass') }}
```

```sql
-- models/iasworld/valclass_view.sql
{{ config(materialized='view') }}

select * from {{ ref('valclass_ephemeral') }}
```

You can see the definition of the view in Athena under `dev_jecochr_iasworld.valclass_view` and test a query against it:

```sql
CREATE OR REPLACE VIEW "valclass_view" AS 
WITH
  __dbt__cte__valclass_ephemeral AS (
   SELECT *
   FROM
     "awsdatacatalog"."iasworld"."valclass"
) 
SELECT *
FROM
  __dbt__cte__valclass_ephemeral
```

```sql
select * from "dev_jecochr_iasworld"."valclass_view" limit 10
```

I'll delete this dummy view from Athena when I pull in this PR, but in the meantime it's up there in case you'd like to confirm with your own testing.

dfsnow (approved):

documentation/design-docs/data-catalog.md
dfsnow: Fine to keep it for now! Let's wait until one of us actually (hopefully) finds something clever.

documentation/design-docs/data-catalog.md
dfsnow: Let's keep it as tacit knowledge for the time being.

documentation/design-docs/data-catalog.md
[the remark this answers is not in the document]
dfsnow: Got it, so it just ends up as a queryable CTE. That works for me!

dfsnow (approved):
Ready to go @jeancochrane! I'm still slightly uneasy about the dbt -> Glue relationship but I think this looks great for now. Will continue to think about better ways to manage all the Glue stuff.

PR #82, opened 2023-08-18
Fix dbt production deploy problems

jeancochrane:
Three problems that emerged out of merging the data catalog branch into master (#78):

1. [Docs are not deploying](https://github.com/ccao-data/data-architecture/actions/runs/5905883640/job/16020814093) because we neglected to set up environment variables properly for the prod context
2. [Updating the dbt cache can fail](https://github.com/ccao-data/data-architecture/actions/runs/5906036003/job/16021252615) due to a naive `mv` call that is not actually performing the correct directory renaming
3. `$MODIFIED_RESOURCES_ONLY` [can be set to `false`](https://github.com/ccao-data/data-architecture/actions/runs/5906092110/job/16021431024) when we want it to be `true` in cases where we have to restore the cache from the `restore-keys` fallback

This PR fixes those three problems.

### DIFF
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -82,6 +82,6 @@ jobs:
         shell: bash
 
       - name: Move dbt manifest directory to update cache
-        run: mv "$TARGET_DIR" "$STATE_DIR"
+        run: rsync -av --delete "$TARGET_DIR/" "$STATE_DIR"
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -33,21 +33,35 @@ jobs:
       - name: Configure dbt environment
         uses: ./.github/actions/configure_dbt_environment
 
-      - name: Cache dbt state directory
+      # We have to use the separate `restore`/`save` actions instead of the
+      # unified `cache` action because only `restore` provides access to the
+      # `cache-matched-key` and `cache-primary-key` outputs as of v3
+      - name: Restore dbt state cache
         id: cache
-        uses: actions/cache@v3
+        uses: actions/cache/restore@v3
         with:
           path: ${{ env.PROJECT_DIR }}/${{ env.STATE_DIR }}
           key: ${{ env.CACHE_NAME }}-${{ env.CACHE_KEY }}
           restore-keys: |
             ${{ env.CACHE_NAME }}-master
 
-      - if: ${{ steps.cache.outputs.cache-hit == 'true' }}
+      # If we restore the cache from the `restore-keys` key, the `cache-hit`
+      # output will be 'false' but the `cache-matched-key` output will be
+      # the name of the `restore-keys` key; we want to count this case as a hit
+      - if: |
+          steps.cache.outputs.cache-hit == 'true' ||
+          steps.cache.outputs.cache-matched-key == format(
+            '{0}-master', env.CACHE_NAME
+          )
         name: Set command args to build/test modified resources
         run: echo "MODIFIED_RESOURCES_ONLY=true" >> "$GITHUB_ENV"
         shell: bash
 
-      - if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+      - if: |
+          steps.cache.outputs.cache-hit != 'true' &&
+          steps.cache.outputs.cache-matched-key != format(
+            '{0}-master', env.CACHE_NAME
+          )
         name: Set command args to build/test all resources
         run: echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
         shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test_dbt.yaml
```diff
@@ -82,6 +82,6 @@ jobs:
         shell: bash
 
       - name: Move dbt manifest directory to update cache
-        run: mv "$TARGET_DIR" "$STATE_DIR"
+        run: rsync -av --delete "$TARGET_DIR/" "$STATE_DIR"
```
jeancochrane, line 85: [The error in question](https://github.com/ccao-data/data-architecture/actions/runs/5906036003/job/16021252615):

```
Run mv "$TARGET_DIR" "$STATE_DIR"
mv: cannot move 'target' to 'state/target': Directory not empty
Error: Process completed with exit code 1.
```

An alternative would be to `rm -Rf "$STATE_DIR" && mv "$TARGET_DIR" "$STATE_DIR"`, but I liked the one-liner suggested by [this StackExchange answer](https://superuser.com/a/623846).

.github/workflows/deploy_dbt_docs.yaml
```diff
@@ -31,6 +31,9 @@ jobs:
           role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
           aws-region: us-east-1
 
+      - name: Configure dbt environment
+        uses: ./.github/actions/configure_dbt_environment
```
jeancochrane, line 35: [The error in question](https://github.com/ccao-data/data-architecture/actions/runs/5905883640/job/16020814093):

```
Run dbt docs generate --target "$TARGET"
18:26:[23](https://github.com/ccao-data/data-architecture/actions/runs/5905883640/job/16020814093#step:6:24)  Running with dbt=1.5.5
18:[26](https://github.com/ccao-data/data-architecture/actions/runs/5905883640/job/16020814093#step:6:27):23  Encountered an error:
Runtime Error
  The profile 'athena' does not have a target named ''. The valid target names for this profile are:
   - dev
   - ci
   - prod
Error: Process completed with exit code 2.
```

Previously we had been hardcoding the target `"ci"` so I hadn't noticed that we were missing this step.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -81,7 +95,8 @@ jobs:
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
 
-      - name: Move dbt manifest directory to update cache
-        run: mv "$TARGET_DIR" "$STATE_DIR"
-        working-directory: ${{ env.PROJECT_DIR }}
-        shell: bash
```
jeancochrane, line 87: [The error in question](https://github.com/ccao-data/data-architecture/actions/runs/5906036003/job/16021252615):

```
Run mv "$TARGET_DIR" "$STATE_DIR"
mv: cannot move 'target' to 'state/target': Directory not empty
Error: Process completed with exit code 1.
```

Saving the cache explicitly with the `actions/cache/save` action resolves this problem.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -33,21 +33,35 @@ jobs:
       - name: Configure dbt environment
         uses: ./.github/actions/configure_dbt_environment
 
-      - name: Cache dbt state directory
+      # We have to use the separate `restore`/`save` actions instead of the
+      # unified `cache` action because only `restore` provides access to the
+      # `cache-matched-key` and `cache-primary-key` outputs as of v3
```
jeancochrane, line 38: This is a frustrating difference! Note [the outputs of the `restore` action](https://github.com/actions/cache/blob/f7ebb81a3f195b4fb88dab7c14e2f7aff52045aa/restore/action.yml#L26-L32) vs. [the outputs of the `cache` action](https://github.com/actions/cache/blob/f7ebb81a3f195b4fb88dab7c14e2f7aff52045aa/action.yml#L12).

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -33,21 +33,35 @@ jobs:
...
 
-      - if: ${{ steps.cache.outputs.cache-hit == 'true' }}
+      # If we restore the cache from the `restore-keys` key, the `cache-hit`
+      # output will be 'false' but the `cache-matched-key` output will be
+      # the name of the `restore-keys` key; we want to count this case as a hit
+      - if: |
+          steps.cache.outputs.cache-hit == 'true' ||
+          steps.cache.outputs.cache-matched-key == format(
+            '{0}-master', env.CACHE_NAME
+          )
```
jeancochrane, line 55: [This workflow run](https://github.com/ccao-data/data-architecture/actions/runs/5906337990) has three attempts where I tested each of the possible cases:

1. [Attempt #1](https://github.com/ccao-data/data-architecture/actions/runs/5906337990/attempts/1): Branch cache exists; exact cache key hit ➡️ this block evaluates True
2. [Attempt #2](https://github.com/ccao-data/data-architecture/actions/runs/5906337990/attempts/2): Deleted branch cache, but `master` branch cache exists; partial cache key hit ➡️ this block evaluates True
3. [Attempt #3](https://github.com/ccao-data/data-architecture/actions/runs/5906337990/attempts/3): Deleted both branch cache and `master` branch cache; no cache key hit ➡️ this block evaluates False

.github/workflows/deploy_dbt_docs.yaml
dfsnow: Nice, I should've caught this too. Tricky tricky

.github/workflows/build_and_test_dbt.yaml
dfsnow: **praise:** Sheesh, this is very confusing behavior. And weird that the actions maintainers don't provide access to the `cache-` vars in `actions/cache`. Nice work on the fast debugging!

dfsnow (approved):

Issue #75, opened 2023-08-17
[Data catalog] Implement source freshness tests for critical iasWorld tables


dfsnow:
Certain iasWorld tables are constantly being updated with new data. When running unit tests and reports, it is critical that these tables be as up-to-date as possible.

dbt builds in tests for [source freshness](https://docs.getdbt.com/docs/build/sources#checking-source-freshness) that we can use to check that the most recent possible data is loaded. Now that all iasWorld tables are sources (as of #71), we should use this functionality as an additional check (in addition to the current CloudWatch checks) to ensure that iasWorld data was successfully updated by [sqoop](https://github.com/ccao-data/service-sqoop-iasworld).

Specifically, we should target the following tables:

- HTPAR
- ASMT_ALL
- PERMIT

dbt source freshness requires a `loaded_at_field`. In our case, we should use the `wen` column. It can also use an optional `filter`, which in our case should constrain the check to the most recent tax year.

PR #84, opened 2023-08-21
Add new scheduled workflow to run source freshness checks every morning

jeancochrane:
This PR implements [source freshness checks](https://docs.getdbt.com/reference/commands/source#dbt-source-freshness) for our data catalog and initializes the checks with tests for three iasWorld tables (`htpar`, `asmt_all`, and `permit`). To run these checks, it adds a new CI workflow and schedules it to run at 9am CST every morning.

While adding the new workflow, I noticed that all of our dbt workflows currently share the same four setup steps. In the spirit of DRY, this PR also factors those steps out into a new `setup_dbt` composite action, and refactors the existing dbt workflows to use it.

Sample successful run [here](https://github.com/ccao-data/data-architecture/actions/runs/5931242660/job/16082679946?pr=84).

Closes https://github.com/ccao-data/data-architecture/issues/75.

### DIFF
--- a/.github/actions/setup_dbt/action.yaml
+++ b/.github/actions/setup_dbt/action.yaml
@@ -1,0 +1,23 @@
+name: Setup dbt
+description: Install dependencies for dbt and configure it for use in CI.
+inputs:
+  role-to-assume:
+    description: AWS IAM role to assume when running dbt operations.
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Install dbt requirements
+      uses: ./.github/actions/install_dbt_requirements
+
+    - name: Load environment variables
+      uses: ./.github/actions/load_environment_variables
+
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v2
+      with:
+        role-to-assume: ${{ inputs.role-to-assume }}
+        aws-region: us-east-1
+
+    - name: Configure dbt environment
+      uses: ./.github/actions/configure_dbt_environment
--- a/.github/workflows/test_dbt_source_freshness.yaml
+++ b/.github/workflows/test_dbt_source_freshness.yaml
@@ -1,0 +1,40 @@
+name: test-dbt-source-freshness
+
+on:
+  schedule:
+    # Every day at 9am CST (3pm UTC)
+    - cron: '0 15 * * *'
+
+jobs:
+  test-dbt-source-freshness:
+    runs-on: ubuntu-latest
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint
+    # so that we can authenticate with AWS
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install dbt requirements
+        uses: ./.github/actions/install_dbt_requirements
+
+      - name: Load environment variables
+        uses: ./.github/actions/load_environment_variables
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+          aws-region: us-east-1
+
+      - name: Configure dbt environment
+        uses: ./.github/actions/configure_dbt_environment
+
+      - name: Test source freshness
+        run: dbt source freshness --target "$TARGET"
+        working-directory: ${{ env.PROJECT_DIR }}
+        shell: bash
+
+
--- a/.github/workflows/test_dbt_source_freshness.yaml
+++ b/.github/workflows/test_dbt_source_freshness.yaml
@@ -4,6 +4,10 @@ on:
   schedule:
     # Every day at 9am CST (3pm UTC)
     - cron: '0 15 * * *'
+  # TODO: Remove this once testing is complete
+  pull_request:
+    branches: [master]
+
 
 jobs:
   test-dbt-source-freshness:
--- a/.github/workflows/test_dbt_source_freshness.yaml
+++ b/.github/workflows/test_dbt_source_freshness.yaml
@@ -17,24 +17,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Install dbt requirements
-        uses: ./.github/actions/install_dbt_requirements
-
-      - name: Load environment variables
-        uses: ./.github/actions/load_environment_variables
-
-      - name: Configure AWS credentials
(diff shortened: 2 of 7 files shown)


.github/actions/setup_dbt/action.yaml
```diff
@@ -0,0 +1,23 @@
+name: Setup dbt
+description: Install dependencies for dbt and configure it for use in CI.
+inputs:
+  role-to-assume:
+    description: AWS IAM role to assume when running dbt operations.
+    required: true
```
jeancochrane, line 6: From an interface perspective, it might make sense to also expose the optional input variables that the nested actions called by this action accept. I decided not to do that here since so far none of our scripts are overriding any of the defaults for those input variables, and I don't yet see a case where it would be useful for them to do so.

.github/actions/setup_dbt/action.yaml
```diff
@@ -0,0 +1,23 @@
...
+      uses: ./.github/actions/load_environment_variables
+
+    - name: Configure AWS credentials
+      uses: aws-actions/configure-aws-credentials@v2
+      with:
+        role-to-assume: ${{ inputs.role-to-assume }}
+        aws-region: us-east-1
+
+    - name: Configure dbt environment
+      uses: ./.github/actions/configure_dbt_environment
```
jeancochrane, line 23: As of this PR, `setup_dbt` is the only action or workflow that's calling these nested actions. Arguably, this is a sign that we should get rid of the nested actions and move their step definitions into this action, but it also seems plausible that it might be useful for an action/workflow to pick and choose among these nested actions in the future for its own purpose, so for now I decided to maintain the factored-out actions and nest them in here.

.github/actions/setup_dbt/action.yaml
jeancochrane: Also: I think you know this since you've already run into this hurdle before, but for extra clarity, the `role-to-assume` argument is exposed here since composite actions currently do not support reading from repository secrets.

.github/workflows/test_dbt_source_freshness.yaml
```diff
@@ -0,0 +1,28 @@
+name: test-dbt-source-freshness
```
jeancochrane, line 1: We had some discussion out of band about where/when would be the best time to run these checks. I'm starting with the simplest approach (scheduled once per day) but it should be easy to refactor this to e.g. allow sqoop to dispatch this workflow itself if we decide to move in that direction in the future.

dbt/models/iasworld/schema.yml
```diff
@@ -3,12 +3,17 @@ version: 2
...
     tables:
       - name: aasysjur
       - name: addn
       - name: addrindx
       - name: aprval
       - name: asmt_all
+        freshness:
+          filter: taxyr = date_format(current_date, '%Y')
+          warn_after: {count: 24, period: hour}
+          error_after: {count: 48, period: hour}
```
jeancochrane, line 16: This config means "warn if the data are 24 hours old, and error if the data are 48 hours old". These are somewhat arbitrary thresholds set based on a quick scan of the data, so I'm definitely open to suggested refinements.

dbt/models/iasworld/schema.yml
```diff
@@ -34,6 +43,10 @@ sources:
       - name: owndat
       - name: pardat
       - name: permit
+        freshness:
+          filter: date_format(date_parse(permdt, '%Y-%m-%d %H:%i:%s.0'), '%Y') = date_format(current_date, '%Y')
```
jeancochrane, line 47: I wasn't totally sure what the best field to use for this filter was, since `permit` doesn't have a `taxyr` column. I went ahead with `permdt` because so far it's consistently never null.

dbt/models/iasworld/schema.yml
```diff
@@ -34,6 +43,10 @@ sources:
...
+          warn_after: {count: 48, period: hour}
+          error_after: {count: 72, period: hour}
```
jeancochrane, line 49: `permdt` appears to be a day out of date, so I bumped the thresholds here assuming that's a normal feature of the data. We can adjust this as necessary if we decide that the underlying data would support a different set of thresholds.

.github/actions/setup_dbt/action.yaml
dfsnow: IMO this is simple enough while still preserving flexibility for the future. Good call.

dbt/models/iasworld/schema.yml
dfsnow: **thought:** These timers look good to me! `ASMT_ALL` is the most active table, and we should be getting sqoop extracts every day, so our period of downtime here shouldn't really ever be longer than 48H. I guess we're doing something a little bit unusual in looking at the "updated date" rather than the "loaded date". We currently don't capture the latter, but it might be something worth adding to sqoop in the future.

dbt/models/iasworld/schema.yml
```diff
@@ -26,6 +31,10 @@ sources:
       - name: htagnt
       - name: htdates
       - name: htpar
+        freshness:
+          filter: taxyr = date_format(current_date, '%Y')
```
dfsnow, line 35: **suggestion (blocking):** I _think_ we actually want all of these to be `taxyr = date_format(current_date, '%Y') - 1`. Since prop taxes are paid in arrears, very often the "current" or "working" year is actually the prior year. Note that:

- sqoop is setup to fetch `current_year - 2` every day
- Rolling over to a new calendar year doesn't necessarily mean rolling over to a new tax year in the system. So it's likely that these tests would start failing on Jan 1, 2024 since there are no records in the filter. This is fixed by using `current_year - 1`.

@wrridgeway Any thoughts here?

dbt/models/iasworld/schema.yml
dfsnow: I don't know enough about this table yet to add much. I think we'll just need to keep an eye on it and update it as we get more info.

dfsnow (commented):
Nice refactor @jeancochrane! We should use some of our time with Mirella to ask about these + other tests.

dbt/models/iasworld/schema.yml
jeancochrane: Makes sense to me! I'll go ahead and bump it back to `current_year - 1` while we discuss, it should be very easy to tweak if we end up deciding on a different threshold.

wrridgeway: Yes, current year, as much as i'd love to use it, would bunk things up come January. When the new year's roll is generated currently does not seem to be automated but rather a manual process. So my vote is `current_year -1` as well.

dfsnow (approved):

PR #88, opened 2023-08-22
70 create reporting view with most recent boundaries

wrridgeway:
need to replace hard-coded table refs with dbt syntax


aws-athena/views/reporting-vw_most_recent_boundary.sql
```diff
@@ -0,0 +1,168 @@
...
+        parcel.town_code AS township_code,
+        parcel.year AS parcel_year,
+        ST_POINT(parcel.x_3435, parcel.y_3435) AS geom
+    FROM {{ source('spatial', 'parcel') }} AS parcel
+    WHERE parcel.year
+        = (
+            SELECT MAX(parcel.year)
+            FROM {{ source('spatial', 'parcel') }} AS parcel
+        )
+        AND parcel.town_code IS NOT NULL
```
dfsnow, line 16: **question (non-blocking):** What's the purpose of this line?

dfsnow (approved):

aws-athena/views/reporting-vw_most_recent_boundary.sql
dfsnow: Resolved offline. In the parcel file, PINs with a null town code don't actually seem to exist.

Issue #89, opened 2023-08-22
Fix broken cache restoration


jeancochrane:
I think we borked dbt cache restoration in https://github.com/ccao-data/data-architecture/pull/82 when we switched to separate `restore`/`save` actions. [This run](https://github.com/ccao-data/data-architecture/actions/runs/5931242647/job/16082679882) is reporting a failure to restore either the master or the branch cache, when it fact both existed at the time of the workflow run according to our [cache list](https://github.com/ccao-data/data-architecture/actions/caches). My current theory is that the `path` argument needs to match between `restore` and `save` in order for the cache to restore properly, but we'll need to test that.

PR #90, opened 2023-08-22
Move dbt cache to remote S3 storage

jeancochrane:
This PR adjusts the way that we store our dbt state caches to get around the fact that GitHub Actions caches cannot be updated once they've been created. By moving our state caches to S3, we can ensure that the cache always gets updated at the end of every successful run of the `build-and-test-dbt` workflow, which will allow subsequent runs to only build and test resources that have been modified since the last successful run.

See [this run](https://github.com/ccao-data/data-architecture/actions/runs/5968306499/job/16194026297) for evidence of the remote cache being used successfully.

Closes https://github.com/ccao-data/data-architecture/issues/89.

### DIFF
--- a/.github/actions/restore_dbt_cache/action.yaml
+++ b/.github/actions/restore_dbt_cache/action.yaml
@@ -30,7 +30,6 @@ outputs:
 runs:
   using: composite
   steps:
-    # Check if there is an exact or fallback match for the cache key
     - name: Check for a cache key match
       id: query-cache-keys
       run: |
@@ -65,7 +64,6 @@ runs:
         RESTORE_KEY: ${{ inputs.restore-key }}
         BUCKET: ${{ inputs.bucket }}
 
-    # In case of a cache hit, copy the contents of the cache to the path
     - if: steps.query-cache-keys.outputs.cache-hit == 'true'
       name: Copy cache to path
       run: aws s3 cp "s3://$BUCKET/$KEY/manifest.json" "$CACHE_PATH/manifest.json"
--- a/.github/actions/restore_dbt_cache/action.yaml
+++ b/.github/actions/restore_dbt_cache/action.yaml
@@ -1,0 +1,73 @@
+name: Restore dbt cache
+description: Attempts to restore dbt cache from S3 storage.
+inputs:
+  path:
+    description: The local path to restore the cache to in case of a hit.
+    required: true
+  key:
+    description: The cache key to query for.
+    required: true
+  restore-key:
+    description: An additional key to use as a fallback for the cache.
+    required: true
+  bucket:
+    description: The S3 bucket that stores caches.
+    required: false
+    default: ccao-dbt-cache-us-east-1
+outputs:
+  cache-hit:
+    description: >-
+      Boolean indicating whether a match was found for the cache key.
+    value: ${{ steps.query-cache-keys.outputs.cache-hit }}
+  exact-match:
+    description: >-
+      Boolean indicating whether a cache hit was an exact match. Always
+      false if cache-hit is false.
+    value: ${{ steps.query-cache-keys.outputs.exact-match }}
+  cache-matched-key:
+    description: The cache key that matched, if any. Empty if cache-hit is false.
+    value: ${{ steps.query-cache-keys.outputs.cache-matched-key }}
+runs:
+  using: composite
+  steps:
+    # Check if there is an exact or fallback match for the cache key
+    - name: Check for a cache key match
+      id: query-cache-keys
+      run: |
+        if aws s3api head-object --bucket "$BUCKET" --key "$KEY/manifest.json"; then
+          echo "Cache hit: Found exact match"
+          {
+            echo "cache-hit=true";
+            echo "exact-match=true";
+            echo "cache-matched-key=$KEY"
+          } >> $GITHUB_OUTPUT
+        else
+          echo "Did not find exact match for cache key, checking fallback"
+          if aws s3api head-object --bucket "$BUCKET" --key "$RESTORE_KEY/manifest.json"; then
+            echo "Cache hit: Found fallback match"
+            {
+              echo "cache-hit=true";
+              echo "exact-match=false";
+              echo "cache-matched-key=$RESTORE_KEY"
+            } >> $GITHUB_OUTPUT
+          else
+            echo "Cache miss: Did not find fallback match for cache key"
+            {
+              echo "cache-hit=false";
+              echo "exact-match=false";
+              echo "cache-matched-key=''";
+            } >> $GITHUB_OUTPUT
+          fi
+        fi
+      env:
+        KEY: ${{ inputs.key }}
+        RESTORE_KEY: ${{ inputs.restore-key }}
+        BUCKET: ${{ inputs.bucket }}
+
+    # In case of a cache hit, copy the contents of the cache to the path
+    - if: steps.query-cache-keys.outputs.cache-hit == 'true'
+      name: Copy cache to path
+      run: aws s3 cp "s3://$BUCKET/$KEY/manifest.json" "$PATH/manifest.json"
+      env:
+        KEY: ${{ steps.query-cache-keys.outputs.cache-matched-key }}
+        PATH: ${{ inputs.path }}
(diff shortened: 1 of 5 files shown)


.github/actions/restore_dbt_cache/action.yaml
```diff
@@ -0,0 +1,76 @@
+name: Restore dbt cache
+description: Attempts to restore dbt cache from S3 storage.
+inputs:
+  path:
+    description: The local path to restore the cache to in case of a hit.
+    required: true
+  key:
+    description: The cache key to query for.
+    required: true
+  restore-key:
```
jeancochrane, line 10: The [GitHub Action](https://github.com/actions/cache) that this action is based on accepts `restore-keys` as a newline-delimited list, but for our purposes I only ever expect we'll need to use one fallback (for the master cache), so we gain some simplicity by switching to accepting a string value.

.github/actions/restore_dbt_cache/action.yaml
```diff
@@ -0,0 +1,76 @@
...
+    description: An additional key to use as a fallback for the cache.
+    required: true
+  bucket:
+    description: The S3 bucket that should store the cache.
+    required: false
+    default: ccao-dbt-cache-us-east-1
```
jeancochrane, line 16: I don't think it should be a problem to have this bucket name be public; the bucket itself is not public, and there shouldn't be anything sensitive in the dbt manifest file anyway.

.github/actions/restore_dbt_cache/action.yaml
```diff
@@ -0,0 +1,76 @@
...
+    description: The cache key that matched, if any. Empty if cache-hit is false.
+    value: ${{ steps.query-cache-keys.outputs.cache-matched-key }}
+runs:
+  using: composite
+  steps:
+    # Check if there is an exact or fallback match for the cache key
+    - name: Check for a cache key match
+      id: query-cache-keys
+      run: |
+        if aws s3api head-object --bucket "$BUCKET" --key "$KEY/manifest.json"; then
```
jeancochrane, line 37: Note that this simple `if ... then` conditional will fail on _any_ nonzero status from `aws s3api head-object`, not just the expected 404. I think this is acceptable, since we want to continue without the cache in case of an unexpected error (e.g. a 403 forbidden due to bad credentials), and [testing confirms](https://github.com/ccao-data/data-architecture/actions/runs/5965925106/job/16184439121#step:4:79) that the `aws s3api head-object` call will print the exit status to the console in these cases so we should be able to debug them.

```
An error occurred (403) when calling the HeadObject operation: Forbidden
```

.github/actions/restore_dbt_cache/action.yaml
```diff
@@ -0,0 +1,76 @@
...
+      shell: bash
+      env:
+        KEY: ${{ inputs.key }}
+        RESTORE_KEY: ${{ inputs.restore-key }}
+        BUCKET: ${{ inputs.bucket }}
+
+    # In case of a cache hit, copy the contents of the cache to the path
+    - if: steps.query-cache-keys.outputs.cache-hit == 'true'
+      name: Copy cache to path
+      run: aws s3 cp "s3://$BUCKET/$KEY/manifest.json" "$CACHE_PATH/manifest.json"
```
jeancochrane, line 71: In order to support listing, pushing, and pulling objects from this bucket, I adjusted policy for the IAM role that the GitHub Actions runner assumes to give it limited read/write access to the `ccao-dbt-cache-us-east-1` bucket.

.github/actions/save_dbt_cache/action.yaml
```diff
@@ -0,0 +1,23 @@
...
+    required: true
+  bucket:
+    description: The S3 bucket that should store the cache.
+    required: false
+    default: ccao-dbt-cache-us-east-1
+runs:
+  using: composite
+  steps:
+    - name: Save dbt cache
+      run: aws s3 cp "$CACHE_PATH/manifest.json" "s3://$BUCKET/$KEY/manifest.json"
```
jeancochrane, line 18: This remote storage naming mostly works well, but in pull requests where the branch name includes a slash, it results in some potentially confusing [nested folder behavior](https://github.com/ccao-data/data-architecture/actions/runs/5968306499/job/16194026297#step:10:59):

```
upload: dbt/target/manifest.json to s3://ccao-dbt-cache-us-east-1/dbt-cache-jeancochrane/89-fix-broken-cache-restoration/manifest.json
```

My guess is that this is probably fine, since the cache bucket will mostly be used by bots and it might even be nice to have higher-level organization (e.g. anything in the `dbt-cache-jeancochrane` subfolder likely corresponds to a cache for a branch that I authored). 

I also wonder if we want to add a [lifecycle configuration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/how-to-set-lifecycle-configuration-intro.html) to this bucket? It'll need a bit of research, since we'll want the `dbt-cache-master` subfolder to persist indefinitely while everything else can be deleted after a period of inactivity. Curious what you think about all this!

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -23,35 +23,20 @@ jobs:
...
 
-      # If we restore the cache from the `restore-keys` key, the `cache-hit`
-      # output will be 'false' but the `cache-matched-key` output will be
-      # the name of the `restore-keys` key; we want to count this case as a hit
-      - if: |
-          steps.cache.outputs.cache-hit == 'true' ||
-          steps.cache.outputs.cache-matched-key == format(
-            '{0}-master', env.CACHE_NAME
-          )
+      - if: steps.cache.outputs.cache-hit == 'true'
```
jeancochrane, line 34: One nice benefit of this change is that we can implement a definition of `cache-hit` that makes more sense to me (i.e. it's still a hit even if we restore from the fallback cache)!

.github/actions/restore_dbt_cache/action.yaml
dfsnow: **praise:** Nice, this is perfect. Super nice simplifications all around.

.github/actions/save_dbt_cache/action.yaml
dfsnow: IMO this is totally fine, since none of us are likely to use this bucket manually. 

**suggestion (non-blocking):** Let's definitely set some lifecycle rules though just to keep things tidy (likewise on the dbt Athena results buckets if there aren't any already 👀). IMO 60 days to mark all non-master keys for deletion with another 30 days to delete is sufficient. Let me know if you want to set it up or if I should grab it.

dfsnow (approved):
Nice work @jeancochrane! Love that this actually simplifies things in many ways and that we can now more easily fetch the manifest for local testing if needed. Merge at your leisure!

.github/actions/save_dbt_cache/action.yaml
jeancochrane: Cool! I had to adjust the cache keys to create a consistent prefix for PR caches that we could use in the lifecycle policy (`pr-caches/`), but I did that in 74031b6b476244103ff676916b803e603c885e26 and added a policy. Does this look right to you? The difference between expiration and deletion is a bit over my head, so perhaps it needs a little tweaking:

image
 
Luckly, dbt-athena creates the query results bucket with an existing lifecycle policy that deletes all query results after 3 days -- nice!

Screenshot 2023-08-25 135118

dfsnow: @jeancochrane That policy looks good to me! 

`ccao-athena-results-us-east-1` is the bucket I setup for general Athena output. I also setup that lifecycle rule. I was talking about `ccao-dbt-athena-ci-us-east-1` and `ccao-dbt-athena-test-us-east-1`, neither of which have lifecycle rules. IMO we probably should setup lifecycle rules for those as well.

jeancochrane: Oops, my bad @dfsnow, I totally misread the bucket name 🤦🏻‍♀️ I went ahead and copied that policy to `ccao-dbt-athena-ci-us-east-1` and `ccao-dbt-athena-test-us-east-1`!

PR #91, opened 2023-08-23
Remove prefix underscore from slugified schema names

dfsnow:
This PR updates our schema name generator so that schemas that start with numbers _are not_ prefixed with an underscore. The prefix underscore is designed to prevent invalid schemas (ones that start with numbers), but it is unnecessary for our use case since all dbt-generated schemas for non-main branches are prefixed with `dev_` or `ci_`.

Including the underscore prefix in the name generator results in schema names with repeated underscores when the feature branch is generated from an issue, i.e. `ci__70-create-reporting-view-with-most-recent-boundaries_census`.


jeancochrane (approved):
Makes sense to me!

PR #93, opened 2023-08-23
86 add uniqueness tests for iasworld tables

wagnerlmichael:
Largely copied from the `dbt/models/default/schema.yml`.

I added the `docs.md` because the `schema.yaml` has the description field with a call to `doc()`. Let me know if I should keep that or not.

Each of the tests passed.

### DIFF
--- a/dbt/models/iasworld/docs.md
+++ b/dbt/models/iasworld/docs.md
@@ -1,0 +1,3 @@
+{% docs addn %}
+test
+{% enddocs %}
--- a/dbt/models/iasworld/docs.md
+++ b/dbt/models/iasworld/docs.md
@@ -1,3 +1,39 @@
 {% docs addn %}
-test
+temp addn docs
+{% enddocs %}
+
+{% docs aprval %}
+temp aprval docs
+{% enddocs %}
+
+{% docs comdat %}
+temp comdat docs
+{% enddocs %}
+
+{% docs dweldat %}
+temp dweldat docs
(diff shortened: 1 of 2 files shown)


dbt/models/iasworld/docs.md
```diff
@@ -0,0 +1,39 @@
+{% docs addn %}
+temp addn docs
+{% enddocs %}
```
jeancochrane, line 3: **[Thought, non-blocking]** Per your note in the PR body:

> I added the docs.md because the schema.yaml has the description field with a call to doc(). Let me know if I should keep that or not.

It's not strictly necessary for us to add temp docs like this -- you'll notice if you navigate to [the docs site](https://ccao-data.github.io/data-architecture/#!/source_list/iasworld) that resources without a `description` attribute get assigned a default that says "This is not currently documented" -- but I think it's fine to leave these placeholders in, we can flesh them out when we get around to https://github.com/ccao-data/data-architecture/issues/26.

Screenshot 2023-08-24 100504

jeancochrane (approved):
Nice work jumping into a new codebase, this looks great! It looks like this branch is out of date with the main branch, but you should be able to pull in the latest changes and resolve any conflicts using the "Update branch" button just above the "Squash and merge" button on the main page of this PR. Once conflicts are resolved, feel free to squash and merge using that second button. Let me know if you're unsure how to do any of this and we can pair on it!

Issue #87, opened 2023-08-22
Add conditional uniqueness tests for iasWorld tables


jeancochrane:
A few of the iasWorld tests in our [data integrity checks sheet](https://cookcounty-my.sharepoint.com/:x:/g/personal/william_ridgeway_cookcountyil_gov/EfSAPa4UHGBDoxCsIpNjNX8BOiQ-YdAKvt4Bf9wf5sNrWg?email=Michael.Wagner2%40cookcountyil.gov&e=PX0SWE) rely on conditional expressions to define the subset of columns that should be unique, namely:

* `asmt_all`
* `sales`
* `htpar`

Update the `unique_combination_of_columns` macro to accept a new `where` parameter that accepts an optional SQL expression and defaults to `null`. When `where` is not null, the macro should add a `WHERE` clause to its query using the expression passed in by the caller. Use this new version of `unique_combination_of_columns` to implement the uniqueness tests for the columns that require conditional uniqueness.

PR #98, opened 2023-08-24
Add conditional uniqueness tests for iasWorld tables

jeancochrane:
This PR adds a new set of uniqueness tests for iasWorld tables, following up on #93. The tests in this PR are only unique for rows that meet a certain set of conditions, which means that we need to use the [`config.where` filter attribute](https://docs.getdbt.com/reference/resource-configs/where) to define the set of rows that should be unique.

Closes https://github.com/ccao-data/data-architecture/issues/87.

### DIFF
--- a/dbt/models/iasworld/schema.yml
+++ b/dbt/models/iasworld/schema.yml
@@ -40,9 +40,10 @@ sources:
               where: >-
                 cur = 'Y' and
                 deactivat is null and
-                procname in ('CCAOVALUE', 'CCAOFINAL', 'BORVALUE')
+                procname in ('CCAOVALUE', 'CCAOFINAL', 'BORVALUE') and
+                valclass is null
               config:
-                error_if: ">5991764"
+                error_if: ">125"
       - name: asmt_hist
       - name: cname
       - name: comdat
--- a/dbt/models/iasworld/schema.yml
+++ b/dbt/models/iasworld/schema.yml
@@ -25,6 +25,8 @@ sources:
                 cur = 'Y' and
                 deactivat is null and
                 procname in ('CCAOVALUE', 'CCAOFINAL', 'BORVALUE')
+              config:
+                error_if: ">5991764"
       - name: asmt_hist
       - name: cname
       - name: comdat
@@ -46,6 +48,17 @@ sources:
           filter: *latest_taxyr
           warn_after: *24_hours
           error_after: *48_hours
+        tests:
+          - unique_combination_of_columns:
+              name: htpar_unique_by_parid_caseno_taxyr_subkey
+              combination_of_columns:
+                - parid
+                - caseno
+                - taxyr
+                - subkey
+              where: cur = 'Y' and deactivat is null
+              config:
+                error_if: ">2"
       - name: land
       - name: legdat
       - name: lpmod
@@ -60,5 +73,12 @@ sources:
           error_after: &72_hours {count: 72, period: hour}
       - name: rcoby
       - name: sales
+        tests:
+          - unique_combination_of_columns:
+              name: sales_unique_by_parid_instruno
+              combination_of_columns:
+                - parid
+                - instruno
+              where: substr(saledt, 1, 4) >= '2023'
       - name: splcom
       - name: valclass
--- a/dbt/models/iasworld/schema.yml
+++ b/dbt/models/iasworld/schema.yml
@@ -12,8 +12,8 @@ sources:
       - name: asmt_all
         freshness:
           filter: &latest_taxyr taxyr >= date_format(current_date - interval '1' year, '%Y')
-          warn_after: {count: 24, period: hour}
-          error_after: {count: 48, period: hour}
+          warn_after: &24_hours {count: 24, period: hour}
+          error_after: &48_hours {count: 48, period: hour}
       - name: asmt_hist
       - name: cname
       - name: comdat
@@ -33,8 +33,8 @@ sources:
       - name: htpar
         freshness:
           filter: *latest_taxyr
-          warn_after: {count: 24, period: hour}
-          error_after: {count: 48, period: hour}
+          warn_after: *24_hours
+          error_after: *48_hours
       - name: land
       - name: legdat
       - name: lpmod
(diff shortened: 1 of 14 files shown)


dbt/tests/generic/test_unique_combination_of_columns.sql
```diff
@@ -10,7 +10,7 @@
 -- Adapted from dbt_utils.unique_combination_of_columns, and adjusted to add the
 -- optional duplicate threshold and to only report one row for each dupe.
 {% test unique_combination_of_columns(
-    model, combination_of_columns, allowed_duplicates=0
+    model, combination_of_columns, allowed_duplicates=0, where=null
```
jeancochrane, line 13: `where` is technically a reserved keyword in SQL, but it doesn't seem to cause problems when we use it as a jinja variable, and I like the simplicity of the interface that it allows for!

jeancochrane:
Oops, converting back to a draft since I think I found a syntax error!

jeancochrane:
Alright, we should be good now 👍🏻

dbt/tests/generic/test_unique_combination_of_columns.sql
```diff
@@ -1,22 +1,26 @@
...
 ) %}
 
     {%- set columns_csv = combination_of_columns | join(", ") %}
 
     select {{ columns_csv }}, count(*) as num_duplicates
     from {{ model }}
+    {% if where %} where ({{ where }}) {% endif %}
```
dfsnow, line 23: This line is really funny if you don't know what's going on.

**question (non-blocking):** Am I correct in my understanding that dbt/jinja interpret any non-null value as truthy here?

dbt/models/iasworld/schema.yml
```diff
@@ -109,21 +134,28 @@ sources:
...
           error_after: &72_hours {count: 72, period: hour}
       - name: rcoby
       - name: sales
+        tests:
+          - unique_combination_of_columns:
+              name: sales_unique_by_parid_instruno
+              combination_of_columns:
+                - parid
+                - instruno
+              where: substr(saledt, 1, 4) >= '2023'
```
dfsnow, line 159: **issue (non-blocking):** This conditional hopefully shouldn't be necessary to make the test pass (sales _should_ be unique by document number). If it is, we may need to borrow some of the more complex logic of the sales view, i.e:

```sql
WHERE sales.instruno IS NOT NULL
    AND sales.deactivat IS NULL
    AND tc.township_code IS NOT NULL
```

@wrridgeway Any thoughts here?

dbt/tests/generic/test_unique_combination_of_columns.sql
```diff
@@ -1,22 +1,26 @@
...
     group by {{ columns_csv }}
     having count(*) > {{ allowed_duplicates }} + 1
```
dfsnow, line 25: **thought:** We may run into lots of cases where this dupe test triggers due to many `null` values, while the actual values are unique. This conditional will be especially useful in those cases.

dbt/models/iasworld/schema.yml
```diff
@@ -30,6 +30,20 @@ sources:
...
+                - parid
+                - procname
+                - taxyr
+              where: >-
+                cur = 'Y' and
+                deactivat is null and
+                procname in ('CCAOVALUE', 'CCAOFINAL', 'BORVALUE') and
+                valclass is null
+              config:
+                error_if: ">125"
```
dfsnow, line 46: **thought (dreadful):** It's going to be so much fun to figure out what all these errors are...

dfsnow (approved):
Looks great @jeancochrane! This will be super useful very quickly as we start building out more complex tests.

**question (non-blocking):** My only question is how does this relate to the existing [where](https://docs.getdbt.com/reference/resource-configs/where) test macro. I assume we can't/shouldn't use that functionality in this case?

dbt/models/iasworld/schema.yml
jeancochrane: I don't fully understand the data context here, but I wanted to confirm that the test doesn't pass if you remove the `where` clause:

```
18:07:39  Failure in test sales_unique_by_parid_instruno (models/iasworld/schema.yml)
18:07:39    Got 2133988 results, configured to fail if != 0      
```

dbt/tests/generic/test_unique_combination_of_columns.sql
jeancochrane: That's true! I would be on board for changing it to `if where is not null`, but I ended up switching to using the builtin `config.where` filter in 8e72252edde3b2292df5d7d1d27f8bdf20efc89a so I think the question is moot.

jeancochrane (commented):
Great job finding the builtin `config.where` attribute @dfsnow, I'm kicking myself for missing that! Turns out it produces identical output, so I switched to using it in 8e72252edde3b2292df5d7d1d27f8bdf20efc89a and undid my changes to `unique_combination_of_columns`.

dbt/models/iasworld/schema.yml
jeancochrane: I'm still curious what @wrridgeway thinks, but I'm going to go ahead and merge in the meantime to unblock test development.

wrridgeway: for sales, yeah, we can just do unique by doc_no, but we have to include is_multisale = FALSE as well. we should get two rows that fail.

jeancochrane: @wrridgeway Does that mean we should remove this test and replace it with one like this?

```yaml
- name: sales
  tests:
      name: sales_unique_by_doc_no
      combination_of_columns:
        - doc_no
      config:
        where: is_multisale = false
        error_if: ">2
```

dfsnow: @jeancochrane That's correct. I'd do:

```yaml
- name: sales
  tests:
      name: sales_unique_by_doc_no
      combination_of_columns:
        - doc_no
      config:
        where: not is_multisale
        error_if: ">2
```

wrridgeway: this is true, but in that case, we do _not_ expect `iasworld.sales` to be unique by `doc_no`. the test i provided jean and which she implemented is the best i can do currently for uniqueness in that table. it _is_ encouraging that for 2023 things are unique by parid and instruno - that's a HUGE improvement. it's also worth noting that i don't think we should _ever_ expect the sales table to be unique only by instruno since multi-pin sales are legitimate.

Issue #96, opened 2023-08-24
[Data catalog] Test neighborhood codes in `iasworld.land`


jeancochrane:
Add a test to ensure that the neighborhood codes in the corresponding column in `iasworld.land` match the universe of available codes as recorded in `spatial.neigbhborhoods.town_nbhd`. If it's possible to do with dbt tests, it would be nice if we could actually join these two tables to ensure that they match; otherwise, we don't expect neighborhood codes to change, so it's good enough to compare them to a hardcoded list.

Blocked while we figure out exactly which `iasworld.land` column needs the test.

PR #103, opened 2023-08-29
Test neighborhood codes in `iasworld.pardat.nbhd`

jeancochrane:
This PR adds a new test to check that the values in `iasworld.pardat.nbhd` are a subset of the universe of neighborhood codes defined in `spatial.neighborhood.town_nbhd`. In order to support this test, it adds a new generic test macro, `column_is_subset_of_external_column`, that we can use for the general case where we want to check that a column is a subset of another column.

Closes https://github.com/ccao-data/data-architecture/issues/96.


dbt/models/iasworld/schema.yml
```diff
@@ -143,6 +143,19 @@ sources:
...
+              name: pardat_nbhd_matches_spatial_town_nbhd
+              column: nbhd
+              external_model: spatial.neighborhood
+              external_column: town_nbhd
+              additional_select_columns:
+                - parid
+                - taxyr
+              config:
+                # Codes ending in 999 are dummy codes used for some purpose,
+                # although we do not yet know what it is
```
jeancochrane, line 156: I thought about putting this in the docs for the `nbhd` column, but I figure I might as well leave that for our big documentation effort (#26) since I don't have much context on the column yet. But happy to move it if you feel strongly!

dbt/tests/generic/test_column_is_subset_of_external_column.sql
```diff
@@ -0,0 +1,28 @@
+-- Test that a given column is a subset of a column in an external relation.
```
jeancochrane, line 1: I'm not totally sure how useful this test macro will be yet, but since the new test would have required a custom query script anyway, I figured I might as well start with the general case and see if it becomes useful.

jeancochrane:
Tagging in @dfsnow, @wrridgeway, and @wagnerlmichael, all of whom have a stake in this test!

dbt/tests/generic/test_column_is_subset_of_external_column.sql
```diff
@@ -0,0 +1,28 @@
...
+    {%- set additional_select_columns_csv = additional_select_columns | join(", ") %}
+    {%- set columns_csv = additional_select_columns_csv ~ ", " ~ column %}
+
+    with
+        distinct_external_values as (
+            select distinct ({{ external_column }}) as external_column
+            from {{ external_model }}
+        )
+    select {{ columns_csv }}
+    from {{ model }}
```
dfsnow, line 23: **question (non-blocking):** Just to check my understanding, `{{ model }}` here is derived from the yaml `name: pardat` and subset by the `config:` where clause, is that correct?

dfsnow (approved):
@jeancochrane Looks great. This will be super useful going forward. I get the feeling we're going to need a lot of these more complex macros.

dbt/tests/generic/test_column_is_subset_of_external_column.sql
jeancochrane: Yes! When you use a `config.where` clause, dbt ends up substituting `{{ model }}` with a subquery like `(select * from {{ model }} where {{ where }}) as dbt_subquery`. Some gory details on this process and ways of customizing it are [here](https://docs.getdbt.com/reference/resource-configs/where#custom-logic). One side effect of this implementation is that you can't add an additional alias to a model if you use a `config.where` clause (i.e. `from {{ model }} as alias` throws an error) which is why we don't namespace the `{{ column }}` reference on line 25 even though I tend to prefer namespacing for explicitness wherever possible.

PR #110, opened 2023-08-31
Fix `default.vw_card_res_char` tests to better match the reality of the view

jeancochrane:
In Teams discussion with Nicole and Billy, I discovered two small problems with our `default.vw_card_res_char` tests:

1. `vw_card_res_char_unique_by_card_and_year` is incorrect, since this table is actually unique by card, year, and _PIN_

> [12:46 PM] Nicole Jardine (Assessor)
> I believe this view is unique to pin, year, and card

2. `vw_card_res_char_renovation_fields_match` is not actually useful, since it's just checking derived data, and the logic that generates these data is much more complicated than a simple test like "`char_renovation` should be the same value as `char_recent_renovation`"

> [1:40 PM] Jean Cochrane (Assessor)
>
> I'm wondering if the test checking whether `char_renovation` matches `char_recent_renovation` in `default.vw_card_res_char` needs some refinement... currently we have a naive check to make sure that these fields match, but I think in theory its plausible for them not to match, right? E.g. if a card was not renovated in a given year, but it was renovated in the prior two years, then we would have char_renovation=0 and char_recent_renovation=true; or vice versa, if a card was renovated in a given year, and it was also renovated in the prior three years, then we would have char_renovation=0 and char_recent_renovation=false. Do either of these seem like plausible scenarios? I don't totally get what `char_renovation` is indicating so I expect my assumptions may be wrong!

> [1:41 PM] William Ridgeway (Assessor)
> you're exactly right that they can have different values, perhaps it's not even worth implementing a test since all it would be doing is reiterating the logic that's already defined in the original view

This PR adjusts these two tests to better reflect these properties of the underlying data, and to fix tests that are currently failing ([example](https://github.com/ccao-data/data-architecture/actions/runs/6036357813)).

We also make one quick tweak to the command that runs dbt tests in the case of a cache hit, which allows us to properly run tests against upstream models.

### DIFF
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -62,7 +62,7 @@ jobs:
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             echo "Running tests on modified/new resources only"
-            dbt test -t "$TARGET" -s state:modified state:new --state "$STATE_DIR"
+            dbt test -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
           else
             echo "Running tests on all resources"
             dbt test -t "$TARGET"
--- a/dbt/models/default/schema.yml
+++ b/dbt/models/default/schema.yml
@@ -69,27 +69,11 @@ models:
     tests:
       # Unique by card and year
       - dbt_utils.unique_combination_of_columns:
-          name: vw_card_res_char_unique_by_card_and_year
+          name: vw_card_res_char_unique_by_pin_card_and_year
           combination_of_columns:
+            - pin
             - year
             - card
-          config:
-            error_if: ">913"
-      # char_recent_renovation correlates with char_renovation
-      - expression_is_true:
-          name: vw_card_res_char_renovation_fields_match
-          # char_renovation can be null, which we count here as falsey
-          expression: >-
-            char_recent_renovation = (
-              case when char_renovation = '1' then true else false end
-            )
-          select_columns:
-            - card
-            - year
-            - char_renovation
-            - char_recent_renovation
-          config:
-            error_if: ">73942"
       # TODO: Characteristics columns should adhere to pre-determined criteria
   - name: default.vw_pin_address
     description: '{{ doc("vw_pin_address") }}'


dbt/models/default/schema.yml
```diff
@@ -69,27 +69,11 @@ models:
...
       # Unique by card and year
       - dbt_utils.unique_combination_of_columns:
-          name: vw_card_res_char_unique_by_card_and_year
+          name: vw_card_res_char_unique_by_pin_card_and_year
           combination_of_columns:
+            - pin
             - year
             - card
-          config:
-            error_if: ">913"
```
jeancochrane, line 77: Adding PIN makes this test pass for all values, which is nice!

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             echo "Running tests on modified/new resources only"
-            dbt test -t "$TARGET" -s state:modified state:new --state "$STATE_DIR"
+            dbt test -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
```
jeancochrane, line 65: This fixes a small bug that I hadn't noticed before working on this PR. Prior to this change, the lack of a `--defer` flag meant that tests would fail when run against views or tables that were not built as part of the CI environment (see [here](https://github.com/ccao-data/data-architecture/actions/runs/6041375559/job/16394210971#step:9:1) for a sample failing workflow run). The lack of certain views or tables in a CI environment is intentional: We don't want to build models in CI unless their definition is new or has changed on the PR branch, which helps us speed up CI by only building and testing models that are being affected by the PR. The `--defer` tells dbt that if a model does not exist in the CI version of the DAG, but it _does_ exist in the prod version of the DAG, then it should refer to the prod version for the purpose of running builds and tests. (The exception to this is models that have changed as part of the PR, which are always built in the CI environment due to the `-s state:modified` flag.)

Detailed docs on dbt's deferral system are [here](https://docs.getdbt.com/reference/node-selection/defer).

dbt/models/default/schema.yml
wrridgeway: Awesome. Good thing to remember since I'm always less explicit/clear/helpful than I should be; card is essentially a sub-PIN, so we'll only ever work with card in tandem with PIN.

wrridgeway: Will these unit tests get tripped up as more rows are added to our tables? This is a distinct possibility due to what we call "divisions" - parcels getting split into more than one piece and new PINs being created. It happens on a pretty regular basis, unfortunately... we might want to choose a predefined time in the past for unit tests (2018, for a random example), if that's the case.

wrridgeway: Cool, just wanted to be sure.

wrridgeway (approved):

Issue #101, opened 2023-08-25
[Data catalog] Move CTAs into dbt DAG


jeancochrane:
Currently, the dbt DAG only manages our views (as `materialized: view` models) and defines the rest of our data as `source` models that are managed outside the DAG. For many of our data sources this will continue to be appropriate, but we would like to move the CTA queries as defined in [`aws-athena/ctas/`](https://github.com/ccao-data/data-architecture/tree/master/aws-athena/ctas) into the DAG as `materialized: table` models.

Implement this change so that we can manage our CTAs with dbt. Continue to follow the pattern established in https://github.com/ccao-data/data-architecture/pull/41 by symlinking from the `dbt/models/` directory to the `aws-athena/ctas/` directory so that we can maintain git history for these models.

PR #111, opened 2023-08-31
Refactor Athena table definitions to move them into the dbt DAG

jeancochrane:
# Overview

This PR adds symlinks and dbt model definitions for the Athena tables defined in the `aws-athena/ctas/` directory. It also tweaks those table definitions to fit them into the DAG and properly define their lineage.

## Additional complexities

Note that this change introduces two new areas of complexity to our dbt environments:

### 1. More resources to clean up

Since Athena `CREATE TABLE AS` statements store their output in S3, and since we expect to occasionally rebuild these tables as part of CI runs, we need to add an additional layer on top of the [`cleanup_dbt_resources.sh` script](https://github.com/ccao-data/data-architecture/blob/0424d980316a53a5403bd722ba749a05d99aaeb9/.github/scripts/cleanup_dbt_resources.sh#L1-L55) that is run on CI in order to clean up S3 resources that may be created on CI as part of the model building process. In order to handle this, I've added a new lifecycle rule to the `ccao-dbt-athena-ci-us-east-1` bucket to delete all old data with the `data/` prefix after 30 days, which should be more than enough time to close any PRs that may make use of those data.

Note that I did not add a corresponding lifecycle rule to the `ccao-dbt-athena-dev-us-east-1` bucket; my reasoning is that it's more likely that developers will want to persist their development table state over the long term.

### 2. More complex state selection for local development

These tables are more complicated to keep in the DAG than the views defined in `aws-athena/views/`, since the processing they perform is CPU intensive and they need to store their table output in S3. We shouldn't need to worry about this on CI since we use [deferral](https://docs.getdbt.com/reference/node-selection/defer) and [state selection](https://docs.getdbt.com/reference/node-selection/methods#the-state-method) to avoid rebuilding models unless they are new or have changed, but we will need an additional layer of safeguards for local development to ensure that developers who are less experienced with dbt don't accidentally waste time and resources rebuilding these tables unnecessarily.

To this end, we plan to introduce a wrapper script around dbt that beginning dbt developers can use to abstract away the complexity of remote state management.

Closes https://github.com/ccao-data/data-architecture/issues/101.

### DIFF
--- a/.github/actions/setup_dbt/action.yaml
+++ b/.github/actions/setup_dbt/action.yaml
@@ -4,6 +4,10 @@ inputs:
   role-to-assume:
     description: AWS IAM role to assume when running dbt operations.
     required: true
+  role-duration-seconds:
+    description: Expiration time for AWS OIDC token. Default is one hour.
+    required: false
+    default: 3600
 runs:
   using: composite
   steps:
@@ -18,6 +22,7 @@ runs:
       with:
         role-to-assume: ${{ inputs.role-to-assume }}
         aws-region: us-east-1
+        role-duration-seconds: ${{ inputs.role-duration-seconds }}
 
     - name: Configure dbt environment
       uses: ./.github/actions/configure_dbt_environment
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -22,6 +22,7 @@ jobs:
         uses: ./.github/actions/setup_dbt
         with:
           role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+          role-duration-seconds: 14400  # Worst-case time for full build
 
       - name: Restore dbt state cache
         id: cache
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -51,7 +51,7 @@ jobs:
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             echo "Running build on modified/new resources only"
-            dbt run -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
+            dbt run -t "$TARGET" --select location.tax --defer --state "$STATE_DIR"
           else
             echo "Running build on all resources"
             dbt run -t "$TARGET"
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -51,7 +51,7 @@ jobs:
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             echo "Running build on modified/new resources only"
-            dbt run -t "$TARGET" --select location.tax --defer --state "$STATE_DIR"
+            dbt run -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
           else
             echo "Running build on all resources"
             dbt run -t "$TARGET"
--- a/aws-athena/ctas/location-access.sql
+++ b/aws-athena/ctas/location-access.sql
@@ -1,27 +1,28 @@
-CREATE TABLE IF NOT EXISTS location.access
-WITH (
-    FORMAT = 'Parquet',
-    WRITE_COMPRESSION = 'SNAPPY',
-    EXTERNAL_LOCATION = 's3://ccao-athena-ctas-us-east-1/location/access',
-    PARTITIONED_BY = ARRAY['year'],
-    BUCKETED_BY = ARRAY['pin10'],
-    BUCKET_COUNT = 1
-) AS (
+{{
+    config(
+        materialized='table',
+        partitioned_by=['year'],
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH access AS (
     WITH distinct_pins AS (
         SELECT DISTINCT
             x_3435,
             y_3435
-        FROM spatial.parcel
+        FROM {{ source('spatial', 'parcel') }}
     ),
 
     distinct_years AS (
         SELECT DISTINCT year
-        FROM spatial.parcel
+        FROM {{ source('spatial', 'parcel') }}
     ),
 
     distinct_years_rhs AS (
         SELECT DISTINCT year
-        FROM spatial.walkability
+        FROM {{ source('spatial', 'walkability') }}
     ),
 
     walkability AS (
@@ -43,12 +44,12 @@ WITH (
                 SELECT
                     dy.year AS pin_year,
                     MAX(df.year) AS fill_year
-                FROM spatial.walkability AS df
+                FROM {{ source('spatial', 'walkability') }} AS df
                 CROSS JOIN distinct_years AS dy
                 WHERE dy.year >= df.year
                 GROUP BY dy.year
             ) AS fill_years
-            LEFT JOIN spatial.walkability AS fill_data
+            LEFT JOIN {{ source('spatial', 'walkability') }} AS fill_data
                 ON fill_years.fill_year = fill_data.year
         ) AS cprod
             ON ST_WITHIN(
@@ -65,10 +66,12 @@ WITH (
         walk.access_cmap_walk_total_score,
         walk.access_cmap_walk_data_year,
         pcl.year
-    FROM spatial.parcel AS pcl
+    FROM {{ source('spatial', 'parcel') }} AS pcl
     LEFT JOIN walkability AS walk
         ON pcl.x_3435 = walk.x_3435
         AND pcl.y_3435 = walk.y_3435
         AND pcl.year = walk.pin_year
     WHERE pcl.year >= (SELECT MIN(year) FROM distinct_years_rhs)
 )
+
+SELECT * FROM access
--- a/aws-athena/ctas/location-census_acs5.sql
+++ b/aws-athena/ctas/location-census_acs5.sql
@@ -2,42 +2,43 @@
 -- current PINs. For example, 2020 ACS data is not yet release, but we do have
 -- current (2021) geographies and PINs. We need to join 2019 geographies to
 -- 2021 PINs to facilitate joining 2019 ACS5 data
-CREATE TABLE IF NOT EXISTS location.census_acs5
-WITH (
-    FORMAT = 'Parquet',
-    WRITE_COMPRESSION = 'SNAPPY',
-    EXTERNAL_LOCATION = 's3://ccao-athena-ctas-us-east-1/location/census_acs5',
-    PARTITIONED_BY = ARRAY['year'],
-    BUCKETED_BY = ARRAY['pin10'],
-    BUCKET_COUNT = 1
-) AS (
+{{
+    config(
+        materialized='table',
+        partitioned_by=['year'],
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH census_acs5 AS (
     WITH distinct_pins AS (
         SELECT DISTINCT
             x_3435,
             y_3435
-        FROM spatial.parcel
+        FROM {{ source('spatial', 'parcel') }}
     ),
 
     distinct_years AS (
         SELECT DISTINCT year
-        FROM spatial.parcel
+        FROM {{ source('spatial', 'parcel') }}
     ),
 
     distinct_years_rhs AS (
         SELECT DISTINCT year
-        FROM spatial.census
+        FROM {{ source('spatial', 'census') }}
     ),
 
     acs5_max_year AS (
         SELECT MAX(year) AS max_year
-        FROM census.acs5
+        FROM {{ source('census', 'acs5') }}
     ),
 
     acs5_year_fill AS (
         SELECT
             dy.year AS pin_year,
             MAX(df.year) AS fill_year
-        FROM census.acs5 AS df
+        FROM {{ source('census', 'acs5') }} AS df
         CROSS JOIN distinct_years AS dy
         WHERE dy.year >= df.year
         GROUP BY dy.year
@@ -81,7 +82,7 @@ WITH (
         FROM distinct_pins AS dp
         LEFT JOIN (
             SELECT *
-            FROM spatial.census
+            FROM {{ source('spatial', 'census') }}
             WHERE year <= (SELECT max_year FROM acs5_max_year)
         ) AS cen
             ON ST_WITHIN(
@@ -105,7 +106,7 @@ WITH (
         ayf.census_acs5_tract_geoid,
         dj.year AS census_acs5_data_year,
         pcl.year
-    FROM spatial.parcel AS pcl
+    FROM {{ source('spatial', 'parcel') }} AS pcl
     LEFT JOIN acs5_year_fill AS ayf
         ON pcl.year = ayf.pin_year
     LEFT JOIN distinct_joined AS dj
@@ -114,3 +115,5 @@ WITH (
         AND pcl.y_3435 = dj.y_3435
     WHERE pcl.year >= (SELECT MIN(year) FROM distinct_years_rhs)
 )
+
+SELECT * FROM census_acs5
--- a/aws-athena/ctas/location-census_acs5.sql
+++ b/aws-athena/ctas/location-census_acs5.sql
@@ -94,16 +94,16 @@ WITH census_acs5 AS (
 
     SELECT
         pcl.pin10,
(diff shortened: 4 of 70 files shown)


.github/actions/setup_dbt/action.yaml
```diff
@@ -4,6 +4,10 @@ inputs:
   role-to-assume:
     description: AWS IAM role to assume when running dbt operations.
     required: true
+  role-duration-seconds:
+    description: Expiration time for AWS OIDC token. Default is one hour.
+    required: false
+    default: 3600
```
jeancochrane, line 10: Now that the maximum expected build time is [about four hours](https://github.com/ccao-data/data-architecture/actions/runs/6086810837) in the worst case where all resources need to be rebuilt, we have to bump the expiration time for the OIDC token, which requires passing this variable through to the underlying `configure-aws-credentials` action.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
         uses: ./.github/actions/setup_dbt
         with:
           role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+          role-duration-seconds: 14400  # Worst-case time for full build
```
jeancochrane, line 25: I debated storing this value as a repo secret, but decided against it since it doesn't feel to me like it makes the role any less secure. Still, if you disagree I'd be happy to go ahead and edit the history to remove this value and move it to a repo secret.

aws-athena/ctas/location-access.sql
```diff
@@ -1,27 +1,28 @@
-CREATE TABLE IF NOT EXISTS location.access
-WITH (
-    FORMAT = 'Parquet',
-    WRITE_COMPRESSION = 'SNAPPY',
-    EXTERNAL_LOCATION = 's3://ccao-athena-ctas-us-east-1/location/access',
```
jeancochrane, line 5: The `FORMAT`, `WRITE_COMPRESSION`, and `EXTERNAL_LOCATION` attributes do not have to be set in the model definition anymore; they are the same for all tables, so we configure them at the model level in `dbt_project.yml` below. On the other hand, we leave `PARTITIONED_BY`, `BUCKETED_BY`, and `BUCKET_COUNT` to be set by individual models, since these values are not identical across all models, or even within schemas.

aws-athena/ctas/location-access.sql
```diff
@@ -1,27 +1,28 @@
...
+{{
+    config(
+        materialized='table',
+        partitioned_by=['year'],
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH access AS (
```
jeancochrane, line 10: Although the addition of this subquery is not strictly necessary, it makes the diff easier to interpret, since otherwise you would have to sift through lots of added indents. I think that's maybe possible on GitHub if you select the "hide whitespace" option...? But I leave it up to you, let me know what you'd prefer!

aws-athena/ctas/location-census_acs5.sql
```diff
@@ -93,19 +94,19 @@ WITH (
...
+        dj.census_acs5_congressional_district_geoid,
+        dj.census_acs5_county_subdivision_geoid,
+        dj.census_acs5_place_geoid,
+        dj.census_acs5_puma_geoid,
+        dj.census_acs5_school_district_elementary_geoid,
+        dj.census_acs5_school_district_secondary_geoid,
+        dj.census_acs5_school_district_unified_geoid,
+        dj.census_acs5_state_representative_geoid,
+        dj.census_acs5_state_senate_geoid,
+        dj.census_acs5_tract_geoid,
```
jeancochrane, line 106: I'm pretty sure this was a mistake in the original table definition; as the query was written initially, Athena complained that no such attributes could be resolved from the `acs5_year_fill` subquery (aliased to `ayf` here). This fix seemed to correct to me, but I could be wrong!

aws-athena/ctas/proximity-dist_pin_to_pin.sql
```diff
@@ -99,38 +101,4 @@ WITH (
...
-        nearest_neighbor_2_pin10,
-        nearest_neighbor_2_dist_ft,
-        nearest_neighbor_3_pin10,
-        nearest_neighbor_3_dist_ft,
-        year
-    FROM proximity.dist_pin_to_pin_temp2
-);
-
-DROP TABLE IF EXISTS proximity.dist_pin_to_pin_temp
-DROP TABLE IF EXISTS proximity.dist_pin_to_pin_temp2
```
jeancochrane, line 136: It didn't seem like this temporary table approach was necessary anymore; I couldn't figure out what it was adding, especially because there didn't seem to be any definition for `proximity.dist_pin_to_pin_temp2` anymore. Does that sound right to you?

dbt/dbt_project.yml
```diff
@@ -28,6 +28,8 @@ clean-targets:
 models:
   athena:
     +materialized: view
+    +write_compression: SNAPPY
+    +format: PARQUET
```
jeancochrane, line 32: Here's where we set `write_compression` and `format` for all tables. The equivalent of `external_location` will be set in `profiles.yml` below.

dbt/models/default/schema.yml
```diff
@@ -115,7 +115,7 @@ models:
             - mail_address_zipcode_1
             - mail_address_zipcode_2
           config:
-            error_if: ">880615"
+            error_if: ">880634"
```
jeancochrane, line 118: More failures 😕

dbt/profiles.yml
```diff
@@ -3,8 +3,9 @@ athena:
   outputs:
     dev:
       type: athena
-      s3_staging_dir: s3://ccao-dbt-athena-test-us-east-1/results/
-      s3_data_dir: s3://ccao-dbt-athena-test-us-east-1/data/
+      s3_staging_dir: s3://ccao-dbt-athena-dev-us-east-1/results/
+      s3_data_dir: s3://ccao-dbt-athena-dev-us-east-1/data/
```
jeancochrane, line 7: I think `ccao-dbt-athena-dev` is clearer than `ccao-dbt-athena-test`, especially in contrast to the `ci` bucket defined below. If you agree, I'll go ahead and delete `ccao-dbt-athena-test` once I merge in this PR.

dbt/profiles.yml
```diff
@@ -3,8 +3,9 @@ athena:
...
+      s3_data_naming: schema_table
```
jeancochrane, line 8: See [the docs](https://dbt-athena.github.io/docs/configuration/table-configuration#table-data-location) for a detailed explanation of the S3 data naming options that `dbt-athena` exposes to us. `schema_table` corresponds to `{s3_data_dir}/{schema}/{table}/`, which matches how production data has historically been stored, so I figured we should move to it over the default `schema_table_unique`. This config renders the `external_location` attribute of the CTAS unnecessary.

aws-athena/ctas/proximity-dist_pin_to_pin.sql
dfsnow: Ah, a bit of complexity/stupidity here. This query running successfully is highly dependent on the size of the buffer defined here:

```sql
INNER JOIN pin_locations AS loc
    ON ST_CONTAINS(
        ST_BUFFER(ST_POINT(dp.x_3435, dp.y_3435), 1000), loc.point
    )
```

The smaller the buffer, the more quickly it runs, but the less likely it is to find distances for all PINs.

Basically, this query is saying "find all PIN distances within 1km of the target PIN centroid". BUT, not all PINs in Cook actually have neighbors within 1km. Expanding the buffer to 10km will cover all PINs, but makes the query take ages to run (and usually fail).

So, my (undocumented) solution was to run the query twice, first using a 1km buffer, then expanding the buffer to 10km _only for PINs which did not have 3 neighbors within the 1km buffer_.

**suggestion:** This might actually be the perfect use-case for dbt, as we could simply define that series of transforms in the DAG. We can create a 1km buffer materialized table, a 10km buffer table, then a simple view that combines them.

Alternatively, it's possible that Athena can now just run this query using the 10km buffer. When we switched to V3 query engine it did speed up some of our other queries, maybe this one will get a boost too. Regardless, I recommend we open a separate issue to deal with this.

dbt/profiles.yml
dfsnow: So, just to confirm, the bucket purposes are:

- `s3://ccao-dbt-athena-dev-us-east-1` - For local development. Materialized tables are written to `data/`, where they live in perpetuity.
- `s3://ccao-dbt-athena-ci-us-east-1` - For CI. Same directory structure but things in `data/` are deleted after 30 days
- `s3://ccao-dbt-athena-test-us-east-1` - Now deprecated. Will be deleted.

If so, that's sensible to me 👍

dbt/profiles.yml
dfsnow: Nice! This is a great simplification. Big fan.

**question (blocking):** What is the materialization strategy used here? As in, if someone is working on a branch and creates a new materialized table, then edits their branch and _re-_ materializes it, will the old table be entirely overwritten? Will it delete all the partitions created by the previous materialization?

dbt/profiles.yml
```diff
@@ -15,17 +16,16 @@ athena:
...
-      # Prefix all generated data by schema, so that we can delete it when the
-      # PR is merged
-      s3_data_naming: schema_table
       threads: 5
     prod:
       type: athena
       s3_staging_dir: s3://ccao-athena-results-us-east-1/
-      s3_data_dir: s3://ccao-athena-data-us-east-1/
+      s3_data_dir: s3://ccao-athena-ctas-us-east-1/
+      s3_data_naming: schema_table
```
dfsnow, line 28: **question (blocking):** Will merging this branch overwrite the existing production tables? If so, we may want to be cautious about a merge until the `dist_pin_to_pin` issue is resolved.

dfsnow (commented):
@jeancochrane See my blocking questions. Mainly concerned about overwriting the current state of the production data + complexity in the `dist_pin_to_pin` CTAS.

I also have another question unrelated to a specific line: How should we update/re-run these CTAS when the underlying source data is updated? Using a single example, we will soon gather new `spatial.park` data for 2024. When we want to update the state of `dist_pin_to_park` using the new data, do we just run dbt locally, targeting `prod`, and then upload the resulting state manifest back to the master cache location? I'm guessing this falls under the complexity mentioned in point 2 of the OP.

aws-athena/ctas/location-access.sql
jeancochrane: Good call, issue open here! https://github.com/ccao-data/data-architecture/issues/120

aws-athena/ctas/proximity-dist_pin_to_pin.sql
jeancochrane: Interesting, thanks for the insight @dfsnow! I also like your proposed solution of using a series of intermediate tables. Perhaps we open a follow-up story to implement that solution, and in the meantime remove this model from the DAG and continue treating it as a source for now?

dbt/profiles.yml
jeancochrane: Under `schema_table`, new materialized tables will indeed write over the old ones! My thinking was that this was desirable in order to save us on S3 space, but we could also use `schema_table_unique` in development if we want tables to be persisted between versions.

dbt/profiles.yml
jeancochrane: It will! See my note in [this thread above](https://github.com/ccao-data/data-architecture/pull/111/files#r1321844067) for a suggestion of how we can handle this.

aws-athena/ctas/proximity-dist_pin_to_pin.sql
dfsnow: @jeancochrane I like that fix. Let's go ahead and do it!

dbt/profiles.yml
dfsnow: @jeancochrane Just to call out some potentially unexpected behavior, some quick testing shows that dbt doesn't necessarily remove the entire old materialization:

Here we create a new test table partitioned by year and stored in S3 as parquet:

```sql
CREATE TABLE location.test_table
WITH (
    format = 'Parquet',
    write_compression = 'SNAPPY',
    external_location = 's3://ccao-dbt-athena-test-us-east-1/test_table',
    partitioned_by = ARRAY['year']
) AS
SELECT b23025_007m, year
FROM census.acs5
```

Dropping the table removes the table _metadata_ defining the location and partitions, but _not_ the data itself. All the data still exists in S3.

```sql
DROP TABLE location.test_table;
```

Re-instantiating the table using a subset of the original partitions correctly creates _new_ parquet files for each year after 2017.

```sql
CREATE TABLE location.test_table
WITH (
    format = 'Parquet',
    write_compression = 'SNAPPY',
    external_location = 's3://ccao-dbt-athena-test-us-east-1/test_table',
    partitioned_by = ARRAY['year']
) AS
SELECT b23025_007m, year
FROM census.acs5
WHERE year >= '2018'

SELECT year, COUNT(*) AS cnt
FROM location.test_table
GROUP BY year
ORDER BY year
```

row | year | cnt
-- | -- | --
1 | 2018 | 17521
2 | 2019 | 17521
3 | 2020 | 17619
4 | 2021 | 17619

HOWEVER, the original pre-2018 partitions/parquet files _remain in S3_, since they aren't overwritten by the new `CREATE TABLE` nor actually dropped by the `DROP TABLE`. This doesn't matter too much since those partitions aren't included in the Athena table definition, but we do use a lot of crawlers (which would definitely add those stale partitions back to the table).

I'm not sure this is actually a problem or that there's anything to actually do about it. I just felt the need to call it out since it makes me uneasy and is not how a typical relational DB would act.

jeancochrane: Hm, interesting! Thanks for the experiment @dfsnow, I'm going to play around a little bit to see if any of the other materialization options perform more closely to expectation here. I'm guessing you ran the `DROP TABLE` statement directly in Athena; did you build `location.test_table` using dbt or using an Athena query? If the former, do you happen to have copies of the dbt models that generated those parquet files?

dfsnow: I didn't build directly with dbt no, but happy to test further! Though I'd be really surprised if `dbt-athena` had some sort of logic to drop the extra partitions. Either way, let's not consider this blocking, just something to keep in mind.

jeancochrane:
>I also have another question unrelated to a specific line: How should we update/re-run these CTAS when the underlying source data is updated? Using a single example, we will soon gather new `spatial.park` data for 2024. When we want to update the state of `dist_pin_to_park` using the new data, do we just run dbt locally, targeting prod, and then upload the resulting state manifest back to the master cache location? I'm guessing this falls under the complexity mentioned in point 2 of the OP.

This is a great question @dfsnow! I think there are a few different ways we could approach this, each of which incurs a tradeoff between convenience and engineering effort:

| Approach | Convenience | Engineering effort |
| --- | --- | --- |
| Use source freshness and [the `source_status` selector](https://docs.getdbt.com/reference/node-selection/syntax#the-source_status-status) in a dispatchable GitHub Actions workflow to rebuild models whose sources have been updated | High | High |
| Write a dispatchable GitHub Actions workflow that accepts a list of models to rebuild as an input variable | Medium | Medium |
| Add docs to the README instructing devs to download remote state, figure out which models are downstream of the updated sources, rebuild downstream models using `--select`, and upload remote state | Low | Low |

The "medium" approach seems like the best option to me, since it's not substantially more effort than testing and documenting the commands locally and I could see it being useful outside of this particular use case. What do you think? If that sounds good to you I can open up an issue for it!

aws-athena/ctas/proximity-dist_pin_to_pin.sql
jeancochrane: Great, `proximity.dist_pin_to_pin` has been reverted to a source in f2afd7d4aa03e034abe203b5fdb1f7a7d0390f09 and I opened https://github.com/ccao-data/data-architecture/issues/122 to track the work of adding it to the DAG!

dfsnow:
> The "medium" approach seems like the best option to me, since it's not substantially more effort than testing and documenting the commands locally and I could see it being useful outside of this particular use case. What do you think? If that sounds good to you I can open up an issue for it!

@jeancochrane Let's do both the medium and low approaches, since as you said the medium approach isn't substantially more effort than the low, and I think it would be useful to have the commands documented somewhere.

dfsnow (approved):

jeancochrane:
Makes sense to me @dfsnow, I opened a follow-up issue here! https://github.com/ccao-data/data-architecture/issues/128

PR #112, opened 2023-08-31
Add dbt wrapper script to facilitate local development

jeancochrane:
This PR adds a wrapper script around `dbt` that developers can use to leverage [deferral](https://docs.getdbt.com/reference/node-selection/defer) and [state selection techniques](https://docs.getdbt.com/reference/node-selection/methods#the-state-method) without having to understand how they work in detail.

Connects https://github.com/ccao-data/data-architecture/pull/111.

### DIFF
--- a/dbt/.gitignore
+++ b/dbt/.gitignore
@@ -5,3 +5,4 @@ target/
 dbt_modules/
 dbt_packages/
 assets/*.svg
+.remote_state/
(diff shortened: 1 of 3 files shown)


jeancochrane:
Closing this for now, but leaving the branch open; this would be nice to do, but I want to gather more data on how our devs use dbt locally before moving forward with dev tooling.

PR #116, opened 2023-09-05
114 add condo units with negative predicted values to declared livable/non livable spaces tables

wrridgeway:
@carolinedavidson discovered some condo units with negative predicted values in her condo report. We believe these units are likely non-livable and should be treated as such during modeling, reporting, etc.

This PR adds this data to ingest scripts (`ccao-condominium_parking.R`), generates an Athena table for the data, and adds it to the Athena condo characteristics view (`default.vw_pin_condo_char`).

### DIFF
--- a/aws-athena/views/default-vw_pin_condo_char.sql
+++ b/aws-athena/views/default-vw_pin_condo_char.sql
@@ -339,7 +339,7 @@ SELECT DISTINCT
                 AND prior_values.oneyr_pri_board_tot BETWEEN 10 AND 5000
             )
             OR prior_values.oneyr_pri_board_tot BETWEEN 10 AND 1000
-            OR nonlivable.flag = "negative pred"
+            OR nonlivable.flag = 'negative pred'
         )
         AND nonlivable.flag != 'questionable'
             THEN 1
--- a/aws-athena/views/default-vw_pin_condo_char.sql
+++ b/aws-athena/views/default-vw_pin_condo_char.sql
@@ -198,7 +198,9 @@ chars AS (
         condo units received (unused) negative predicted values. These units
         were non-livable units incorrectly classified as livable. They received
         negative predicted values due to their very low % of ownership. This CTE
-        excludes them from the model going forward. */
+        excludes them from the model going forward. 3) Questionable garage units
+        are those that have been deemed nonlivable by some part of our
+        nonlivable detection, but upon human review have been deemed livable. */
         LEFT JOIN {{ source('ccao', 'pin_nonlivable') }} AS nonlivable
             ON par.parid = nonlivable.pin
     )
--- a/aws-athena/views/default-vw_pin_condo_char.sql
+++ b/aws-athena/views/default-vw_pin_condo_char.sql
@@ -27,10 +27,11 @@ questionable_gr AS (
     FROM {{ source('ccao', 'pin_questionable_garage_units') }}
 ),
 
--- In the process of QC'ing condo data, we discovered that some condo units
--- received (unused) negative predicted values. These units were non-livable units
--- incorrectly classified as livable. They received negative predicted values due to 
--- their very low % of ownership. This CTE excludes them from the model going forward.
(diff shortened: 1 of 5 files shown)


wrridgeway:
@wagnerlmichael once this PR is merged, we'll need to change your sql syntax where it interacts with `is_question_garage_unit`. It will work how you initially assumed it should (i.e. `NULL` values will become `FALSE`).

aws-athena/views/default-vw_pin_condo_char.sql
```diff
@@ -27,6 +27,14 @@ questionable_gr AS (
     FROM {{ source('ccao', 'pin_questionable_garage_units') }}
 ),
 
+-- Some condo PINs have been detected as non-livable through modeling
```
dfsnow, line 30: **suggestion:** Can we add a little more context here i.e:

```suggestion
-- In the process of QC'ing condo data, we discovered that some condo units
-- received (unused) negative predicted values. These units were non-livable units
-- incorrectly classified as livable. They received negative predicted values due to 
-- their very low % of ownership. This CTE excludes them from the model going forward.
```

aws-s3/scripts-ccao-data-raw-us-east-1/ccao-condominium_parking.R
```diff
@@ -16,36 +13,37 @@ output_bucket <- file.path(AWS_S3_RAW_BUCKET, "ccao", "condominium")
...
 ##### QUESTIONABLE GARAGE UNITS #####
 
-# Get local file addresses
-source_paths <- c("O:/Condo Worksheets/Condo 2022/")
-
-source_files <- list.files(
-  source_paths,
-  recursive = TRUE, full.names = TRUE
+openxlsx::read.xlsx(
+  "O:/Condo Worksheets/Condo 2022/2022 Desk Review - Questionable Garage Units.xlsx", # styler: off
```
dfsnow, line 17: **suggestion:** We really shouldn't be ingesting much from the network drives. Can we move these files to the raw data S3 bucket?

dfsnow (changes requested):

aws-athena/views/default-vw_pin_condo_char.sql
```diff
@@ -341,6 +349,7 @@ SELECT DISTINCT
                 AND prior_values.oneyr_pri_board_tot BETWEEN 10 AND 5000
             )
             OR prior_values.oneyr_pri_board_tot BETWEEN 10 AND 1000
+            OR negative_preds.is_negative_pred = TRUE
```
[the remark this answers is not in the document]
wrridgeway: personally, i think `= TRUE ` is more explicit and doesn't detract from anything?

dfsnow (changes requested):
@wrridgeway Let's generalize this change a bit more to cover any future flags of non-livable condo units uncovered by the Data Dept. I would call the table something like `pin_non_livable_flags_data_team` or similar. Add an extra column that includes the reason for the flag (in this case, negative predicted values).

wrridgeway:
@jeancochrane I'm kind of at a loss here as to why DBT has started throwing an error that it wasn't throwing before... any clue what I'm doing wrong?

jeancochrane:
@wrridgeway Hmm, that's weird; seems like something may be off with the deferral to the production state. I'll investigate, but in the meantime you should be able to fix it by clearing the cache (i.e. deleting the object `s3://ccao-dbt-cache-us-east-1/pr-caches/114-add-condo-units-with-negative-predicted-values-to-declared-livablenon-livable-spaces-tables/manifest.json` in S3) and rerunning the build.

wrridgeway:
Thanks @jeancochrane worked like a charm!

wrridgeway:
Once this is merged, we'll need to delete `ccao.pin_399_garage_units`, `ccao.pin_negative_predicted_value`, and `ccao.pin_questionable_garage_units` from s3/athena

dfsnow (approved):

PR #118, opened 2023-09-08
Add rich descriptions to column definitions in dbt

jeancochrane:
This PR updates our dbt model definitions in `models/*/schema.yml` to add rich descriptions taken from the `var_notes` column of the [`ccao` vars dict](https://github.com/ccao-data/ccao/blob/master/data-raw/vars_dict.csv).

There are two different ways in which this PR adds rich descriptions:

1. Descriptions that are pithy and that summarize the resource are stored in the `<model>.columns.<column>.description` attribute, which shows up in the `Description` column of the `Columns` table in the docs ([example here](https://ccao-data.github.io/data-architecture/#!/model/model.athena.default.vw_pin_address#columns))
    * Motivation: Provide a scannable summary of each column for docs users as they browse the `Columns` table
2. Descriptions that are complex, or that document assumptions or lineage of a column, are stored in the `<model>.columns.<column>.meta.notes` attribute, which shows up in the `More?` column of the `Columns` docs table and only gets displayed when a row is clicked
    * Motivation: Provide more detail for docs users who are diving deep on a particular column

The changes have been deployed to https://ccao-data.github.io/data-architecture for the purposes of testing.

Connects https://github.com/ccao-data/ccao/issues/2.

<details>
<summary>Click for details on how these docs were populated</summary>
<br/>

The process of populating these docs was largely scripted, but those scripts aren't being kept under version control since their usefulness is tightly coupled to this particular one-time task. Nevertheless, I think it's important to describe how I did it to ensure that my assumptions about the source data are correct:

1. I used [csvkit](https://csvkit.readthedocs.io/en/latest/) to filter the [vars dict](https://github.com/ccao-data/ccao/blob/master/data-raw/vars_dict.csv) and produce an output CSV `column_descriptions.csv` containing:
    * The columns `vars_from_view`, `var_name_athena`, and `var_notes`;
    * For all rows where `vars_from_view` and `var_notes` were not empty;
    * With a minor manual edit to [row 469](https://github.com/ccao-data/ccao/blob/master/data-raw/vars_dict.csv#L469) to split `default.vw_card_res_char` and `default.vw_pin_condo_char` onto two separate lines
2. I then wrote a quick Python script `populate_descriptions.py` (committed in 8965668b8804038ef1c07d879eda4409cac21451 but then reverted after running) to:
    * Iterate each row of `column_descriptions.csv`
    * Open up the `schema.yml` file corresponding to the variable's schema
    * Search and replace the `- name: <model_name>` line to add a block containing a `columns.<column>.description` key
    * Write the `schema.yml` file back to disk
3. Finally, I went through the `schema.yml` files manually to consolidate duplicate `columns:` definitions and move descriptions to the `meta.notes` field where appropriate

I initially tried an alternate approach for the script in step 2 where I read and wrote `schema.yml` using pyyaml, which allowed for insertion of column definitions without duplication; this didn't work, however, because `yaml.dump()` was not smart enough to preserve the original formatting of `schema.yml`, and so would collapse file formatting and generate a horrendous diff. Given this behavior, a little bit of manual work to clean up duplication seemed fine, especially because manual work was required to distinguish between description and notes fields anyway.

</details>


dbt/models/census/schema.yml
```diff
@@ -12,3 +12,135 @@ sources:
...
+      - name: percent_income_household_received_snap_past_year
+        description: Percent of households that received SNAP in the past 12 months
+        meta:
+          notes: ACS variable B22003_002E / B22003_001E
+      - name: percent_income_below_poverty_level
+        description: Percent of people below poverty level
+        meta:
+          notes: ACS variable B17001_002E / B17001_001E
+      - name: percent_household_total_occupied_w_sel_cond
+        description: "Selected conditions, including: incomplete plumbing or kitchens, overcrowding, 30% or more of the household income spent on rent or monthly owner costs"
```
jeancochrane, line 58: I wasn't entirely sure whether this description should go in the top-level `description` field or the more hidden `meta.notes` field for this and the variables that follow. What do you think?

dbt/models/default/schema.yml
```diff
@@ -122,13 +258,29 @@ models:
   - name: default.vw_pin_land
     description: '{{ doc("vw_pin_land") }}'
   - name: default.vw_pin_sale
-    description: '{{ doc("vw_pin_sale") }}'
```
jeancochrane, line 125: This isn't actually getting deleted, it's just being moved below the `columns` definition.

dbt/models/census/schema.yml
```diff
@@ -12,3 +12,135 @@ sources:
...
+      - name: percent_race_asian
+        meta:
+          notes: Percentages will NOT sum to 1 since the Census allows more than 1 race per respondent
+      - name: percent_race_aian
+        meta:
+          notes: Percentages will NOT sum to 1 since the Census allows more than 1 race per respondent
+      - name: percent_mobility_no_move
+        description: Percent of people (older than 1 year) who have not moved in the past 12 months
+        meta:
+          notes: ACS variable B07003_004E / B07003_001E
```
jeancochrane, line 40: One gripe that I have about this organizational scheme (summaries in `description`, details in `meta.notes`) is that when you click on a row in a dbt `Columns` table, the description is actually rendered _after_ the notes:

Screenshot 2023-09-08 at 3 32 29 PM

I'm not sure if there's anything we can do about this (e.g. swap the pattern so the summaries are in `meta.notes`) since the description is shown in truncated form in the table overview, and we probably want to maintain that behavior for the summaries. But I want to call it out as something I'm aware of and ambivalent about!

dbt/models/default/schema.yml
```diff
@@ -151,6 +303,270 @@ models:
...
+        meta:
+          notes: For CPS, secondary school attendance boundaries are used. Taken from Cook County GIS shapefiles
+      - name: school_elementary_district_geoid
+        description: Elementary school district ID for a given PIN
+        meta:
+          notes: For CPS, elementary school attendance boundaries are used. Taken from Cook County GIS shapefiles
+      - name: prop_address_zipcode_2
+        description: Property street address
+        meta:
+         notes: Not the address of the taxpayer/property owner
```
dfsnow, line 342: **issue (blocking):** The single-spacing of those `notes` field starting here is causing all columns below to parse incorrectly in the docs.

```suggestion
      - name: prop_address_zipcode_2
        description: Property street address
        meta:
          notes: Not the address of the taxpayer/property owner
```

dbt/models/default/schema.yml
```diff
@@ -151,6 +303,270 @@ models:
...
+        meta:
+         notes: First 2 digits are township code, last 3 digits are neighborhood code
+      - name: misc_unincorporated_area_bool
+        description: Indicator for a PIN within an unincorporated part of Cook County
+        meta:
+         notes: Taken from Cook County GIS shapefiles
+      - name: misc_subdivision_id
+        description: Subdivision ID for a given PIN
+        meta:
+          notes: Nearly all PINs are within one subdivision. Each subdivision has a unique ID with the PageSubref column. Page being the area and section and subref being the letter of that sub on a page. The subrefs go A B C D and so on, however new subs can kill old subs, so a page could have subrefs B C G M L R S T, with each letter being a unique subdivision
```
dfsnow, line 480: Good example of a super long and ugly line in the docs. I couldn't get it to parse with line breaks no matter what I tried. 

**question:** If we aren't going to parse these programmatically (i.e. to build a separate data dictionary), maybe it's worth just moving everything into the `description` field, separating the summary like this:

```suggestion
      - name: misc_subdivision_id
        description: | 
        Subdivision ID for a given PIN
        
        Nearly all PINs are within one subdivision. Each subdivision has a unique ID with the PageSubref column. Page being the area and section and subref being the letter of that sub on a page. The subrefs go A B C D and so on, however new subs can kill old subs, so a page could have subrefs B C G M L R S T, with each letter being a unique subdivision
```

Or just with a period:

```suggestion
      - name: misc_subdivision_id
        description: | 
        Subdivision ID for a given PIN. Nearly all PINs are within one subdivision. Each subdivision has a unique ID with the PageSubref column. Page being the area and section and subref being the letter of that sub on a page. The subrefs go A B C D and so on, however new subs can kill old subs, so a page could have subrefs B C G M L R S T, with each letter being a unique subdivision
```

@jeancochrane What do you think?

dfsnow (changes requested):

dbt/models/default/schema.yml
jeancochrane: Oops, good catch! I fixed this in 4026e44ef0186a8f1e75639f108f6f95cac22edd and also did a quick grep to confirm there are no other instances of it.

dbt/models/default/schema.yml
```diff
@@ -151,6 +303,270 @@ models:
...
+        description: Airport noise calculated via kriging noise monitor data from CDA
+        meta:
+          notes: See https://gitlab.com/ccao-data-science---modeling/models/ccao_res_avm/-/issues/70 for more information
+      - name: cook_municipality_name
+        description: Municipality name for a given PIN
+        meta:
+          notes: Taken from Cook County GIS shapefiles
+      - name: census_acs5_tract_geoid
+        meta:
+          notes:  Census ACS5 columns provide the geographic IDs (GEOIDs) of Census geography with the latest ACS5 data available. For example, only 2019 ACS5 data is available in 2022, so the ACS5 tract GEOID is the 2019 Census geography joined to the 2022 PIN
```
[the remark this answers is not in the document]
jeancochrane: Fixed in f9af2fc2d5ea25b93cb35203cb735b121a01959f along with a grep to confirm there are no more leading spaces in `meta.notes` fields.

dbt/models/default/schema.yml
jeancochrane: This is definitely a bummer @dfsnow! I played around with it a little bit too, and it seems like whatever Angular code they're using to generate the docs site is creating the `Columns` table summary by just concatenating everything from the `description` field into a single string. It's nice that the multiline formatting makes the detail view look better once you click on a row, but I wish the summary view would make use of that formatting too.

After trying a few things, I think my favorite is actually a combination of the two approaches you suggested -- period plus newline:

```suggestion
      - name: misc_subdivision_id
        description: |
        Subdivision ID for a given PIN.
        
        Nearly all PINs are within one subdivision. Each subdivision has a unique ID with the PageSubref column. Page being the area and section and subref being the letter of that sub on a page. The subrefs go A B C D and so on, however new subs can kill old subs, so a page could have subrefs B C G M L R S T, with each letter being a unique subdivision.
```

To me, this formatting gives the best experience for both the summary and detail views, since the period makes the summary easier to read while the newline makes the details easier to read:

Screenshot 2023-09-12 112509
Screenshot 2023-09-12 112606

(I don't love the blank `Details` line plus horizontal scrollbar in the details view, but based on the raw HTML I think that's an artifact generated by the fact that other rows have a `meta.notes` attribute while `misc_subdivision_id` does not, and I expect it to go away if and when we strip out `meta.notes` from the other rows.)

Does this sound reasonable to you? If so, I'm curious whether you think we should make this change now, or put it on the backburner until you flesh out the rest of the docs? It doesn't seem like it would make sense to make this change for fields that currently have a `meta.notes` attribute but no `description`, but I could be convinced otherwise.

dfsnow: @jeancochrane This totally makes sense to me. Adding the period also makes the descriptions machine-parseable in case we ever need to split them between summary and details. IMO, let's just move everything into the description field, with line breaks for fields that have `meta.notes` only. I'll populate the summary line when I flesh out the docs.

jeancochrane: Sounds good @dfsnow, done in [0c6c213](https://github.com/ccao-data/data-architecture/pull/118/commits/0c6c213fa90cb9bc5b42162ce3fd3f3bc9cf4fbf)!

dfsnow (approved):
Looks great! Nice work @jeancochrane

PR #119, opened 2023-09-08
Damon refactor ratio stats mki

Damonamajor:
This adds MKI into the existing ratio stats. It also deletes to code for the base functions as well as the _met and _ci functions to directly access assesspy.

- things like cod_boot are included in the assesspy package, but can't be accessed from the package to use outside of it. 
- The outputs are still set to the test folder. It needs to be shifted back to the original folder. 
- I've been having issues removing the ap. before the functions since it also renames the column names. Because of that they are imported separately. I have a structure in place and am working through it, but it's just time consuming running checks on it in AWS. I'll send a separate push if this is necessary to fix.


aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -12,80 +34,69 @@
...
-s3_ratio_stats = 's3://'+ s3_bucket + '/' + s3_prefix + 'ratio_stats.parquet'
+athena_db = "iasworld"
+
+s3_bucket = "ccao-data-warehouse-us-east-1"
+s3_prefix = "reporting/ratio_stats/"
+s3_output = "s3://" + s3_bucket + "/" + s3_prefix
+s3_ratio_stats = "s3://" + s3_bucket + "/" + s3_prefix + "ratio_stats.parquet"
+s3_ratio_stats_test = (
+    "s3://" + s3_bucket + "/" + "reporting/ratio_stats_test/" + "ratio_stats.parquet"
+)
```
jeancochrane, line 53: **[Nitpick]** In the PR body, you mentioned that the outputs should be reset to the prod folder:

> The outputs are still set to the test folder. It needs to be shifted back to the original folder.

I agree with this, but I think that now that we're proposing an official change to the script via this pull request, there's no need for the test bucket anymore, so we might as well return to using `s3_ratio_stats` now:

```suggestion
```

If we need to test any further changes, we can always switch it back!

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -110,329 +121,314 @@ def run_query_get_result(
...
-    report_summarise(pull, 'township_code', 'Town')
-    ]).to_parquet(
-        s3_ratio_stats
-        )
+pd.concat(
+    [
+        report_summarise(pull, "triad", "Tri"),
+        report_summarise(pull, "township_code", "Town"),
+    ]
+).to_parquet(s3_ratio_stats_test)
```
jeancochrane, line 431: **[Nitpick]** Same as above, let's change this back to the production bucket:

```suggestion
).to_parquet(s3_ratio_stats)
```

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -1,4 +1,26 @@
...
+from assesspy import prb as prb
+from assesspy import cod as cod
+from assesspy import mki_met as mki_met
+from assesspy import mki as mki
+from assesspy import detect_chasing as detect_chasing
+from assesspy import prb_met as prb_met
+from assesspy import prd_met as prd_met
+from assesspy import cod_met as cod_met
+from assesspy import prd as prd
+import assesspy as ap
```
jeancochrane, line 12: **[Nitpick]** A couple notes here about imports:

1. The `as` keyword is only necessary if we want to create an alias for the package we're importing (e.g. `import assesspy as ap`). Since in these cases we're not changing the name of the packages we're importing, we don't need to include aliases.
2. If we want to import multiple symbols from one package, we can consolidate them into one `import` statement by separating the symbols with commas, e.g.:

```python
from assesspy import cod_ci, boot_ci, prb, ...
```

3. The symbol imports on lines 1-11 are duplicative with the `import assesspy as ap` call on line 12, and I can't tell if the import on line 12 is being used anywhere. We should either import these symbols from the package, or import the package directly, but not both.

Putting this all together, my suggestion would be:

```suggestion
from assesspy import (
    boot_ci,
    cod,
    cod_ci,
    cod_met,
    detect_chasing,
    mki,
    mki_met,
    prb,
    prb_met,
    prd
    prd_met,
)
```

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -12,80 +34,69 @@
 job = Job(glueContext)
 
 # Import necessary libraries
```
jeancochrane, line 36: **[Nitpick, non-blocking]** Looks like this comment is no longer in the right place? But I also think it was never really necessary to begin with, we should expect the reader to know what an `import` is for.

```suggestion
```

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -110,329 +121,314 @@ def run_query_get_result(
...
+    ci = ", ".join([str(element) for element in ci])
 
     return ci
 
-# Formula specific bootstrapping functions
-def cod_boot(ratio, nboot = 100, alpha = 0.05):
 
-    return boot_ci(cod, ratio, nboot = nboot, alpha = alpha)
+# Formula specific bootstrapping functions
+def cod_boot(ratio, nboot=100, alpha=0.05):
```
jeancochrane, line 167: **[Question, non-blocking]** You mentioned in the PR body that this function is defined in `assesspy`, but isn't available for use by package consumers:

> things like cod_boot are included in the assesspy package, but can't be accessed from the package to use outside of it.

 Can you point me to the lines in `assesspy` where it's defined? I couldn't find it, but if it's true that it's defined in `assesspy`, we should try to figure out how to make it available to this script too.

aws-glue/jobs/reporting-ratio_stats.py
jeancochrane: Reading the PR body again, it seems like this might be a result of difficulty refactoring to use the `assesspy` import:

> I've been having issues removing the ap. before the functions since it also renames the column names. Because of that they are imported separately. I have a structure in place and am working through it, but it's just time consuming running checks on it in AWS. I'll send a separate push if this is necessary to fix.

I don't quite understand what the trouble is, perhaps we could pair on it together?

jeancochrane (changes requested):
Looking great so far! Just a few nitpicks below. I'm not totally sure why `black` started failing on CI, seeing as it seems like the formatting it was complaining about pre-dated your changes; but I think it's good to switch to `black` formatting for this script anyway so we might as well do it now.

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -1,91 +1,85 @@
...
+import time
+
+import boto3
+import numpy as np
+import pandas as pd
+from assesspy import boot_ci, cod
+from assesspy import cod_ci as cod_boot
+from assesspy import cod_met, detect_chasing, mki, mki_met, prb, prb_met, prd
+from assesspy import prd_ci as prd_boot
+from assesspy import prd_met
```
jeancochrane, line 11: **[Nitpick, non-blocking]** We could collapse this all into one import

```suggestion
from assesspy import (
    boot_ci,
    cod,
    cod_ci as cod_boot,
    cod_met,
    detect_chasing,
    mki,
    mki_met,
    prb,
    prb_met,
    prd,
    prd_ci as prd_boot,
    prd_met
)
```

jeancochrane (approved):
Looks great to me! Let's get functional review from @dfsnow before we merge 👍🏻

aws-glue/jobs/reporting-ratio_stats.py
dfsnow: It turns out that `isort` really hates the more pythonic syntax, so sticking with Damon's original imports.

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -110,329 +104,246 @@ def run_query_get_result(
...
 
-# General boostrapping function
-def boot_ci(fun, *args, nboot = 100, alpha = 0.05):
+    median_n = sum(no_outliers)
 
-    num_args = len(args)
-    args = pd.DataFrame(args).T
-    n = len(args)
+    median_val = np.median(x_no_outliers)
+    median_ci = median_boot(x, nboot=1000)
```
dfsnow, line 140: @wrridgeway This is copied from the original code, but shouldn't this line be:

```suggestion
    median_ci = median_boot(x_no_outliers, nboot=1000)
```

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -110,329 +104,246 @@ def run_query_get_result(
...
+        out = [None, None, mki_n]
 
-def median_boot(ratio, nboot = 100, alpha = 0.05):
+    return out
 
-    return boot_ci(np.median, ratio, nboot = nboot, alpha = alpha)
 
-# Fairness metrics functions that comply with CCAO's SOPs
 def ccao_cod(ratio):
     """ """
```
dfsnow, line 174: **nitpick:** Drop this unused docstring.

dfsnow (commented):
Nice work @Damonamajor! This is nearly good to go, see my few comments, one of which is for @wrridgeway.

@wrridgeway Can you give this a final PR review so we can merge it this week?

aws-glue/jobs/reporting-ratio_stats.py
wrridgeway: Yes, it should be.

wrridgeway:
Could we run the new and the old script and compare their output if this hasn't already been done?

Damonamajor:
> @Damonamajor Could we run the new and the old script and compare their output if this hasn't already been done?

Yup, I'll do it after this meeting.

aws-glue/jobs/reporting-ratio_stats.py
Damonamajor: @wrridgeway @dfsnow I'm not sure why, but this change seems to make it not run on AWS.

Damonamajor:
> > @Damonamajor Could we run the new and the old script and compare their output if this hasn't already been done?
> 
> Yup, I'll do it after this meeting.

I ran and downloaded the files. The dimensions are the same (aside from adding mki things), as well as the outputs. The one exception is the confidence intervals, which are very minorly off. Billy said this was a result of bootstrapping.

wrridgeway (changes requested):

aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -1,91 +1,85 @@
...
-import sys
-from awsglue.transforms import *
-from awsglue.utils import getResolvedOptions
-from pyspark.context import SparkContext
+import re
+import time
+
+import boto3
+import numpy as np
+import pandas as pd
```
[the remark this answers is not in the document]
Damonamajor: @wrridgeway The tables are named along with the functions. Doing this, without extra steps, would result in title names being ap.cod_ci.

wrridgeway (approved):

PR #127, opened 2023-09-14
Simplify and de duplicate condo characteristics view

wrridgeway:
This PR addresses two primary concerns:

1. Joining `pardat` with `comdat` and `oby` can lead to duplicates due to multiple landlines within PINs and cards. We only need a few columns from `oby`/`comdat` that don't vary within PIN (year built, essentially), so we can effectively limit our selection of rows from these "core" tables as the first row when sorted by `lline`.
2. There _shouldn't_ be duplicates in the condo characteristics data, but if there are, they will be filtered out during ingest in the future to avoid creating duplicates in out modelling view.

We also replaced calculating prior year values for nonlivable detection with values from `vw_pin_history`.


dfsnow (changes requested):
@wrridgeway The SQL updates look great, awesome simplification. But it looks like the line endings on `aws-s3/scripts-ccao-data-warehouse-us-east-1/ccao-condominium-pin_condo_char.R` got changed, making the diff unusable. Can you use `unix2dos` to swap the line endings then re-request review?

wrridgeway:
Should be fixed

dfsnow (approved):

Issue #129, opened 2023-09-14
[Data catalog] Defer to master dbt state on CI to resolve intermittent build failures


jeancochrane:
There are currently two scenarios for remote state [deferral](https://docs.getdbt.com/reference/node-selection/defer) on CI:

1. Prior to the first successful run, PR runs pull remote state for the master branch and defer to that state
    * This means that deferred models will point to views and tables in the production schemas (e.g. `location`, `default`)
2. Following the first successful run, PR runs pull remote state specific to the branch
    * Deferred models will point to views/tables in the CI schemas (e.g. `ci_feature-branch-name_location`, `ci__feature-branch-name_default`)

The goal of this design was to allow us to defer building resources unless the branch diff requires we refresh them, while also allowing us to only build and test models that have changed in a given commit. By deferring to branch-specific state, we aimed to only have to test changes to a model in a given PR _for the specific commit that changed that model_, rather than having to rerun every build and test for every model that has been changed in every commit to the branch.

However, this design has demonstrated a tendency to cause state mismatch on CI on runs following the first successful run; `ref()` calls on subsequent runs can point to views/tables in the CI schemas that have not been created, since the first successful run actually deferred to views/tables in the production schemas instead. As an example, see [this PR run](https://github.com/ccao-data/data-architecture/actions/runs/6178064304/job/16770545222) where we attempted to recreate `vw_pin_condo_char`, but failed due to the fact that its dependency model `vw_pin_land` did not exist in the CI schema as it had instead deferred to production in the first successful run on that branch.

To resolve this problem, we should change the design of remote deferral on CI so that we _always defer to the master branch state_ instead of managing branch-specific state. The downside of this change is that we will have to start building and testing every model that has been changed in that branch on every run, even if a previous commit already ran the test; but the upside is that we will begin to finally start deferring correctly to production models and we will simplify our state management in the process.

PR #132, opened 2023-09-14
Always defer to master dbt state on CI to resolve intermittent build failures

jeancochrane:
We've been seeing some intermittent build failures on CI due to the fact that we're attempting to defer to branch-specific state rather than master state on CI in some cases. (More details in #129.) This PR simplifies that logic by always deferring to master branch state on CI, which means that we will run more builds and tests than we otherwise would, but we will also experience a more stable CI pipeline.

I tested these changes a few different ways via temporary commits:

* See [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/6191129053/job/16808775960#step:8:1) for a test confirming that this deferral behavior works in a simple case where we modify `default.vw_pin_universe` and want to build/test it without building/testing its dependencies.
* See [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/6191206527/job/16809028111#step:4:88) for a test confirming that the altered `restore-cache` action works when no exact match is found and no fallback is configured
* See [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/6191223729/job/16809080729#step:4:92) for a test confirming that `restore-cache` works when no exact match is found and no fallback match is found either

Closes https://github.com/ccao-data/data-architecture/issues/129.

### DIFF
--- a/.github/actions/configure_dbt_environment/action.yaml
+++ b/.github/actions/configure_dbt_environment/action.yaml
@@ -11,16 +11,12 @@ runs:
           echo "On pull request branch, setting dbt env to CI"
           {
             echo "TARGET=ci";
-            echo "CACHE_KEY=pr-caches/$GITHUB_HEAD_REF";
-            echo "CACHE_RESTORE_KEY=master-cache"
             echo "HEAD_REF=$GITHUB_HEAD_REF"
           } >> "$GITHUB_ENV"
         elif [[ $GITHUB_REF_NAME == 'master' ]]; then
           echo "On master branch, setting dbt env to prod"
           {
             echo "TARGET=prod";
-            echo "CACHE_KEY=master-cache";
-            echo "CACHE_RESTORE_KEY=master-cache"
           } >> "$GITHUB_ENV"
         else
           echo "CI context did not match any of the expected environments"
--- a/.github/actions/restore_dbt_cache/action.yaml
+++ b/.github/actions/restore_dbt_cache/action.yaml
@@ -8,8 +8,8 @@ inputs:
     description: The cache key to query for.
     required: true
   restore-key:
-    description: An additional key to use as a fallback for the cache.
-    required: true
+    description: An optional additional key to use as a fallback for the cache.
+    required: false
   bucket:
     description: The S3 bucket that should store the cache.
     required: false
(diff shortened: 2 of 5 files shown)


.github/actions/restore_dbt_cache/action.yaml
```diff
@@ -41,16 +41,26 @@ runs:
...
+                echo "cache-matched-key=$RESTORE_KEY"
+              } >> $GITHUB_OUTPUT
+            else
+              echo "Cache miss: Did not find fallback match for cache key"
+              {
+                echo "cache-hit=false";
+                echo "exact-match=false";
+                echo "cache-matched-key=''";
+              } >> $GITHUB_OUTPUT
+            fi
```
jeancochrane, line 61: This block has not changed, it was simply indented further and gated to match the specific case where an exact `key` match was not found and a `restore-key` fallback is configured. Previously, `restore-key` was required, so this was the only possible logic that could be executed.

.github/actions/configure_dbt_environment/action.yaml
```diff
@@ -11,16 +11,12 @@ runs:
           echo "On pull request branch, setting dbt env to CI"
           {
             echo "TARGET=ci";
-            echo "CACHE_KEY=pr-caches/$GITHUB_HEAD_REF";
```
jeancochrane, line 14: `CACHE_KEY` is moving to the `dbt.env` file, since it's no longer different between master and PR environments.

.github/actions/configure_dbt_environment/action.yaml
```diff
@@ -11,16 +11,12 @@ runs:
...
-            echo "CACHE_RESTORE_KEY=master-cache"
```
jeancochrane, line 15: `CACHE_RESTORE_KEY` is no longer necessary since we're not using fallback caches anymore.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -29,7 +29,6 @@ jobs:
         with:
           path: ${{ env.PROJECT_DIR }}/${{ env.STATE_DIR }}
           key: ${{ env.CACHE_KEY }}
-          restore-key: ${{ env.CACHE_RESTORE_KEY }}
```
dfsnow, line 32: **question:** I'm slightly unclear: if we're not passing the `restore-key` to the composite action and it's not a required input, when will the fallback behavior ever be triggered?

dfsnow (commented):
Looks good to me @jeancochrane! Only question for me is about the need for keeping the fallback restore behavior now that we have branch-specific caches.

.github/workflows/build_and_test_dbt.yaml
jeancochrane: Good question @dfsnow! We don't expect the fallback behavior to ever be triggered as the workflow is currently written. I decided to leave it in anticipating that we may want to implement other workflows that use our custom `restore_cache` action with fallbacks in the future, but seeing as we could always remove it and reintroduce it later if we decide we need it, I'm also down to strip it out in this PR; let me know what you'd prefer!

dfsnow: @jeancochrane I'm fine with leaving it in! Will approve this PR in a sec.

dfsnow (approved):

PR #134, opened 2023-09-14
Add SNS notification for sqoop-bot

### LINKED ISSUE #109: Fix failure notifications for sqoop-bot
The `test-dbt-models` workflow has been failing the past couple of days due to some malformed vw_card_res_char renovation data: https://github.com/ccao-data/data-architecture/actions/workflows/test_dbt_models.yaml.

However, since GitHub only notifies whoever triggered the workflow, no one has receives GitHub or email notifications. We need to figure out how to route these failure notifications from `sqoop-bot` initiated workflows to the correct human.

dfsnow:
This pull request adds a composite action to send an AWS SNS email when the `test-dbt-models` workflow fails. This additional notification system is necessary because there doesn't seem to be a way to notify people when scheduled or automated jobs fail using GitHub's built-in notification system. See [this SO post](https://stackoverflow.com/questions/62304258/github-actions-notifications-on-workflow-failure) that is similar to our use case.

Note that I added a GitHub secret for the SNS topic ARN. I also updated the role policy to allow SNS publishing to only the relevant topic ARN.

Closes #109.


.github/actions/send_notification/action.yml
```diff
@@ -0,0 +1,39 @@
...
+        TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%S")
+        {
+          echo "NOTIFICATION_SUBJECT=dbt tests failed for workflow run: $GITHUB_RUN_ID UTC"
+          echo "NOTIFICATION_BODY<<EOF"
+          echo "dbt tests failed for workflow $GITHUB_RUN_ID, run on $TIMESTAMP"
+          echo ""
+          echo "Link to failing workflow: \
+        https://github.com/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
+          echo EOF
+        } >> "$GITHUB_ENV"
```
dfsnow, line 30: This constructs a little multi-line string for the email body. Here's a sample output:

> dbt tests failed for workflow 6191942634, run on 2023-09-14T23:32:02Z
>
> Link to failing workflow: https://github.com/ccao-data/data-architecture/actions/runs/6191942634

.github/actions/setup_dbt/action.yaml
```diff
@@ -14,7 +14,7 @@ runs:
       uses: ./.github/actions/load_environment_variables
 
     - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@v2
+      uses: aws-actions/configure-aws-credentials@v4
```
dfsnow, line 17: I figured we might as well bump this action version since we started getting deprecation warning for v2 in the log output. Also v4 came out [just a few days ago](https://github.com/aws-actions/configure-aws-credentials/releases/tag/v4.0.0) 🤷

.github/workflows/test_dbt_models.yaml
```diff
@@ -23,3 +23,10 @@ jobs:
         run: dbt test --target "$TARGET"
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
+
+      - name: Send failure notification
+        if: failure()
```
dfsnow, line 28: I _think_ this works and runs only if the workflow fails. I found the docs a little confusing here. Will test it more once it's merged and I can run workflow dispatch manually.

dfsnow:
Note that I cancelled the `build-and-test-dbt` workflow because the main branch is still in the process of building the CTAS. Should be able to re-run manually in the morning to update the state and get the build to run.

.github/actions/send_notification/action.yml
jeancochrane: **[Suggestion, non-blocking]** Nice and simple! I think that [the `$GITHUB_OUPUT` parameter](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) might be a more appropriate place for these variables to live, since configuring them as output variables would allow subsequent steps to opt into reading them (in contrast to the `$GITHUB_ENV` environment variables, which all subsequent steps have to inherit, including steps in the workflow that calls this action).

jeancochrane (approved):
Looks great, thanks for taking this on! I'm fine with us testing the `failure()` logic in production 👍🏻 

One (non-blocking) thought about the design: Based on the fact that notification logic is factored out into an action, and that the action has a general name like `send_notification`, I would expect it to be designed as a general-purpose, reusable action for sending notifications; instead, by hardcoding the body/message, we tightly couple it to the specific case of a dbt test failure. This might be fine given that we don't yet have another use-case in mind for sending notifications, but I would suggest one of the following approaches to make the module boundaries more clear:

* **Option 1**: Don't use a composite action and instead define the logic directly in `test_dbt_models`
    * This seems reasonable to me given that the action is abstracting away a pretty small amount of logic, so maybe we want to keep things simple by defining a custom `aws sns` call in whichever workflows need it, but it offers the least reusability (i.e. none)
* **Option 2:** Rename the action to something that indicates its tight coupling, like `send_dbt_test_failure_notification`
    * This approach is most similar to the current solution, but it leaves unresolved the question of why factor out a composite action if it's so tightly coupled to one use-case
* **Option 3**: Factor out the body and message into input variables to the action, and tweak the name/description to make it clear that the action can send arbitrary notifications
    * This would offer the most reusability, but is the biggest change from the current design, and I'm a bit uncertain whether we'll be able to preserve the multiline string

I'm happy to move forward with whatever option makes the most sense to you, or to just merge this in and decide on an approach later once we have a better idea of whether we want to reuse the `send_notification` logic!

.github/actions/send_notification/action.yml
dfsnow: Good point! Fixed in 98134ad99655d7c9d14b44453289c8b8ffde5f37.

dfsnow:
> * **Option 1**: Don't use a composite action and instead define the logic directly in `test_dbt_models`

Good points @jeancochrane! I think we should go with option 1 and I refactored as such in 98134ad99655d7c9d14b44453289c8b8ffde5f37. I don't think we're likely to use this action elsewhere; we should really be relying on the built-in notification system wherever possible.

PR #136, opened 2023-09-15
Update dbt docs with more detail on handling table models and upgrade to dbt 1.6

jeancochrane:
The introduction of our spatial queries as materialized tables into the dbt DAG in https://github.com/ccao-data/data-architecture/pull/111 creates an additional layer of complexity for local dbt development, since the time to build the full DAG has increased from a couple of minutes to a couple of hours.

This PR updates the dbt README to provide detailed instructions on how to avoid building compute-intensive tables using [`dbt clone`](https://docs.getdbt.com/reference/commands/clone) and [the `--exclude` option](https://docs.getdbt.com/reference/node-selection/exclude). In order to support this workflow, we upgrade to the v1.6.x lines of dbt Core and the dbt-athena adapter, which are the first versions to support `dbt clone` in addition to being the latest stable minor versions of both libraries.

We also take this opportunity to sneak in one small improvement to developer experience by reducing the number of times that dbt-athena will retry a failed query from 5 to 1; in my experience, intermittent Athena failures are very rare compared to SQL syntax and compilation failures, and eliminating the retry logic allows for faster debugging cycles when trying to resolve the latter class of failure.

### DIFF
--- a/dbt/.gitignore
+++ b/dbt/.gitignore
@@ -5,3 +5,4 @@ target/
 dbt_modules/
 dbt_packages/
 assets/*.svg
+master-cache/
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -125,56 +125,96 @@ aws-mfa
 
 #### Build tables and views
 
-Build all the tables and views in our Athena warehouse:
+We use the [`dbt run` command](https://docs.getdbt.com/reference/commands/run)
+to build tables and views (called
+[models](https://docs.getdbt.com/docs/build/models) in dbt jargon) in our
+Athena data warehouse. See the following sections for specific instructions
+on how to build [development](#build-tables-and-views-in-development) and
+[production](#build-tables-and-views-in-production) models.
+
+#### Build tables and views in development
+
+When passed no arguments, `dbt run` will default to building _all_ tables
+and views in development schemas dedicated to your user. The full build takes
+about three hours and thirty minutes, so we don't recommend running it
+from scratch.
+
+Instead, start by copying the production dbt state file (also known as the
+[manifest file](https://docs.getdbt.com/reference/artifacts/manifest-json)):
 
 ```
-dbt run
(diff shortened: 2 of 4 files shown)


dbt/.gitignore
```diff
@@ -5,3 +5,4 @@ target/
 dbt_modules/
 dbt_packages/
 assets/*.svg
+master-cache/
```
jeancochrane, line 8: The docs now recommend storing the main branch state in this subdirectory, so we add it to `.gitignore` to ensure that it does not get committed accidentally.

dbt/README.md
```diff
@@ -125,57 +125,97 @@ aws-mfa
...
+aws s3 cp s3://ccao-dbt-cache-us-east-1/master-cache/manifest.json master-cache/manifest.json
 ```
 
-To build a subset of the models, use the `--select` option:
+Then, use [`dbt clone`](https://docs.getdbt.com/reference/commands/clone) to
+clone the production tables and views into your development environment:
 
 ```
-dbt run --select location.vw_pin10_location default.vw_pin_universe
+dbt clone --state master-cache
```
dfsnow, line 153: **praise:** This is so cool! <3

dbt/README.md
```diff
@@ -404,3 +444,34 @@ It usually means that dbt's state has unresolvable conflicts with the current
...
+note that this will rebuild _all_ dependency models, even ones that already
+exist in your development environment, so if your model depends on another model
+that is compute-intensive (basically, anything in the `location` or `proximity`
+schemas) you should use [the `--exclude`
+option](https://docs.getdbt.com/reference/node-selection/exclude) to exclude
+these compute-intensive models from being rebuilt:
+
+```
+dbt run --select +model.vw_pin_shared_input --exclude location.* proximity.*
+```
```
dfsnow, line 477: Extremely cool. Wild that dbt is this flexible.

dfsnow (approved):
Looks great @jeancochrane! Very helpful addition.

Issue #120, opened 2023-09-11
[Data catalog] Fast follow: Fix indentation in CTAS


jeancochrane:
In https://github.com/ccao-data/data-architecture/pull/111, we added semantically unnecessary subqueries to the queries in `aws-athena/ctas/` in order to make the diff easier to review. Once #111 comes in, we should remove these subqueries to keep the queries clean.

See this comment thread for context: https://github.com/ccao-data/data-architecture/pull/111/files#r1317607121

Blocked by #111.

PR #137, opened 2023-09-15
Remove unnecessary subqueries in CTAS models

jeancochrane:
This PR implements a simple change that produces an absolutely horrendous diff: Removing the semantically-unnecessary CTAS subqueries that we introduced to make the diff in #111 simpler. The goal was to keep this PR as simple as possible, changing nothing except formatting, so that the reviewer can skim the diff and be confident that they won't miss anything important.

The diff is much easier to read if you enable the "hide whitespace" feature in the `Files changed` view:

Screenshot 2023-09-17 at 4 40 56 PM

Closes https://github.com/ccao-data/data-architecture/issues/120.

### DIFF
--- a/aws-athena/ctas/location-access.sql
+++ b/aws-athena/ctas/location-access.sql
@@ -7,71 +7,67 @@
     )
 }}
 
-WITH access AS (
-    WITH distinct_pins AS (
-        SELECT DISTINCT
-            x_3435,
-            y_3435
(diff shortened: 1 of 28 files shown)


dfsnow (approved):
Reviewing this with hidden whitespace makes the diff much less scary! Looks great @jeancochrane.

Issue #130, opened 2023-09-14
Refactor spatial queries to reduce build times


jeancochrane:
There is a decent amount of duplication between in our spatial queries, and a lot of stable but compute-heavy operations. We should refactor this queries with an eye toward two related improvements:

1. Factoring out any duplicated logic that's easy to factor out
2. Pulling stable, compute-heavy operations out into intermediate models so that we can rerun them as infrequently as possible

jeancochrane:
@dfsnow I think this was completed in https://github.com/ccao-data/data-architecture/pull/217, is that right?

dfsnow:
> @dfsnow I think this was completed in #217, is that right?

@jeancochrane There are still a few very long running `cnt_` queries in the `proximity` DB that need to be optimized. Was keeping this open as a reminder to someday get to those.

PR #138, opened 2023-09-18
Add `proximity.dist_pin_to_pin` to the dbt DAG using intermediate transforms

### LINKED ISSUE #122: [Data catalog] Move `proximity.dist_pin_to_pin` CTAS query into the dbt DAG
While we moved just about all of the CTAS queries into the DAG in https://github.com/ccao-data/data-architecture/pull/111, we left out `proximity.dist_pin_to_pin` because it requires a more complex solution. Create a separate PR to move `proximity.dist_pin_to_pin` into the DAG by defining a series of intermediate transformations that gradually expand the radius of the buffer.

See this comment thread for more context: https://github.com/ccao-data/data-architecture/pull/111#discussion_r1321844067

Blocked by #111.

jeancochrane:
This PR changes `proximity.dist_pin_to_pin` from a source to a ref so that it can be tracked in the dbt DAG. In order to build the table in a reasonable amount of time, we define a series of intermediate transformations, documented in `dbt/models/proximity/docs.md`, that expand the PIN-to-PIN buffer gradually so that we only compute the 10km buffer for PINs that do not have neighbors within 1km.

As evidenced by [this successful GitHub workflow run](https://github.com/ccao-data/data-architecture/actions/runs/6251873412/job/16973914989), this new design takes about 14 minutes to build `dist_pin_to_pin`.

Closes #122.


.sqlfluff
```diff
@@ -35,3 +35,6 @@ extended_capitalisation_policy = upper
 
 [sqlfluff:rules:convention.casting_style]
 preferred_type_casting_style = cast
+
+[sqlfluff:templater:jinja]
+load_macros_from_path = dbt/macros
```
jeancochrane, line 40: As of this PR, we're now using a custom macro in the `aws-athena/` subdirectory, so we need to update the `.sqlfluff` config to let SQLFluff know where the macros are defined. SQLFluff has [builtin support](https://docs.sqlfluff.com/en/stable/configuration.html#builtin-jinja-macro-blocks) for native dbt macros, but custom macros need to be configured using `load_macros_from_path`. (Sidebar: somehow we've gotten away with using custom macros in the `dbt/macros/tests` directory, in particular the `assert_equal` macro we use for unit testing, but I'm not sure how! I spent some time trying to figure it out but gave up after about an hour, since it doesn't seem necessary to figure out.)

aws-athena/ctas/proximity-dist_pin_to_pin_10km.sql
```diff
@@ -0,0 +1,36 @@
...
+{{
+    config(
+        materialized='table',
+        partitioned_by=['year'],
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH missing_matches AS (  -- noqa: ST03
```
jeancochrane, line 13: We ignore [ST03](https://docs.sqlfluff.com/en/stable/rules.html#rule-structure.unused_cte) here because we technically _do_ reference `missing_matches` on line 31 below, but since it's passed into a macro as a string, SQLFluff isn't able to recognize it as a reference. (This context might make sense as an inline comment, but it seemed borderline to me so I left it out; let me know if you disagree!)

dbt/macros/nearest_pin_neighbors.sql
```diff
@@ -0,0 +1,95 @@
...
+-- another PIN that is within `radius_km` of the given PIN.
+--
+-- The `source_model` must contain four required columns, and is modeled
+-- after spatial.parcel, which is the primary source for this macro:
+--
+-- * pin10
+-- * year
+-- * x_3435
+-- * y_3435
+{% macro nearest_pin_neighbors(source_model, num_neighbors, radius_km) %}
```
jeancochrane, line 13: This macro is nearly identical to the pre-existing `aws-athena/ctas/proximity-dist_pin_to_pin.sql` query, with three main changes:

1. Parameterization of `source_model`, `num_neigbors`, and `radius_km`
2. Addition of `pin10` to the `distinct_pins` subquery per our missing data discussion on Teams ([diff here](https://github.com/ccao-data/data-architecture/pull/138/commits/8aaf932db4e83748a485727216a7e7b81a5c002c))
3. Reformatting to match dbt styles using `sqlfmt`

Everything else should be identical!

dbt/models/proximity/docs.md
```diff
@@ -0,0 +1,21 @@
...
+This type of recursive spatial query is expensive, however, and some PINs are
+quite far (>1km) from the nearest three PINs, so we use intermediate tables
+to strike a balance between data completeness and computational efficiency.
+
+To compute the full set of distances in `proximity.dist_pin_to_pin`, we first
+generate PIN-to-PIN distances using a 1km buffer and store the results in the
+`proximity.dist_pin_to_pin_1km` table. Then, we query for PINs that did not have
+any matches within 1km and redo the distance query with an expanded 10km buffer,
+storing the results in the `proximity.dist_pin_to_pin_10km` table. Finally, the
+10km table is aliased to the `proximity.dist_pin_to_pin` view for ease of
```
jeancochrane, line 19: Note that while `dist_pin_to_pin` is now a view, it doesn't make use of the `vw_*` prefix we typically use for views. This seemed OK in this particular case, since the new view is intended to be a light layer of abstraction on top of the new underlying intermediate tables, but if you think it's worth adding the prefix and refactoring the consumers of the view to point to the new ref for clarity then I would be happy to do so.

.sqlfluff
dfsnow: Fascinating. And also very convenient/fortunate that the stack of tools we've chosen continue to play so nicely together!

aws-athena/ctas/proximity-dist_pin_to_pin_10km.sql
dfsnow: I think this is fine without being a comment, since it's pretty clear from the code itself.

aws-athena/ctas/proximity-dist_pin_to_pin.sql
```diff
@@ -1,136 +1,2 @@
...
-        nearest_neighbor_3_pin10,
-        nearest_neighbor_3_dist_ft,
-        year
-    FROM proximity.dist_pin_to_pin_temp2
-);
-
-DROP TABLE IF EXISTS proximity.dist_pin_to_pin_temp
-DROP TABLE IF EXISTS proximity.dist_pin_to_pin_temp2
+-- View that finds the 3 nearest neighbor PINs for every PIN for every year
+SELECT * FROM {{ ref('proximity.dist_pin_to_pin_10km') }}
```
dfsnow, line 2: **issue (blocking):** If `dist_pin_to_pin_10km.sql` only contains the values _missing_ from `dist_pin_to_pin_1km.sql`, then shouldn't this be the union of `dist_pin_to_pin_1km.sql` and `dist_pin_to_pin_10km.sql`?

dbt/macros/nearest_pin_neighbors.sql
```diff
@@ -0,0 +1,95 @@
...
+            select
+                pcl.pin10,
+                {% for idx in range(1, num_neighbors + 1) %}
+                    max(
+                        case when pd.row_num = {{ idx + 1 }} then pd.neighbor_pin10 end
+                    ) as nearest_neighbor_{{ idx }}_pin10,
+                    max(
+                        case when pd.row_num = {{ idx + 1 }} then pd.dist end
+                    ) as nearest_neighbor_{{ idx }}_dist_ft,
+                {% endfor %}
```
dfsnow, line 84: **praise:** This is really neat. I bet a bunch of our other views can be condensed using jinja templating.

dbt/models/proximity/docs.md
dfsnow: I agree it's fine to keep the current name since it's such a light abstraction (just combining two tables, vs combining many tables + cleaning, as the other views do).

dfsnow, line 19: **praise:** Thank you for adding clear docs for my completely undocumented hack! Much appreciated.

dfsnow (commented):
@jeancochrane Great work! When we visit #130 it may be worth it to see what else we can abstract into macros. See my one blocking issue, but otherwise looks good.

jeancochrane:
@dfsnow Looks like the new table doesn't have quite the same contents as the existing one:

```sql
select count(*)
from "ci_jeancochrane-122-data-catalog-move-proximitydist-pin-to-pin-ctas-query-into-the-dbt-dag_proximity"."dist_pin_to_pin"

Results: 32370578
```

```sql
select count(*)
from "proximity"."dist_pin_to_pin"

Results: 32420636
```

That's a difference of about 50k rows; I would be fine with this if the new version had _more_ rows, but it seems concerning that there are fewer.

Further queries confirm that the new version of the table is a subset of the old version:

```sql
select count(*)
from "proximity"."dist_pin_to_pin" as prod
full outer join "ci_jeancochrane-122-data-catalog-move-proximitydist-pin-to-pin-ctas-query-into-the-dbt-dag_proximity"."dist_pin_to_pin" as ci
on prod.pin10 = ci.pin10 and prod.year = ci.year
where prod.pin10 is null

Results: 0
```

```sql
select count(*)
from "proximity"."dist_pin_to_pin" as prod
full outer join "ci_jeancochrane-122-data-catalog-move-proximitydist-pin-to-pin-ctas-query-into-the-dbt-dag_proximity"."dist_pin_to_pin" as ci
on prod.pin10 = ci.pin10 and prod.year = ci.year
where ci.pin10 is null

Results: 53610
```

Here's a version of the query grouped by year:

```sql
select prod.year, count(*)
from "proximity"."dist_pin_to_pin" as prod
full outer join "ci_jeancochrane-122-data-catalog-move-proximitydist-pin-to-pin-ctas-query-into-the-dbt-dag_proximity"."dist_pin_to_pin" as ci
on prod.pin10 = ci.pin10 and prod.year = ci.year
where ci.pin10 is null
group by prod.year
order by prod.year desc
```

| year | count of missing PINs |
| --- | --- |
| 2022 | 2931 |
| 2021 | 2369 |
| 2020 | 2885 |
| 2019 | 2870 |
| 2018 | 2677 |
| 2017 | 2629 |
| 2016 | 2488 |
| 2015 | 2632 |
| 2014 | 2715 |
| 2013 | 2760 |
| 2012 | 2862 |
| 2011 | 2759 |
| 2010 | 2760 |
| 2009 | 2791 |
| 2008 | 2304 |
| 2007 | 2143 |
| 2006 | 2038 |
| 2005 | 1836 |
| 2004 | 1630 |
| 2003 | 1508 |
| 2002 | 1423 |
| 2001 | 1331 |
| 2000 | 1269 |

What do you think @dfsnow? Is this worth investigating before we merge and overwrite the existing table?

dfsnow:
> What do you think @dfsnow? Is this worth investigating before we merge and overwrite the existing table?

@jeancochrane Do a quick anti-join on pin and let's see if it's easy to diagnose. If it takes more than 30 minutes I'm comfortable just merging it as is, since the 3 neighbors are only used for filling.

jeancochrane:
@dfsnow Not much closer to diagnosing it, but here's some interesting information for you -- only about 5k PINs are missing data for multiple years:

```sql
with missing_pins as (
    select prod.pin10 as pin10, prod.year as year
    from "proximity"."dist_pin_to_pin" as prod
    full outer join "ci_jeancochrane-122-data-catalog-move-proximitydist-pin-to-pin-ctas-query-into-the-dbt-dag_proximity"."dist_pin_to_pin" as ci
    on prod.pin10 = ci.pin10 and prod.year = ci.year
    where ci.pin10 is null
),

missing_pins_grouped_by_year as (
    select pin10, array_agg(year) as missing_years
    from missing_pins
    group by pin10
    having cardinality(array_agg(year)) > 1
)

select count(*)
from missing_pins_grouped_by_year

Results: 4940
```

And only 153 PINs are missing data for _all_ years:

```sql
select count(*)
from missing_pins_grouped_by_year
where cardinality(missing_years) = 23

Results: 153
```

This at least makes me feel better that we're not just dropping an entire group of PINs! If you feel fine with this, I'll go ahead and merge.

dfsnow:
@jeancochrane Likely has something to do with PIN changes across years or fake "999" PINs. Let's just go ahead and merge it.

jeancochrane:
Sounds good, thanks @dfsnow! Mind hitting the Approve button so I can merge?

dfsnow (approved):

PR #141, opened 2023-09-22
Remove deactivated iasWorld rows from Athena views

wrridgeway:
We discovered recently that several tables include `cur` and `deactivat` columns that should be invoked regularly; while ignoring these columns shouldn't lead to duplication, it will lead to old PINs that are no longer part of the universe being included in things like modelling and reporting.

This PR adds conditionals across the `default`, `modelling`, and `reporting` Athena DBs.


wrridgeway:
Code is ready for review, but I need to spend a little time testing these changes on athena.

wrridgeway:
After working through default, modeling, and reporting, the changes implemented in this PR appear to be working as expected.

wrridgeway:
After working through default, modeling, and reporting, the changes implemented in this PR appear to be working as intended.

dfsnow (approved):
@wrridgeway, this looks set to go, pending double checking with Mirella.

PR #142, opened 2023-09-26
Update build and test workflow with manual dispatch option

### LINKED ISSUE #128: [Data catalog] Write a dispatchable workflow for rebuilding models
Per [this PR comment thread](https://github.com/ccao-data/data-architecture/pull/111#issuecomment-17160958250), we need a way to rebuild our output models when their source data change, e.g. when we rerun our R extraction scripts once per year. We could theoretically do this by running one-off `dbt run` commands locally, but that would require downloading and then reuploading the remote master state in a way that is risky and fragile. There are a number of automated ways we could accomplish the same goal (see the PR comment thread for some discussion), but the fastest and most reliable will be to just define a new GitHub Action workflow that takes a list of selectors as input and uses them to build the selected models.

Workflow requirements:

* Is named `build-dbt-models` and stored in `.github/workflows/build_dbt_models.py`
* Can be [dispatched manually](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#configuring-a-workflow-to-run-manually) (see [`test-dbt-models`](https://github.com/ccao-data/data-architecture/blob/21b40aa4e4e38dceadac6db8d30d9140c36a0385/.github/workflows/test_dbt_models.yaml#L3) for an example of this)
* Accepts a list of strings as an [input variable](https://medium.com/@geralexgr/parameters-and-variables-github-workflows-8bc86b641a4b), where each string represents a symbol that can be passed to a [dbt `--select` option](https://docs.getdbt.com/reference/node-selection/syntax#specifying-resources)
* Has the same [checkout, setup, and remote state download behavior](https://github.com/ccao-data/data-architecture/blob/21b40aa4e4e38dceadac6db8d30d9140c36a0385/.github/workflows/build_and_test_dbt.yaml#L9-L32) as `build-and-test-dbt`
* Runs the build via `dbt run -t "$TARGET" -s <state selectors based on input variables> --defer --state "$STATE_DIR"` after downloading remote state
* [Updates the dbt cache](https://github.com/ccao-data/data-architecture/blob/21b40aa4e4e38dceadac6db8d30d9140c36a0385/.github/workflows/build_and_test_dbt.yaml#L73-L77) in the same fashion as `build-and-test-dbt` if the build succeeds

dfsnow:
This PR adds a manual dispatch option for the `build-and-test-dbt` workflow. It takes a space-separated list of model names as an input, pulls the `master` branch cache, builds and tests the specified models, then re-uploads the cache to S3.

Closes #128.

### DIFF
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -45,8 +45,6 @@ jobs:
         run: |
           if [[ $CACHE_HIT == 'true' ]]; then
             echo "MODIFIED_RESOURCES_ONLY=true" >> "$GITHUB_ENV"
-          else
-            echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
           fi
           if [[ $EVENT_NAME == 'workflow_dispatch' ]]; then
               echo "MANUALLY_DISPATCHED=true" >> "$GITHUB_ENV"
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -41,7 +41,7 @@ jobs:
           path: ${{ env.PROJECT_DIR }}/${{ env.STATE_DIR }}
           key: ${{ env.CACHE_KEY }}
 
-        name: Set command args for building/testing resources
+      - name: Set command args for building/testing resources
         run: |
           if [[ ${{ steps.cache.outputs.cache-hit == 'true' }} ]]; then
             echo "MODIFIED_RESOURCES_ONLY=true" >> "$GITHUB_ENV"
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -9,10 +9,11 @@ on:
   workflow_dispatch:
     inputs:
       models:
+        required: true
         description: >
           A space-separated list of dbt models to build.
           Prefix with the + sign to rebuild upstream models
-        default: default.vw_pin_sale default.vw_pin_universe
+        default: 'default.vw_pin_sale default.vw_pin_universe'
         type: string
 
 jobs:
@@ -40,19 +41,17 @@ jobs:
           path: ${{ env.PROJECT_DIR }}/${{ env.STATE_DIR }}
           key: ${{ env.CACHE_KEY }}
 
-      - if: steps.cache.outputs.cache-hit == 'true'
-        name: Set command args to build/test modified resources
-        run: echo "MODIFIED_RESOURCES_ONLY=true" >> "$GITHUB_ENV"
-        shell: bash
+        name: Set command args for building/testing resources
+        run: |
+          if [[ ${{ steps.cache.outputs.cache-hit == 'true' }} ]]; then
+            echo "MODIFIED_RESOURCES_ONLY=true" >> "$GITHUB_ENV"
+          else
+            echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
+          fi
 
-      - if: steps.cache.outputs.cache-hit != 'true'
-        name: Set command args to build/test all resources
-        run: echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
-        shell: bash
-
-      - if: inputs.models != '' && github.event_name == 'workflow_dispatch'
-        name: Set command args for manually dispatched workflow
-        run: echo "MANUALLY_DISPATCHED=true" >> "$GITHUB_ENV"
+          if [[ ${{ github.event_name == 'workflow_dispatch' }} ]]; then
+              echo "MANUALLY_DISPATCHED=true" >> "$GITHUB_ENV"
+          fi
         shell: bash
 
       - name: Test dbt macros
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -6,6 +6,15 @@ on:
   push:
     branches: [master]
 
+  workflow_dispatch:
+    inputs:
+      models:
+        description: >
+          A space-separated list of dbt models to build.
+          Prefix with the + sign to rebuild upstream models
+        default: default.vw_pin_sale default.vw_pin_universe
+        type: string
+
 jobs:
   build-and-test-dbt:
     runs-on: ubuntu-latest
@@ -41,6 +50,11 @@ jobs:
         run: echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
         shell: bash
 
+      - if: inputs.models != '' && github.event_name == 'workflow_dispatch'
+        name: Set command args for manually dispatched workflow
+        run: echo "MANUALLY_DISPATCHED=true" >> "$GITHUB_ENV"
+        shell: bash
+
       - name: Test dbt macros
         run: dbt run-operation test_all
         working-directory: ${{ env.PROJECT_DIR }}
@@ -49,8 +63,13 @@ jobs:
       - name: Build models
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
-            echo "Running build on modified/new resources only"
-            dbt run -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
+            if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
+              echo "Running build on manually selected resources"
+              dbt run -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR"
+            else
+              echo "Running build on modified/new resources only"
+              dbt run -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
+            fi
           else
             echo "Running build on all resources"
             dbt run -t "$TARGET"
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test_dbt.yaml
```diff
@@ -6,6 +6,16 @@ on:
   push:
     branches: [master]
 
+  workflow_dispatch:
+    inputs:
+      models:
+        required: true
```
dfsnow, line 12: I _think_ this will prevent an empty string from being passed (the docs are sparse on this). Going to let dbt handle any errors from passing malformed model names.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -31,14 +41,17 @@ jobs:
...
+          else
+            echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
+          fi
 
-      - if: steps.cache.outputs.cache-hit != 'true'
-        name: Set command args to build/test all resources
-        run: echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
+          if [[ ${{ github.event_name == 'workflow_dispatch' }} ]]; then
+              echo "MANUALLY_DISPATCHED=true" >> "$GITHUB_ENV"
+          fi
```
dfsnow, line 54: I condensed these into a single step just for the sake of brevity, but if there's a reason to have them separate let me know!

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -61,16 +79,21 @@ jobs:
...
+            fi
           else
             echo "Running tests on all resources"
             dbt test -t "$TARGET"
           fi
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
 
-      - if: github.ref == 'refs/heads/master'
+      - if: github.ref == 'refs/heads/master' && success()
```
dfsnow, line 96: I _think_ this is change is already implicit, but we might as well make the success dependency explicit.

.github/workflows/build_and_test_dbt.yaml
jeancochrane: **[Thought, non-blocking]** I think requiring this variable and letting dbt Core handle validation makes sense! An alternative interface we could consider would be to allow this variable to be empty and to set a `default` of `path:models`, which will build everything. But building everything is currently a pretty intensive operation, so I also think it would make sense to keep the current design and require users to ask for that explicitly.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -6,6 +6,16 @@ on:
...
     branches: [master]
 
+  workflow_dispatch:
+    inputs:
+      models:
+        required: true
+        description: >
+          A space-separated list of dbt models to build.
+          Prefix with the + sign to rebuild upstream models
+        default: 'default.vw_pin_sale default.vw_pin_universe'
```
jeancochrane, line 16: **[Question, non-blocking]** Any particular reason to default to these two models? And will this default ever be executed, or is it a no-op given that the variable is required?

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -6,6 +6,16 @@ on:
   push:
     branches: [master]
 
+  workflow_dispatch:
+    inputs:
+      models:
+        required: true
+        description: >
+          A space-separated list of dbt models to build.
+          Prefix with the + sign to rebuild upstream models
```
jeancochrane, line 15: **[Suggestion, non-blocking]** Do you feel confident that everyone on the team will understand the proper format for a dbt model (particularly, that the schema is required)? If not, perhaps we should include an example?

```suggestion
          A space-separated list of dbt models to build.
          Prefix with the + sign to rebuild upstream models.
          For example, "default.vw_pin_sale default.vw_pin_universe".
```

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -31,15 +41,20 @@ jobs:
...
-      - if: steps.cache.outputs.cache-hit != 'true'
-        name: Set command args to build/test all resources
-        run: echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
+      - name: Set command args for building/testing resources
+        run: |
+          if [[ $CACHE_HIT == 'true' ]]; then
+            echo "MODIFIED_RESOURCES_ONLY=true" >> "$GITHUB_ENV"
+          else
+            echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
+          fi
```
jeancochrane, line 50: **[Praise]** Nice consolidation here 👍🏻

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -49,8 +64,13 @@ jobs:
...
+            if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
+              echo "Running build on manually selected resources"
+              dbt run -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR"
+            else
+              echo "Running build on modified/new resources only"
+              dbt run -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
+            fi
           else
             echo "Running build on all resources"
             dbt run -t "$TARGET"
```
jeancochrane, line 76: **[Question, non-blocking]** I don't expect this fallback conditional branch to ever be reached, but in the rare case that it is reached, do we also need to handle manual dispatch?

```suggestion
            if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
              dbt run -t "$TARGET" -s ${{ inputs.models }}
            else
              dbt run -t "$TARGET"
            fi
```

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -61,16 +81,21 @@ jobs:
...
+            if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
+              echo "Running tests on manually selected resources"
+              dbt test -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR"
+            else
+              echo "Running tests on modified/new resources only"
+              dbt test -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR"
+            fi
           else
             echo "Running tests on all resources"
             dbt test -t "$TARGET"
```
jeancochrane, line 93: **[Question, non-blocking]** Same question here as above RE: fallback behavior!

.github/workflows/build_and_test_dbt.yaml
jeancochrane: **[Praise]** I also think it's implicit, but I agree that making it explicit seems reasonable anyway!

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -31,15 +41,20 @@ jobs:
...
+      - name: Set command args for building/testing resources
+        run: |
+          if [[ $CACHE_HIT == 'true' ]]; then
+            echo "MODIFIED_RESOURCES_ONLY=true" >> "$GITHUB_ENV"
+          else
+            echo "MODIFIED_RESOURCES_ONLY=false" >> "$GITHUB_ENV"
+          fi
+          if [[ $EVENT_NAME == 'workflow_dispatch' ]]; then
+              echo "MANUALLY_DISPATCHED=true" >> "$GITHUB_ENV"
+          fi
```
jeancochrane, line 53: **[Thought, non-blocking]** The lack of an `else` branch where we set `MANUALLY_DISPATCHED=false` seems like it doesn't mess up the check on line 67, probably because these bash blocks inherit the default behavior of continuing execution on unset variables; this makes me wonder whether the corresponding branch to set `MODIFIED_RESOURCES=false` on lines 48-50 above is also unnecessary 🤔 But that's not the focus of this PR so I'm fine punting on it until later!

jeancochrane (approved):
Looks great! I like the simplicity of this solution and the way that it continues to keep our CI build/test logic centralized. A couple thoughts below, but nothing blocking from my perspective.

.github/workflows/build_and_test_dbt.yaml
dfsnow: Hard to test since I can't seen the dispatch button on non-master branches, but I think this just auto-populates the field with those views, while the `required` arg just won't let you submit the dispatch with an empty string.

jeancochrane: Ah, OK! I think it makes sense to store an example value here as a prompt then.

.github/workflows/build_and_test_dbt.yaml
dfsnow: My thinking here was:

- The only way to reach this conditional is the `master-cache` missing from S3
- In which case, we would probably want to rebuild the full DAG anyways
- Rebuilding the cache with a subset of models would lead to errors, as the workflow expects `master-cache` to encompass the full DAG

jeancochrane: Makes sense to me!

.github/workflows/build_and_test_dbt.yaml
jeancochrane: I think [this comment](https://github.com/ccao-data/data-architecture/pull/142#discussion_r1338794842) resolves my concern here 👍🏻

.github/workflows/build_and_test_dbt.yaml
dfsnow: Fixed in [164b58a](https://github.com/ccao-data/data-architecture/pull/142/commits/164b58a00d3e02ea1090269d22469109f03b4d30)!

PR #145, opened 2023-09-28
Fix symlink instructions in new-dbt-model.md

jeancochrane:
@wagnerlmichael noticed while following the instructions to add a new dbt model that the command for creating a new symlink resulted in a broken link due to path resolution problems. This PR tweaks those docs to ensure that we give readers a symlink command that will work.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-dbt-model.md
+++ b/.github/ISSUE_TEMPLATE/new-dbt-model.md
@@ -140,18 +140,30 @@ models:
 
 - [ ] Add a symlink from the appropriate subfolder of the `dbt/models/`
   directory to the SQL query you created in the `aws-athena/` directory.
-  Make sure to change into the subdirectory of `dbt/models/` that will
-  store the symlink before running this command so that the link can
-  properly follow the relative path to the target.
+  Note that the path for the target file in the `aws-athena/` directory
+  needs to be relative to the location of the symlink, _not_ to the
+  location of the working directory where you run the `ln` command;
+  as a result, the path to the target file in the `aws-athena/` directory
+  will not autocomplete in your shell as it usually does when you use
+  the tab key.
 
 ```bash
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/new-dbt-model.md
```diff
@@ -140,13 +140,30 @@ models:
...
+  needs to be relative to the location of the symlink, _not_ to the
+  location of the working directory where you run the `ln` command;
+  as a result, the path to the target file in the `aws-athena/` directory
+  will not autocomplete in your shell as it usually does when you use
+  the tab key.
 
 ```bash
 # View example
-ln -s aws-athena/views/default-vw_new_model.sql dbt/models/default/default.vw_new_model.sql
+ln -s ../../../aws-athena/views/default-vw_new_model.sql dbt/models/default/default.vw_new_model.sql
```
wagnerlmichael, line 152: I think the command I used was `ln -s ../../../aws-athena/views/sale-vw_ias_salesval_upload.sql sale.vw_ias_salesval_upload.sql`. Not sure if it matters or not whether the location of the symlink is absolute relative to the root of the repo.

jeancochrane: Thanks for double-checking this @wagnerlmichael! I think it should be OK like this -- the difference with your command was that we changed into the `dbt/models/sale` subdirectory before running the command, but if we rant that command from the root of the `data-architecture` directory as specified in the new command displayed in the diff, it would have worked as well (I confirmed this locally).

wagnerlmichael: Got it, sounds good!

wagnerlmichael (approved):

dfsnow (approved):

PR #147, opened 2023-10-02
Add view for iasworld sales val join

### LINKED ISSUE #146: sale_key integration with sales val
When we want to implement an upload into iasWorld, we will need some sort of mechanism that allows us to perform a join. Since `doc_no` is not a unique identifier in the sales table in iasWorld, we will need to use the `sale_key` column. We will have to incorporate this in some way to our sales-val process.

wagnerlmichael:
I've added a view based on discussions with @dfsnow to help with the iasworld sales table upload. I tried to follow [this template](https://github.com/ccao-data/data-architecture/blob/master/.github/ISSUE_TEMPLATE/new-dbt-model.md) as best as I could. Let me know what needs fixing! @jeancochrane.

Closes #146

### DIFF
--- a/aws-athena/views/sale-vw_ias_salesval_upload.sql
+++ b/aws-athena/views/sale-vw_ias_salesval_upload.sql
@@ -2,7 +2,7 @@ WITH ias_sales AS (
     SELECT 
         salekey,
         NULLIF(REPLACE(instruno, 'D', ''), '') AS instruno_clean
-    FROM {{'iasworld', 'sales'}}
+    FROM {{ source('iasworld', 'sales') }}
     WHERE cur = 'Y'
     AND deactivat IS NULL
 )
--- a/aws-athena/views/sale-vw_ias_salesval_upload.sql
+++ b/aws-athena/views/sale-vw_ias_salesval_upload.sql
@@ -1,0 +1,17 @@
+WITH ias_sales AS (
+    SELECT 
+        salekey,
+        NULLIF(REPLACE(instruno, 'D', ''), '') AS instruno_clean
+    FROM {{'iasworld', 'sales'}}
+    WHERE cur = 'Y'
+    AND deactivat IS NULL
+)
+
+SELECT 
+    ias_sales.salekey,
+    sf.sv_is_outlier,
+    sf.sv_outlier_type,
+    CAST(NULL AS VARCHAR) AS analyst_override, -- Should we assume this has been created already in ias world and pull this column from iasworld.sale?
+    sf.run_id
+FROM ias_sales 
+JOIN {{ ref('sale.flag') }} sf ON ias_sales.instruno_clean = sf.meta_sale_document_num;
--- a/aws-athena/views/sale-vw_ias_salesval_upload.sql
+++ b/aws-athena/views/sale-vw_ias_salesval_upload.sql
@@ -1,28 +1,28 @@
 WITH ias_sales AS (
-    SELECT 
+    SELECT
         salekey,
         NULLIF(REPLACE(instruno, 'D', ''), '') AS instruno_clean
     FROM {{ source('iasworld', 'sales') }}
     WHERE cur = 'Y'
-    AND deactivat IS NULL
+        AND deactivat IS NULL
 ),
 
 max_version AS (
-    SELECT 
+    SELECT
         meta_sale_document_num,
         MAX(version) AS max_version
     FROM {{ source('sale', 'flag') }}
     GROUP BY meta_sale_document_num
 )
 
-SELECT 
+SELECT
     ias_sales.salekey,
     sf.sv_is_outlier,
     sf.sv_outlier_type,
     sf.run_id
-FROM ias_sales 
-INNER JOIN {{ source('sale', 'flag') }} sf 
+FROM ias_sales
+INNER JOIN {{ source('sale', 'flag') }} AS sf
     ON ias_sales.instruno_clean = sf.meta_sale_document_num
-INNER JOIN max_version mv
+INNER JOIN max_version AS mv
     ON sf.meta_sale_document_num = mv.meta_sale_document_num
     AND sf.version = mv.max_version;
--- a/aws-athena/views/sale-vw_ias_salesval_upload.sql
+++ b/aws-athena/views/sale-vw_ias_salesval_upload.sql
@@ -13,4 +13,4 @@ SELECT
     sf.sv_outlier_type,
     sf.run_id
 FROM ias_sales 
-INNER JOIN {{ ref('sale.flag') }} sf ON ias_sales.instruno_clean = sf.meta_sale_document_num;
+INNER JOIN {{ source('sale', 'flag') }} sf ON ias_sales.instruno_clean = sf.meta_sale_document_num;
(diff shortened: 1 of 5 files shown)


wagnerlmichael:
Per @dfsnow's request, I'm attaching summary stats in this issue. The following shows **raw counts** for the total number of sales that are touched by the flagging scripts and those that are not. Within the sales touched by the flagging script, we look at total # of non-outliers and total # of outliers, per year. We also include percentage figures. These flags are based on the current `.yaml` configurations in `model-sales-val`. @dfsnow let me know if I can include any other information.

The `iasworld.sales` table is filtered as follows prior to the the left join:

``` 
WITH ias_sales AS (
    SELECT 
        salekey,
        saledt,
        NULLIF(REPLACE(instruno, 'D', ''), '') AS instruno_clean
    FROM "iasworld"."sales"
    WHERE cur = 'Y'
    AND deactivat IS NULL
)
```

#### Raw Numbers

|sale_year|not_touched_by_flagging|touched_by_flagging|is_outlier|is_not_outlier|
|---------|-----------------------|-------------------|----------|--------------|
|2014     |39459                  |62582              |3678      |58904         |
|2015     |43868                  |59499              |3453      |56046         |
|2016     |65178                  |68864              |3783      |65081         |
|2017     |63618                  |70037              |3568      |66469         |
|2018     |63387                  |72053              |3325      |68728         |
|2019     |79488                  |68511              |2978      |65533         |
|2020     |71818                  |68736              |2483      |66253         |
|2021     |80182                  |75600              |2309      |73291         |
|2022     |78747                  |70439              |2537      |67902         |
|2023     |39147                  |35434              |797       |34637         |

#### Same Data, Percentages

|sale_year|pct_not_touched_by_flagging|pct_touched_by_flagging|pct_is_outlier|pct_is_not_outlier|
|---------|---------------------------|-----------------------|--------------|------------------|
|2014     |38.67                      |61.33                  |5.88          |94.12             |
|2015     |42.44                      |57.56                  |5.8           |94.2              |
|2016     |48.63                      |51.37                  |5.49          |94.51             |
|2017     |47.6                       |52.4                   |5.09          |94.91             |
|2018     |46.8                       |53.2                   |4.61          |95.39             |
|2019     |53.71                      |46.29                  |4.35          |95.65             |
|2020     |51.1                       |48.9                   |3.61          |96.39             |
|2021     |51.47                      |48.53                  |3.05          |96.95             |
|2022     |52.78                      |47.22                  |3.6           |96.4              |
|2023     |52.49                      |47.51                  |2.25          |97.75             |

aws-athena/views/sale-vw_ias_salesval_upload.sql
```diff
@@ -0,0 +1,16 @@
+WITH ias_sales AS (
+    SELECT 
+        salekey,
+        NULLIF(REPLACE(instruno, 'D', ''), '') AS instruno_clean
+    FROM {{'iasworld', 'sales'}}
```
jeancochrane, line 5: **[Nitpick]** I think we'll need to wrap these strings in a [`source()` call](https://docs.getdbt.com/docs/build/sources) here to instruct dbt to properly reference the table! The [workflow](https://github.com/ccao-data/data-architecture/actions/runs/6384372564/job/17326897801?pr=147) isn't failing for this specific reason, but I expect if we resolved the error that the workflow run is reporting we would see another error get raised for this line.

```suggestion
    FROM {{ source('iasworld', 'sales') }}
```

aws-athena/views/sale-vw_ias_salesval_upload.sql
```diff
@@ -0,0 +1,16 @@
...
+    AND deactivat IS NULL
+)
+
+SELECT 
+    ias_sales.salekey,
+    sf.sv_is_outlier,
+    sf.sv_outlier_type,
+    sf.run_id
+FROM ias_sales 
+INNER JOIN {{ ref('sale.flag') }} sf ON ias_sales.instruno_clean = sf.meta_sale_document_num;
```
jeancochrane, line 16: The [`build-and-test-dbt` workflow](https://github.com/ccao-data/data-architecture/actions/runs/6384372564/job/17326897801?pr=147) is complaining that the `sale.flag` model does not exist:

```
Compilation Error
Model 'model.athena.sale.vw_ias_salesval_upload' (models/sale/sale.vw_ias_salesval_upload.sql) depends
  on a node named 'sale.flag' which was not found
```

My guess is that `sale.flag` is generated by a Glue job and not by any transformations stored in dbt, is that right? If so, we'll need to treat it as a source (i.e. a table created outside the DAG) instead of a ref (i.e. a table created inside the DAG).

To make this a source, we first need to adjust the call here:

```suggestion
INNER JOIN {{ source('sale', 'flag') }} sf ON ias_sales.instruno_clean = sf.meta_sale_document_num;
```

Then, we'll need to add a `source` configuration to the `models/sale/schema.yml` file below, which I will sketch out in a separate comment.

dbt/models/sale/schema.yml
```diff
@@ -0,0 +1,8 @@
+models:
```
jeancochrane, line 1: **[Nitpick]** One quick change to support treating `sale.flag` as a source:

```suggestion
sources:
  - name: sale
    tables:
      - name: flag
      
models:
```

jeancochrane (changes requested):
Looking good so far, great job figuring out how to add a new view! A couple of nitpicks below, which should help resolve the [failures](https://github.com/ccao-data/data-architecture/actions/runs/6384372564/job/17326897801?pr=147#step:6:29) in the `build-and-test-dbt` workflow. It seems like `pre-commit` [is also failing](https://github.com/ccao-data/data-architecture/actions/runs/6384372574/job/17326897751?pr=147), although the output is less helpful there; try running it locally to see if the output is any clearer, and if not we can pair on it!

dbt/models/sale/schema.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  - name: sale
+    tables:
+      - name: flag
```
jeancochrane, line 4: **[Nitpick]** I totally forgot, but there's one last piece here! We need to add a model entry for `sale.vw_ias_salesval_upload` so that it shows up in our docs:

```suggestion
sources:
  - name: sale
    tables:
      - name: flag
      
models:
  - name: sale.vw_ias_salesval_upload
    description: '{{ doc("vw_ias_salesval_upload") }}'
```

dfsnow (approved):
Looks great @wagnerlmichael. We'll eventually need to update this to take only not-yet-uploaded flags (by inner joining `iasworld.sales`), but don't worry about that quite yet.

dbt/models/sale/docs.md
```diff
@@ -0,0 +1,3 @@
+{% docs vw_ias_salesval_upload.sql %}
```
jeancochrane, line 1: **[Nitpick, non-blocking]** I missed this on the the first go-round, but to make this match our existing docs we should remove the `.sql` prefix! Here's an example of another docs block:

https://github.com/ccao-data/data-architecture/blob/ba9eadfed62f62f24dcbe73b8ccb903a14d6bae1/dbt/models/default/docs.md?plain=1#L40

Not entirely necessary, but maybe good for consistency?

```suggestion
{% docs vw_ias_salesval_upload %}
```

jeancochrane (changes requested):
A couple tiny changes for things I missed the first time, my bad!

jeancochrane (approved):
Good to go!

PR #154, opened 2023-10-05
Lower reporting minimum group sales threshold

wrridgeway:
Lowering this limit allows us to report on geographies with few sales and as long as it's above 2 doesn't throw an error in glue.


dfsnow (commented):
**question:** We can't actually calculate any ratio stats for groups with the few sales, so what stats _does_ this actually calculate?

dfsnow (approved):

PR #155, opened 2023-10-10
Fix links in dbt README

jeancochrane:
This PR fixes the header links in the dbt README, which seem to be broken due to manual `<h2>` tags with `id` attributes not producing reliable links like I thought they did. I'm not sure if there has been a recent change to the GitHub rendering engine to make this trick not work anymore, or if it's just never worked the way I expect; but in this PR we remove the reliance on the `<h2>` tags and instead fall back to GitHub's autogenerated anchors. These anchors are ugly, and they have the downside of requiring that they be updated whenever header text changes, but they work nonetheless.

The changes in this PR can be tested by viewing the README file in this branch and clicking the links in the `Quick links` section: https://github.com/ccao-data/data-architecture/blob/jeancochrane/fix-links-in-dbt-readme/dbt/README.md

### DIFF
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -7,11 +7,11 @@ This directory stores the configuration for building our data catalog using
 
 ### In this document
 
-* [🖼️ Background: What does the data catalog do?](#background-what-does-the-data-catalog-do)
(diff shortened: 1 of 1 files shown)


dfsnow (approved):
Looks good to me! Thanks for the quick fix @jeancochrane.

Issue #115, opened 2023-09-05
Create a script to compare row counts on the Open Data portal to Athena


wrridgeway:
Our historical values asset completed its automated refresh on Sept 1, 2023, but for some reason only had ~9 million rather than ~ 40 million rows. We were not notified of a failure by email. This needs investigating.

## UPDATE 10/5/2023 (from Dan)

We're hearing from some folks that the Socrata agent may be failing silently in some cases, leading to the Open Data portal being incomplete or not up-to-date. Since the Open Data portal is a critical asset, we should build checks to guard against such failures.

I propose that we create a short Python script that pulls the row count by year from each Open Data asset and compares it to the count of the source Athena view. Row counts can be gathered directly from the Socrata SoQL API, e.g.:

```bash
curl 'https://datacatalog.cookcountyil.gov/resource/vgzx-68gb.json?$query=SELECT%20COUNT(*),year%20GROUP%20BY%20year' | jq
```

We should run the script on a schedule using GitHub Actions, and use the same notification setup used in [`test-dbt-models`](https://github.com/ccao-data/data-architecture/blob/master/.github/workflows/test_dbt_models.yaml).

Since current year row counts may not match the Open Data Portal exactly, we should use a "fuzzy" match for the most recent 2 years (+/- 2%) and an exact match for all years prior to that.

wrridgeway:
uploaded w/o issue on oct 1st, just FYI

dfsnow:
@jeancochrane @wrridgeway See updated issue description and title.

PR #156, opened 2023-10-11
Add script and GitHub workflow to test open data assets on a schedule

jeancochrane:
This PR adds a new scheduled workflow `test-open-data-assets` that runs a Python script once a month to confirm that the row counts of our open data assets match the row counts of the views that produce them. The goal here is to create a system that can alert us if the Socrata agent is silently failing to publish rows for some data sets, which is behavior we've experienced in the past.

In order to implement this alerting system, this PR makes two major refactors:

* The logic that the `test-dbt-models` workflow uses to push failure notifications to SNS is factored out into a composite action, `publish_sns_topic`, so that it can be reused by the new workflow that runs open data tests.
* [Exposures](https://docs.getdbt.com/docs/build/exposures) are added to the dbt DAG for all of our open data assets except for [Neighborhood Boundaries](https://datacatalog.cookcountyil.gov/Property-Taxation/Assessor-Neighborhood-Boundaries/pcdw-pxtg) (which doesn't have a `year` field, so I'm unsure if it should be tested). These exposure configurations allow the Python script to programmatically parse the metadata for our open data assets and compare it to the dbt models that produced it.

Note that my [test workflow run](https://github.com/ccao-data/data-architecture/actions/runs/6487851859/job/17619059598) did not actually pass, but we agreed in an out-of-band discussion that the failures are most likely due to the fact that data was last refreshed nearly two weeks ago, so we're going to wait till these tests run on 11/3 to see if we need to make any further adjustments.

Closes https://github.com/ccao-data/data-architecture/issues/115.

### DIFF
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -28,7 +28,9 @@ ASSET_API_QUERY_PARAMS = {"$query": "SELECT COUNT(*),year GROUP BY year"}
 DBT = dbtRunner()
 
 # Data for the most recent two years are permitted to be slightly different,
-# according to this buffer value
+# according to this buffer value. This is because we expect some level of
+# change in the Athena data (which is updated daily) compared to the Open Data
+# portal (which is updated bi-weekly or monthly).
 BUFFER = 0.02
 
 
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -18,7 +18,7 @@ DBT = dbtRunner()
 
 # Data for the most recent two years are permitted to be slightly different,
 # according to this buffer value
-BUFFER = 0.2
+BUFFER = 0.02
 
 
 def main() -> None:
@@ -93,7 +93,9 @@ def main() -> None:
             # they're strings (as in our data). This is super annoying but
             # it's still the fastest way we know to print a clean table
             # in Python :\
-            diff_table.print_table(max_rows=None)
+            diff_table.print_table(
+                max_rows=None, max_columns=None, max_column_width=None
+            )
         print()
         raise ValueError("Open data asset test failed")
 
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -4,10 +4,10 @@ import contextlib
 import datetime
 import io
 import json
-import pprint
 import sys
 import typing
 
+import agate
 import requests
 from dbt.cli.main import dbtRunner
 
@@ -27,7 +27,7 @@ def main() -> None:
     with open(dbt_manifest_path) as dbt_manifest_file:
         dbt_manifest_json = json.loads(dbt_manifest_file.read())
 
-    errors: typing.List[typing.Dict] = []
+    diffs: typing.List[typing.List[typing.Dict]] = []
 
     for exposure in dbt_manifest_json["exposures"].values():
         asset_name = exposure["label"]
@@ -62,7 +62,7 @@ def main() -> None:
                     "--args",
                     f"{{model: '{source_model_name}', print: true}}",
                     "--target",
-                    "prod"
+                    "prod",
                 ]
             )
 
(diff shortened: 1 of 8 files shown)


.github/scripts/test_open_data_assets.py
```diff
@@ -0,0 +1,212 @@
...
+        asset_url = exposure["url"]
+        asset_id = asset_url.split("/")[-1]
+
+        # For now, assume just one dependency table, since all of our open
+        # data assets currently have a 1:1 relationship with a view.
+        fq_source_model_name = exposure["depends_on"]["nodes"][0]
+
+        # Dependency tables in the DAG are fully qualified, so we have to
+        # strip the prefix to get the schema and tablename.
+        source_model_name = ".".join(fq_source_model_name.split(".")[2:])
```
jeancochrane, line 54: Taken together, you can see that this test requires we make a lot of assumptions about the formatting of our exposures. Another big one is that we're assuming that every exposure needs to be tested, which may or may not be the case in the future. I think this is probably not a big deal given that we can always use the `meta` configuration for our exposures to set more fine-grained configs if in the future we decide we want more flexibility in this system.

.github/scripts/test_open_data_assets.py
```diff
@@ -0,0 +1,212 @@
...
+import typing
+
+import agate
+import requests
+from dbt.cli.main import dbtRunner
+
+# API URL and params that we can use to grab counts by year for each asset
+ASSET_API_URL = "https://datacatalog.cookcountyil.gov/resource/{asset_id}.json"
+ASSET_API_QUERY_PARAMS = {"$query": "SELECT COUNT(*),year GROUP BY year"}
+DBT = dbtRunner()
```
jeancochrane, line 28: TIL that dbt has an [interface](https://docs.getdbt.com/reference/programmatic-invocations) for running CLI commands from Python scripts!

.github/scripts/test_open_data_assets.py
```diff
@@ -0,0 +1,212 @@
...
+        asset_api_url = ASSET_API_URL.format(asset_id=asset_id)
+        response = requests.get(asset_api_url, params=ASSET_API_QUERY_PARAMS)
+        if not response.ok:
+            print("Encountered error in request to asset endpoint")
+            response.raise_for_status()
+
+        asset_row_counts_by_year = response.json()
+
+        dbt_output = io.StringIO()
+        with contextlib.redirect_stdout(dbt_output):
```
jeancochrane, line 67: This is not a super clean way of returning data from the underlying query, but as far as I can tell, there is not yet a way of returning rows to a macro invoked via `dbtRunner.invoke()`, so we have to rely on printing to stdout and then capturing that output.

.github/scripts/test_open_data_assets.py
```diff
@@ -0,0 +1,212 @@
...
+            "The following view/asset pairs had mismatching row counts "
+            "by year:"
+        )
+        for diff in diffs:
+            print()
+            diff_table = agate.Table.from_object(diff)
+            # Note that agate adds thousands separators to years, even when
+            # they're strings (as in our data). This is super annoying but
+            # it's still the fastest way we know to print a clean table
+            # in Python :\
```
jeancochrane, line 106: Any ideas about better ways to do this? Agate is a dependency of dbt so it was the easy option, but I'm sure there's another way!

.github/workflows/test_open_data_assets.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup dbt
+        uses: ./.github/actions/setup_dbt
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+
+      - name: Install Python requirements
+        run: pip install requests
```
jeancochrane, line 30: It would probably be better practice to install from a requirements file, but given that there's only one dependency and we're using a super minimal and stable version of its API (`requests.get()`), I figure an inline install is probably good enough for now.

dbt/models/default/schema.yml
```diff
@@ -706,3 +706,130 @@ models:
       # TODO: Data completeness correlates with availability of spatial data
       # by year
   - name: default.vw_pin_exempt
+
+exposures:
```
jeancochrane, line 710: See the docs on [exposures](https://docs.getdbt.com/docs/build/exposures) for more details about what these properties mean.

.github/workflows/test_open_data_assets.yaml
dfsnow: **question:** It [looks like](https://github.com/dbt-labs/dbt-core/blob/main/core/setup.py#L91) `requests` is a dependency of `dbt-core`, so shouldn't this be handled by `setup_dbt`?

.github/workflows/test_open_data_assets.yaml
```diff
@@ -0,0 +1,58 @@
...
+  test-open-data-assets:
+    runs-on: ubuntu-latest
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint
+    # so that we can authenticate with AWS
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
dfsnow, line 22: **nitpick:** I think we bumped to `checkout@v4` for other repos.

dbt/macros/row_count_by_year.sql
```diff
@@ -0,0 +1,24 @@
...
+-- If the `print` argument is set to True (default is False), the macro will
+-- print the results of the query to stdout, which allows this macro to be used
+-- by scripts to return data.
+{% macro row_count_by_year(model, year_column="year", ordering="asc", print=False) %}
+    {% set query %}
+        select count(*) as COUNT, {{ year_column }}
+        from {{ model }}
+        group by {{ year_column }}
+        order by {{ year_column }} {{ ordering }}
+    {% endset %}
```
dfsnow, line 17: **suggestion:** It might be worth just making this generalizable i.e. `row_count_by_group()`, since this is a super common pattern that we may want to re-use.

dfsnow (commented):
Awesome work getting this done so fast @jeancochrane. See my comments for minor edits. Neighborhood boundaries need not be tested. They are static and not likely to change any time soon.

.github/workflows/test_open_data_assets.yaml
jeancochrane: Good catch, I removed this block in [dd01ef6](https://github.com/ccao-data/data-architecture/pull/156/commits/dd01ef6ee608379335167b9249f2261bc3b26c49)!

.github/scripts/test_open_data_assets.py
jeancochrane: Makes sense! I added a new exposure attribute `meta.test_row_count` to enable this test in [4dd4714](https://github.com/ccao-data/data-architecture/pull/156/commits/4dd4714624cbca478483b55adf33c457ed160107).

dbt/macros/row_count_by_year.sql
jeancochrane: Good idea, done in [64f056b](https://github.com/ccao-data/data-architecture/pull/156/commits/64f056bd3ec0933468aa59962dfebc997ce8e15f) 👍🏻

.github/workflows/test_open_data_assets.yaml
jeancochrane: Oops, good catch! Done in [cbba969](https://github.com/ccao-data/data-architecture/pull/156/commits/cbba969b8f311c9db3e08b680d51c92a387dd070).

jeancochrane:
@dfsnow I made a bunch of changes based on your comments so I think this is ready for a final pass!

dfsnow (approved):
Looks great @jeancochrane! Nice work.

Issue #123, opened 2023-09-12
Add Python black config to ensure line length formatting matches flake8


jeancochrane:
The default max line length for the `black` autoformatter is [88 characters](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length), but the default max line length for the `flake8` linter is [79 characters](https://flake8.pycqa.org/en/2.5.5/config.html#settings). This can pose a problem for Python files in this repo, since [Super-Linter](https://github.com/super-linter/super-linter) uses both `flake8` and `black`, and if you want to run `black` locally to clean up errors that are reported by the linter it can sometimes extend lines beyond 79 characters in a way that interferes with `flake8`.

We should resolve this discrepancy by adding a `black` config file to the repo that we can use on both local and CI environments. The steps here are:

1. Create a new file `aws-glue/pyproject.toml` with the following contents:

```toml
[tool.black] line-length = 88
```

2. Add a new environment variable to the [Super-Linter workflow configuration](https://github.com/ccao-data/data-architecture/blob/ed1f35dffc94c93e88905d9d8b33476deb204f17/.github/workflows/lint.yaml#L26-L35) to point `black` to the new config file:

```diff
diff --git a/.github/workflows/lint.yaml b/.github/workflows/lint.yaml
index 0d44318..afa1be9 100644
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -33,3 +33,4 @@ jobs:
           VALIDATE_MARKDOWN: false
           LINTER_RULES_PATH: /
           YAML_CONFIG_FILE: .yamllint.yml
+          PYTHON_BLACK_CONFIG_FILE: aws-glue/pyproject.toml
```

3. Create a new test python file like `test.py` in the `aws-glue` subdirectory with a line >79 characters
4. Run `black` locally in the context of the `aws-glue` subdirectory and confirm it reformats `test.py`
5. Revert `test.py` so that it once again has >79 characters
6. Push `test.py` and confirm that Super-Linter fails for both `flake8` and `black`
7. Remove `test.py`, push changes, and open a pull request

jeancochrane:
@Damonamajor is handling this as part of https://github.com/ccao-data/data-architecture/pull/119, so we can close once that comes in 👍

PR #157, opened 2023-10-12
Move Python autoformatter and linter configs from SuperLinter to pre-commit

jeancochrane:
SuperLinter is currently configured to use a number of Python linters and autoformatters, but since these have not been added to our pre-commit config yet, resolving the errors locally can be annoying. This PR moves the following tools from SuperLinter to our pre-commit config so that developers editing Python code can more easily lint and format their code locally:

* [Black](https://github.com/psf/black)
* [flake8](https://flake8.pycqa.org/en/latest/)
* [isort](https://pycqa.github.io/isort/)

To me, this PR raises a bigger-picture question: When is it appropriate to move a tool from SuperLinter to pre-commit? My thinking is that SuperLinter has the benefit of minimal configuration, while pre-commit has the benefit of local integration, so a rough heuristic might be that once a lint/format tool starts to cause repeated friction with local development of frequently-touched files we should move it into pre-commit. I'm open to pushback on that heuristic thought!

Closes https://github.com/ccao-data/data-architecture/issues/123.


aws-glue/jobs/reporting-ratio_stats.py
```diff
@@ -171,7 +171,6 @@ def ccao_mki(fmv, sale_price):
 
 
 def ccao_cod(ratio):
-
```
jeancochrane, line 174: From here down, the rest of the changes in this PR represent formatting tweaks to appease Black/flake8/isort and thehy should not have any effect on code behavior.

dfsnow (approved):
Looks good to me @jeancochrane. I agree with your heuristic. Whenever we move to doing lots of local development with a lintr/formatter, let's build it into pre-commit.

PR #162, opened 2023-10-17
161 add sales val tables to dbt schema file

### LINKED ISSUE #161: Add sales val tables to dbt schema file
We need to add the parameter, metadata, and group_means tables to the dbt schema tile before updating the sales view to incorporate outlier flags.

wagnerlmichael:
Adding three extra sources from the sales val program.

Closes #161.


jeancochrane (approved):
Good to go, thanks for this!

PR #163, opened 2023-10-17
Speed up distance queries using an expanding buffer

jeancochrane:
This PR refactors the distance queries in the `proximity` schema to use an expanding buffer approach, similar to what we implemented in https://github.com/ccao-data/data-architecture/pull/138.

In testing, this approach reduced the execution time of our most expensive query (`dist_pin_to_water`) from roughly 2 hours to roughly 15 minutes.


jeancochrane:
Closing since I wasn't able to get this to work in a reasonable amount of time -- see here for more details: https://github.com/ccao-data/data-architecture/issues/130#issuecomment-1769443401

jeancochrane:
OK, never mind 😅

Issue #113, opened 2023-09-01
Document and test DBT models


jeancochrane:
@dfsnow is going to take a pass at adding tests and docs for DBT models.

PR #164, opened 2023-10-17
Add iasWorld column descriptions to dbt docs

dfsnow:
This PR adds column descriptions to (almost) all iasWorld columns in our dbt documentation. The descriptions were pulled from the system tables of the iasWorld DB, then cleaned, filtered, and amended where appropriate.

Descriptions/columns that are shared across many tables are filtered out into doc blocks, which are stored in a model-level `columns.md` file.

Related to #113.


dbt/models/iasworld/columns.md
```diff
@@ -0,0 +1,983 @@
+## acres
```
dfsnow, line 1: Headers are added here in order to support folding in editors, as well as to more easily differentiate blocks if they become multiple lines long.

dbt/models/iasworld/columns.md
dfsnow, line 1: Columns that are shared across 3 or more tables (and have the same description across tables) are stored here. All columns referenced from this file are prefixed with `column_` in there jinja call to differentiate them from more general docs.

dbt/models/iasworld/schema.yml
```diff
@@ -6,8 +6,251 @@ sources:
...
+            description: Date of last subscription server extract
+          - name: taxdist_req
+            description: Flag to indicate if `LEGDAT.TAXDIST` is a required field
+          - name: thisyear
+            description: Default year for users
+          - name: wen
+            description: '{{ doc("column_wen") }}'
+          - name: who
+            description: '{{ doc("column_who") }}'
+
```
dfsnow, line 88: I added spaces between models here in order to assist with navigation. Planning to do the same for other large models with many column/test definitions.

dbt/models/iasworld/columns.md
jeancochrane: **[Praise]** Good thinking, I threw together a quick PR (https://github.com/ccao-data/data-architecture/pull/165) to codify this pattern in the docs!

dbt/models/iasworld/schema.yml
jeancochrane: **[Suggestion, non-blocking]** Makes sense to me! since 4k lines is a length at which a file can become difficult to navigate in some IDEs (and also in the GitHub UI), we might also consider taking advantage of the fact that [schemas can be loaded from one or more arbitrarily-named YAML files](https://docs.getdbt.com/reference/configs-and-properties#where-can-i-define-properties) and move toward a pattern where each iasWorld table gets its own `<table>.yaml` file, perhaps in a dedicated subdirectory like `dbt/models/iasworld/schema/`. I'll leave this decision up to you, since I'm expecting the shape of these files will change even further once we start adding tests.

jeancochrane (approved):
Very impressed that you did this so fast 👏🏻 I mostly skimmed this to confirm that the patterns look good, as long as the documentation compiles I'm happy!

dbt/models/iasworld/schema.yml
dfsnow: I like that design for the large/unwieldy sources like `iasworld`. I'll open a separate PR for the split!

PR #165, opened 2023-10-17
Update guidance on defining model descriptions

jeancochrane:
In https://github.com/ccao-data/data-architecture/pull/164, we're moving to a new pattern where we only use doc blocks for descriptions if they're reused by >2 entities. This PR updates our guidance to ensure that this pattern is documented.

PR #166, opened 2023-10-17
Reorganize dbt schema files

dfsnow:
This PR splits our dbt `schema.yml` files for certain models/sources into separate `schema/$NAME.yml` files. The intention here is to make very large schema files a bit less unwieldy.

It also drops the top-level `version: 2` key from our model definitions, as it is [optional as of dbt 1.5](https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.5#quick-hits).

Apologies for the truly horrifying diff, but rest assured that basically everything here is a structural/location change other than https://github.com/ccao-data/data-architecture/commit/4c15ff7c9353da8317ddbd99805085e2dbed0723 (dropping the `version` key).

Note that I'll separately update the README in #165 to reflect these changes.


dbt/dbt_project.yml
```diff
@@ -1,6 +1,5 @@
 name: 'athena'
 version: '0.1.0'
-config-version: 2
```
dfsnow, line 3: This option and the `version: 2` key are [optional as of dbt 1.5](https://docs.getdbt.com/guides/migration/versions/upgrading-to-v1.5#quick-hits). Exclude 4c15ff7 to see only schema structure changes.

dbt/models/default/exposures.yml
dfsnow, line 1: I put exposures in their own separate yaml file. Open to any ideas on where the best place is for these to live!

dbt/models/default/schema/default.vw_card_res_char.yml
dfsnow, line 1: What do we think about prefixing with the schema here AND with sources (i.e. `iasworld.sales.yml` below)? I think it makes sense for consistency between files but it's a little weird since sources don't have the schema in their actual node name.

dbt/models/iasworld/schema/iasworld.aasysjur.yml
```diff
@@ -0,0 +1,85 @@
+sources:
+  - name: iasworld
+    loaded_at_field: date_parse(wen, '%Y-%m-%d %H:%i:%s.0')
```
dfsnow, line 3: One downside of this approach is that top-level keys such as `loaded_at_field` and `tags` need to be replicated across each file.

jeancochrane: **[Praise]** That doesn't seem to bad to me! If we have to choose between duplicating these fields and requiring that our teammates navigate 5k+ lines of code to find them, I think duplication is probably the better choice anyway.

jeancochrane (approved):
I think this file structure will be much easier to navigate, thanks for this!

PR #167, opened 2023-10-18
Add dbt tags for sources

dfsnow:


dbt/models/ccao/schema.yml
```diff
@@ -1,11 +1,32 @@
 sources:
   - name: ccao
+    tags:
+      - load_manual
     tables:
       - name: class_dict
+
```
dfsnow, line 7: Pre-adding spaces here between resources since they'll soon have additional documentation and keys.

dfsnow:
Looking for reviews from both @jeancochrane (implementation) and @wrridgeway (domain knowledge) on this one.

jeancochrane (approved):
Implementation looks great to me! Should we go ahead and add your description of the tags to dbt/README.md?

wrridgeway:
I think you can add iasworld.oby to type_condo? it's not just class 299s, but for our purposes it's used exclusively for condos.

dfsnow:
> I think you can add iasworld.oby to type_condo? it's not just class 299s, but for our purposes it's used exclusively for condos.

Done!

PR #173, opened 2023-10-19
Fix proration typo

wrridgeway:


dfsnow (approved):
Thanks for the quick turnaround @wrridgeway. Can you check/update any open data assets that depend on these?

PR #178, opened 2023-10-23
Correct nopar calculation

wrridgeway:
Because there can be duplicates within pin and document number, we need to make the calculation of number of parcels associated with a sale invoke DISTINCT

### DIFF
--- a/aws-athena/views/default-vw_pin_sale.sql
+++ b/aws-athena/views/default-vw_pin_sale.sql
@@ -76,12 +76,14 @@ unique_sales AS (
             -- We remove the letter 'D' that trails some document numbers in
             -- iasworld.sales since it prevents us from joining to mydec sales.
             -- This creates one instance where we have duplicate document
-            -- numbers, so we sort by the original document number within the
+            -- numbers, so we sort by sale date (specifically to avoid conflicts
+            -- with detecting the easliest duplicate sale when there are
+            -- multiple within one document number, within a year) within the
             -- new doument number to identify and remove the sale causing the
             -- duplicate document number.
             ROW_NUMBER() OVER (
                 PARTITION BY NULLIF(REPLACE(sales.instruno, 'D', ''), '')
-                ORDER BY sales.instruno
+                ORDER BY sales.saledt
             ) AS bad_doc_no,
             -- Some pins sell for the exact same price a few months after
             -- they're sold. These sales are unecessary for modeling and may be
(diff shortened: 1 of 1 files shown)


dfsnow (commented):
@wrridgeway How does this change the output of `vw_pin_sale`, if at all? Can you describe the results of a diff?

wrridgeway:
For example, previously, this could have been considered a multisale:

```
select
  parid,
  instruno,
  saledt
from iasworld.sales
where instruno in ('1834733034')
  and cur = 'Y'
  and deactivat is null
```
parid | instruno | saledt
-- | -- | --
24254150171010 | 1834733034 | 11/1/2018
24254150171010 | 1834733034 | 11/28/2018
24254150171010 | 1834733034 | 11/28/2018
24254150171010 | 1834733034 | 11/28/2018
24254150171010 | 1834733034 | 11/28/2018
24254150171010 | 1834733034 | 11/28/2018

Multisales are treated a bit more generously when it comes to de-duplication, so we lose a few observations. I have never really looked into multisales stuff too much since we don't use them for anything, but they're gonna make a difference when it comes to replicating the old view with an unfiltered view + legacy column filters. Here are year-by-year diffs:

year | old | new | diff
-- | -- | -- | --
2023 | 55364 | 55364 | 0
2022 | 98239 | 98239 | 0
2021 | 104000 | 104000 | 0
2020 | 91225 | 91225 | 0
2019 | 92103 | 92089 | -14
2018 | 97157 | 97153 | -4
2017 | 95673 | 95672 | -1
2016 | 97973 | 97973 | 0
2015 | 82148 | 82148 | 0
2014 | 88741 | 88740 | -1
2013 | 92508 | 92508 | 0
2012 | 72989 | 72988 | -1
2011 | 57937 | 57918 | -19
2010 | 58500 | 58499 | -1
2009 | 59730 | 59730 | 0
2008 | 61607 | 61607 | 0
2007 | 86767 | 86767 | 0
2006 | 110040 | 110040 | 0
2005 | 132874 | 132874 | 0
2004 | 129456 | 129456 | 0
2003 | 120105 | 120105 | 0
2002 | 114439 | 114439 | 0
2001 | 105056 | 105056 | 0
2000 | 102957 | 102957 | 0
1999 | 806 | 806 | 0
1973 | 1 | 1 | 0

we still maintain uniqueness by doc_no outside of multisales:

```
docs <- dbGetQuery(
  conn = AWS_ATHENA_CONN_NOCTUA,
  "select doc_no
from \"ci_177-correct-nopar-calculation-in-vw-pin-sale_default\".vw_pin_sale
where is_multisale = FALSE"
)

> table(duplicated(docs$doc_no))

  FALSE 
1755441
```

dfsnow (approved):
@wrridgeway Got it, I think I understand. Regardless, I'm comfortable with that level of change. Probably saves me from debugging some dbt tests too.

PR #180, opened 2023-10-24
Address randomness caused by athena tiebreak in bad_doc_no

wrridgeway:
The new `bad_doc_no` column was not sorted enough to avoid tiebreakers and was causing different row counts each time the view was queried.


dfsnow (approved):
**praise:** This kind of stuff is such a pain to debug. Nice catch.

PR #183, opened 2023-10-25
Revert "Remove pandas index from glue job output"

wrridgeway:
Reverts ccao-data/data-architecture#160


dfsnow:
Note that this is needed due to Tableau freaking out about the removal of an arbitrary and unused column. We will revert this reversion after reporting for the year is finished.

dfsnow (approved):

ccao-jardine:
This reversion worked nicely, thanks team for fixing this so responsively! 

This cycle, let's refrain from updates to any production views and tables in the reporting db until November 10 at the earliest. 

In general, for the future, please notify the Director of Data Analytics (currently vacant, currently me 😅) of upcoming changes to production assets in the reporting db well in advance so as to not disrupt any of these more fragile workflows.

PR #187, opened 2023-10-25
Create legacy view, point reporting at legacy view

wrridgeway:
We want to push out some pretty big changes to vw_pin_sale, but not disrupt reporting. Eventually reporting will incorporate the changes to vw_pin_sale, but for he next few weeks if we point reporting views at a legacy version of vw_pin_sale we can move forward with our changes and keep reporting consistent for the tri. Not a permanent solution.


dfsnow (approved):
Looks good @wrridgeway. @ccao-jardine we should be set to finish out reporting using this copy of the view. We'll delete it once reporting is done for the year.

ccao-jardine:
Sounds good, thanks!

Issue #107, opened 2023-08-31
Fix hard boundaries in airport noise surface


dfsnow:
Previously, the Data Department created an airport noise feature based on kriging the noise values recorded from ground-based sensors. This approach created a continuous surface showing airport noise, with a fallback value for PINs not covered by the surface. The "surface" covers most of Chicago and Cook, but purposefully does not cover _all_ of Cook (it was limited due to computational expense).

However, per @mbjackson-capp, this implementation may have some strange effects, as there is a discontinuity in the noise values when moving from the edge of the surface to the fallback value.

We should either:

- Update the airport noise feature to use a kriging surface that covers all parcels in Cook County (no need for a fallback value)
- Replace the feature with a simpler model based on the physical falloff of noise ` 1 / r^2 ` (inverse square)

stevevance:
The Chicago Department of Aviation has a program to replace windows in houses affected by noise. I presume they have geographic standards as to which houses are eligible. Would this be an alternative to using the "surface"?

PR #193, opened 2023-10-26
Upload point-source airport noise model

mbjackson-capp:
Create model to replace prior model based on kriging surface, modeling sound intensity from each airport as a point source at that airport's centroid and calculating expected DNL based on physical law of inverse-square falloff.

#107 
[point source airport sound model summary.pdf](https://github.com/ccao-data/data-architecture/files/13183060/point.source.airport.sound.model.summary.pdf)

### DIFF
--- a/aws-s3/scripts-ccao-data-raw-us-east-1/spatial-point_source_air_noise.R
+++ b/aws-s3/scripts-ccao-data-raw-us-east-1/spatial-point_source_air_noise.R
@@ -1,0 +1,168 @@
+# Create model of point-source inverse-square airport noise falloff
+
+# Matt Jackson, October 2023
+
+library(arrow)
+library(aws.s3)
+library(ccao)
+library(DBI)
+library(geoarrow)
+library(ggplot2)
+library(ggpubr)
+library(glue)
+library(gstat)
+library(here)
+library(noctua)
+library(stars)
+library(sf)
+library(sp)
+# "The legacy packages maptools, rgdal, and rgeos, underpinning the sp package,
+# which was just loaded, will retire in October 2023.
+# Please refer to R-spatial evolution reports for details, especially
+# https://r-spatial.org/r/2023/05/15/evolution4.html.
+# It may be desirable to make the sf package available;
+# package maintainers should consider adding sf to Suggests:.
+# The sp package is now running under evolution status 2
+# (status 2 uses the sf package in place of rgdal)"
+library(tidycensus)
+library(tidyverse)
+library(tmap)
+# "Breaking News: tmap 3.x is retiring. Please test v4, e.g. with
+# remotes::install_github('r-tmap/tmap')"
+library(units)
+library(viridis)
+
+## READ DATA ===================================================================
+
+# Optionally enable query caching
+noctua_options(cache_size = 10)
+
+# Establish connection
+AWS_ATHENA_CONN_NOCTUA <- dbConnect(noctua::athena())
+print("noctua connection established")
+
+midway <- read_parquet(
+  "s3://ccao-data-warehouse-us-east-1/spatial/environment/midway_noise_monitor/midway_noise_monitor.parquet"
+  )
+
+ohare_noise <- dbGetQuery(
+  conn = AWS_ATHENA_CONN_NOCTUA,
+  "SELECT site, year, noise, modeled_omp_build_out_values,
+  ST_GeomFromBinary(geometry_3435) as geometry
+  FROM spatial.ohare_noise_monitor;"
+)
+
+ohare_noise$geometry <- st_as_sfc(ohare_noise$geometry, crs = 3435)
+ohare_noise <- st_as_sf(ohare_noise)
+
+## CLEAN DATA ==================================================================
+
+ohare_noise <- mutate(ohare_noise, airport = "ohare")
+
+midway_noise <- midway %>%
+  rename("site" = "location", "noise" = "avg_noise_level") %>%
+  select(-address) %>%
+  mutate(modeled_omp_build_out_values = NA,
+         airport = "midway") %>%
+  select(site, year, noise, modeled_omp_build_out_values, airport, geometry) %>%
+  mutate(noise = as.numeric(noise),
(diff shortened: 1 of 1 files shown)


aws-s3/scripts-ccao-data-raw-us-east-1/spatial-point_source_air_noise.R
```diff
@@ -0,0 +1,168 @@
...
+
+## PREDICT AIRPORT NOISE LEVELS ================================================
+
+# Set intercept at 0, i.e., when linear distance from airport is infinite,
+# the absolute intensity of sound emanating from it will be 0
+# (Decibel measure will NOT drop to 0; it can keep going to -Inf since 0 dB is 
+# just a a cutoff of human *perception* of sound)
+
+ohare_model_intensity <- lm(added_intensity ~ 0 + isq_dist_ohare, data = airport_clean %>% filter(airport == "ohare")) 
+# Adjusted R-squared:  0.4002
```
dfsnow, line 150: **question:** This regression is minimizing squared errors for the average of _all_ noise readings, 2011 - 2019, correct? 

If so, I have two thoughts/suggestions:

1. Make sure you're using the OMP (O'Hare Modernization Project) values for O'Hare, as those are the "current" set of monitor results.
2. I expect that the noise intensity has changed over time. Perhaps we should include `year` as an independent variable in the regression instead of using the average. This complicates the modeling and feature construction a little bit but would probably produce a better fit.

aws-s3/scripts-ccao-data-raw-us-east-1/spatial-point_source_air_noise.R
```diff
@@ -0,0 +1,168 @@
...
+# package maintainers should consider adding sf to Suggests:.
+# The sp package is now running under evolution status 2
+# (status 2 uses the sf package in place of rgdal)"
+library(tidycensus)
+library(tidyverse)
+library(tmap)
+# "Breaking News: tmap 3.x is retiring. Please test v4, e.g. with
+# remotes::install_github('r-tmap/tmap')"
+library(units)
+library(viridis)
```
dfsnow, line 33: **suggestion:** Let's drop any dependencies not used directly by the script.
```suggestion
library(arrow)
library(ccao)
library(DBI)
library(geoarrow)
library(glue)
library(noctua)
library(sf)
library(sp)
library(dplyr)
library(units)
```

aws-s3/scripts-ccao-data-raw-us-east-1/spatial-point_source_air_noise.R
```diff
@@ -0,0 +1,168 @@
...
+
+# Optionally enable query caching
+noctua_options(cache_size = 10)
+
+# Establish connection
+AWS_ATHENA_CONN_NOCTUA <- dbConnect(noctua::athena())
+print("noctua connection established")
+
+midway <- read_parquet(
+  "s3://ccao-data-warehouse-us-east-1/spatial/environment/midway_noise_monitor/midway_noise_monitor.parquet"
```
dfsnow, line 45: **nitpick:** Both Midway and O'Hare have monitor tables in Athena, let's just use the Athena method of grabbing them for both datasets.

aws-s3/scripts-ccao-data-raw-us-east-1/spatial-point_source_air_noise.R
dfsnow, line 1: **issue:** Let's run this entire script through styler `styler::style_file()` and lintr `lintr::lint()` to get rid of the linting errors and formatting inconsistencies.

dfsnow (changes requested):
Great job overall @mbjackson-capp. This is very clear and concise, particularly the included report. I appreciate the inclusion of graphics and context for those not already familiar with this issue.

Just a few minor tweaks around the edges and some minor questions to resolve. Once they are addressed, request re-review and I'll approve the merge.

aws-s3/scripts-ccao-data-raw-us-east-1/spatial-airport_noise_point_source_new_models.R
```diff
@@ -0,0 +1,357 @@
...
+airport_recent$isq_dist_ohare <- (airport_recent$distance_ohare)^-2
+airport_recent$isq_dist_midway <- (airport_recent$distance_midway)^-2
+
+airport_recent$total_intensity <- decibel_to_intensity(airport_recent$noise)
+airport_recent$added_intensity <- decibel_to_intensity(airport_recent$noise -
+                                                         BASELINE_DNL)
+
+ohare_model1_intensity <- lm(added_intensity ~ 0 + isq_dist_ohare,
+                             data = airport_recent %>%
+                               filter(airport == "ohare"))
```
dfsnow, line 199: **nitpick:** Make the same formatting fix from before for all the `lm()` calls.

aws-s3/scripts-ccao-data-raw-us-east-1/spatial-airport_noise_point_source_new_models.R
```diff
@@ -0,0 +1,357 @@
...
+summary(ohare_model1_intensity)
+# Adjusted R-squared: 0.6612
```
dfsnow, line 201: **thought:** Looks like we have a winner. It also seems most defensible to me to use the latest values from each airport and consider them steady into the future. I say we go with this model. What do you think?

dfsnow: Choosing values from a single year also makes the resulting feature easier to update since we only have to change the resulting model coefficients if we get new monitor values.

dfsnow (commented):
@mbjackson-capp Looks great! Just a few formatting nitpicks and some minor questions to address. Once those are dealt with we can merge this and you can fully move on to featurizing the regression results.

dfsnow (approved):
@mbjackson-capp This looks good to go now! I was able to run the full script and reproduce your results. Nice work on the reporting.

PR #197, opened 2023-10-27
Add model view column definitions

dfsnow:
This PR uses shared column definitions to add documentation to the model input views.


dbt/models/default/schema/default.vw_pin_condo_char.yml
```diff
@@ -7,9 +7,7 @@ models:
 
     columns:
       - name: bldg_is_mixed_use
-        description: |
-          The 10-digit PIN (building) contains a 14-digit PIN that is
-          neither class 299 nor 399.
+        description: '{{ doc("shared_column_char_bldg_is_mixed_use") }}'
```
dfsnow, line 10: All these definitions were moved into `shared_columns.md`.

dbt/models/model/schema.yml
```diff
@@ -55,20 +55,541 @@ sources:
...
+        description: '{{ doc("column_num_school_with_rating_in_half_mile") }}'
+
   - name: model.vw_card_res_input
     config:
       tags:
         - type_res
 
     columns:
-      - name: char_bldg_sf_95_percentile
+      # Columns inherited from vw_pin_shared_input
```
dfsnow, line 388: Unfortunately, there d[oesn't seem to be a great way](https://stackoverflow.com/questions/24090177/how-to-merge-yaml-arrays/30770740#30770740) to merge YAML arrays using only built-in language features, hence this very messy anchor usage.

I suspect the only way around this would be to do some kind of pre-processing on the YAML schema files, but I'm not sure if it's worth doing. @jeancochrane, what do you think?

jeancochrane: **[Praise]** Yeah, to the best of my knowledge, anchors are the best way of performing this kind of reference. I think your instinct is right; the duplication is a bit annoying, but I think any type of preprocessing would introduce a huge layer of complexity that I don't think is worth it unless/until we can identify something that's truly prohibitive without it.

jeancochrane (approved):
Excellent!

PR #199, opened 2023-10-30
Drop sales val outputs from legacy sale view

dfsnow:
This PR drops the (unused) sales val model outputs from the temporary legacy sales view. The view is currently throwing the following error due to a change in the underlying datatype of the sales val output:

```
Failed analyzing stored view 'awsdatacatalog.reporting.vw_ratio_stats': line 79:8: View 
'awsdatacatalog.default.legacy_vw_pin_sale' is stale or in invalid state: column 
[sv_is_outlier] of type boolean projected from query view at position 45 cannot be coerced 
to column [sv_is_outlier] of type bigint stored in view definition
```

This error is blocking reporting jobs from running. Removing the sales val outputs and redefining the view should be a quick fix. Pinging @wrridgeway for awareness.


jeancochrane (approved):
Seems reasonable to me!

PR #208, opened 2023-11-01
Testing unique buyer and seller name within instruno

wrridgeway:
There are instances where iasworld.sales is missing seller and buyer names when mydec has them. We might as well fill missing names with mydec data if we can. This PR also changes how mydec data is joined to iasworld data, now by document number rather than document number and pin. This allows multisales to have complete mydec data without changing our row count.


wrridgeway:
```
select count(*) as rows from "ci_198-ensure-integrity-of-buyerseller-information-for-vw-pin-salesql_default".vw_pin_sale
UNION ALL
select count(*) as rows from default.vw_pin_sale
```

view | rows
-- | --
new | 2477667
old | 2477667

dfsnow (approved):
Awesome, looks good to me. Nice catch @Damonamajor.

PR #212, opened 2023-11-02
Add dbt daily job and update res_report_summary materialization

wrridgeway:
This PR adds a GitHub workflow for running daily table materializations for dbt models. It uses `res_report_summary` (previously a Glue-created table) as an example of the method.


wrridgeway:
ran the action and glue job this morning and then queried both:

```
new <- dbGetQuery(
  conn = AWS_ATHENA_CONN_NOCTUA,
  "select
    *
  from \"ci_206-replace-vw-res-report-summary-and-res-report-summary-with-dbt-materialization_reporting\".res_report_summary
"
)

old <- dbGetQuery(
  conn = AWS_ATHENA_CONN_NOCTUA,
  "select * from reporting.res_report_summary
"
)

> summary(new$fmv_median)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.    NA's 
  10000  153590  246660  310725  386276 6517199     955 
> summary(old$fmv_median)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.    NA's 
  10000  153590  246660  310726  386388 6509198     955
```

wrridgeway:
Differences here seem to be only a product of median tie-breaking, all other columns agree

dfsnow:
@wrridgeway @jeancochrane I think I cracked this. It seems like most of these run failures were due to race conditions when building the S3 outputs. Basically, when watching the S3 paths and logs closely it looks like:

1. `dbt-athena` deletes the target S3 keys from a bucket at the beginning of a run. This seems to be the very first thing it does. It does it using direct S3 and Glue API calls.
2. It then executes the CTAS statement, which directly populates the target S3 bucket.

So if a second query finishes a CTAS _after_ the first query has already populated the bucket, then we'll get the `TABLE ALREADY EXISTS` error.

Simply re-running the failed job should be enough to fix this. If we want a longer term solution, we should switch to [high-availablity](https://github.com/dbt-athena/dbt-athena#highly-available-table-ha) mode which uses a different update mechanism that would likely solve this issue. However, I don't think that's worth doing quite yet.

jeancochrane (approved):
Looks good to me! And agreed about high-availability tables being a good option for resolving the race condition issue if it becomes too much of a problem.

.github/workflows/build_daily_dbt_models.yaml
```diff
@@ -0,0 +1,58 @@
+name: build-daily-dbt-models
+
+on:
+  schedule:
+    # Every day at 8am (2pm UTC-5)
```
dfsnow, line 5: Daylight savings will change the time that this actually runs but 🤷

.github/workflows/build_daily_dbt_models.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Build daily models
+        run: dbt run -t "$TARGET" -s tag:daily --defer --state "$STATE_DIR"
+        working-directory: ${{ env.PROJECT_DIR }}
+        shell: bash
+
+      - name: Test daily models
+        run: dbt test -t "$TARGET" -s tag:daily --defer --state "$STATE_DIR"
+        working-directory: ${{ env.PROJECT_DIR }}
+        shell: bash
+
```
dfsnow, line 42: @jeancochrane We weren't sure if we needed to update the state cache here. I'm not sure that anything in the cache would change since the asset is already supposed to be built.

```suggestion

      - if: github.ref == 'refs/heads/master' && success()
        name: Update dbt state cache
        uses: ./.github/actions/save_dbt_cache
        with:
          path: ${{ env.PROJECT_DIR }}/${{ env.TARGET_DIR }}
          key: ${{ env.CACHE_KEY }}
          
```

.github/workflows/test_dbt_source_freshness.yaml
```diff
@@ -2,8 +2,8 @@ name: test-dbt-source-freshness
 
 on:
   schedule:
-    # Every day at 9am CST (3pm UTC)
-    - cron: '0 15 * * *'
+    # Every day at 8am (2pm UTC-5)
+    - cron: '0 13 * * *'
```
dfsnow, line 6: Incidentally just updating some of the scheduled workflows in this PR.

aws-athena/ctas/reporting-res_report_summary.sql
```diff
@@ -2,20 +2,31 @@
...
 */
+
+{{
+    config(
+        materialized='table',
+        table_type='hive',
+        format='parquet',
+        write_compression='zstd',
+        bucketed_by=['year'],
+        bucket_count=1
```
dfsnow, line 26: This config creates a single compressed Parquet file instead of lots of uncompressed little files. I'm pretty confident this will increase query speed, if only slightly. I'm going to make a separate PR to apply these settings to other CTAS.

dbt/models/reporting/schema.yml
```diff
@@ -6,10 +6,13 @@ sources:
...
 
-      - name: res_report_summary
-        description: '{{ doc("table_res_report_summary") }}'
-
 models:
+  - name: reporting.res_report_summary
+    description: '{{ doc("table_res_report_summary") }}'
+    config:
+      tags:
+        - daily
```
dfsnow, line 14: This tag can be applied to any CTAS-based model we want to rebuild daily. No more dedicated Glue jobs for scheduled rebuilds.

dfsnow (approved):

Issue #195, opened 2023-10-27
Update major road feature OSM tags


dfsnow:
Currently, the `major_road` feature is constructed by querying line shapefiles from OpenStreetMap. The current query uses only three tags `highway/motorway`, `highway/primary`, and `highway/trunk`.

There are two issues with this:

1. OSM tags are inconsistent, even within Cook County. Often things that most people would consider major roads are not tagged as such.
2. There are a variety of link tag types that we should probably include in this road definition, e.g. `highway/primary_link`

As such, we should update the [loader script](https://github.com/ccao-data/data-architecture/blob/master/aws-s3/scripts-ccao-data-warehouse-us-east-1/spatial-environment-major_road.R) and rebuild the feature using the expanded tag set.

dfsnow:
Pinging @ccao-jardine on this since I know there's interest from other departments as well.

ccao-jardine:
Ha, great minds think alike! I have a related issue drafted up that we'll discuss today.

wagnerlmichael:
It turns out these tags may have been more consistent than originally thought. It could be that they updated their data. I looked at the[ tags at OSM](https://wiki.openstreetmap.org/wiki/Map_features#Highway), and it didn't seem like there were a lot of solid options for potential additions. 

We did try including the `links` tags for `motorway`,  `primary`, and `trunk` roads, but it changed very little, as shown below

![existing_main_roads](https://github.com/ccao-data/data-architecture/assets/93889413/dfc352f0-af7b-4706-ae70-2a555409f4d8)

![existing_primary_motor_trunk_links](https://github.com/ccao-data/data-architecture/assets/93889413/f6ef570d-2ef8-4987-8c70-3cd8e6e72050)

We will close the issue and re-run the script, in case they did update their data.

PR #216, opened 2023-11-06
Update osmdata package

wagnerlmichael:
During the exploratory analysis in #195, we discovered that the `osmdata` package was out of date. The package update fixes the following error:

```
> osm_roads <- opq(bbox = "Cook County, IL") %>%
+   add_osm_feature(
+     key = "highway",
+     value = c("motorway", "trunk", "primary")
+   ) %>%
+   osmdata_sf() %>%
+   .$osm_lines %>%
+   select(osm_id, name) %>%
+   st_transform(4326) %>%
+   mutate(geometry_3435 = st_transform(geometry, 3435))
Error in `httr2::req_perform()`:
! HTTP 405 Method Not Allowed.
Run `rlang::last_trace()` to see where the error occurred.
```

### DIFF
--- a/aws-s3/renv.lock
+++ b/aws-s3/renv.lock
@@ -2030,7 +2030,7 @@
     },
     "osmdata": {
       "Package": "osmdata",
-      "Version": "0.2.2",
+      "Version": "0.2.5",
       "Source": "Repository",
       "Repository": "CRAN",
       "Requirements": [
--- a/aws-s3/renv.lock
+++ b/aws-s3/renv.lock
@@ -2047,7 +2047,7 @@
         "utils",
         "xml2"
       ],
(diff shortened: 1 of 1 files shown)


aws-s3/renv.lock
```diff
@@ -2030,7 +2030,7 @@
     },
     "osmdata": {
       "Package": "osmdata",
-      "Version": "0.2.2",
+      "Version": "0.2.5",
```
dfsnow, line 2033: **issue:** The hash value for the package has almost certainly changed too. Usually `renv.lock` is updated with `renv::snapshot()`, which would also change the hash version.

wagnerlmichael: Good catch! I now know I should use renv always instead of manually editing the file.

dfsnow (approved):

PR #223, opened 2023-11-08
Add `workflow_dispatch` fix for CI dbt builds

dfsnow:
This PR adds a small tweak to detect and use PR context for dbt when using the `workflow_dispatch` option. The fix is needed because using `GITHUB_HEAD_REF` to detect PR context (and thus to set the dbt target to `ci`) doesn't work for manually dispatched workflows. The issue was discovered during these workflow runs:

- https://github.com/ccao-data/data-architecture/actions/runs/6804924433/job/18503445904
- https://github.com/ccao-data/data-architecture/actions/runs/6804785306/job/18503010735

### DIFF
--- a/.github/actions/configure_dbt_environment/action.yaml
+++ b/.github/actions/configure_dbt_environment/action.yaml
@@ -9,11 +9,17 @@ runs:
         # present, we must be in a PR context. Occasionally, we need to
         # manually rebuild assets generated during a PR. In that case, we
         # detect the PR context using the branch name and event
-        if [[ -n "$GITHUB_HEAD_REF" ]] || [[ "$GITHUB_REF_NAME" != 'master' && "$GITHUB_EVENT_NAME" == 'workflow_dispatch' ]]; then
+        if [[ -n "$GITHUB_HEAD_REF" ]]; then
           echo "On pull request branch, setting dbt env to CI"
           {
             echo "TARGET=ci";
             echo "HEAD_REF=$GITHUB_HEAD_REF"
+          } >> "$GITHUB_ENV"
+        elif [[ "$GITHUB_REF_NAME" != 'master' && "$GITHUB_EVENT_NAME" == 'workflow_dispatch' ]]; then
+          echo "Manually dispatched from pull request branch, setting dbt env to CI"
+          {
+            echo "TARGET=ci";
+            echo "HEAD_REF=$GITHUB_REF_NAME"
           } >> "$GITHUB_ENV"
         elif [[ "$GITHUB_REF_NAME" == 'master' ]]; then
           echo "On master branch, setting dbt env to prod"
--- a/.github/actions/configure_dbt_environment/action.yaml
+++ b/.github/actions/configure_dbt_environment/action.yaml
@@ -6,14 +6,16 @@ runs:
     - name: Configure dbt environment
       run: |
         # GITHUB_HEAD_REF is only set on pull_request events, so if it's
-        # present, we must be in a PR context
-        if [ -n "$GITHUB_HEAD_REF" ]; then
+        # present, we must be in a PR context. Occasionally, we need to
+        # manually rebuild assets generated during a PR. In that case, we
+        # detect the PR context using the branch name and event
+        if [[ -n "$GITHUB_HEAD_REF" ]] || [[ "$GITHUB_REF_NAME" != 'master' && "$GITHUB_EVENT_NAME" == 'workflow_dispatch' ]]; then
           echo "On pull request branch, setting dbt env to CI"
           {
             echo "TARGET=ci";
             echo "HEAD_REF=$GITHUB_HEAD_REF"
           } >> "$GITHUB_ENV"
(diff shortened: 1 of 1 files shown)


.github/actions/configure_dbt_environment/action.yaml
```diff
@@ -6,14 +6,16 @@ runs:
...
-        # present, we must be in a PR context
-        if [ -n "$GITHUB_HEAD_REF" ]; then
+        # present, we must be in a PR context. Occasionally, we need to
+        # manually rebuild assets generated during a PR. In that case, we
+        # detect the PR context using the branch name and event
+        if [[ -n "$GITHUB_HEAD_REF" ]] || [[ "$GITHUB_REF_NAME" != 'master' && "$GITHUB_EVENT_NAME" == 'workflow_dispatch' ]]; then
           echo "On pull request branch, setting dbt env to CI"
           {
             echo "TARGET=ci";
             echo "HEAD_REF=$GITHUB_HEAD_REF"
```
jeancochrane, line 16: **[Question, blocking]** Hmm, the one thing I'm confused about is that if `$GITHUB_HEAD_REF` is unset, then how are we able to reference it here? I feel like maybe I'm missing something obvious due to it being the end of the day 😅

jeancochrane (commented):
One quick question before we merge!

.github/actions/configure_dbt_environment/action.yaml
dfsnow: We want it to be either:

- Set (by the PR context) and the set `target = ci`
- OR not set (since not triggered by a PR push, but instead by workflow dispatch) but still on set `target = ci` since the triggering branch isn't master

jeancochrane: I still don't quite understand 😅  in the spirit of total clarity, here's my mental model of what will happen here on a non-main-branch workflow dispatch:

* `[[ -n "$GITHUB_HEAD_REF" ]]` will evaluate to `false`, but `[[ "$GITHUB_REF_NAME" != 'master' && "$GITHUB_EVENT_NAME" == 'workflow_dispatch' ]]` will evaluate to `true`, so we continue into the conditional block on line 13
* `$TARGET` will be set to `"ci"`
* `$HEAD_REF` will be set to the empty string, since `$GITHUB_HEAD_REF` is unset
    * It's possible that the GitHub bash environment has the `-u` option set, in which case this line would raise an undefined variable exception; I doubt this is the case, so I think `$HEAD_REF` being empty is more likely
* When the `generate_schema_name` macro gets run as part of pretty much every `dbt` command, `$HEAD_REF` will be empty, so this block will set the schema name to `ci__<schema>`:

https://github.com/ccao-data/data-architecture/blob/e862947e29fecc3ead3db37d2d0f831e81e627e1/dbt/macros/generate_schema_name.sql#L29-L35

* Shenanigans may ensue due to `ci__<schema>` not matching the schema that is typically used for pull requests. Here's an example showing `dbt test` raising an error because the `ci__default` schema does not exist:

```
(venv) jecochr@21CCAO-LAPTOP54:~/code/data-architecture/dbt$ HEAD_REF=$GITHUB_HEAD_REF dbt test --target ci --select vw_pin_universe_class_count_is_consistent_by_year
16:15:32  Running with dbt=1.6.5
16:15:33  Registered adapter: athena=1.6.3
16:15:33  Found 61 models, 29 tests, 136 sources, 10 exposures, 0 metrics, 525 macros, 0 groups, 0 semantic models
16:15:33
16:15:37  Concurrency: 5 threads (target='ci')
16:15:37
16:15:37  1 of 1 START test vw_pin_universe_class_count_is_consistent_by_year ............ [RUN]
16:15:42  1 of 1 ERROR vw_pin_universe_class_count_is_consistent_by_year ................. [ERROR in 4.71s]
16:15:42
16:15:42  Finished running 1 test in 0 hours 0 minutes and 9.30 seconds (9.30s).
16:15:42
16:15:42  Completed with 1 error and 0 warnings:
16:15:42
16:15:42    Runtime Error in test vw_pin_universe_class_count_is_consistent_by_year (models/default/schema/default.vw_pin_universe.yml)
  SCHEMA_NOT_FOUND: line 11:18: Schema 'ci__default' does not exist
16:15:42
16:15:42  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1

```

Does this seem right to you, or am I missing something that makes this work in practice? If the above logic is right, I think we'll need to adjust this code slightly so that we set `HEAD_REF=$GITHUB_REF_NAME` instead of `HEAD_REF=$GITHUB_HEAD_REF` on `workflow_dispatch`.

dfsnow: 🤦 I didn't see the further use of `$GITHUB_HEAD_REF`. Truly end of day error.

Will fix and re-ping you.

jeancochrane (approved):
Thanks for your patience with my review, this looks good to go!

PR #224, opened 2023-11-09
Add distance to secondary roads feature

wagnerlmichael:
Creates `proximity.dist_pin_to_secondary_road` table. Uses the raw -> warehouse bucket workflow and simplifies linestrings. We also employ various techniques to make sure there is a little overlap as possible between major roads and secondary roads due to changing osm tags.

Once this is merged we can add this feature to open data.


dfsnow (approved):
This looks good @wagnerlmichael! Nice work. Happy to have this done.

PR #229, opened 2023-11-13
Mitigate incorrect values in htpar.user104

wrridgeway:
Assuming this (`htpar.user104`) data won't get fixed anytime soon, addressing it in `default.vw_pin_appeal` seems to be our best bet for now.


jeancochrane (approved):
Looks great to me! Do you mind removing this `error_if` config before merging so that we can be notified if any new unexpected values pop up (the default number of acceptable failures if `error_if` is not set is 0)?

https://github.com/ccao-data/data-architecture/blob/d83f88246411e360baf220b48ed63a6ab24d11b5/dbt/models/default/schema/default.vw_pin_appeal.yml#L104-L105

jeancochrane (approved):
Good to go!

PR #230, opened 2023-11-14
Simplify major roads feature

### LINKED ISSUE #219: Simplify major roads feature
The process of building the 'distance to major road' feature is currently very time-consuming, which seems to be a result of the high complexity of the linestrings representing major roads.

It's suspected that the major road linestrings contain an excessive number of nodes for each curve, contributing to the increased build times. This complexity may not be necessary for our use case and could be causing inefficient processing. We want to simplify the query here: `spatial-environmental-major_road.R`

wagnerlmichael:
The main task of this PR was originally to use `st_simplify` to reduce the complexity of the linestrings representing major roads. 

It has since expanded in scope to accomplish the following:

- Add historical data back to 2014
- Implement an additive approach. Starting from 2014, each year's data builds upon the previous year, with new major road additions being added to the existing dataset
- Previously the data was written with one script directly to our warehouse bucket. We have added an intermediary script that writes the raw data to our unprocessed bucket, and then we utilize the existing script to perform the simplification, which writes the data to our warehouse bucket

Closes #219


aws-s3/scripts-ccao-data-raw-us-east-1/spatial-environment-major_road.R
```diff
@@ -0,0 +1,47 @@
...
+      ) %>%
+      osmdata_sf() %>%
+      .$osm_lines %>%
+      select(osm_id, name) %>%
+      st_transform(4326) %>%
+      mutate(geometry_3435 = st_transform(geometry, 3435))
+
+    geoarrow::write_geoparquet(osm_roads, remote_file)
+  }
+}
```
dfsnow, line 47: **nitpick:** Run styler and lintr on both your scripts here to fix any listing issues.

aws-s3/scripts-ccao-data-raw-us-east-1/spatial-environment-major_road.R
```diff
@@ -0,0 +1,47 @@
+library(aws.s3)
+library(dplyr)
+library(geoarrow)
+library(glue)
+library(noctua)
```
dfsnow, line 5: **nitpick:** Drop any libraries that aren't actually used:

```suggestion
```

dfsnow (commented):
@wagnerlmichael This looks good, but let's wait for https://github.com/ccao-data/data-architecture/actions/runs/6867260306 to finish before merging.

Also, can you quickly check the difference between the roads in each year? I see there's a large difference in the raw Parquet file size between 2014 and 2023. I recommend making a small multiples map showing the roads and posting it in the PR comments.

dfsnow:
> This is interesting. In the past it seems like there were a lot more roads classified as major roads in the downtown area, and it seems to cut off the roads earlier in the newer 2023 data. Will need to click on the figure and make full screen to get a good view.

Ugh, this is what I was worried about. Clearly someone retagged a bunch of roads at some point, so the road definitions aren't going to be consistent historically. Let's do this:

- Find the earliest year where the tags are mostly consistent with the current (2023) tags
- Change the script such that we're only gathering data from that year on
- Re-run the CTAS again

Forward filling here will be handled by the CTAS, and backward filling should be handled by the model views, where appropriate (so you don't have to do anything to handle filling).

wagnerlmichael:
We ended up taking the 2023 data from the January 1st historical data. This is consistent with our major roads (`"motorway", "trunk", "primary"`) all the way back to 2014. Next year, we will need to handle the [ OSM highway change](https://github.com/ccao-data/data-architecture/pull/230#issuecomment-1811176599), since later in the year in 2023.

wagnerlmichael:
Due to the major roads changing from year to year, we have decided to go with an additive approach. Starting from 2014, each year's data builds upon the previous year, with new major road additions being added to the existing dataset.

dfsnow (approved):
Looks good @wagnerlmichael. Thanks for the iterations on this.

PR #236, opened 2023-11-17
Add iasWorld dbt tests and namespace all tests

dfsnow:
This PR adds 250+ dbt tests covering major iasWorld tables and columns. Tests definitions are drawn from internal QC queries (via Inquire) where possible. However, most tests are checking "common sense" things like no assessed values less than 0, all PINs in a table exist in `PARDAT`, etc.

Many of the tests in this PR are currently failing (**99 out of 294 are failing**). This is by design. In most cases, the failures reveal some kind of data error that needs a correction or decision. Some of them may be false positives, but many of them I've checked by hand to ensure they work.

I'm not exactly sure what the process should be to make corrections/decisions, given there are so many failures. We could:

1. Append an `error_if` threshold to every failing test, then slowly work our way through them ourselves.
2. Collect all the failures in a spreadsheet (or multiple sheets), then send the sheet to Valuations.
3. Use this PR to work through all the failing tests, and try to fix/address each one.

Curious to hear thoughts @jeancochrane @jeancochrane @wrridgeway.

> [!NOTE]
This PR also adds name spacing to test names in the form of `$SCHEMA_$TABLE_$COLUMN_$TESTNAME`. This makes the test output easier to parse. I also truncated the test names so they don't overflow the default test output width. All file changes to `location/`, `reporting/`, and `default/` are related to this change and can be largely ignored.

I apologize in advance for the massive diff. I should've PR'd this sooner but got carried away. I don't think it's actually necessary to review this whole PR, happy to chat about how to manage it.

[Link to full test output](https://github.com/ccao-data/data-architecture/actions/runs/6937355156/job/18871251597#step:8:45)

### DIFF
--- a/dbt/models/default/schema/default.vw_card_res_char.yml
+++ b/dbt/models/default/schema/default.vw_card_res_char.yml
@@ -109,7 +109,7 @@ models:
 
     tests:
       - unique_combination_of_columns:
-          name: vw_card_res_char_unique_by_pin_card_and_year
+          name: default_vw_card_res_char_unique_by_pin_card_and_year
           combination_of_columns:
             - pin
             - year
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -64,7 +64,7 @@ sources:
             description: Percent good from tables (based on age+cdu)
             tests:
               - dbt_utils.accepted_range:
-                  name: iasworld_addn_depr_between_0_and_1
+                  name: iasworld_addn_depr_between_0_and_100
                   min_value: 0
                   max_value: 100
                   config:
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -16,6 +16,8 @@ sources:
               - dbt_utils.accepted_range:
                   min_value: 0
                   max_value: 1500000
+                  config:
+                    where: cur = 'Y' AND deactivat IS NULL
           - name: area_sqm
             description: Area of this line in meters
           - name: areaid
@@ -39,7 +41,7 @@ sources:
                   to: source('ccao', 'class_dict')
                   field: class_code
                   config:
-                    where: class != 'EX'
+                    where: class != 'EX' AND cur = 'Y' AND deactivat IS NULL
           - name: convaddn
             description: Converted addition total
           - name: cur
@@ -55,6 +57,8 @@ sources:
               - dbt_utils.accepted_range:
                   min_value: 0
                   max_value: 100
+                  config:
+                    where: cur = 'Y' AND deactivat IS NULL
           - name: eff_area
             description: Support value per line
           - name: effageovr
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -41,7 +41,11 @@ sources:
                   to: source('ccao', 'class_dict')
                   field: class_code
                   config:
-                    where: class != 'EX' AND cur = 'Y' AND deactivat IS NULL
+                    where: |
+                      taxyr >= '2022'
+                      AND class != 'EX'
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
           - name: convaddn
             description: Converted addition total
           - name: cur
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -14,6 +14,7 @@ sources:
             description: '{{ doc("column_area") }}'
             tests:
               - dbt_utils.accepted_range:
+                  name: iasworld_addn_area_between_0_and_1.5M
                   min_value: 0
                   max_value: 1500000
                   config:
@@ -27,8 +28,10 @@ sources:
           - name: card
             description: '{{ doc("column_card") }}'
             tests:
-              - not_null
+              - not_null:
+                  name: iasworld_addn_card_not_null
               - dbt_utils.accepted_range:
+                  name: iasworld_addn_card_gte_1
                   min_value: 1
           - name: cdu
             description: '{{ doc("column_cdu") }}'
@@ -38,6 +41,7 @@ sources:
             description: '{{ doc("shared_column_class") }}'
             tests:
               - relationships:
+                  name: iasworld_addn_class_in_ccao_class_dict
                   to: source('ccao', 'class_dict')
                   field: class_code
                   config:
@@ -52,6 +56,7 @@ sources:
             description: '{{ doc("column_cur") }}'
             tests:
               - accepted_values:
+                  name: iasworld_addn_cur_in_accepted_values
                   values: ['Y', 'D']
           - name: deactivat
             description: '{{ doc("column_deactivat") }}'
@@ -59,6 +64,7 @@ sources:
             description: Percent good from tables (based on age+cdu)
             tests:
               - dbt_utils.accepted_range:
+                  name: iasworld_addn_depr_between_0_and_1
                   min_value: 0
                   max_value: 100
                   config:
@@ -94,7 +100,8 @@ sources:
           - name: lline
             description: '{{ doc("shared_column_lline") }}'
             tests:
-              - not_null
+              - not_null:
+                  name: iasworld_addn_lline_not_null
           - name: lower
             description: Lower level addition code
           - name: mktadj
@@ -114,8 +121,10 @@ sources:
           - name: parid
             description: '{{ doc("shared_column_pin") }}'
             tests:
-              - not_null
+              - not_null:
+                  name: iasworld_addn_parid_not_null
               - relationships:
+                  name: iasworld_addn_parid_in_pardat_parid
                   to: source('iasworld', 'pardat')
                   field: parid
           - name: pctcomp
@@ -164,6 +173,7 @@ sources:
             description: '{{ doc("shared_column_seq") }}'
             tests:
               - dbt_utils.sequential_values:
+                  name: iasworld_addn_seq_all_sequential_exist
                   group_by_columns:
                     - parid
                     - taxyr
@@ -174,7 +184,8 @@ sources:
           - name: taxyr
             description: '{{ doc("shared_column_year") }}'
             tests:
-              - not_null
+              - not_null:
+                  name: iasworld_addn_taxyr_not_null
           - name: third
             description: Third floor addition code
           - name: trans_id
@@ -197,9 +208,10 @@ sources:
             description: '{{ doc("column_whocalc") }}'
           - name: yrblt
             description: '{{ doc("column_yrblt") }}'
+
         tests:
           - unique_combination_of_columns:
-              name: addn_unique_by_parid_taxyr_card_lline
+              name: iasworld_addn_unique_by_parid_taxyr_card_lline
               combination_of_columns:
                 - parid
                 - taxyr
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -3,6 +3,7 @@ sources:
     loaded_at_field: date_parse(wen, '%Y-%m-%d %H:%i:%s.0')
     tags:
       - load_auto
+      - test_qc_iasworld
       - type_res
 
     tables:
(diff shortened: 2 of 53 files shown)


dbt/models/default/schema/default.vw_card_res_char.yml
```diff
@@ -109,7 +109,7 @@ models:
 
     tests:
       - unique_combination_of_columns:
-          name: vw_card_res_char_unique_by_pin_card_and_year
```
dfsnow, line 112: All changes in `default/` are related to namespace changes. See PR body for more details.

dbt/tests/generic/test_column_length.sql
```diff
@@ -11,14 +11,14 @@
     {%- set columns_csv = additional_select_columns | join(", ") %}
 
     {%- set length_columns = [] %}
-    {% for column in columns %}
+    {%- for column in columns %}
```
dfsnow, line 14: The changes here are just to control whitespace in the compiled SQL output.

dbt/tests/generic/test_expression_is_true.sql
```diff
@@ -4,10 +4,10 @@
...
 -- select for failing rows. If no `select_columns` array is provided, defaults
 -- to selecting 1 for failing rows.
-{% test expression_is_true(model, expression, select_columns=[1]) %}
+{% test expression_is_true(model, expression, column_name, select_columns=[1]) %}
     {%- set columns_csv = select_columns | join(", ") %}
 
     select {{ columns_csv }}
     from {{ model }}
-    where not ({{ expression }})
+    where not ({{ column_name }} {{ expression }})
```
dfsnow, line 12: I added this in order to use `expression_is_true` in column-level tests, i.e.:

```yaml
          - name: rmbed
            tests:
              - expression_is_true:
                  name: iasworld_dweldat_rmbed_lte_rmtot
                  expression: <= rmtot
                  select_columns:
                    - parid
                    - taxyr
                    - card
```

It still works fine in the original context of passing a full expression.

dbt/tests/generic/test_no_extra_whitespace.sql
dfsnow, line 1: The changes in here are to facilitate testing for leading and trailing whitespace in strings such as addresses, i.e. we want `123 FISH STREET` to pass but `123 FISH STREET ` to fail.

dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -0,0 +1,30 @@
...
+    {%- set join_columns_csv = join_columns | join(", ") -%}
+
+    {%- if "." in column -%} {%- set model_col = column -%}
+    {%- else -%} {%- set model_col = "model" ~ "." ~ column -%}
+    {%- endif -%}
+
+    {%- if "." in external_column -%} {%- set external_model_col = external_column -%}
+    {%- else -%}
+        {%- set external_model_col = "external_model" ~ "." ~ external_column -%}
+    {%- endif -%}
```
dfsnow, line 18: The yucky formatting here is a result of `sqlfmt` 🤷

dbt/tests/generic/test_row_values_match_after_join.sql
dfsnow, line 1: This test is almost explicitly for testing that the class column matches across different tables i.e. a PIN's class in `PARDAT` matches its class in `LEGDAT`. One subtlety is that this will not return a row if _any_ row has a matching class after the join. So, if a `PARDAT` PIN is class 211 and it joins to a `DWELDAT` record with a two buildings on the same PIN - one class 211, one class 204 - then no records are returned (since the `DWELDAT` record has a 211).

dbt/models/reporting/schema.yml
```diff
@@ -7,31 +7,31 @@ sources:
...
-                  AND triad IS NOT NULL
-                  AND geography_type IS NOT NULL
-                  AND property_group IS NOT NULL
-                  AND assessment_stage IS NOT NULL
-                  AND sale_year IS NOT NULL
+                AND triad IS NOT NULL
+                AND geography_type IS NOT NULL
+                AND property_group IS NOT NULL
+                AND assessment_stage IS NOT NULL
+                AND sale_year IS NOT NULL
```
dfsnow, line 17: All the deindenting and changes to `|` are just to get a cleaner compiled output i.e. one with line breaks.

dbt/models/iasworld/schema/iasworld.addn.yml
```diff
@@ -20,20 +27,48 @@ sources:
...
+              - relationships:
+                  name: iasworld_addn_class_in_ccao_class_dict
+                  to: source('ccao', 'class_dict')
+                  field: class_code
+                  config:
+                    where: |
+                      taxyr >= '2022'
+                      AND class != 'EX'
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
```
dfsnow, line 52: This test is on basically every `class` column in iasWorld now and is designed to catch issues like people adding a dash to classes, e.g. `2-12`.

dbt/models/iasworld/schema/iasworld.aprval.yml
```diff
@@ -17,12 +17,35 @@ sources:
...
             description: '`APRTOT` calc date'
           - name: aprland
             description: Appraised land value
+            tests:
+              - dbt_utils.accepted_range:
+                  name: iasworld_aprval_aprland_between_0_and_1B
+                  min_value: 0
+                  max_value: 1000000000
+                  config: &unique-conditions
+                    where: cur = 'Y' AND deactivat IS NULL
```
dfsnow, line 40: Since many tests require the same conditionals, I used YAML anchors a lot to repeat the same conditions across tests in the same schema file.

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -15,8 +15,25 @@ sources:
...
+            tests:
+              - relationships:
+                  name: iasworld_asmt_all_class_in_ccao_class_dict
+                  to: source('ccao', 'class_dict')
+                  field: class_code
+                  config:
+                    where: |
+                      taxyr >= '2022'
+                      AND class NOT IN ('EX', 'RR', '999')
+                      AND NOT REGEXP_LIKE(class, '[0-9]{3}[A|B]')
```
dfsnow, line 27: This is excluding any subclasses such as `693A` or similar.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -42,6 +42,20 @@ sources:
...
+            tests:
+              - not_null:
+                  name: iasworld_dweldat_attic_not_null
+                  # Excludes non-regression classes, as defined in Assessor's
+                  # online PDF of all class definitions
+                  config: &unique-conditions
+                    where: | 
+                      cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class NOT IN ('201', '213', '218', '219', '220', '221', '224', '225', '236', '240', '241', '290', '294', '297')
```
dfsnow, line 54: `DWELDAT` has a bunch of non-regression classes that don't actually have any characteristic data, hence the exclusion here.

dbt/models/iasworld/schema/iasworld.owndat.yml
```diff
@@ -185,12 +208,63 @@ sources:
...
+              config: *unique-conditions
+          - expression_is_true:
+              name: iasworld_owndat_addr1_address_is_well_formed
+              expression: |
+                CASE
+                  WHEN LENGTH(REGEXP_EXTRACT(addr1, '^[N|S|E|W]{0,2}\s{0,1}([0-9]{1,5})',1)) <= 5 THEN TRUE
+                  WHEN REGEXP_LIKE(UPPER(addr1), '^[PO BOX|P\.O\. BOX|P O BOX|BOX|P O BX|PO BX|PO BX|P.O.BOX]') THEN TRUE
+                  WHEN REGEXP_LIKE(UPPER(addr1), '^[C\/O|ONE|TWO|THREE|FOUR|FIVE|SIX|SEVEN|EIGHT|NINE]') THEN TRUE
+                  ELSE FALSE
+                END
```
dfsnow, line 261: Basically just testing whether the address starts with a street number less than 5 digits or is a PO Box.

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -53,12 +95,381 @@ sources:
...
           - name: upd_status
             description: '{{ doc("column_upd_status") }}'
+          - name: val01
+            # Test that all val* columns are not negative. Pulled from
+            # the following Inquire queries via Will C:
+            # - FP Checklist - Negative ASMT Values
+            # - FP Checklist - Non-EX, RR parcels with 0 land value
+            # - FP Checklist - Non-EX, RR parcels with 0 value
+            tests:
+              - dbt_utils.accepted_range: &test-non-negative
```
jeancochrane, line 114: **[Praise]** Nice use of anchors here!

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -75,16 +486,25 @@ sources:
             description: '{{ doc("column_whocalc") }}'
 
         tests:
+          # Note that this test is NOT actually the unique primary key of
+          # this table, since there doesn't seem to BE a unique combination
+          # of identifying columns
```
jeancochrane, line 491: **[Question, non-blocking]** Hm, does that mean we should just remove this test then?

dbt/models/iasworld/schema/iasworld.owndat.yml
```diff
@@ -91,6 +95,8 @@ sources:
             description: Code for predefined note
           - name: own1
             description: '{{ doc("shared_column_owner_name") }}'
+
+
```
jeancochrane, line 99: **[Question, non-blocking]** Any particular meaning to these newlines?

jeancochrane (approved):
Awesome job with these tests! I'm surprised they execute so quickly (8m9s in the test run you linked); not bad for 294 queries.

> I'm not exactly sure what the process should be to make corrections/decisions, given there are so many failures. We could:
>
> 1. Append an `error_if` threshold to every failing test, then slowly work our way through them ourselves.
> 2. Collect all the failures in a spreadsheet (or multiple sheets), then send the sheet to Valuations.
> 3. Use this PR to work through all the failing tests, and try to fix/address each one.

So far we've been following approach 1, but I don't necessarily think that means it's the best option. Ideally we would do 2 or 3, but I'm concerned about spamming ourselves with failures for the duration of the time it takes us to correct mistakes, particularly given that tests might end up running on PRs if someone were to edit an underlying model. (I suppose that's not an issue for the iasWorld tests, though, since we don't actually edit those tables directly.)

In an ideal world, I think the approach we discussed last week is probably the right one:

1. Tag all unit tests that are intended to test that our transformations do what we expect
2. Update the `build_and_test_dbt` workflow to only run unit tests
3. Allow integration tests to fail while we fix them, but insist that unit tests pass

I don't think this work would be particularly difficult, but I'm not actually sure which of our existing tests (if any) correspond to unit tests, so I don't have high confidence that this would make sense for our current set of tests.

All that being said, I'm comfortable with all of the options you outlined, as long as we all understand which approach we're deciding to take.

dbt/models/iasworld/schema/iasworld.asmt_all.yml
dfsnow: I think it's still useful, since an error firing would indicate an increasing number of duplicate values.

dbt/models/iasworld/schema/iasworld.owndat.yml
dfsnow: Fixed in 70b5902!

dfsnow:
> In an ideal world, I think the approach we discussed last week is probably the right one:
> 
> 1. Tag all unit tests that are intended to test that our transformations do what we expect
> 2. Update the `build_and_test_dbt` workflow to only run unit tests
> 3. Allow integration tests to fail while we fix them, but insist that unit tests pass

Okay I think I've landed on a good approach based on this. I tagged all QC/source tests and updated the `build_and_test_dbt` workflow to _exclude_ all QC tests. This way we won't be blocked by QC test failures but can still see them happen in the `test_dbt_models` output. We can then work to get the `test_dbt_models` output "back to green" by correcting the tests (it can serve as a running log of things to fix). 

Thoughts @jeancochrane?

Here [are the commits since last time](https://github.com/ccao-data/data-architecture/pull/236/files/2e48f66ee68cef7859299b38554043ce086c5754..2ec996a46b713d403028f0328d1976987a729c09) if you want to review.

PR #237, opened 2023-11-17
Initial sale view dbt tests

dfsnow:


wrridgeway:
what's the reason for switching from `>-` to `|`? Does it have to do with printing?

dfsnow:
> what's the reason for switching from `>-` to `|`? Does it have to do with printing?

Just for consistency with the other views, though it does technically affect the line breaks for multi-line strings.

Issue #232, opened 2023-11-15
Materialize model views


jeancochrane:
Our `model` views are heavy and cause queries to take a long time to execute. To speed up model iteration, we should materialize these views to improve runtime performance. If we do this by adding the `daily` tag to their schema file definitions we can ensure that they get materialized once per day, which should be enough to ensure we're staying up to date with changes to ias.

CC @dfsnow

dfsnow:
One additional to-do for this: let's add row count tests to these views as seen in #246.

PR #246, opened 2023-11-22
Fix where clause logic to deactivate PINs

dfsnow:
PR #141 introduced a significant regression that resulted in incorrect row counts for important views. The regression primarily affected the experimental `default.vw_pin_appeal` view and resulted in over 90% of its values being missing. Other models break down like so:

| Database | Model               | Cur. Count | PR Count |   Diff. | Pct. Missing |
|----------|---------------------|-----------:|---------:|--------:|-------------:|
| default  | vw_card_res_char    |   27347262 | 27347262 |       0 |        0.00% |
| default  | vw_pin_address      |   45071159 | 45079937 |    8778 |        0.02% |
| default  | vw_pin_appeal       |     743601 |  8407667 | 7664066 |       91.16% |
| default  | vw_pin_exempt       |     184845 |   184845 |       0 |        0.00% |
| default  | vw_pin_history      |   45079082 | 45079095 |      13 |        0.00% |
| default  | vw_pin_sale         |    2477674 |  2477674 |       0 |        0.00% |
| default  | vw_pin_universe     |   45079937 | 45079937 |       0 |        0.00% |
| model    | vw_card_res_input   |   27306493 | 27306493 |       0 |        0.00% |
| model    | vw_pin_condo_input  |   10075360 | 10075360 |       0 |        0.00% |
| model    | vw_pin_shared_input |   45094310 | 45094310 |       0 |        0.00% |

## Cause

The regression is the result of some slightly tricky conditional logic. Basically, by using a column from a joined table in a `WHERE` clause _after a join_, you implicitly require that the joined table have a non-NULL record. `default.vw_pin_appeal` is the best example of this. Here's a reprex:

```sql
SELECT *
FROM iasworld.htpar AS htpar
LEFT JOIN iasworld.htagnt AS htagnt
    ON htpar.cpatty = htagnt.agent
WHERE htpar.cur = 'Y'
    AND htpar.caseno IS NOT NULL
    AND htpar.deactivat IS NULL
    AND htagnt.cur = 'Y' -- REQUIRES that htagnt have a corresponding record, basically making this an INNER JOIN
    AND htagnt.deactivat IS NULL
```

## Solution

- By moving the conditionals for `cur` and `deactivat` into each join, the conditionals are applied _before_ the join rather than after. This is also how Tyler's Inquire constructs queries, according to the [Inquire Manual (p. 34)](https://www.tylertech.com/CLT/IASProductUsers/Enterprise-Assessment-Tax-Documentation/Inquire.pdf).
- I added a note to the [Appeals Open Data asset](https://datacatalog.cookcountyil.gov/Property-Taxation/Assessor-Appeals/y282-6ig3) to let users know that rows were missing within a certain time window. We should update the note once this PR is merged.
- I added dbt tests for fixed row counts on each major view. Row counts should really never go _down_, so this should help prevent similar logic mistakes in the future.

## Follow-ups

- Update all affected Open Data assets + update the appeals asset note
- Add row count tests to model views in #232

### DIFF
--- a/aws-athena/views/default-vw_card_res_char.sql
+++ b/aws-athena/views/default-vw_card_res_char.sql
@@ -135,6 +135,8 @@ FROM {{ source('iasworld', 'dweldat') }} AS dwel
 LEFT JOIN {{ source('iasworld', 'pardat') }} AS pardat
     ON dwel.parid = pardat.parid
     AND dwel.taxyr = pardat.taxyr
+    AND pardat.cur = 'Y'
+    AND pardat.deactivat IS NULL
 LEFT JOIN multicodes
     ON dwel.parid = multicodes.parid
     AND dwel.taxyr = multicodes.taxyr
@@ -146,5 +148,3 @@ LEFT JOIN townships
     AND dwel.taxyr = townships.taxyr
 WHERE dwel.cur = 'Y'
     AND dwel.deactivat IS NULL
-    AND pardat.cur = 'Y'
-    AND pardat.deactivat IS NULL
--- a/aws-athena/views/default-vw_pin_address.sql
+++ b/aws-athena/views/default-vw_pin_address.sql
@@ -52,12 +52,12 @@ FROM {{ source('iasworld', 'pardat') }} AS par
 LEFT JOIN {{ source('iasworld', 'legdat') }} AS leg
     ON par.parid = leg.parid
     AND par.taxyr = leg.taxyr
+    AND leg.cur = 'Y'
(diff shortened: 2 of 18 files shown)


wrridgeway:
This makes sense to me - I've run into this issue a bazillion times and should have realized it'd show up here. I'm totally fine with this being merged in, but it would be interesting to know why the address view has those kinds of discrepancies when it's using three tables that should theoretically match up.

wrridgeway (approved):

PR #252, opened 2023-11-29
Update dbt docs with warnings about mailing address staleness

wrridgeway:
We have learned that owner information has _not been regularly updated_ in owndat since 2017. This PR makes note of that in the dbt docs for `iasworld.owndat`, `default.vw_pin_address`, and `default.vw_pin_exempt`.


jeancochrane (approved):
Looks good!

PR #254, opened 2023-11-30
Add time filter to QC tests

dfsnow:
This PR adds a time filter to most `tag:test_qc*` dbt tests related to testing raw iasWorld data. Currently, it limits all tests to only the current year, as set in `dbt_project.yml`. This is because we can't actually correct errors from prior years in most cases, so they aren't worth actually testing.

### DIFF
--- a/dbt/dbt_project.yml
+++ b/dbt/dbt_project.yml
@@ -22,6 +22,18 @@ clean-targets:
   - "target"
   - "dbt_packages"
 
+vars:
+  # Start year for testing data using test_qc_* tagged dbt tests. Typically
+  # this should be the current year, since errors in past data cannot usually
+  # be amended once records are closed. Set as an integer for compatibility
+  # with comparison operators and SQL's BETWEEN
+  test_qc_year_start: 2023
+
+  # End year for testing data using test_qc_* tagged dbt tests. Typically set
+  # to a date in the future, but can also be use to select specific time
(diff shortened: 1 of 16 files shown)


dbt/dbt_project.yml
```diff
@@ -22,6 +22,18 @@ clean-targets:
   - "target"
   - "dbt_packages"
 
+vars:
+  # Start year for testing data using test_qc_* tagged dbt tests. Typically
+  # this should be the current year, since errors in past data cannot usually
+  # be amended once records are closed. Set as an integer for compatibility
+  # with comparison operators and SQL's BETWEEN
+  test_qc_year_start: 2023
```
dfsnow, line 30: I figure it's no big deal to change these once per year. This also gives us the flexibility to test prior years if we ever want to.

dbt/models/default/schema/default.vw_pin_address.yml
```diff
@@ -73,5 +74,4 @@ models:
       - row_count:
           name: default_vw_pin_address_row_count
           above: 45079937 # as of 2023-11-22
-      # TODO: Mailing address changes after validated sale(?)
```
dfsnow, line 76: Removed since we know this doesn't actually happen :/

jeancochrane (approved):
Looks great! I like the use of dbt variables here.

PR #255, opened 2023-11-30
Materialize main `model` views

### LINKED ISSUE #232: Materialize model views

dfsnow:
Closes #232.

### DIFF
--- a/aws-athena/views/model-vw_card_res_input.sql
+++ b/aws-athena/views/model-vw_card_res_input.sql
@@ -10,9 +10,19 @@ filled with the following steps:
    PIN (ONLY for things that don't vary at the property level, such as census
    tract statistics. Property characteristics are NOT filled)
 
-WARNING: This is a very heavy view. Don't use it for anything other than making
-extracts for modeling
+This view is "materialized" (made into a table) daily in order to improve
+query performance and reduce data queried by Athena. The materialization
+is triggered by sqoop-bot (runs after Sqoop grabs iasWorld data)
 */
+{{
+    config(
+        materialized='table',
+        partitioned_by=['year'],
+        bucketed_by=['pin'],
+        bucket_count=1
+    )
+}}
+
 WITH uni AS (
     SELECT * FROM {{ ref('model.vw_pin_shared_input') }}
     WHERE meta_class IN (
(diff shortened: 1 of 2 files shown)


.github/workflows/test_dbt_models.yaml
```diff
@@ -32,8 +32,10 @@ jobs:
         run: echo "TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%S")" >> "$GITHUB_ENV"
         shell: bash
 
+      # Only triggered when called by a bot. Otherwise, notifications
+      # go to whoever called the workflow
       - name: Send failure notification
-        if: failure()
+        if: github.event_name == 'workflow_dispatch' && github.triggering_actor == 'sqoop-bot[bot]' && failure()
```
dfsnow, line 38: There's really no reason for this SNS topic to be used if the workflow is called by a human, since they'll get a GH notification through the normal channels. Thus, I've gated it to only notify us when triggered by `sqoop-bot` will test once this is merged by manually triggering a `sqoop-bot` rebuild.

aws-athena/views/model-vw_card_res_input.sql
```diff
@@ -10,9 +10,19 @@ filled with the following steps:
...
+is triggered by sqoop-bot (runs after Sqoop grabs iasWorld data)
 */
+{{
+    config(
+        materialized='table',
+        partitioned_by=['year'],
+        bucketed_by=['meta_pin'],
+        bucket_count=1
+    )
+}}
```
jeancochrane, line 24: **[Thought, non-blocking]** Now that these views are being materialized, I wonder if it's misleading that they have the `vw_*` prefix and live in `aws-athena/views` instead of `aws-athena/ctas`? The tradeoff of moving/renaming them is we would probably have to tweak a bunch of downstream consumers like the model pipelines. I could go either way, just curious if you have thoughts about long-term organization.

jeancochrane (approved):
Excellent! I like how simple this diff is, and I think gating the SNS notifications to the sqoop-bot actor makes sense.

aws-athena/views/model-vw_card_res_input.sql
dfsnow: Let's keep it as-is for the moment. We can always refactor these later since there's only the two of them.

PR #247, opened 2023-11-28
Update views to include new airport DNL feature

mbjackson-capp:
Update five views; model-vw_pin_condo_input.sql remains unchanged

PR #256, opened 2023-11-30
Prefix non-prod databases with `z_` for better sorting

dfsnow:
This PR amends changes the prefixes used to identify non-prod databases in Athena. It changes:

- `dev_` -> `z_dev_`
- `ci_` -> `z_ci_`

This sorts all the non-prod databases to the bottom of the Athena dropdown menu (where they belong).

Successful run to create a prefixed DB here: https://github.com/ccao-data/data-architecture/actions/runs/7052325320/job/19197128258?pr=256

I also created a local `z_dev_` environment.

### DIFF
--- a/aws-athena/views/default-vw_pin_universe.sql
+++ b/aws-athena/views/default-vw_pin_universe.sql
@@ -20,7 +20,6 @@ SELECT
     sp.y_3435,
 
     -- PIN locations from spatial joins
-    vwl.census_block_group_geoid,
     vwl.census_block_geoid,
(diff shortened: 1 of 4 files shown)


dfsnow:
@jeancochrane I think we should just merge this after #255 and #247. That way they'll clean up themselves. No need to migrate the `dev_` databases IMO. People can recreate them if they need to.

jeancochrane (approved):
Love it, thanks for updating docs too! I agree RE: your migration strategy.

PR #257, opened 2023-11-30
Materialize model views daily

dfsnow:
I forgot we have a workflow to build models every morning. Moved the materialization of `model` views to that workflow for simplicity.


jeancochrane (approved):
Makes sense!

PR #258, opened 2023-11-30
Re-add `meta_year` column to `model.vw_card_res_input`

dfsnow:
I accidentally removed/moved this column when I needed to use it as a partition key in #232. This PR re-adds it to its proper place.


jeancochrane (approved):
Oops!

PR #260, opened 2023-12-01
Remove stray debugging statements and add comments to corner lots script

jeancochrane:
This PR makes some minor cleanup tweaks to the corner lots script on the `108-create-corner-lot-indicator` branch in order to make the code easier to read.

Once this PR is merged, I'll put up another PR to move the script without changes to the `aws-s3/` directory so that we can merge it into the main branch of the repo and start making changes using [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow).

### DIFF
--- a/Map Version Edits.R
+++ b/Map Version Edits.R
@@ -1,3 +1,18 @@
+# Script to generate shapefiles with indicators for corner lots.
+#
+# To run, set the township name variables below for the township you're
+# interested in, then run the script. The script will output shapefiles to the
+# `Base Data/` directory. These files should then be uploaded to the following
+# S3 path:
+#
+# s3://ccao-data-raw-us-east-1/location/corner_lot/year={year}/*.shp
+#
+# At some point we may want to automate this upload, but it's manual for now.
+#
+# Note that the script can take upwards of 24 hours to complete, so you may
+# want to run it as a Background Job in RStudio or via a tmux session on
+# the server to ensure that your connection doesn't get killed.
+
 library(osmdata)
 library(tidyverse)
 library(sf)
@@ -16,22 +31,20 @@ library(geosphere)
 library(profvis)
 library(furrr)
 
-townshipa_list <- c("Barrington", "Berwyn", "Bloom", "Bremen", "Calumet", "Cicero", 
-                    "Elk Grove", "Evanston", "Hanover", "Hyde Park", "Jefferson", "Lake", 
-                    "Lake View", "Lemont", "Leyden", "Lyons", "Maine", "New Trier", "Niles",
-                    "North Chicago", "Northfield", "Norwood Park", "Oak Park", "Orland",
-                    "Palatine", "Palos", "Proviso", "Rich", "River Forest", "Riverside", "Rogers Park", 
-                    "Schaumburg", "South Chicago", "Stickney", "Thornton", "West Chicago",
-                    "Wheeling", "Worth")
-
+# We need two slightly different representations of the township name in order
+# to satisfy the requirements of the different external systems we query.
+# townshipa is used to filter our shapefiles, and should be the same as the
+# township_name in ccao::town_dict
 townshipa <- "Bloom"
+# township is used to query Cook County open data, which sometimes has a slightly
+# different format. Unfortunately there's no canonical crosswalk, so we
+# typically just use trial and error, and most of the time township is
+# equivalent to townshipa
 township <- "Bloom"
 
 bbox <- ccao::town_shp %>%
   filter(township_name == townshipa) %>%
   st_bbox()
-
-bbox
 
 bbox[1] <- bbox[1] -.0001
 bbox[2] <- bbox[2] -.0001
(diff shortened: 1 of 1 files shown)


Map Version Edits.R
```diff
@@ -1,3 +1,18 @@
...
+# `Base Data/` directory. These files should then be uploaded to the following
+# S3 path:
+#
+# s3://ccao-data-raw-us-east-1/location/corner_lot/year={year}/*.shp
+#
+# At some point we may want to automate this upload, but it's manual for now.
+#
+# Note that the script can take upwards of 24 hours to complete, so you may
+# want to run it as a Background Job in RStudio or via a tmux session on
+# the server to ensure that your connection doesn't get killed.
```
jeancochrane, line 14: I haven't yet tried the Background Job approach in RStudio yet, but it seems like it might be easier than running the job in the foreground and manually refreshing the connection; want to give it a try for your next township export and see if it works OK?

Map Version Edits.R
```diff
@@ -16,23 +31,21 @@ library(geosphere)
 library(profvis)
 library(furrr)
 
-townshipa_list <- c("Barrington", "Berwyn", "Bloom", "Bremen", "Calumet", "Cicero", 
-                    "Elk Grove", "Evanston", "Hanover", "Hyde Park", "Jefferson", "Lake", 
-                    "Lake View", "Lemont", "Leyden", "Lyons", "Maine", "New Trier", "Niles",
-                    "North Chicago", "Northfield", "Norwood Park", "Oak Park", "Orland",
-                    "Palatine", "Palos", "Proviso", "Rich", "River Forest", "Riverside", "Rogers Park", 
-                    "Schaumburg", "South Chicago", "Stickney", "Thornton", "West Chicago",
-                    "Wheeling", "Worth")
```
jeancochrane, line 25: This vector wasn't being used for anything, and since it seems identical to the vector of `township_name`s in `ccao::town_dict`, I went ahead and removed it in favor of a comment pointing to the latter township dict.

Map Version Edits.R
```diff
@@ -71,21 +92,18 @@ network <- suppressWarnings({
 })
 
 
-# Assuming 'network' is your sfnetwork
```
jeancochrane, line 74: This comment didn't seem like it was adding anything meaningful, so I removed it.

Map Version Edits.R
```diff
@@ -71,21 +92,18 @@ network <- suppressWarnings({
...
 network_sf <- st_as_sf(network)
 
-
-ggplot() +
-  geom_sf(data = st_as_sf(town_osm_center), aes(col = highway))
```
jeancochrane, line 79: I'm assuming this plot was just used for debugging purposes, but let me know if you think it should be a necessary part of running the script!

Map Version Edits.R
```diff
@@ -173,37 +193,43 @@ dan_func_single_obs <- function(x, parcels_full, parcels_buffered, network) {
...
     mutate(is_neighboring = cross_int) %>%
     mutate(id = 1:4) %>%
     filter(!is_neighboring)
-  
-  # Step 6: Calculate the angle of cross segments
+
+  # Step 4: Calculate the angle of cross segments in order to filter out
+  # segments that are not at right angles. This is important because we don't
+  # want parallel segments at opposite sides of the parcel, which would not
+  # indicate a corner
```
jeancochrane, line 226: I recall us chatting about how the angle filter may not work properly; is that the case? If so, have we checked in with Dan about that? He didn't seem aware of any such bug when I last discussed the algorithm with him.

Map Version Edits.R
```diff
@@ -217,30 +243,35 @@ dan_func_single_obs <- function(x, parcels_full, parcels_buffered, network) {
...
-  
+
   cross_corner_number <- nrow(cross_corner)
-  
+
   aspect_ratio <- max(cross$aspect_ratio, na.rm = TRUE)
-  
+
+  # Check for an edge case where two corner segments are actually on opposing
+  # sides of the parcel
```
jeancochrane, line 262: To clarify my understanding: Is the following block attempting to accomplish roughly the same thing as the angle filter above?

Map Version Edits.R
Damonamajor: It was for debugging.

Map Version Edits.R
Damonamajor: Background jobs have worked (and what I use during the workday), but there have been trouble with them timing out when on the VPN.

Map Version Edits.R
Damonamajor: I talked with @dfsnow about it, and mentioned that I was building in another indicator (which is beginning at line 263).

Map Version Edits.R
Damonamajor: I'm pretty sure that the aspect ratio is the ratio for the length of two sides of a parcel, that the length is not significantly longer than the width.

jeancochrane: I meant to reference this block in my comment:

https://github.com/ccao-data/data-architecture/blob/0f2c63a33677e5ccc8b210d811c8c3380e80655b/Map%20Version%20Edits.R#L225-L245

Which seems different from the aspect ratio calculation:

https://github.com/ccao-data/data-architecture/blob/0f2c63a33677e5ccc8b210d811c8c3380e80655b/Map%20Version%20Edits.R#L186-L197

But I think your response here answers my question! https://github.com/ccao-data/data-architecture/pull/260#discussion_r1414368160

Map Version Edits.R
jeancochrane: OK cool! I added a quick note on this in 3e25b91506ddbd813c4a84aad8bb84c5471126a6.

PR #261, opened 2023-12-01
 Add APRVAL.reascd field to default.vw_pin_value and dbt docs

wrridgeway:
The field `aprval.reascd` contains the reason codes for changes in assessed values. It is used heavily within Inquire queries to append the change code to any query that uses `ASMT_ALL`. We should update our views to include this column where appropriate, as well as create a `CASE WHEN` statement or table containing the human-readable change reasons.

Additionally, we should add the change reasons to the dbt documentation for this column.

### DIFF
--- a/dbt/models/ccao/docs.md
+++ b/dbt/models/ccao/docs.md
@@ -1,3 +1,11 @@
+# aprval_reascd
+
+{% docs table_aprval_reascd %}
+Reason codes and descriptions for changes to assessed value.
+
+**Primary Key**: `reascd`
+{% enddocs %}
+
 # class_dict
 
 {% docs table_class_dict %}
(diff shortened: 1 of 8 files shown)


wrridgeway:
```
select count(*) as count from "z_ci_243-add-aprvalreascd-field-to-defaultvw-pin-value-and-dbt-docs_default".vw_pin_value
union all
select count(*) as count from "default".vw_pin_value

#	count
1	45079095
2	45079095
```

dbt/models/ccao/docs.md
```diff
@@ -1,3 +1,11 @@
+# aprval_reascd
+
+{% docs table_aprval_reascd %}
+Reason codes and descriptions for changes to assessed value.
```
dfsnow, line 4: **suggestion:** Let's add a quick note here to make clear that this is just a static Parquet file in the warehouse.

dfsnow (commented):
Let's move the definitions in `iasworld/columns.md` to `shared_columns.md` + update the `doc()` calls to point the shared version. Keep the numeric values in the column definition.

dfsnow (approved):

PR #265, opened 2023-12-05
Add script to compute corner lots indicator

jeancochrane:
This PR consolidates the work done on the [`108-create-corner-lot-indicator` branch](https://github.com/ccao-data/data-architecture/tree/108-create-corner-lot-indicator) to merge the corner lots script into the main branch of this repo. It uses a fresh branch instead of `108-create-corner-lot-indicator` since that branch is very far behind the main branch and contains a bunch of files that we don't want to merge.

PR #267, opened 2023-12-06
Update docstring for `spatial-corner_lot.R`

jeancochrane:
I took a brief look at `aws-s3/scripts-ccao-data-raw-us-east-1/spatial-corner_lot.R` after merging #265 and realized that there were a couple of pieces of outdated info in the docstring. This tiny PR fixes that info to ensure that future users have the correct instructions on where output files will get generated and how to upload them to S3.

Note that I'm not 100% confident in where these files should live in S3; I originally had thought that `location.corner_lot` made the most sense, but then refactored to `spatial.corner_lot` after noticing that the [`108-create-corner-lot-indicator` branch](https://github.com/ccao-data/data-architecture/tree/108-create-corner-lot-indicator) had previously used the `spatial` prefix. I'm open to either location as long as it's consistent, so let me know if you have strong preferences!

### DIFF
--- a/aws-s3/scripts-ccao-data-raw-us-east-1/spatial-corner_lot.R
+++ b/aws-s3/scripts-ccao-data-raw-us-east-1/spatial-corner_lot.R
@@ -2,10 +2,10 @@
 #
 # To run, set the township name variables below for the township you're
 # interested in, then run the script. The script will output shapefiles to the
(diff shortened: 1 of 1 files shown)


dfsnow (approved):
Looks good, thanks @jeancochrane!

PR #270, opened 2023-12-14
Update source freshness tests to publish to SNS on scheduled job failure

### LINKED ISSUE #269: Update `test_source_freshness` workflow to send SNS failure notifications
`test_dbt_models` and `build_daily_dbt_models` both send failure notifications to SNS, but `test_source_freshness` does not. Let's bring this workflow up to parity with the other two.

jeancochrane:
This PR updates the `test_dbt_source_freshness` workflow to publish a notification to the dbt testing SNS topic in case of failure. This behavior is only configured for scheduled runs, where GitHub will otherwise [notify the user who last edited the schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).

Workflow run testing the notification: https://github.com/ccao-data/data-architecture/actions/runs/7213700182/job/19654135786

Closes #269.

### DIFF
--- a/.github/workflows/test_dbt_source_freshness.yaml
+++ b/.github/workflows/test_dbt_source_freshness.yaml
@@ -41,7 +41,7 @@ jobs:
       # types should notify the GitHub user who triggered the event
       - name: Send failure notification
         if: github.event_name == 'workflow_dispatch' && failure()
-        uses: ./.github/actions/publist_sns_topic
+        uses: ./.github/actions/publish_sns_topic
         with:
           sns_topic_arn: ${{ secrets.AWS_SNS_NOTIFICATION_TOPIC_ARN }}
           subject: "dbt source freshness tests failed for workflow run: ${{ github.run_id }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/test_dbt_source_freshness.yaml
```diff
@@ -27,3 +27,22 @@ jobs:
         run: dbt source freshness --target "$TARGET"
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
+
+      - name: Get current time
+        if: failure()
```
jeancochrane, line 32: In theory it might make sense to also gate this step to only run when `github.event_name == "schedule"` as in the `Send failure notification` step below, but I went with the bare `failure()` condition so that this SNS behavior matches our other workflows, e.g. `test_dbt_models`:

https://github.com/ccao-data/data-architecture/blob/cb828dfc6d8ba6178b8e20192bd2fccfc30be3c3/.github/workflows/test_dbt_models.yaml#L30-L33

It also has the benefit of simplicity, in that if we want to change the conditions under which SNS notifications are sent, we only have to edit one line (as far as I can tell there's no way to wrap conditions like these in abstractions that could be shared between steps). In any case, I'm happy to be more specific here and adjust the other workflows to match if you feel strongly about changing the pattern.

dfsnow: Eh, I think this is totally fine.

dfsnow (approved):
@jeancochrane Looks good! Thanks for adding this.

Issue #253, opened 2023-11-29
Update `test_dbt_models` workflow to output failure workbook


jeancochrane:
We want to get the ball rolling on a workflow for fixing the data errors that our dbt tests are uncovering. We expect the workflow to look something like this:

1. Nightly scheduled dbt tests [persist failures](https://docs.getdbt.com/reference/resource-configs/store_failures) to a custom schema
2. A helper script queries the persisted failures and generates one or more Excel workbooks with failing rows
    * Still unclear what the ideal format of this sheet is; perhaps custom groupings of tests defined by tags, with one grouping per workbook and one sheet per test in that workbook? Make a decision on this format as part of this issue
3. The script is run manually every so often, and the output workbooks are shared with other teams so they can use them for review and correction

A couple of tasks we need to do in order to validate this workflow:

* Nail down the ideal format for the workbook(s)
* Pull together an MVP of a workbook with all of our currently-failing tests
    * This might involve writing a script per step 2 above, but the priority right now is on validating the workflow and not on engineering the perfect solution
* Schedule a meeting with other teams to discuss the workbook and get feedback

### Open tasks

- [x] Add a script to parse test failure results and output an Excel workbook for human review (#272)
- [x] Update dbt test metadata to ensure that each test has a `meta.description` attribute and returns a useful set of columns for human review in case of failure
- [ ] Update the `test_dbt_models` workflow to run the script on failure, upload the output workbook to S3, and send an SNS notification with a link to the workbook.

jeancochrane:
As of https://github.com/ccao-data/data-architecture/pull/351, metadata is fully updated and the workbook output works; I think we're not going to automate workbook generation just yet, so I'm going to count this issue as closed.

PR #272, opened 2023-12-18
Add script to generate Excel workbook of failed dbt tests

jeancochrane:
This PR is the first in a sequence of three PRs that adds a script `.github/scripts/format_dbt_test_failures.py` to process results of tests run by the command `dbt test --store-failures` and output those results to an Excel workbook for human review.

This first PR adds the `.github/scripts/format_dbt_test_failures.py` script. Upcoming PRs will:

* Update dbt test metadata to ensure that each test has a `meta.description` attribute and returns a useful set of columns for human review in case of failure
* Update the `test_dbt_models` workflow to run the script on failure, upload the output workbook to S3, and send an SNS notification with a link to the workbook.

Connects #253.

## Testing

I ran these commands to run a few failing tests, generate the workbook, and then examine the workbook to make sure it follows the expected format:

```
cd dbt
source venv/bin/activate
pip install -r ../.github/scripts/format_dbt_test_failures_requrements.txt 
dbt test --select source:iasworld.legdat —store-failures
```


.github/scripts/format_dbt_test_failures.py
```diff
@@ -0,0 +1,256 @@
...
+        run_results = json.load(run_results_fobj)
+
+    with open(manifest_filepath) as manifest_fobj:
+        manifest = json.load(manifest_fobj)
+
+    failed_tests_by_category = get_failed_tests_by_category(
+        run_results, manifest
+    )
+    if not failed_tests_by_category:
+        raise ValueError(f"{run_results_filepath} contains no failed rows")
```
dfsnow, line 132: **question:** Won't this also throw an error when there are no tests that fail?

.github/scripts/format_dbt_test_failures.py
```diff
@@ -0,0 +1,256 @@
...
+    @property
+    def fieldnames(self) -> typing.List[str]:
+        """Get a list of fieldnames that encapsulates all of the fieldnames
+        for all of the rows of failing tests tracked by this group."""
+        fieldnames = []
+        for row in self._rows:
+            for column in row.keys():
+                if column not in fieldnames:
+                    fieldnames.append(column)
+        return fieldnames
```
dfsnow, line 84: **suggestion:** We should add a way to order this list with an arbitrary array of strings. ID columns should always be first in the workbook, so being able to pass `["taxyr", "parid", "card"]` and have those be the first 3 columns would be very useful. Caveat here is that not every test/group will have a `card` column.

.github/scripts/format_dbt_test_failures.py
```diff
@@ -0,0 +1,256 @@
...
+                    f"Missing relation_name attribute for test {test_name}. "
+                    "Did you run `dbt test` with the --store-failures flag?"
+                )
+
+            print(f"Querying failed rows from {test_results_relation_name}")
+            cursor.execute(f"select * from {test_results_relation_name}")
+            query_results = cursor.fetchall()
+            if len(query_results) == 0:
+                raise ValueError(
+                    f"Test {test_name} has status 'fail' but no failures"
```
dfsnow, line 206: For clarity:

```suggestion
                    f"Test {test_name} has status 'fail' but no failing rows in Athena"
```

.github/scripts/format_dbt_test_failures.py
```diff
@@ -0,0 +1,256 @@
...
+                )
+
+            # Add custom fields to query results that we don't expect to be
+            # included in the response
+            failed_tests = [
+                {
+                    "iasworld_table": table_name,
```
dfsnow, line 213: **issue:** We also need to make this export compatible with the test _views_ tagged with `test_qc_inquire`. These are multi-table tests pulled from the Inquire system. I think the code already here should work for them, we just have to generalize the naming a bit, i.e.:

```suggestion
                    "source_table": table_name,
```

.github/scripts/format_dbt_test_failures.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def update(self, new_rows: typing.List[typing.Dict]) -> None:
+        """Add a list of new_rows to the rows of failing tests tracked by
+        this group. The new_rows list should be formatted like the rows
+        returned by a csv.DictReader or a DictCursor, i.e. a list of
+        dicts mapping `{column_name: row_value}`."""
+        self._rows = [*self._rows, *new_rows]
+
+    @property
+    def fieldnames(self) -> typing.List[str]:
```
dfsnow, line 76: **issue (maybe):** One potential hiccup here is that since the field names are only populated from failing tests, the actual columns present in the workbook could change a _lot_ between runs. I'm worried that might be kind of confusing for an end user.

One possibility is to pull the metadata from every table in a given category then use it to create the columns, regardless of whether the test failed or not. Not tied to this idea and curious to hear your thoughts!

dbt/dbt_project.yml
```diff
@@ -59,3 +59,5 @@ models:
       +schema: sale
     qc:
       +schema: qc
+tests:
+  +schema: test_failure
```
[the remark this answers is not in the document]
dfsnow: That works for me!

.github/scripts/format_dbt_test_failures.py
```diff
@@ -0,0 +1,256 @@
...
+
+            node = manifest["nodes"].get(unique_id)
+            if node is None:
+                raise ValueError(
+                    f"Missing dbt manifest node with id {unique_id}"
+                )
+
+            meta = node.get("meta", {})
+            category = meta.get("category", DEFAULT_TEST_CATEGORY)
+            test_description = meta.get("description")
```
dfsnow, line 171: **suggestion:** Two other fields we may want to add here:

1. The actual test name, as shown in the `dbt test` output. We can hide this column in the worksheet if needed, but it's useful to have in there.
2. A link to the dbt docs page for the test, which shows the exact SQL definition.

dfsnow (commented):
This is awesome @jeancochrane. Thanks for pulling this together so quickly.

dfsnow:
One more random aside, we probably need to write all the Athena data to the sheet as strings. Otherwise Excel will start converting all the big numbers to scientific notation 🙄

.github/scripts/format_dbt_test_failures.py
jeancochrane: Good call, done in b1ee6eb1fb230a114d984063e79ce8f73fd7f8bc! It's possible we may want ways to change this order from sheet to sheet in the future, but for now the order is hardcoded and we can add an `__init__()` override later if we want.

.github/scripts/format_dbt_test_failures.py
dfsnow: That sounds good to me!

.github/scripts/format_dbt_test_failures.py
dfsnow: Got it, that makes total sense then. Thanks for clarifying.

dfsnow (approved):

PR #276, opened 2024-01-03
Update corner lot data output and script locations

wrridgeway:
Currently our corner lot script lives in the aws-s3/raw `data-architecture` folder, though it doesn't collect and upload any raw data. It should probably be moved to the warehouse folder, and should also be renamed from a `spatial` prefix to `ccao` since it doesn't upload any spatial data.  I've included some rough code to help upload the data in its current form to the warehouse.


dfsnow:
@wrridgeway What's going on with this PR?

wrridgeway:
Nothing. We can kill the PR, but the issue should still be worked at some point.

PR #278, opened 2024-01-03
Add corner lot indicator to modeling views

wrridgeway:
Corner lots live in `ccao.corner_lot` but have not been added to any of the views used for modelling. Let's add it to `default.vw_card_res_char` and `default.vw_pin_condo_char` so that we can use it in the modeling pipelines.

### DIFF
--- a/dbt/models/ccao/schema.yml
+++ b/dbt/models/ccao/schema.yml
@@ -14,6 +14,16 @@ sources:
         tags:
           - type_ic
 
+      - name: corner_lot
+        description: '{{ doc("table_corner_lot") }}'
+        columns:
+          - name: pin10
+            description: '{{ doc("shared_column_pin10") }}'
+          - name: is_corner_lot
+            description: Corner lot indicator.
+          - name: township
+            description: '{{ doc("shared_column_township_code") }}'
+
       - name: hie
         description: '{{ doc("table_hie") }}'
         tags:
--- a/dbt/models/ccao/schema.yml
+++ b/dbt/models/ccao/schema.yml
@@ -20,7 +20,7 @@ sources:
           - name: pin10
             description: '{{ doc("shared_column_pin10") }}'
           - name: is_corner_lot
-            description: Corner lot indicator.
+            description: '{{ doc("shared_column_is_corner_lot") }}'
           - name: township
             description: '{{ doc("shared_column_township_code") }}'
 
--- a/dbt/models/default/schema/default.vw_pin_universe.yml
+++ b/dbt/models/default/schema/default.vw_pin_universe.yml
@@ -252,6 +252,6 @@ models:
             - year
       - row_count:
           name: default_vw_pin_universe_row_count
-          above: 45079937 # as of 2023-11-22
+          above: 45079936 # as of 2024-01-03
       # TODO: Data completeness correlates with availability of spatial data
       # by year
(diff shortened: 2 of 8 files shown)


wrridgeway:
This seems to be working, though there are NULL values for certain PINs:

```
SELECT meta_year, count(*) AS count
FROM "z_ci_277-add-corner-lots-to-athena-views_model".vw_pin_shared_input
WHERE ccao_is_corner_lot IS NULL
GROUP BY meta_year
```
| meta_year | count |
|-----------|-------|
| 2023 | 9769 |
| 2022 | 5317 |
| 2021 | 707 |
| 2020 | 3849 |
| 2019 | 8708 |
| 2018 | 13439 |
| 2017 | 17682 |
| 2016 | 20878 |
| 2015 | 24384 |
| 2014 | 28851 |
| 2013 | 33058 |
| 2012 | 37847 |
| 2011 | 43814 |
| 2010 | 49517 |
| 2009 | 51785 |
| 2008 | 53500 |
| 2007 | 55929 |
| 2006 | 57394 |
| 2005 | 63144 |
| 2004 | 66598 |
| 2003 | 68292 |
| 2002 | 70310 |
| 2001 | 70368 |
| 2000 | 73355 |
| 1999 | 74747 |

dbt/models/ccao/schema.yml
```diff
@@ -14,6 +14,16 @@ sources:
         tags:
           - type_ic
 
+      - name: corner_lot
+        description: '{{ doc("table_corner_lot") }}'
+        columns:
+          - name: pin10
+            description: '{{ doc("shared_column_pin10") }}'
+          - name: is_corner_lot
+            description: Corner lot indicator.
```
dfsnow, line 23: **suggestion:** Let's add a `shared_column` definition for the corner lot indicator.

dfsnow (approved):
1. IMO, we should also add this to PIN universe as well. You can the rework the queries a bit for shared input to pull from there.
2. I'm guessing we have null values because we used the 2021 parcel shape file as a base, where we should've used all distinct XY coordinates over all years.
3. We need a way to "amend" corners over time. I would keep a CSV of the raw data in the raw bucket, which is versioned, then write a short loading script to populate the warehouse bucket with parquet. That way we can flip a corner indicator in the CSVs if necessary.

wrridgeway:
I've got a separate issue/PR for refactoring this workflow, so I'll address buckets, etc. there. I'll work on getting this into pin_universe.

wrridgeway:
I'm going to refrain from referencing `vw_pin_universe` in `vw_pin_shared_input` for now and just grab it straight from `ccao.corner_lot`. We haven't referenced it there historically and there really isn't much need.

dbt/models/default/schema/default.vw_pin_universe.yml
```diff
@@ -250,6 +252,6 @@ models:
             - year
       - row_count:
           name: default_vw_pin_universe_row_count
-          above: 45079937 # as of 2023-11-22
+          above: 45079936 # as of 2024-01-03
```
wrridgeway, line 255: Not sure where this row went, but I miss it already :-(

dbt/models/shared_columns.md
```diff
@@ -786,6 +786,12 @@ percentage of fair cash value at which a property is assessed for taxing
 purposes. See `ccao.class_dict` for more information
 {% enddocs %}
 
+## is_corner_lot
+
+{% docs shared_column_is_corner_lot %}
+Corner lot indicator.
```
dfsnow, line 792: **suggestion:** Can you open an issue to improve the docs here (and for the table) describing the algorithm used?

dfsnow (approved):
One small request, otherwise this looks good to go.

dbt/models/shared_columns.md
wrridgeway: Can do.

PR #281, opened 2024-01-03
Update metadata in failing dbt tests for clearer workbook output

jeancochrane:
This PR updates all dbt tests that are currently failing to add the following metadata that is used to populate the workbook in `.github/scripts/format_dbt_test_failures.py`:

* `meta.description`: A rich description for the test; defaults to the empty string
* `meta.category`: The category (i.e. workbook sheet name) for the test; defaults to the name of the test generic
* `meta.table_name`: The name of the table that the test is concerned with; defaults to the name of the table the test is defined on
* `select_columns`: Additional columns to select when storing test failures
    * All failing iasworld tests are updated to include `taxyr`, `parid`, `who`, `wen`, and optionally `card` in the selected columns

Test these changes by running tests and generating a workbook -- run these commands in the `dbt` subdirectory:

```
dbt test --store-failures
```

The workbook will be output to your working directory.


dbt/models/iasworld/schema/iasworld.htpar.yml
```diff
@@ -303,6 +312,9 @@ sources:
                 - cpaddr3
               allow_interior_space: true
               config: *unique-conditions
+              meta:
+                description: >
+                  Address columns should have no leading or trailing whitespace
```
jeancochrane, line 317: Whitespace failures will be particularly confusing to read in the workbook unless we come up with a way of highlighting them in the output; I wonder if it would be better to just remove these tests entirely?

dbt/models/iasworld/schema/iasworld.owndat.yml
```diff
@@ -250,6 +261,12 @@ sources:
                 - user27
               allow_interior_space: true
               config: *unique-conditions
+              meta:
+                description: >
+                  The following columns should have no leading or trailing
+                  whitespace: own1, own2, addr1, addr2, addr3, addr4, addr5,
+                  adrpre, adrdir, adrstr, adrsuf, unitdesc, unitno, cityname,
+                  statecode, zip1, zip2, user27
```
jeancochrane, line 269: Same here RE: whitespace tests being hard to debug. If we keep them, we should probably split these out into separate tests for every field.

dbt/models/qc/schema.yml
```diff
@@ -1,28 +1,52 @@
 models:
-  - name: qc.vw_class_mismatch
+  - name: qc.vw_oby_class_mismatch
```
jeancochrane, line 2: I refactored a bunch of these QC tests to make wider use of generics and simplify the configurations, since I needed to refactor them to select columns and add descriptions anyway.

dbt/models/qc/schema.yml
```diff
@@ -495,12 +629,32 @@ models:
     tests:
       - expression_is_true:
           name: qc_vw_change_in_high_low_value_sales_ratio_high
-          expression: |
-            price_less_than_10k_growth_status != 'No significant change'
```
jeancochrane, line 499: Note that I think this test previously had the logic backwards; `expression` should be treated [as an assert statement](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#expression_is_true-source), but here it seems to be the opposite. There are a few places in this file where I had to make similar switches; another argument against using `expression_is_true`, since clearly it's confusing to remember how the expression works.

dbt/tests/generic/test_accepted_range.sql
```diff
@@ -0,0 +1,44 @@
...
+    max_value=none,
+    inclusive=true,
+    select_columns=[]
+) %}
+
+    {%- set columns_csv = select_columns | join(", ") %}
+
+    with
+        meet_condition as (
+            select {{ column_name }} {%- if columns_csv %}, {{ columns_csv }}{% endif %}
```
jeancochrane, line 16: Now that I've refactored our generic tests to accept a `select_columns` param, I'm wondering whether we should standardize on `additional_select_columns` instead... these tests are all written to select `column_name` by default if it exists in the test, so perhaps `additional_select_columns` makes it clearer what the param does? (The semantics is important since if a column is included twice in the `SELECT` statement, Athena will raise a somewhat confusing error.) What do you think is the clearer interface?

dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -250,9 +253,14 @@ sources:
           - row_values_match_after_join:
               name: iasworld_comdat_class_matches_pardat_class
               column: class
+              column_name: comdat_classes
               external_model: source('iasworld', 'pardat')
               external_column: class
+              external_column_name: pardat_classes
```
dfsnow, line 259: **nitpick:** The `class_code` sheet already has a `pardat_class` column. Can we rename these to `comdat_class` and `pardat_class`? Same for `asmt_all_classes` in the other test.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -636,26 +688,32 @@ sources:
...
+                  select_columns: *select-columns
                   config:
                     where: |
                       CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
                       AND class NOT IN ('208', '209', '211', '212', '236', '297')
                       AND cur = 'Y'
                       AND deactivat IS NULL
-              - dbt_utils.accepted_range:
+                  meta:
+                    description: sfla should be between 1 and 10000
```
dfsnow, line 703: **nitpick** Technically this should be something like:

```suggestion
                    description: sfla (bldg. square feet) should be between 1 and 10000 for single-family properties (excl. 208 or 209)
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -636,26 +688,32 @@ sources:
...
                   max_value: 20000
+                  select_columns: *select-columns
                   config:
                     where: |
                       CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
                       AND class IN ('208', '209', '211', '212')
                       AND cur = 'Y'
                       AND deactivat IS NULL
+                  meta:
+                    description: sfla should be between 1 and 20000
```
dfsnow, line 716:
```suggestion
                    description: sfla (bldg. square feet) should be between 1 and 20000 for multi-family and large single-family
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -46,14 +46,17 @@ sources:
...
                   # online PDF of all class definitions
                   config: &unique-conditions
-                    where: | 
+                    where: |
                       CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
                       AND cur = 'Y'
                       AND deactivat IS NULL
                       AND class NOT IN ('201', '213', '218', '219', '220', '221', '224', '225', '236', '240', '241', '290', '294', '297')
+                  meta:
+                    description: attic should not be null
```
dfsnow, line 59: **issue:** I think it's important to add labels to most of the `dweldat` tests indicating what the underlying column is, since most people won't be familiar with the DB column names. This is especially true for `user` columns. I suggest something like:

```suggestion
                    description: attic (attic type) should not be null
```

dbt/tests/generic/test_accepted_range.sql
dfsnow: I do think `additional_select_columns` is a bit clearer for column-level tests.

dfsnow (commented):
Nice job getting this done @jeancochrane. I was able to build the workbook with no issue. I think it needs one or two more rounds of changes before it's set for the Valuations folks, but it's really close.

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Makes sense! I went through the file and added labels to most of the dweldat columns in 7ab148ef5729656b368e5e35515ce3a6513a358d.

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Fixed in 7ab148ef5729656b368e5e35515ce3a6513a358d.

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Fixed in 7ab148ef5729656b368e5e35515ce3a6513a358d.

dbt/models/iasworld/schema/iasworld.htpar.yml
jeancochrane: Cool! I'll leave that for a follow-up PR, I think we should probably have a broader discussion with the other teams about what tests to keep or hide since I expect there may be more that they want to get rid of.

dbt/tests/generic/test_accepted_range.sql
jeancochrane: I think that makes sense, done in f0a0586bc123834ae3d615668bf96cb56341493e! The only gray area is `expression_is_true`, which is sometimes a column-level test and sometimes a table-level test. Since we use it much more often as a table-level test, and since the column-level versions are probably ripe for refactoring for other generic test functions anyway, I went ahead and kept its param as `select_columns`, even though there's one or two instances where we use it on a column-level test where we already select `column_name` by default.

jeancochrane:
This should be ready for another look @dfsnow! Let me know if I missed any of your comments from last round.

dfsnow (approved):
Overall this looks a lot better! However, I still see a few minor issues to clear up. We could address in a separate PR if you think that would be easier, so approving this for now.

- No description for tab `duplicate_records` table `asmt_all`
- The `sales` tab has seemingly incorrectly matched columns starting on row 647, see `deed_type` and `price` columns
- Same issue as above for `expression_is_true` tab, in last few rows
- The `expression_is_true` tab probably needs a new name

Something else to address in a separate PR: we should add formatting and filters to these sheets to make them easier to work with. I'd start by just setting some column widths and filters on major columns.

jeancochrane:
Agreed @dfsnow! I opened https://github.com/ccao-data/data-architecture/issues/288 to encompass those edits since this PR is already so huge. I'll merge this one in and then get started on #288 shortly.

PR #285, opened 2024-01-08
Correct `env_airport_noise_data_year` fill misjoin

wrridgeway:
'omp' is a value for `env_airport_noise_data_year` in our location data crosswalk, but can't be joined to our generic `year`column. Airport noise is already filled during ingest so we don't need to use `env_airport_noise_data_year` when joining data for `location.vw_pin10_fill`


wrridgeway:
Please don't laugh at how long this took me.

dfsnow (approved):

PR #293, opened 2024-01-16
 Add card-level class to res chars and modeling views

wrridgeway:
If we want to use class as a feature for splitting modeling groups we need it to be card-level for the res pipeline. Currently we're only using parcel-level class in the res modeling view.

Let's add a card-level `char_class` column to `vw_card_res_char` and `vw_card_res_input`.


dfsnow (approved):
Nice @wrridgeway. Thanks for getting this together so quickly.

PR #294, opened 2024-01-18
Add distance to major university

Damonamajor:
This adds the distances between parcels and the following universities in feet

'Illinois Institute of Technology Chicago-Kent College of Law', 
'Columbia College', 
'University of Illinois at Chicago College of Medicine',
 'Rush University Medical Center', 
'University of Chicago' 
'University of Illinois at Chicago',
 'Northwestern University',
 'Loyola University of Chicago',
 'De Paul University', 
'Chicago State University', 
'Moraine Valley Community College'

PR #295, opened 2024-01-19
Create distance to vacant land feature

Damonamajor:
This PR:
- Adds distance to vacant lot as a proximity CTAS. A vacant lot in this case is any class 100 land larger than 1,000 sqft.
- Also adds our first ever [ephemeral](https://docs.getdbt.com/docs/build/materializations#ephemeral) table - `ephemeral_vacant_land`, which gets used by the CTAS. @jeancochrane let me know if my schema setup here makes sense.
- Adds missing doc block calls for the distance to university feature

A couple notes on this:

- The vacant land feature takes about 2.5 hours to generate :/
- Since we're matching parcel file to parcel file, class 100 parcels match with themselves and are the "nearest" vacant land, but that's okay for now


aws-athena/ctes/ephemeral_vacant_land.sql
dfsnow, line 1: This is effectively just a CTE that gets interpolated into the larger query for distance to vacant land. I propose that we start using these more often where appropriate, they're super hand.

dbt/models/ephemeral/schema.yml
```diff
@@ -0,0 +1,3 @@
+models:
+  - name: ephemeral_vacant_land
```
dfsnow, line 2: Note the slightly different schema: `ephemeral_vacant_land` instead of `ephemeral.vacant_land`, as CTEs can't have periods in them. I think this is perfectly fine, but we can always refactor (to keep the period) later if we really want.

dbt/models/proximity/schema.yml
```diff
@@ -185,6 +188,20 @@ models:
...
       - name: nearest_secondary_road_osm_id
         description: '{{ doc("column_nearest_secondary_road_osm_id") }}'
+      - name: nearest_university_data_year
+        description: '{{ doc("shared_column_data_year") }}'
+      - name: nearest_university_dist_ft
+        description: '{{ doc("column_nearest_university_dist_ft") }}'
+      - name: nearest_university_gnis_code
+        description: '{{ doc("column_nearest_university_gnis_code") }}'
+      - name: nearest_university_name
+        description: '{{ doc("column_nearest_university_name") }}'
```
dfsnow, line 198: These were missing from the last PR (#294), oops.

wrridgeway (approved):
This looks good to me. Spent a bit of time querying the new tables in Athena and didn't spot anything suspicious.

aws-athena/ctes/ephemeral_vacant_land.sql
jeancochrane: **[Questions, non-blocking]** A couple questions here:

1. What's the particular reason that an ephemeral model makes sense in this case? My naive intuition would be that ephemeral models are useful for cases where we want to reuse a chunk of query logic across multiple models, but where that query logic is static and/or tightly coupled to specific models/sources such that a macro is overkill. That doesn't seem to be the case here, so is it just that it makes the argument to `dist_to_nearest_geometry` clearer? I'm pretty sure (although not totally confident) that a named subquery defined within the query context of `proximity.dist_pin_to_vacant_land` would also work with `dist_to_nearest_geometry`, and my hunch is that might be clearer than an ephemeral model since it would reduce the number of files a reader has to browse to understand what's going on here. But that opinion is weakly-held and I'm open to this pattern if we can make a good case for it being necessary or at least clearer than the alternative.
2. If we do decide to move forward with ephemeral models, I wonder if they might make more sense in a new `aws-athena/ephemeral/` subdirectory rather than `aws-athena/ctes/`? They aren't quite CTEs since they don't actually get materialized to tables, and they're not quite views, so I wonder if it makes sense for them to have their own subdirectory. Again, this is all pretty weakly held!

dbt/dbt_project.yml
```diff
@@ -59,5 +59,7 @@ models:
       +schema: sale
     qc:
       +schema: qc
+    ephemeral:
+      +schema: ephemeral
```
jeancochrane, line 63: **[Suggestion, non-blocking]** If we're going to define an `ephemeral` schema anyway, it might make sense to use this opportunity to set `materialized: ephemeral` for all models in this schema, which would eliminate the need for setting `config(materialized = 'ephemeral')` in every ephemeral query file. But I could also see an argument that the explicitness of the `config(materialized = 'ephemeral')` call might be helpful for readers, given that this type of materialization is relatively uncommon compared to `view` and `table`.

jeancochrane (approved):
Looking good! I think ephemeral models are a neat tool and I can see their usefulness in the abstract, but I'm not yet totally convinced it's the right choice here. Still, if you feel strongly about it, go ahead and merge it and we can continue discussion as more opportunities for ephemeral models arise.

aws-athena/ctes/ephemeral_vacant_land.sql
dfsnow: > **[Questions, non-blocking]** A couple questions here:
> 
> 1. What's the particular reason that an ephemeral model makes sense in this case? My naive intuition would be that ephemeral models are useful for cases where we want to reuse a chunk of query logic across multiple models, but where that query logic is static and/or tightly coupled to specific models/sources such that a macro is overkill. That doesn't seem to be the case here, so is it just that it makes the argument to `dist_to_nearest_geometry` clearer? I'm pretty sure (although not totally confident) that a named subquery defined within the query context of `proximity.dist_pin_to_vacant_land` would also work with `dist_to_nearest_geometry`, and my hunch is that might be clearer than an ephemeral model since it would reduce the number of files a reader has to browse to understand what's going on here. But that opinion is weakly-held and I'm open to this pattern if we can make a good case for it being necessary or at least clearer than the alternative.

I could not for the life of me get a subquery to work within `dist_to_nearest_geometry`. It doesn't work as a CTE or directly nested in the macro call, but for whatever reason an ephemeral view works. @jeancochrane any thoughts?

> 2. If we do decide to move forward with ephemeral models, I wonder if they might make more sense in a new `aws-athena/ephemeral/` subdirectory rather than `aws-athena/ctes/`? They aren't quite CTEs since they don't actually get materialized to tables, and they're not quite views, so I wonder if it makes sense for them to have their own subdirectory. Again, this is all pretty weakly held!

Yep, this makes sense. Will make this change.

dbt/dbt_project.yml
dfsnow: Agreed! I think it's totally sensible to just make everything under that schema ephemeral by default. Changed.

aws-athena/ctes/ephemeral_vacant_land.sql
jeancochrane: > I could not for the life of me get a subquery to work within dist_to_nearest_geometry. It doesn't work as a CTE or directly nested in the macro call, but for whatever reason an ephemeral view works. @jeancochrane any thoughts?

@dfsnow This diff seemed to generate a query that was running OK in Athena (I stopped it so I didn't see if it completed, but it ran for ~1min without raising an error):

```diff
diff --git a/aws-athena/ctas/proximity-dist_pin_to_vacant_land.sql b/aws-athena/ctas/proximity-dist_pin_to_vacant_land.sql
index 30966b6..f04bd4a 100644
--- a/aws-athena/ctas/proximity-dist_pin_to_vacant_land.sql
+++ b/aws-athena/ctas/proximity-dist_pin_to_vacant_land.sql
@@ -7,6 +7,21 @@
     )
 }}

+WITH ephemeral_vacant_land AS (
+    SELECT
+        parcel.pin10,
+        ST_ASBINARY(ST_POINT(parcel.x_3435, parcel.y_3435)) AS geometry_3435,
+        parcel.year
+    FROM {{ source('spatial', 'parcel') }} AS parcel
+    INNER JOIN {{ source('iasworld', 'pardat') }} AS pardat
+        ON parcel.pin10 = SUBSTR(pardat.parid, 1, 10)
+        AND parcel.year = pardat.taxyr
+        AND pardat.cur = 'Y'
+        AND pardat.deactivat IS NULL
+        AND pardat.class = '100'
+        AND ST_AREA(ST_GEOMFROMBINARY(geometry_3435)) >= 1000
+)
+
 SELECT
     pcl.pin10,
     ARBITRARY(xy.pin10) AS nearest_vacant_land_pin10,
@@ -15,7 +30,7 @@ SELECT
     pcl.year
 FROM {{ source('spatial', 'parcel') }} AS pcl
 INNER JOIN
-    ( {{ dist_to_nearest_geometry(ref('ephemeral_vacant_land')) }} ) AS xy
+    ( {{ dist_to_nearest_geometry('ephemeral_vacant_land') }} ) AS xy
     ON pcl.x_3435 = xy.x_3435
     AND pcl.y_3435 = xy.y_3435
     AND pcl.year = xy.pin_year
```

```
$ dbt run --select proximity.dist_pin_to_vacant_land
```

Does that seem like it'd work for you? Again, I'm up for an ephemeral model if you think it's better than this alternative, but I kinda like the CTE.

PR #296, opened 2024-01-21
Add exemptions feature subquery/CTE

dfsnow:
This PR adds two exemption-related features to the residential model input view:

1. Whether or not a PIN has an exemption active in a given year
2. How many consecutive years a PIN has had an exemption active

### DIFF
--- a/aws-athena/views/model-vw_pin_shared_input.sql
+++ b/aws-athena/views/model-vw_pin_shared_input.sql
@@ -279,6 +279,10 @@ SELECT
     sdrs.school_district_avg_rating
         AS other_school_district_secondary_avg_rating,
 
+    -- Exemption features
+    exemption_features.ccao_is_exe_homeowner_active,
+    exemption_features.ccao_n_years_exe_homeowner_active,
+
     -- Corner lot indicator
     lot.is_corner_lot AS ccao_is_corner_lot,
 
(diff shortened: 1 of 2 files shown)


dfsnow:
@wrridgeway @jeancochrane I'm going to merge this for the sake of expediency, but please give it a quick glance when y'all get the chance.

aws-athena/views/model-vw_pin_shared_input.sql
```diff
@@ -103,6 +103,32 @@ school_district_ratings AS (
...
+    SELECT
+        year,
+        pin,
+        SIGN(exe_homeowner) AS ccao_is_active_exe_homeowner,
+        act AS ccao_n_years_exe_homeowner
+    FROM (
+        SELECT
+            *,
+            SUM(SIGN(exe_homeowner))
+                OVER (PARTITION BY pin, grp ORDER BY year)
```
jeancochrane, line 118: **[Question, non-blocking]** I don't totally get what the `grp` subquery is doing for us here. Without a detailed understanding of the underlying data model, my naive assumption would be that `SUM(SIGN(exe_homeowner)) OVER (PARTITION BY pin ORDER BY year)` would be enough to get the history of the `exe_homeownrer` column. Is this the "down-up fill" you mention in the comment above?

dfsnow: It's not intuitive at all, but the purpose of `grp` is to reset the cumulative count of years whenever the `exe_homeowner` value is 0 (i.e. whenever someone loses the exemption).

PR #297, opened 2024-01-22
Move conditional for nearest dist macro to CTE

dfsnow:
Per @jeancochrane's example in #295, I've refactored the distance to nearest geometry macro to use a CTE instead of a conditional passed as a macro argument. This is significantly simpler and more intuitive in my opinion while getting the same results.


jeancochrane (approved):
Nice cleanup!

Issue #288, opened 2024-01-09
dbt test workbook improvements


jeancochrane:
As a fast follow to https://github.com/ccao-data/data-architecture/pull/281, we want to make a few improvements to the dbt test workbook:

- [x] No description for tab `duplicate_records` table `asmt_all`
- [ ] The `sales` tab has seemingly incorrectly matched columns starting on row 647, see `deed_type` and `price` columns
- [ ] Same issue as above for `expression_is_true` tab, in last few rows
- [ ] The `expression_is_true` tab probably needs a new name
- [ ] Add formatting and filters to the sheets
    - [x] Set column widths
    - [ ] Set column filters on ID columns
    - [ ] Add top-level labels for column groups
    - [ ] Color code each column group
- [ ] Add `test_qc_exclude_from_workbook` tag to enable excluding specific tests from the workbook output

PR #298, opened 2024-01-23
Clean up and format dbt test workbook

jeancochrane:
This PR makes a number of quality-of-life improvements to the dbt test workbook that gets output by the `.github/scripts/format_dbt_test_failures.py` script. It is the basis of the workbook that was sent over to the rest of the internal stakeholders for review.

Improvements include:

* Adds a test caching system to speed up workbook development
* Writes test results using a write-only openpyxl workbook in order to reduce memory use
* Adds groupings for worksheet columns, with dedicated colors and top-level headers
* Adds support for a new `test_qc_exclude_from_workbook` tag that can be used to exclude certain tests or models
* Adds a few missing descriptions and columns for failing tests

Changes were tested by generating the workbook with the following commands run in the `dbt/` subdirectory:

```
dbt test --select test_qc_* --exclude test_qc_exclude_from_workbook --store-failures
```

Closes https://github.com/ccao-data/data-architecture/issues/288.

### DIFF
--- a/.github/scripts/format_dbt_test_failures.py
+++ b/.github/scripts/format_dbt_test_failures.py
@@ -46,13 +46,13 @@ import sys
 import typing
 
 import openpyxl
+import openpyxl.cell
 import openpyxl.styles
+import openpyxl.styles.colors
 import openpyxl.utils
 import pyathena
 import pyathena.cursor
 import simplejson as json
-from openpyxl.worksheet.worksheet import Worksheet
-from openpyxl.styles import Alignment
 
 # Tests without a config.meta.category property will be grouped in
 # this default category
@@ -95,14 +95,14 @@ class FailedTestGroup:
     convenience methods for formatting those results for output to a report."""
 
     # Names of fields that are used for debugging
-    _debugging_field_names = [TEST_NAME_FIELD, DOCS_URL_FIELD]
+    _debugging_fieldnames = [TEST_NAME_FIELD, DOCS_URL_FIELD]
     # Names of fields that identify the failing test
-    _test_metadata_field_names = [
+    _test_metadata_fieldnames = [
         *[SOURCE_TABLE_FIELD, DESCRIPTION_FIELD],
-        *_debugging_field_names,
+        *_debugging_fieldnames,
     ]
     # Names of fields that are used for diagnostics
-    _diagnostic_field_names = [
+    _diagnostic_fieldnames = [
         TAXYR_FIELD,
         PARID_FIELD,
         CARD_FIELD,
@@ -111,9 +111,9 @@ class FailedTestGroup:
         WEN_FIELD,
     ]
     # The complete set of fixed fields
-    _fixed_field_names = [
-        *_test_metadata_field_names,
-        *_diagnostic_field_names,
+    _fixed_fieldnames = [
+        *_test_metadata_fieldnames,
+        *_diagnostic_fieldnames,
     ]
 
     def __init__(
@@ -141,9 +141,7 @@ class FailedTestGroup:
         # Remove any fixed fieldnames from the ordered list that are not
         # present in this group
         fixed_field_order = [
-            field
-            for field in self._fixed_field_names
-            if field in fieldnames
+            field for field in self._fixed_fieldnames if field in fieldnames
         ]
 
         # Reorder the fieldnames so that diagnostic fields are presented in the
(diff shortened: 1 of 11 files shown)


dbt/models/qc/schema.yml
```diff
@@ -794,10 +798,11 @@ models:
     config:
       tags:
         - test_qc_sales
+        - test_qc_exclude_from_workbook
     tests:
       - expression_is_true:
           name: qc_vw_iasworld_sales_day_of_month_lte_half_observations
-          expression: sales_by_day > 0.5 * total_sales
+          expression: sales_by_day < 0.5 * total_sales
```
jeancochrane, line 805: This logic was previously reversed, leading to erroneous reports of failures.

.github/scripts/format_dbt_test_failures.py
```diff
@@ -91,53 +95,58 @@ class FailedTestGroup:
...
+        *_test_metadata_fieldnames,
+        *_diagnostic_fieldnames,
+    ]
 
-    def __init__(self) -> None:
-        self._rows: typing.List[typing.Dict] = []
+    def __init__(
+        self, rows: typing.Optional[typing.List[typing.Dict]] = None
+    ) -> None:
+        self.raw_rows: typing.List[typing.Dict] = rows or []
```
dfsnow, line 122: **question (dumb):** Is this just so we can use tests without any returned rows?

.github/scripts/format_dbt_test_failures.py
```diff
@@ -151,19 +160,116 @@ def rows(self) -> typing.List[typing.List]:
...
+        only those fields that exist in the test group, returning the
+        names of the fields (e.g. ["foo", "bar"])."""
+        existing_fieldnames = self.fieldnames
+        return [
+            field
+            for field in possible_fieldnames
+            if field in existing_fieldnames
+        ]
+
+    def __filter_for_existing_field_indexes(
```
dfsnow, line 179: **question:** Is the dunderscore use here intentional?

.github/scripts/format_dbt_test_failures.py
```diff
@@ -189,6 +295,93 @@ def main() -> None:
...
+        save_failed_tests_by_category_to_file(
+            failed_tests_by_category,
+            failed_test_cache_path,
+        )
+
+    print("Generating the output workbook")
+    # It's important to use a write-only workbook here because otherwise
+    # the metadata required to store cell info about a large number of failing
+    # tests can cause the process to run out of memory
+    workbook = openpyxl.Workbook(write_only=True)
```
dfsnow, line 327: **question:** Is there a way to "convert" the workbook to read-write on final output? Solely to get rid of the dialog box when initially opening the output.

.github/scripts/format_dbt_test_failures.py
```diff
@@ -361,37 +549,180 @@ def add_sheet_to_workbook(
...
+        sheet.append(data_row)
+
+    # Merge cells in the grouping headers. This approach is a bit of a hack
+    # since merged cells are not fully supported in write-only workbooks, hence
+    # why it takes place _after_ rows have been added to the sheet whereas most
+    # formatting options for write-only workbooks need to happen _before_
+    # data is added. See here for details: https://stackoverflow.com/a/66159254
+    for cell_range in merged_cell_range:
+        sheet.merged_cells.ranges.add(cell_range)
+
```
dfsnow, line 725: **praise:** Thank you for the very clear commenting in this section!

.github/scripts/format_dbt_test_failures.py
```diff
@@ -189,6 +295,93 @@ def main() -> None:
...
+    # tests can cause the process to run out of memory
+    workbook = openpyxl.Workbook(write_only=True)
+    for sheet_name, failed_test_group in failed_tests_by_category.items():
+        print(f"Adding sheet for {sheet_name}")
+        add_sheet_to_workbook(workbook, sheet_name, failed_test_group)
+    workbook.save(output_filepath)
+    print(f"Output workbook saved to {output_filepath}")
+
+
+def get_failed_test_cache_path(
```
dfsnow, line 335: Just to check my understanding, you're basically just caching the Athena results and using them as long as they exist to speed up testing? Then if they don't exist, query Athena? Goal being to not query Athena every time we build the workbook?

dfsnow (approved):
Looks great @jeancochrane, nice work. Makes me realize I need to brush up on/find a case for OOP.

.github/scripts/format_dbt_test_failures.py
jeancochrane: By "this", do you mean the default initialization to an empty list? That's for convenience when extracting tests from Athena -- initializing with an empty list allows us to create an empty group and then use the same `update()` call to add the tests to the group whether or not the group already exists:

https://github.com/ccao-data/data-architecture/blob/1559fdc119826760c3d5a8e941ec5e3b220f6716/.github/scripts/format_dbt_test_failures.py#L491-L493

But we could rewrite this pretty easily if we wanted to require that rows exist:
```suggestion
        self, rows: typing.List[typing.Dict]
    ) -> None:
        self.raw_rows: typing.List[typing.Dict] = rows
```

```python
 if not failed_tests_by_category.get(category):
    failed_tests_by_category[category] = FailedTestGroup(failed_tests) 
else:
    failed_tests_by_category[category].update(failed_tests) 
```

.github/scripts/format_dbt_test_failures.py
jeancochrane: Nope, good catch! Fixed in 3642e5d05f7b5b6842c2354f1459b19b0d53b1ec.

.github/scripts/format_dbt_test_failures.py
jeancochrane: Yup, exactly! Otherwise, the queries take up a bunch of time unnecessarily.

.github/scripts/format_dbt_test_failures.py
jeancochrane: Hm, what dialog box are you seeing? I don't see it when I open the file locally, but perhaps something changes during transfer or upload to OneDrive.

jeancochrane:
I left some responses to your questions @dfsnow, this is probably worth one more look just to make sure we're on the same page!

.github/scripts/format_dbt_test_failures.py
dfsnow: Nope fine as-is, I'm just a python dummy

.github/scripts/format_dbt_test_failures.py
dfsnow: Hunh, I thought I remembered a dialog box but don't see anything now after downloading the latest version sent via email. All good!

dfsnow:
> I left some responses to your questions @dfsnow, this is probably worth one more look just to make sure we're on the same page!

I think we're set to go! Let's merge

PR #300, opened 2024-01-23
Remove non-numeric characters from `nbhd` column in `vw_pin_sale`

wrridgeway:


dfsnow (approved):
Eeesh, hopefully this is caught in the current dbt tests. If not, let's add a new one that catches the culprit you found @wrridgeway .

PR #312, opened 2024-02-08
Add exemption features to condo model view

wrridgeway:
We [added exemption features](https://github.com/ccao-data/data-architecture/commit/6b9b31b69185b62329901e581c4cdd3489ad327b) to the res pipeline. This PR replicates these features for condos.


dfsnow (approved):
Woof, I wish there was a cleaner way to do this, but I suppose this does work.

PR #314, opened 2024-02-09
Add proper dbt source freshness tests

dfsnow:
This PR updates our iasWorld [dbt source freshness tests](https://docs.getdbt.com/docs/build/sources#snapshotting-source-data-freshness) to use an actual `loaded_at` column (added by https://github.com/ccao-data/service-sqoop-iasworld/pull/10), rather than the date of last change for a record.

It adds freshness checks for _all_ iasWorld tables pulled by sqoop. Since not all tables are updated daily, some checks expect weekly updates.

I can't test this properly until sqoop does a full DB refresh, which will happen tomorrow (2/10), but code review in the meantime would be appreciated!

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.aasysjur.yml
+++ b/dbt/models/iasworld/schema/iasworld.aasysjur.yml
@@ -65,6 +65,8 @@ sources:
             description: Flag for Landisc `TAXYR`
           - name: legadd_inc
             description: Increment for new lines in table `LEGADD`
+          - name: loaded_at
+            description: '{{ doc("shared_column_loaded_at") }}'
           - name: maildat_ind
             description: Flag for maildat functionality on `AA11`
           - name: max_taxyr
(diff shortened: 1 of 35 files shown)


dbt/models/iasworld/schema/iasworld.aasysjur.yml
```diff
@@ -1,13 +1,16 @@
...
     tags:
       - load_auto
       - test_qc_iasworld
 
     tables:
       - name: aasysjur
         description: '{{ doc("table_aasysjur") }}'
+        freshness:
+          warn_after: {count: 192, period: hour} # 8 days
+          error_after: {count: 360, period: hour} # 15 days
```
dfsnow, line 13: Not all tables get daily updates; some get pulled every Saturday. We expect the latter to be loaded within the last 8 days.

dbt/models/iasworld/schema/iasworld.addn.yml
```diff
@@ -9,6 +9,10 @@ sources:
     tables:
       - name: addn
         description: '{{ doc("table_addn") }}'
+        freshness:
+          filter: taxyr >= date_format(current_date - interval '1' year, '%Y')
```
dfsnow, line 13: Not all tables have a `taxyr` column, but for those that do I'm filtering the table to only scan the latest two years to limit data scanned.

dfsnow:
As of [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/7859098065), all sources are updated with the new `loaded_at` field and passing checks!

Issue #308, opened 2024-02-07
Changes to dbt tests following review from other teams


jeancochrane:
Some changes that we need to make to our dbt tests:

- [x] Update `class_mismatch` tests so that we're only looking at res (i.e. no entry in `comdat`)
- [x] `iasworld.comdat` and excluding any PINs that have a match in `comdat`, is that right?
- [x] Change matching logic so that we're looking for at least one `dweldat` class that matches `pardat`
- [ ] Add a test to ensure there are no minor classes on PINs where there's a major class component
- [ ] Reach out to Brian and discuss tests we can add focused on commercial
- [x] Work with Rodrigo to add tests for land codes and CALP model <-> neighborhood mismatches
    - This needs to check a source spreadsheet against configured neighborhood parameters, but I'm not totally sure if we ingest those parameters into our warehouse yet
    - Spun off here: https://github.com/ccao-data/data-architecture/issues/319
- [x] Update land major class tests to accommodate EE/EX classes
- [x] Work with Tia to expand list of accepted values for a number of characteristics:
  * Exterior walls: 1,2,3,4,6,7,8,9
  * Basement type: 1,2.3,4,5,6,7
  * Basement finish: 1,2,3,4,5,6
  * Attic type: 1,2,3,4,5
  * Garage size: 1-10
- [x] Work with Mirella to add tests to ensure that new occupancy and proration fields are not 0
- [x] Work with Mirella to tweak tests that search for duplicates to screen for role type
    - Fail any record that has role type `RR`
- [x] Update the description for `nbhd must be contained in town_nbhd` test to make the purpose of the test clearer
- [x] Add test to ensure that prorated res PINs with cards have at least one of the cards in both PINs associated with the tieback
    - Spun off here: https://github.com/ccao-data/data-architecture/issues/322
- [ ] Work with Mirella to add test to confirm that `addn` changes match HIEs
    - Test that `addn.value` matches `asmt.tot30` when `addn.lline > 0`
        * Question: How to join these two tables?
- [x] Update address tests to check against `legdat` instead of `owndat`
- [x] Add test to ensure entries in `legdat` with missing addresses have at least a 0 in address number or municipality
- [x] Add test to ensure `legdat` street directionals are correct
- [x] Add test to ensure that 234s have partial finished basements

jeancochrane:
> - Update `class_mismatch` tests so that we're only looking at res (i.e. no entry in `comdat`)

[...]

PR #323, opened 2024-02-27
Changes to dbt tests following review from other teams

jeancochrane:
This PR makes a number of different tweaks to our dbt tests following review from other teams (#308):

- Refactor the tests that check for class codes in `pardat` to ensure that they filter for residential parcels only using a new generic test called `res_class_matches_pardat`
- Expand the list of accepted values for a number of characteristics:
  * Exterior walls: 1,2,3,4,6,7,8,9
  * Basement type: 1,2.3,4,5,6,7
  * Basement finish: 1,2,3,4,5,6
  * Attic type: 1,2,3,4,5
  * Garage size: 1-10
- Add tests to ensure that new calculated occupancy and proration fields are not 0
- Add a test to fail any records in `asmt_all` that have a rolltype of `RR`
- Update the description for `nbhd must be contained in town_nbhd` test to make the purpose of the test clearer
- Update address tests to check against `legdat` instead of `owndat`
- Add tests to ensure entries in `legdat` with missing addresses have at least a 0 in address number and a municipality
- Add test to ensure that 234s have partial finished basements

A couple to-do items from the issue still remain:

* I haven't yet figured out how to test that `addn.value` matches `asmt_all.tot30` when `addn.lline > 0`. The required join between these two tables is just too complex for my current understanding of the data model. If you know how to do it I'd be happy to add that test as part of this PR, and otherwise I'm going to spin it out into a separate issue and work with Mirella to clarify exactly how best to perform the join.
* I'm still waiting to hear back from Ray about what "no minor classes on PINs where there's a major class component" means, but I don't want that to hold up this PR

Connects https://github.com/ccao-data/data-architecture/issues/308.


.github/scripts/format_dbt_test_failures.py
```diff
@@ -524,24 +526,39 @@ def get_tablename_from_node(node: typing.Dict) -> str:
...
 
     # Search for the model that is implicated in this test via the
-    # depends_on key
-    models = node.get("depends_on", {}).get("nodes", [])
-    if not models:
+    # test_metadata.kwargs.model attribute. Note that it is common to use the
+    # elements in the depends_on array for this purpose, but this approach
+    # is fraught, since the order of parents for tests with multiple
+    # dependencies is not clear and can differ:
+    # https://github.com/dbt-labs/dbt-core/issues/6746#issuecomment-1829860236
```
jeancochrane, line 533: Something about the way that `res_class_matches_pardat` performs its join breaks my previous assumption that the last element of `depends_on` always represents the model that the test is attached to (the workbook was reporting the source table for these tests as `pardat`). Using `test_metadata.kwargs.model` is a little bit more hacky, because it requires using a regex to extract the tablename from a string representing a `get_where_subquery()` call, but ultimately I think it's a more correct pointer to the source model than `depends_on`.

aws-athena/views/qc-vw_comdat_class_mismatch.sql
jeancochrane, line 1: I figure the test that uses this view is no longer useful because we do not fully understand the requirements for commercial classes, but I'm down to reinstate it and hide it if you'd prefer to keep it around.

aws-athena/views/qc-vw_oby_class_mismatch.sql
jeancochrane, line 1: This view is no longer necessary to test `oby` ➡️ `pardat` mismatch because we are now using the `res_class_matches_pardat` generic test for that.

dbt/models/iasworld/schema/iasworld.aprval.yml
```diff
@@ -55,11 +55,14 @@ sources:
           - name: aprtot
             description: '{{ doc("column_aprtot") }}'
             tests:
-              - dbt_utils.accepted_range:
+              - accepted_range:
```
jeancochrane, line 58: This test started failing recently, so I switched it to use our new custom macro that allows selecting extra columns.

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -86,6 +97,18 @@ sources:
             description: Reason code for change in assessed value
           - name: rolltype
             description: '{{ doc("column_rolltype") }}'
+            tests:
+              - not_accepted_values:
+                  name: iasworld_asmt_all_rolltype_not_rr
```
jeancochrane, line 102: Mirella suggested this new test.

dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -49,6 +49,26 @@ sources:
             description: Total business living area
           - name: calcexclude
             description: Exclude this record from income calculations
+          - name: calc_meth
+            description: '{{ doc("shared_column_calc_meth") }}'
+            tests:
+              - accepted_values:
+                  name: iasworld_comdat_calc_meth_accepted_values
```
jeancochrane, line 56: This test, along with the `external_*` tests below, represent the tests Mirella requested to check the values of new calculated columns.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -437,11 +518,21 @@ sources:
...
+                    - wen
+                    - mktadj
                   config:
                     where: |
                       CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
                       AND cur = 'Y'
                       AND deactivat IS NULL
+                  meta:
+                    category: incorrect_values
+                    description: mktrsn should only be 5 if mktadj is not null
```
jeancochrane, line 535: This test has also started failing recently, hence we add a category, description, and additional select columns.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1166,3 +1243,26 @@ sources:
               meta:
                 category: missing_values
                 description: user24 (proration rate) should not be null
+          - expression_is_true:
+              name: iasworld_dweldat_bsmt_and_user12_match_234_class
```
jeancochrane, line 1247: This test came out of a desk review discussion.

dbt/models/iasworld/schema/iasworld.land.yml
```diff
@@ -42,6 +42,21 @@ sources:
             description: Agricultural flag
           - name: allocpct
             description: Allocation percentage
+            tests:
+              - not_accepted_values:
+                  name: iasworld_land_allocpct_not_0
```
jeancochrane, line 47: This is another one of the new calculated fields, but it is specific to `land`.

dbt/models/iasworld/schema/iasworld.owndat.yml
```diff
@@ -221,30 +221,8 @@ sources:
             description: '{{ doc("column_whocalc") }}'
           - name: zip1
             description: '{{ doc("shared_column_mail_address_zipcode_1") }}'
-            tests:
-              - not_accepted_values:
-                  name: iasworld_owndat_zip1_has_no_00000_values
```
jeancochrane, line 226: My understanding is that these tests are no longer necessary, since address tests should be defined on `legdat` (which they already are).

dbt/models/iasworld/schema/iasworld.owndat.yml
```diff
@@ -286,27 +264,3 @@ sources:
...
                 - test_qc_exclude_from_workbook
-          - expression_is_true:
-              name: iasworld_owndat_addr1_address_is_well_formed
-              expression: |
-                CASE
-                  WHEN LENGTH(REGEXP_EXTRACT(addr1, '^[N|S|E|W]{0,2}\s{0,1}([0-9]{1,5})',1)) <= 5 THEN TRUE
-                  WHEN REGEXP_LIKE(UPPER(addr1), '^[PO BOX|P\.O\. BOX|P O BOX|BOX|P O BX|PO BX|PO BX|P.O.BOX]') THEN TRUE
-                  WHEN REGEXP_LIKE(UPPER(addr1), '^[C\/O|ONE|TWO|THREE|FOUR|FIVE|SIX|SEVEN|EIGHT|NINE]') THEN TRUE
-                  ELSE FALSE
-                END
```
jeancochrane, line 297: I couldn't really figure out how to translate this to a `legdat` test, given that there's no equivalent of `owndat.addr1` on `legdat` and I can't quite tell what these tests are trying to glean. There's already a version of the first case statement defined on `legdat` (`iasworld_legdat_adrno_length_lte_5`), and the other two statements just seem to be covering alternative formats that can be true if `addr1` doesn't start with the address number, which doesn't really make much sense under a more normalized schema. I'm open to suggestions if you have ideas for how to adapt this, though!

dbt/models/shared_columns.md
```diff
@@ -210,6 +210,14 @@ This is the value after the first round of appeals at the Assessor's
 Office **and** the second round of appeals at the Board of Review.
 {% enddocs %}
 
+## calc_meth
+
+{% docs shared_column_calc_meth %}
+Calculation parameter.
+
+If present, must be `E`.
+{% enddocs %}
```
jeancochrane, line 219: These docs are not the clearest, but they correspond to what Mirella sent me over email.

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -0,0 +1,62 @@
...
+
+    with
+        filtered_model as (
+            select child.*
+            from (select * from {{ model }}) as child
+            left join
+                {{ source("iasworld", "comdat") }} as comdat
+                on child.parid = comdat.parid
+                and child.taxyr = comdat.taxyr
+                {%- if filter_for_res_by_card %} and child.card = comdat.card{% endif %}
```
jeancochrane, line 30: Reading this back is making me concerned that we could potentially generate some duplicates in the `child` model by joining it against `comdat` without `card`. Does that make sense? I can adjust this to group by `parid` and `taxyr` if so.

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -0,0 +1,62 @@
...
+            where comdat.parid is null
+        )
+
+    select
+        array_agg(distinct(filtered_model.{{ column_name }})) as {{ column_name }},
+        {%- for col in additional_select_columns %}
+            max(filtered_model.{{ col }}) as {{ col }},
+        {%- endfor %}
+        max(pardat.class) as pardat_class
```
jeancochrane, line 39: Grabbing the `max` of `pardat.class` should be fine, since `pardat` should only have one entry for each `(parid, taxyr)` pair, but grabbing the `max` of the `filtered_model` fields introduces more of a possibility for misrepresenting the metadata (particularly `who` and `wen`). The alternative here would be to `array_agg` these fields as well, but that felt like it might end up making the data harder to read, so I went with the simplicity of `max` at the risk of losing some information. I'm open to pushback on this choice though!

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -0,0 +1,62 @@
...
+-- `filter_for_res_by_card=false`.
+{% test res_class_matches_pardat(
+    model,
+    column_name,
+    filter_for_res_by_card=true,
+    major_class_only=false,
+    additional_select_columns=[]
+) %}
+
+    {%- set num_class_digits = 1 if major_class_only else 3 %}
```
jeancochrane, line 20: Note that while the original `asmt_all` test compared the first 3 digits of the class codes, the original `dweldat` test didn't (`oby` and `land` both look at major classes). I'm assuming it's fine to perform the `dweldat` test using the first 3 digits of each class for simplicity, but I could be wrong!

dbt/tests/generic/test_res_class_matches_pardat.sql
dfsnow: You actually can't join on card since `comdat` may have different cards than the other tables. I would do an anti-join with any PIN present in `comdat`, so join by `parid` and `taxyr`. This will eliminated mixed-use PINs, so we'll need an extra test for those.

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -0,0 +1,62 @@
...
+    with
+        filtered_model as (
+            select child.*
+            from (select * from {{ model }}) as child
+            left join
+                {{ source("iasworld", "comdat") }} as comdat
+                on child.parid = comdat.parid
+                and child.taxyr = comdat.taxyr
+                {%- if filter_for_res_by_card %} and child.card = comdat.card{% endif %}
+            where comdat.parid is null
```
dfsnow, line 31: **issue:** Need to add the where the clauses used in other tables/tests.

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+    select
+        array_agg(distinct(filtered_model.{{ column_name }})) as {{ column_name }},
+        {%- for col in additional_select_columns %}
+            max(filtered_model.{{ col }}) as {{ col }},
+        {%- endfor %}
```
dfsnow, line 38: **issue:** I would `array_agg()` the class column _and_ the `card`. Drop the `distinct()`, it's okay to have duplicate classes.

dfsnow (commented):
A couple things to fix:

- Card is missing in the class tab for all tables
- The land table test in the class tab is missing all ID info
- Land actually can be exempt on a non-exempt parcel e.g. you can have a `EX` land code on a `211` pardat parcel. However, the reverse is _not_ true (can't have an `EX` pardat parcel and a 200 land class). We can add a separate issue for the latter.
- The duplicate records test looks strange. I think we're missing a filter on the `ASMT` table

Also, I can handle your first outstanding to-do specified in the PR body.

aws-athena/views/qc-vw_comdat_class_mismatch.sql
jeancochrane: I opened up https://github.com/ccao-data/data-architecture/issues/329 to track that work 👍🏻

dbt/models/iasworld/schema/iasworld.owndat.yml
jeancochrane: Makes sense, thanks for the extra context!

dbt/tests/generic/test_res_class_matches_pardat.sql
jeancochrane: Cool, I think the extra test should fall under the purview of https://github.com/ccao-data/data-architecture/issues/329.

dbt/tests/generic/test_res_class_matches_pardat.sql
jeancochrane: Good catch, done in be72b141fcab8524626afaca3ba18ceaaecc5c43!

dbt/tests/generic/test_res_class_matches_pardat.sql
jeancochrane: Done in be72b141fcab8524626afaca3ba18ceaaecc5c43.

dbt/models/iasworld/schema/iasworld.sales.yml
```diff
@@ -202,6 +202,19 @@ sources:
...
+                  agg_func: max
+                - column: substr(saledt, 1, 4)
+                  label: sale_year
+                  agg_func: array_agg
+                - column: who
+                  label: who
+                  agg_func: array_agg
+                - column: wen
+                  label: wen
+                  agg_func: array_agg
```
jeancochrane, line 217: Which of these two ways of specifying different aggregation functions for `additional_select_columns` do you like better? (The other option being function-specific params like `select_columns_aggregated_with_{func}`, as used in `res_class_matches_pardat` above.) I kind of like the flexibility of this one more, but whichever you prefer, I'll put up a fast-follow PR refactoring the other generic test to use the favored approach.

dfsnow: Oh I definitely prefer this one more. In the refactor PR, we may want to include some logic for handling when a mapping is passed with the incorrect columns i.e. column + label, but no agg_func.

dfsnow (approved):
Looks good @jeancochrane. Thanks for crushing this.

dbt/models/iasworld/schema/iasworld.sales.yml
jeancochrane: Great, I just opened an issue to track this here: https://github.com/ccao-data/data-architecture/issues/330

PR #327, opened 2024-02-29
Add `is_active_exe_homeowner`, `n_years_exe_homeowner`, `airport_dnl_total` to dbt

wrridgeway:
These columns are missing descriptions in dbt. Also need to fix the description for ccao_is_corner_lot.

### DIFF
--- a/dbt/models/shared_columns.md
+++ b/dbt/models/shared_columns.md
@@ -800,7 +800,7 @@ purposes. See `ccao.class_dict` for more information
 ## is_corner_lot
 
 {% docs shared_column_is_corner_lot %}
-Corner lot indicator.
+Corner lot indicator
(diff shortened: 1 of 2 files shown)


dbt/models/shared_columns.md
```diff
@@ -797,10 +797,15 @@ percentage of fair cash value at which a property is assessed for taxing
 purposes. See `ccao.class_dict` for more information
 {% enddocs %}
 
+## is_active_exe_homeowner
```
wrridgeway, line 800: I'm not sure if these should be shared columns, but we don't have a `columns.md` in the model... model

dfsnow: They can be shared, that's fine.

PR #338, opened 2024-03-13
Add stage-level classes to `default.vw_pin_history`

wrridgeway:
`default.vw_pin_history` currently has one `class` column, but values for three stages. Since class can change between stages, this makes it impossible to know whether a stage's values should be excluded from a ratio study, if SOPs were to dictate such exclusions.


wrridgeway:
Currently non-stage specific class from pardat is still in `vw_pin_history`. It's probably unnecessary since it only disagrees with board-stage class 3 times in all of our data. These instances are likely errors.

```
select
    pin, year, class, board_class
from "z_ci_337-add-stage-level-classes-to-defaultvw-pin-history_default".vw_pin_history
where class != board_class
```

pin | year | class | board_class
-- | -- | -- | --
18074080240000 | 2022 | 209 | 100
18044020300000 | 2022 | 278 | 100
15261090070000 | 2023 | 2 | EX

aws-athena/views/default-vw_pin_value.sql
```diff
@@ -11,6 +11,13 @@ WITH stage_values AS (
         parid AS pin,
         taxyr AS year,
         -- Mailed values
+        MAX(
```
dfsnow, line 14: **nitpick:** You could probably replace `MAX()` with `ARBITRARY()` for some potential speed gains. But I would test to see if `ARBITRARY()` can return `NULL` first.

aws-athena/views/default-vw_pin_history.sql
```diff
@@ -2,15 +2,18 @@
 SELECT
     vwpv.pin,
     vwpv.year,
-    par.class,
+    REGEXP_REPLACE(par.class, '([^0-9EXR])', '') AS class,
```
dfsnow, line 5: **question (blocking):** What was the motivation for this regex? Were there non `EX`, `RR`, or numeric classes in the data?

dbt/models/shared_columns.md
```diff
@@ -193,6 +193,16 @@ Board of Review assessed value of building from year specified by column
 prefix (or year of observation if not prefixed)
 {% enddocs %}
 
+## board_class
```
dfsnow, line 196:
```suggestion
## board_class

```

dbt/models/shared_columns.md
```diff
@@ -218,6 +228,39 @@ Calculation parameter.
...
 {% enddocs %}
 
+## certified_bldg
+
+{% docs shared_column_certified_bldg %}
+Certified assessed value of building from year specified by column
+prefix (or year of observation if not prefixed)
+{% enddocs %}
+
+## certified_class
```
dfsnow, line 238:
```suggestion
## certified_class

```

dbt/models/shared_columns.md
```diff
@@ -324,6 +344,16 @@ Mailed assessed value of building from year specified by column
 prefix (or year of observation if not prefixed)
 {% enddocs %}
 
+## mailed_class
```
dfsnow, line 347:
```suggestion
## mailed_class

```

dfsnow (commented):
Let's see some dbt tests in here

aws-athena/views/default-vw_pin_history.sql
wrridgeway: yes, '2-99'

wrridgeway:
ARBITRARY seems to work just fine:

```
select count(*)
from "z_ci_337-add-stage-level-classes-to-defaultvw-pin-history_default".vw_pin_value new
left join default.vw_pin_value old
on new.pin = old.pin and new.year = old.year
where new.mailed_bldg != old.mailed_bldg
    OR new.mailed_land != old.mailed_land
    OR new.mailed_tot != old.mailed_tot
    OR new.certified_bldg != old.certified_bldg
    OR new.certified_land != old.certified_land
    OR new.certified_tot != old.certified_tot
    OR new.board_bldg != old.board_bldg
    OR new.board_land != old.board_land
    OR new.board_tot != old.board_tot
```

returns 0

aws-athena/views/default-vw_pin_history.sql
```diff
@@ -2,15 +2,18 @@
 SELECT
     vwpv.pin,
     vwpv.year,
-    par.class,
+    REGEXP_REPLACE(par.class, '[^[:alnum:]]', '') AS class,
```
wrridgeway, line 5: There *are* classes in pardat with hyphens in them. This new filtering should avoid stripping out A & B suffixes from our class codes.

aws-athena/views/default-vw_pin_value.sql
```diff
@@ -11,57 +11,78 @@ WITH stage_values AS (
...
+            END
+        ) AS mailed_class,
+        ARBITRARY(
             CASE
                 WHEN procname = 'CCAOVALUE' AND taxyr < '2020' THEN ovrvalasm2
                 WHEN procname = 'CCAOVALUE' AND taxyr >= '2020' THEN valasm2
             END
         ) AS mailed_bldg,
-        MAX(
+        ARBITRARY(
```
wrridgeway, line 27: Thanks @dfsnow for the suggestion.

dbt/models/default/schema/default.vw_pin_value.yml
```diff
@@ -43,8 +49,48 @@ models:
         description: '{{ doc("shared_column_year") }}'
 
     tests:
+      - not_accepted_values:
+          name: default_vw_pin_value_mailed_class_no_hyphens
+          column_name: mailed_class
+          values: "2-99"
```
wrridgeway, line 55: I'm not entirely sure if wrapping this in double quotes is correct. The test ran without erring with no quotes as well...

wrridgeway:
> Let's see some dbt tests in here

I've added some for `vw_pi_value`. For `vw_pin_history` it's a little difficult since it's primarily a product of `vw_pin_history`. I spoke with @jeancochrane about implementing a test like:

```
select *
from default.vw_pin_value as value
left join default.vw_pin_history as history
    on value.pin = history.pin
    and cast(value.year as int) = cast(history.year as int) - 1
where value.mailed_tot != history.oneyr_pri_mailed_tot
    or value.certified_tot != history.oneyr_pri_certified_tot
    or value.board_tot != history.oneyr_pri_board_tot
```

but it doesn't really fit into our current generics. She offered to update `row_values_match_after_join` to make it able to handle something like this, or I could write a specific test that doesn't depend on a generic. Thoughts @dfsnow ?

aws-athena/views/default-vw_pin_history.sql
dfsnow: **question:** Are other class fields in other views treated the same way? If not, we need to standardize this across views.

aws-athena/views/default-vw_pin_value.sql
dfsnow: TIL about all the other crazy Trino aggregate functions, like [`max_by()`](https://trino.io/docs/current/functions/aggregate.html#max_by).

dbt/models/default/schema/default.vw_pin_value.yml
dfsnow: If there's a dash in it then it's going to be converted into a string anyways, so this should be fine.

dbt/models/default/schema/default.vw_pin_value.yml
```diff
@@ -43,8 +49,48 @@ models:
...
           name: default_vw_pin_value_unique_by_14_digit_pin_and_year
           combination_of_columns:
             - pin
             - year
+      - not_null:
+          name: default_vw_pin_value_mailed_class_not_null
+          column_name: mailed_class
+          config:
+            where: CAST(year AS int) < {{ var('test_qc_year_start') }}
+            error_if: ">289" # as of 2024-03-15
```
dfsnow, line 66: I don't love this system of ignoring nulls. Is there are better way to handle them? Should we include them in the QC sheets? @jeancochrane do you have thoughts here?

aws-athena/views/default-vw_pin_history.sql
wrridgeway: Yes, we will need to update class code cleaning in other views. I've already made an [issue](https://github.com/ccao-data/data-architecture/issues/341) for it.

dbt/models/default/schema/default.vw_pin_value.yml
jeancochrane: @dfsnow If the failure is caused by errors in an underlying iasWorld table, I do think it's worth adding a test to that table so that it gets added to the worksheet and fixed. It's a little bit harder to think about how to handle tests that are intended to test for incorrect logic in our views, but that can be led to fail based on issues in the underlying data; ideally we would be able to fully separate logic tests from data tests, probably by running our logic tests using a set of fixtures that populate tables with test data, but that's a much bigger lift and likely not possible in some cases. My first thought at an 80/20 solution would be to restrict the date range for the test to a time period where we know there are no errors, but we can discuss more during our infrastructure discussion if that option doesn't work.

dfsnow: Alright, let's discuss more during Infrastructure Week.

dfsnow:
> > Let's see some dbt tests in here
> 
> I've added some for `vw_pin_value`. For `vw_pin_history` it's a little difficult since it's primarily a product of `vw_pin_history`. I spoke with @jeancochrane about implementing a test like:
> 
> ```
> select *
> from default.vw_pin_value as value
> left join default.vw_pin_history as history
>     on value.pin = history.pin
>     and cast(value.year as int) = cast(history.year as int) - 1
> where value.mailed_tot != history.oneyr_pri_mailed_tot
>     or value.certified_tot != history.oneyr_pri_certified_tot
>     or value.board_tot != history.oneyr_pri_board_tot
> ```
> 
> but it doesn't really fit into our current generics. She offered to update `row_values_match_after_join` to make it able to handle something like this, or I could write a specific test that doesn't depend on a generic. Thoughts @dfsnow ?

Eh, I think what you've added so far is probably enough.

dbt/models/shared_columns.md
```diff
@@ -324,6 +346,17 @@ Mailed assessed value of building from year specified by column
...
+
+{% docs shared_column_mailed_class %}
+Stage-level property type and/or use at the time of CCAO mailing.
+
+Designates the property type, such as vacant, residential, multi-family,
+agricultural, commercial or industrial. The classification determines the
+percentage of fair cash value at which a property is assessed for taxing
+purposes. See `ccao.class_dict` for more information
```
dfsnow, line 357: **mega-nitpick:** Multi-line doc strings should have a period ending:

```suggestion
purposes. See `ccao.class_dict` for more information.
```

dfsnow (approved):
Looks good other than my tiny nitpick @wrridgeway.

Issue #330, opened 2024-03-01
Refactor `additional_select_columns` to support more configurable naming and aggregation


jeancochrane:
In https://github.com/ccao-data/data-architecture/pull/323 we modified the `additional_select_columns` attributes for both the `unique_combination_of_columns` and `res_class_matches_pardat` test generics to add similar behavior, but we did so in different ways. We should align these approaches, and optionally extend the new functionality to other test generics that support `additional_select_columns`.

See https://github.com/ccao-data/data-architecture/pull/323#discussion_r1509641732 for context.

PR #339, opened 2024-03-13
Refactor `additional_select_columns` dbt test params to support more configurable naming and aggregation

jeancochrane:
Most of our generic dbt test functions support an `additional_select_columns` parameter that instructs the test to select additional columns from the base table for use in test failure reporting. There are a few problems with the current implementation, however:

* Each test generic implements `additional_select_columns` separately
* The `expression_is_true` test uses the older name `select_columns` for the same parameter
* `unique_combination_of_columns` and `res_class_matches_pardat` both permit an object structure for aliasing and aggregation, but this structure is not supported in other generics, and the two generics that do support it implement it differently

This PR represents an attempt to resolve these issues and standardize our implementation of `additional_select_columns` attributes using a new macro `format_additional_select_columns`.

Closes https://github.com/ccao-data/data-architecture/issues/330.

### DIFF
--- a/aws-athena/views/qc-vw_iasworld_sales_price_diff_sale_mydec.sql
+++ b/aws-athena/views/qc-vw_iasworld_sales_price_diff_sale_mydec.sql
@@ -3,6 +3,7 @@
 SELECT
     REGEXP_REPLACE(iasworld.instruno, '[^0-9]', '') AS document_number,
     SUBSTR(iasworld.saledt, 1, 4) AS taxyr,
+    iasworld.parid,
     iasworld.price AS price_iasworld,
     SUBSTR(mydec.year_of_sale, 1, 4) AS year_mydec,
     mydec.line_11_full_consideration AS price_mydec
--- a/dbt/macros/format_additional_select_columns.sql
+++ b/dbt/macros/format_additional_select_columns.sql
@@ -3,15 +3,17 @@
 -- statement.
 --
 -- If an element of the list is a string, the column it represents will
--- be selected using the name of the column as an alias. If instead the
--- element is a dictionary, it can contain the following key-value
--- pairs:
+-- be selected as-is. If instead the element is a dictionary, it can contain
+-- the following key-value pairs:
+--
 -- * `column` (required string): The name of the column to select
 -- * `alias` (optional string): The alias to use for the column
 -- (defaults to `column`)
 -- * `agg_func` (optional string): An aggregation function to use to
 -- select the column (defaults to no aggregation)
 {% macro format_additional_select_columns(additional_select_columns) %}
+    -- Pass execution off to a helper function with a configurable error
+    -- handler, to make it possible to unit test exceptions
     {{
         return(
             _format_additional_select_columns(
@@ -36,11 +38,16 @@
                     )
                 -}}
             {%- else -%}
-                {%- set label = col.label if col.label else col.column -%}
                 {%- if col.agg_func -%}
-                    {{- col.agg_func }} ({{ col.column }}) as {{ label }}
+                    {%- set alias = col.alias if col.alias else col.column -%}
+                    {{- col.agg_func }} ({{ col.column }}) as {{ alias }}
                     {{- trailing_comma }}
-                {%- else -%} {{- col.column }} as {{ label }}{{ trailing_comma }}
+                {%- else -%}
+                    {%- if col.alias -%}
+                        {{- col.column }} as {{ col.alias }}{{ trailing_comma }}
+                    {%- else -%}
+                        {{- col.column }}{{ trailing_comma }}
+                    {%- endif -%}
                 {%- endif -%}
             {%- endif -%}
         {%- else -%} {{ col }}{{ trailing_comma }}
--- a/dbt/macros/format_additional_select_columns.sql
+++ b/dbt/macros/format_additional_select_columns.sql
@@ -27,7 +27,6 @@
     additional_select_columns, raise_error_func
 ) %}
     {%- for col in additional_select_columns -%}
-        {%- set trailing_comma = "" if loop.last else "," %}
         {%- if col is mapping -%}
             {%- if "column" not in col -%}
                 {{-
@@ -41,15 +40,14 @@
                 {%- if col.agg_func -%}
                     {%- set alias = col.alias if col.alias else col.column -%}
                     {{- col.agg_func }} ({{ col.column }}) as {{ alias }}
-                    {{- trailing_comma }}
                 {%- else -%}
-                    {%- if col.alias -%}
-                        {{- col.column }} as {{ col.alias }}{{ trailing_comma }}
-                    {%- else -%} {{- col.column }}{{ trailing_comma }}
+                    {%- if col.alias -%} {{- col.column }} as {{ col.alias }}
+                    {%- else -%} {{- col.column }}
                     {%- endif -%}
                 {%- endif -%}
             {%- endif -%}
-        {%- else -%} {{ col }}{{ trailing_comma }}
+        {%- else -%} {{ col }}
         {%- endif -%}
+        {{- "" if loop.last else "," }}
     {%- endfor -%}
 {% endmacro %}
--- a/dbt/macros/format_additional_select_columns.sql
+++ b/dbt/macros/format_additional_select_columns.sql
@@ -45,8 +45,7 @@
                 {%- else -%}
                     {%- if col.alias -%}
                         {{- col.column }} as {{ col.alias }}{{ trailing_comma }}
-                    {%- else -%}
-                        {{- col.column }}{{ trailing_comma }}
+                    {%- else -%} {{- col.column }}{{ trailing_comma }}
                     {%- endif -%}
                 {%- endif -%}
             {%- endif -%}
--- a/dbt/macros/format_additional_select_columns.sql
+++ b/dbt/macros/format_additional_select_columns.sql
@@ -8,7 +8,7 @@
 --
 -- * `column` (required string): The name of the column to select
 -- * `alias` (optional string): The alias to use for the column
--- (defaults to `column`)
+-- (defaults to `column` or `{agg_func}_{column}`)
 -- * `agg_func` (optional string): An aggregation function to use to
 -- select the column (defaults to no aggregation)
 {% macro format_additional_select_columns(additional_select_columns) %}
@@ -38,7 +38,11 @@
                 -}}
             {%- else -%}
                 {%- if col.agg_func -%}
-                    {%- set alias = col.alias if col.alias else col.column -%}
+                    {%- set alias = (
+                        col.alias
+                        if col.alias
+                        else col.agg_func ~ "_" ~ col.column
+                    ) -%}
                     {{- col.agg_func }} ({{ col.column }}) as {{ alias }}
                 {%- else -%}
                     {%- if col.alias -%} {{- col.column }} as {{ col.alias }}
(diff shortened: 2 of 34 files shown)


aws-athena/views/qc-vw_iasworld_sales_price_diff_sale_mydec.sql
```diff
@@ -3,6 +3,7 @@
 SELECT
     REGEXP_REPLACE(iasworld.instruno, '[^0-9]', '') AS document_number,
     SUBSTR(iasworld.saledt, 1, 4) AS taxyr,
+    iasworld.parid,
```
jeancochrane, line 6: Neglecting to select this column was resulting in output for this test that was missing a parid, as well as township (since that's queried off of parid and taxyr). Not directly related to this PR, but I went ahead and fixed it here anyway.

dbt/macros/format_additional_select_columns.sql
```diff
@@ -0,0 +1,55 @@
...
+--
+-- If an element of the list is a string, the column it represents will
+-- be selected as-is. If instead the element is a dictionary, it can contain
+-- the following key-value pairs:
+--
+-- * `column` (required string): The name of the column to select
+-- * `alias` (optional string): The alias to use for the column
+-- (defaults to `column`)
+-- * `agg_func` (optional string): An aggregation function to use to
+-- select the column (defaults to no aggregation)
```
jeancochrane, line 13: An additional subtlety that is left out of this docstring: If not specifying an `agg_func`, the `alias` attribute is actually not necessary, meaning there's no real reason to use the dict structure unless you need aggregation. Nonetheless, it's supported here for the sake of maximum flexibility.

As an example, these two configurations will produce identical results:

```yaml
additional_select_columns:
  - foo as bar
  - baz as qux
```

```yaml
additional_select_columns:
  - column: foo
    alias: bar
  - column: baz
    alias: qux
```

It's necessary to support `alias` in the dict structure in order to allow for aggregation with a configurable name; we could consider failing if a dict structure is used but `agg_func` is not present, which would have the benefit of enforcing one standard approach, but I figure making the interface forgiving is more important than enforcing a specific standard. This also leaves open the possibility of supporting additional transformations on top of `agg_func` in the future while ensuring that such a change would be easy to make.

dbt/macros/tests/test_generate_schema_name.sql
```diff
@@ -12,10 +12,6 @@
     {% endif %}
 {% endmacro %}
 
-{% macro mock_raise_compiler_error(_error) %}
-    {{ return("Compiler error raised") }}
-{% endmacro %}
```
jeancochrane, line 17: I realized in the course of developing this PR that it's trivial to adjust `mock_raise_compiler_error` to return the error string for the purposes of making sure that a specific error was raised, so I did this and moved the updated macro to `dbt/macros/tests/utils.sql` so that it could be used by the new `format_additional_select_columns` test.

dbt/tests/generic/test_column_length.sql
```diff
@@ -8,14 +8,16 @@
...
+        additional_select_columns
+    ) %}
 
     {%- set length_columns = [] %}
     {%- for column in columns %}
         {%- set length_columns = length_columns.append([column, "len_" + column]) %}
     {%- endfor %}
 
-    {%- set select_columns = additional_select_columns %}
+    {%- set select_columns = [] %}
```
jeancochrane, line 20: The previous version of this code was somewhat convoluted; we set `columns_csv` to a parsed version of `additional_select_columns` on line 11, but then redefined `columns_csv` on line 36 to a list of transformed strings including (among others) columns from `additional_select_columns`. I've gone ahead and simplified this logic so that we only parse `additional_select_columns` once, and then template it directly in the `SELECT` query alongside `columns_csv` on lines 41-49 below.

dbt/tests/generic/test_expression_is_true.sql
```diff
@@ -1,13 +1,17 @@
...
--- for failing rows.
-{% test expression_is_true(model, column_name, expression, select_columns=[]) %}
-    {%- set select_columns_csv = select_columns | join(", ") -%}
+-- an optional `additional_select_columns` option representing an array of
+-- columns to select for failing rows in addition to `column_name`. If no
+-- `additional_select_columns` array is provided, defaults to selecting the
+-- column represented by `column_name`; if `column_name` is also missing, falls
+-- back to selecting 1 for failing rows.
+{% test expression_is_true(
+    model, column_name, expression, additional_select_columns=[]
```
jeancochrane, line 10: Previously, we had named this attribute `select_columns` rather than `additional_select_columns` because it was rare for `expression_is_true` to be defined on a column as opposed to on a model; when defined on a model, `additional_select_columns` might be a confusing name because no other columns are being selected by default.  (When defined on a column, however, the column corresponding to `column_name` is selected by default, so the parameter name is clear.)

However, I think standardization of the generic interfaces is more important than maximum clarity for `expression_is_true` in particular; there is also no guarantee that the balance of model vs. column tests won't change for `expression_is_true` in the future. As such, I've gone ahead and refactored the parameter name here to align it with the rest of the generics.

jeancochrane:
Oops, I marked this PR as ready for review before running the final check (generating the output workbook and making sure it still looks correct). Doing that now; no need to review just yet.

dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -88,21 +88,35 @@ sources:
                   name: iasworld_comdat_chgrsn_eq_5_mktadj_not_null
                   expression: |
                     != '5' OR mktadj IS NOT NULL
+                  additional_select_columns:
+                    - parid
+                    - taxyr
+                    - card
+                    - who
+                    - wen
+                    - mktadj
```
jeancochrane, line 97: This test, along with `iasworld_comdat_class_in_ccao_class_dict` and `iasworld_dweldat_class_in_ccao_class_dict`, have started failing since the last time we published a workbook, so they need descriptions and additional select columns. I went ahead and bundled them into this PR, since this test makes use of the new format for `additional_select_columns` when using `expression_is_true`.

dbt/macros/tests/test_format_additional_select_columns.sql
```diff
@@ -0,0 +1,83 @@
...
+        )
+    }}
+{% endmacro %}
+
+{% macro test_format_additional_select_columns_dict_element_array_agg() %}
+    {{
+        assert_equals(
+            "test_format_additional_select_columns_dict_element_array_agg",
+            format_additional_select_columns([{"column": "foo", "agg_func": "max"}]),
+            "max (foo) as foo",
```
dfsnow, line 68: **suggestion (non-blocking):** I'm not sure that I like this as default behavior since it's not how SQL behaves in most dialects. IMO, we should raise an exception if there's an aggregation function and no alias. Open this behavior if you think it's more flexible though!

dbt/macros/format_additional_select_columns.sql
```diff
@@ -0,0 +1,55 @@
...
+                {%- if col.agg_func -%}
+                    {%- set alias = col.alias if col.alias else col.column -%}
+                    {{- col.agg_func }} ({{ col.column }}) as {{ alias }}
+                    {{- trailing_comma }}
+                {%- else -%}
+                    {%- if col.alias -%}
+                        {{- col.column }} as {{ col.alias }}{{ trailing_comma }}
+                    {%- else -%} {{- col.column }}{{ trailing_comma }}
+                    {%- endif -%}
+                {%- endif -%}
```
dfsnow, line 50: **nitpick:** Can you just move `{{ trailing_comma }}` to here and remove it from the other lines?

```suggestion
                {%- endif -%}
                {{ trailing_comma }}
```

dbt/macros/tests/test_format_additional_select_columns.sql
```diff
@@ -0,0 +1,83 @@
...
+{% macro test_format_additional_select_columns_string_element() %}
+    {{
+        assert_equals(
+            "test_format_additional_select_columns_string_element",
+            format_additional_select_columns(["foo"]),
+            "foo",
+        )
+    }}
+{% endmacro %}
+
```
dfsnow, line 20: **suggestion:** Could probably add a test case for the `foo as bar` case:

```suggestion

{% macro test_format_additional_select_columns_string_element() %}
    {{
        assert_equals(
            "test_format_additional_select_columns_string_element",
            format_additional_select_columns(["foo as bar"]),
            "foo as bar",
        )
    }}
{% endmacro %}

```

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -39,7 +39,11 @@ sources:
                     description: class code must be valid
               - res_class_matches_pardat:
                   name: iasworld_asmt_all_class_matches_pardat_class
-                  select_columns_aggregated_with_max: *select-columns
+                  additional_select_columns:
+                    - {"column": "taxyr", "agg_func": "max"}
+                    - {"column": "parid", "agg_func": "max"}
+                    - {"column": "who", "agg_func": "max"}
+                    - {"column": "wen", "agg_func": "max"}
```
dfsnow, line 46: **question (blocking):** Is this just a compacted way of doing the following?

```yaml
- column: taxyr
  agg_func: max
```

If so, I think I prefer the YAML case even if it's more verbose. If we do want to keep the JSON string we should add a test case for it.

dfsnow (approved):
This looks awesome @jeancochrane, see comments for minor edits.

dbt/macros/tests/test_format_additional_select_columns.sql
jeancochrane: Could you say more about what you mean by this not being how SQL behaves in most dialects? My thinking is that it's convenient to have `alias` default to `column` since in most of the cases where we're using it we don't need an alias that's different from the column name, but if there's something about that behavior that's confusing or contrary to the SQL standard I'm happy to raise an error instead.

dbt/models/iasworld/schema/iasworld.asmt_all.yml
jeancochrane: Yup, and I switched to the more verbose form for this case and a few other instances in 3109ebac03f62c9bfaca78c28c2151b41d474b65! I'll start expanding these out going forward, I agree that it's easier to read.

dbt/macros/tests/test_format_additional_select_columns.sql
jeancochrane: That makes sense, added in 3109ebac03f62c9bfaca78c28c2151b41d474b65.

dbt/macros/tests/test_format_additional_select_columns.sql
dfsnow: Something like:

```sql
SELECT MAX(major_class_code)
FROM ccao.class_dict
```

will return a column named `_col0` in Athena or `max(major_class_code)` in other SQL dialects. In other words, it _doesn't_ return the column name that is input to the aggregation function, as this macro does. IMO the the macro behavior is slightly confusing/not explicit since it doesn't match SQL.

jeancochrane: I see, thanks for the clarification! What do you think about defaulting to `{agg_func}_{column}`, e.g. `max_major_class_code` in this case? I think that might more closely match the typical SQL behavior you've described here as opposed to raising an exception.

jeancochrane: I went ahead and implemented this approach in 763d1ebd1ca827a3ba798152eea4504f55d22198.

PR #343, opened 2024-03-13
Add `location.neighborhood_group` table via seed

jeancochrane:
This PR updates the dbt DAG to add a new [seed table](https://docs.getdbt.com/docs/build/seeds), `location.neighborhood_group`, that can be used by the sales val flagging script to group neighborhoods into submarkets.

This is a companion PR to https://github.com/ccao-data/model-sales-val/pull/107 and should be merged before that PR is merged.


.gitignore
```diff
@@ -6,10 +6,9 @@
 .DS_Store
 *.geojson
 *.parquet
-*.csv
```
jeancochrane, line 9: I removed this global restriction since we do in fact want CSVs in the `dbt/seeds` directory to be stored in version control. We could also consider configuring an exception for the `dbt/seeds/` directory, but that seemed a little more complicated and I didn't feel like messing with it right now; instead I just ignored a specific expected CSV file in the `dbt/.gitignore` file below.

dbt/dbt_project.yml
```diff
@@ -61,3 +61,8 @@ models:
       +schema: qc
 tests:
   +schema: test_failure
+
+seeds:
+  athena:
+    location:
+      +schema: location
```
jeancochrane, line 68: It'd be nice if we could have seeds inherit the schema config from the models, so that we don't have to duplicate them. Perhaps we could rig something up with anchors and aliases? I didn't bother with it here, however.

dbt/seeds/location/docs.md
```diff
@@ -0,0 +1,23 @@
...
+  state of the neighborhood groups at a specific moment in time. To do so, query
+  for all records where `updated_at` is below a certain date, group by `nbhd`,
+  and select the record with the highest value for `version` within the group.
+- Instructions for how to update the seed that generates this table:
+  - When **adding new neighborhoods**, append the rows to the CSV file with a
+    `version` of 1
+  - When **updating neighborhoods that already have a group**, append the rows
+    to the CSV file and increment the `version` by 1
+  - In the case of **all new rows**, set the `updated_at` column to the current
+    date
```
jeancochrane, line 20: This versioning scheme is similar to the type 2 slowly-changing dimension approach that dbt uses for [snapshots](https://docs.getdbt.com/docs/build/snapshots). I considered implementing versioning with snapshots instead -- the idea was something like an additional `location.neighborhood_group_history` table generated and updated by a snapshot that we could use to record the history of the groups whenever we change them -- but I didn't like the fact that the snapshots wouldn't be totally replicable from scratch in our DAG without the history of the table.

This all ties into deeper concerns about incremental tables that I'm hoping we can discuss during our infrastructure chat next week, but for now I went with the simplest option: Require the code author to manually implement the versioning any time they make a change. This is not ideal but I'm hoping that these groupings will change infrequently enough that it won't be a big deal (or that we'll have resolved our snapshot/incremental table design issue and we can refactor this by the time the next group of neighborhoods gets defined).

dbt/seeds/location/schema.yml
```diff
@@ -0,0 +1,12 @@
...
+    description: '{{ doc("seed_neighborhood_group") }}'
+    columns:
+      - name: nbhd
+        description: '{{ doc("shared_column_nbhd_code") }}'
+      - name: group_name
+        description: Name of the group that this neighborhood is in.
+      - name: updated_at
+        description: Timestamp of the last record update.
+      - name: version
+        description: Incrementing version of the record.
```
jeancochrane, line 12: The goal of this data model is to be flexible enough to allow us to adapt to speculative new features, without committing ourselves up front to features that turn out to be unnecessary. Some of the features that we have in mind are:

* We might need to change groupings based on feedback from Valuations or to reflect changes in the underlying markets. In that case, the `version` and `updated_at` columns will help preserve the history of the table so that we can always determine the exact set of groupings that were used to generate a set of flags without having to match timestamps on git commits.
* At some point we may need different groupings for res vs. condos. In this case, I expect we could add a `type` column to distinguish the two groups.
* We might decide to set a group to be equivalent to a township, as was the case in the original sales val runs. In that case, we could add a record for each neighborhood in a township using the same `group_name`.

dbt/seeds/location/schema.yml
```diff
@@ -0,0 +1,47 @@
...
+        tests:
+          - not_null:
+              name: seed_neighborhood_group_version_not_null
+              meta:
+                description: version should not be null
+          - accepted_range:
+              name: seed_neighborhood_group_version_gte_1
+              min_value: 1
+              meta:
+                description: version should be >= 1
```
jeancochrane, line 38: The goal of this data model is to be flexible enough to allow us to adapt to speculative new features, without committing ourselves up front to features that turn out to be unnecessary. Some of the features that we have in mind are:

* We might need to change groupings based on feedback from Valuations or to reflect changes in the underlying markets. In that case, the `version` and `updated_at` columns will preserve the history of the table so that we can always determine the exact set of groupings that were used to generate a set of flags without having to match timestamps on git commits.
* At some point we may need different groupings for res vs. condos. In this case, I expect we could add a `type` column to distinguish the two groups.
* We might decide to set a group to be equivalent to a township, as was the case in the original sales val runs. In that case, we could add a record for each neighborhood in a township using the same `group_name`.

jeancochrane: Also, see [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/8285811333/job/22674356192) for evidence of these tests passing.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -62,14 +62,14 @@ jobs:
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
               echo "Running build on manually selected resources"
-              dbt run -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR"
+              dbt build -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR" --resource-types model seed
```
jeancochrane, line 65: [`dbt run`](https://docs.getdbt.com/reference/commands/run) only builds models, but now that our DAG includes seeds, we need to switch to [`dbt build`](https://docs.getdbt.com/reference/commands/build) instead. Note that the `--resource-types` flag is quite important for us, since our DAG is big enough that we don't want to run tests when we rebuild resources.

dbt/dbt_project.yml
dfsnow: I think it's fine for it to be a little bit redundant!

dbt/seeds/location/docs.md
```diff
@@ -0,0 +1,29 @@
+# neighborhood_group
+
+{% docs seed_neighborhood_group %}
+A mapping of neighborhood codes to groups that are used for sales
+validation.
+
+These mappings are defined by the Valuations team and then delivered to the
+Data team in the form of an Excel workbook. The data team then manually
```
dfsnow, line 8:
```suggestion
Data team in the form of an Excel workbook. The Data team then manually
```

dfsnow (approved):
This looks great @jeancochrane, nice idea! We can definitely talk more about the versioning and incremental table updates next week. I also have a few other tables that we should probably transform into seeds. I'll make a separate PR to do so.

.gitignore
jeancochrane: Thanks for figuring that out, done in 406bf885a2f78007d4435c8f31623b8138ff389d!

dbt/seeds/location/docs.md
jeancochrane: Good catch, fixed in 8c4fcde6bf6c7b5cb7ca769b21983047852b9f56.

PR #350, opened 2024-03-18
Change `geography_id` type to string in `ratio_stats.py`

wrridgeway:
Switching `geography_id` to string in [ratio_stats.py](https://github.com/ccao-data/data-architecture/blob/master/aws-glue/jobs/reporting-ratio_stats.py) would make reporting easier for @ccao-jardine


ccao-jardine:
Works beautifully on my end, thanks!

jeancochrane (approved):
Good by me!

Issue #344, opened 2024-03-14
Add descriptions and selected columns for all non-failing tests


jeancochrane:
So far we've added descriptions and select columns for all _failing_ tests (https://github.com/ccao-data/data-architecture/pull/281) but we don't have descriptions or select columns for tests that aren't failing. This means that when a test starts failing, it silently appears in the workbook with garbled metadata, and we have to notice it and update its metadata manually in order to not produce a corrupted workbook.

Let's go through and make sure all our tests have these attributes so that any tests that start failing will appear correctly in the workbook.

Issue #263, opened 2023-12-05
Tweak dbt tests metadata to support output workbook


jeancochrane:
Following up on #253, we will need to adjust dbt test metadata to ensure that the output is useable by external teams. A couple changes we need to make:

* Addition of a `meta.display_name` attribute with a human-readable name for the test (used to populate that Excel sheet name)
* Addition of a `meta.description` attribute with a short description of the test (used to populate the optional first row of the Excel sheet)
* Addition of extra columns to test queries to ensure that the output data can be read by a reviewer and acted upon

PR #351, opened 2024-03-19
Add descriptions and selected columns for all non failing tests

jeancochrane:
This PR adds `meta.description` and `additional_select_columns` attributes to all remaining dbt QC tests that are not fully configured following https://github.com/ccao-data/data-architecture/pull/281. This mostly represents non-failing tests, and I've confirmed via inspection of the workbook that the test output remains the same following this change. Note that since these attributes are only needed for outputting the QC test workbook, we only add them to tests that are configured with the `test_qc_*` tag and do not also have the `test_qc_exclude_from_workbook` tag, since these are the only tests that get output to the workbook.

Closes https://github.com/ccao-data/data-architecture/issues/344 and closes https://github.com/ccao-data/data-architecture/issues/263.

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -91,7 +91,10 @@ sources:
                   name: iasworld_addn_cur_in_accepted_values
                   values: ['Y', 'D']
                   additional_select_columns: *select-columns
-                  config: *unique-conditions
+                  config:
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND deactivat IS NULL
                   meta:
                     description: cur should be 'Y' or 'D'
           - name: deactivat
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -94,7 +94,6 @@ sources:
                   config:
                     where: |
                       CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
-                      AND deactivat IS NULL
                   meta:
                     description: cur should be 'Y' or 'D'
           - name: deactivat
--- a/dbt/models/iasworld/schema/iasworld.addn.yml
+++ b/dbt/models/iasworld/schema/iasworld.addn.yml
@@ -46,11 +46,21 @@ sources:
             tests:
               - not_null:
                   name: iasworld_addn_card_not_null
+                  additional_select_columns: &select-columns-no-card
+                    - taxyr
+                    - parid
+                    - who
+                    - wen
                   config: *unique-conditions
-              - dbt_utils.accepted_range:
+                  meta:
+                    description: card should not be null
+              - accepted_range:
                   name: iasworld_addn_card_gte_1
                   min_value: 1
+                  additional_select_columns: *select-columns-no-card
                   config: *unique-conditions
+                  meta:
+                    description: card should be >= 1
           - name: cdu
             description: '{{ doc("column_cdu") }}'
           - name: chgrsn
@@ -80,17 +90,23 @@ sources:
               - accepted_values:
                   name: iasworld_addn_cur_in_accepted_values
                   values: ['Y', 'D']
+                  additional_select_columns: *select-columns
                   config: *unique-conditions
+                  meta:
+                    description: cur should be 'Y' or 'D'
           - name: deactivat
             description: '{{ doc("column_deactivat") }}'
           - name: depr
             description: Percent good from tables (based on age+cdu)
             tests:
-              - dbt_utils.accepted_range:
+              - accepted_range:
                   name: iasworld_addn_depr_between_0_and_100
                   min_value: 0
                   max_value: 100
+                  additional_select_columns: *select-columns
                   config: *unique-conditions
+                  meta:
+                    description: depr should be between 0 and 100
           - name: eff_area
             description: Support value per line
           - name: effageovr
@@ -124,7 +140,10 @@ sources:
             tests:
               - not_null:
                   name: iasworld_addn_lline_not_null
+                  additional_select_columns: *select-columns
                   config: *unique-conditions
+                  meta:
+                    description: lline should not be null
           - name: loaded_at
             description: '{{ doc("shared_column_loaded_at") }}'
           - name: lower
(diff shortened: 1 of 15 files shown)


dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1254,8 +1343,9 @@ sources:
                 - user20
               config: *unique-conditions
               meta:
-                category: missing_values
-                description: user24 (proration rate) should not be null
```
jeancochrane, line 1258: This test was previously miscategorized due to a copy/paste typo.

dbt/models/iasworld/schema/iasworld.sales.yml
```diff
@@ -42,6 +42,10 @@ sources:
                     - parid
                     - who
                     - wen
+                  config:
+                    where: |
+                      substr(saledt, 1, 4) >= '2011'
+                      AND deactivat IS NULL
```
jeancochrane, line 48: This date filter does not match the typical pattern of `unique-conditions`, but it matches how `unique-conditions` is configured for the rest of the `sales` tests:

https://github.com/ccao-data/data-architecture/blob/2d81055921bde290a794efd116296bcb5d5ccb9a/dbt/models/iasworld/schema/iasworld.sales.yml#L68-L72

dbt/tests/generic/test_sequential_values.sql
```diff
@@ -0,0 +1,53 @@
+-- Override dbt_utils.sequential_values to support `additional_select_columns` parameter
```
jeancochrane, line 1: See the original test definition for comparison: https://github.com/dbt-labs/dbt-utils/blob/23da1f424398696e25656dc6f89066b9a5d13794/macros/generic_tests/sequential_values.sql#L1-L42

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -221,6 +232,14 @@ sources:
               - accepted_values:
                   name: iasworld_dweldat_cur_in_accepted_values
                   values: ['Y', 'D']
+                  additional_select_columns: *select-columns
+                  config:
+                    where: |
+                        CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                        AND deactivat IS NULL
+                        AND class NOT IN ('201', '213', '218', '219', '220', '221', '224', '225', '236', '240', '241', '290', '294', '297')
```
dfsnow, line 240: **question (non-blocking):** What's the need for the additional class filter on `cur` in `dweldat`?

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -920,7 +968,11 @@ sources:
               - accepted_values:
                   name: iasworld_dweldat_char_renovation_in_accepted_values
                   values: ['0', '1']
+                  additional_select_columns: *select-columns
                   config: *unique-conditions
+                  meta:
+                    description: >
+                      user3 (property renovation indicator) should be '0' or '1'
```
dfsnow, line 975: **suggestion:** It might be useful to update the user field tests to reflect how they're displayed on the iasWorld dwellings tab. For example, this one would be:

```suggestion
                      user3 (Renovated) should be '0' or '1'
```

dbt/models/iasworld/schema/iasworld.sales.yml
dfsnow: That's fine. `SALES` is one of the few critical tables that doesn't have a `taxyr` per record, hence the weird filter.

dbt/models/qc/schema.yml
```diff
@@ -189,35 +189,56 @@ models:
                 - who
                 - wen
               config: *time-filter
+              meta:
+                table_name: asmt
+                description: val01 should not be negative
```
dfsnow, line 194: **suggestion:** Similar to the fields for `dweldat` it might be handy here to add the corresponding text from the `Asmt` tab of iasWorld. For example, I think this would be:

```suggestion
                description: val01 (FARM LAND) should not be negative
```

dfsnow (commented):
Great job @jeancochrane. I know this must've been a slog, thanks for grinding through it!

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: This is just copied from the `unique-conditions` definition for `dweldat`:

https://github.com/ccao-data/data-architecture/blob/d7c7fc05df3ea1635b5efbbe593288b7f5a090c6/dbt/models/iasworld/schema/iasworld.dweldat.yml#L61-L66

But it seems potentially unnecessary in this case, so I'll plan to remove it along with the `deactivat` filter.

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Good idea, I opened up https://github.com/ccao-data/data-architecture/issues/354 to track this work as a separate issue since it'll be a bit of a slog.

dbt/models/qc/schema.yml
jeancochrane: This is included in the scope of #354.

dbt/models/iasworld/schema/iasworld.addn.yml
```diff
@@ -199,21 +232,34 @@ sources:
...
                   name: iasworld_addn_seq_all_sequential_exist
                   group_by_columns:
                     - parid
                     - taxyr
                     - card
                     - lline
+                  additional_select_columns:
+                    - who
+                    - wen
+                  config: *unique-conditions
```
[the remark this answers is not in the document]
jeancochrane: Update: I tested it out, and it seems like we actually don't currently have any failures even with the `unique-conditions` config. I'm going to leave it on for the sake of consistency, then, and we can re-evaluate later if it starts to fail in unpredictable ways.

dbt/models/iasworld/schema/iasworld.addn.yml
```diff
@@ -80,17 +90,26 @@ sources:
               - accepted_values:
                   name: iasworld_addn_cur_in_accepted_values
                   values: ['Y', 'D']
-                  config: *unique-conditions
```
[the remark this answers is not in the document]
jeancochrane: Fixed in 72cb53c188175b3b53e5f12326e902837270fc39.

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Done in 72cb53c188175b3b53e5f12326e902837270fc39.

dfsnow (approved):

Issue #354, opened 2024-03-20
Update test descriptions to use iasWorld column names


jeancochrane:
We often make use of `config.description` in dbt tests to hint at the underlying meaning of columns with opaque names. These column hints are included as parentheticals that follow the technical name for the column. For example, `attic type` is the column hint for the `attic` column in this test:

https://github.com/ccao-data/data-architecture/blob/d7c7fc05df3ea1635b5efbbe593288b7f5a090c6/dbt/models/iasworld/schema/iasworld.dweldat.yml#L73-L75

It would be clearer to end users if these hints used the name of the column as it is displayed in the iasWorld UI, rather than the internal description of the column that we use for our purposes.

This change will be most useful for `dweldat` tests and tests in the `qc` schema that refer to the `asmt` test, but we should audit the rest of the tests as well.

PR #355, opened 2024-03-20
Add iasWorld colnames to descriptions of tests in qc, dweldat, comdat, and oby

jeancochrane:
This PR is a fast follow on https://github.com/ccao-data/data-architecture/pull/351 to make the `config.description` attributes easier to read for tests in the `qc`, `dweldat`, `comdat`, and `oby` models. The more opaque fields on these models now include a hint indicating the iasWorld name of the field in the `config.description` attribute which gets templated into the dbt test failure workbook. This should make it a little bit easier for workbook users to act on problems in iasWorld when they are reported in the failure workbook.

Closes https://github.com/ccao-data/data-architecture/issues/354.

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.comdat.yml
+++ b/dbt/models/iasworld/schema/iasworld.comdat.yml
@@ -74,7 +74,8 @@ sources:
                       AND deactivat IS NULL
                       AND calc_meth IS NOT NULL
                   meta:
-                    description: calc_meth should be 'E' if not null
+                    description: >
+                      calc_meth (Calculation Method) should be 'E' if not null
           - name: card
             description: '{{ doc("column_card") }}'
             tests:
@@ -203,7 +204,7 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: >
-                      external_calc_rcnld (calculated full NMV) should not be 0
+                      external_calc_rcnld (Net Market Value) should not be 0
           - name: external_occpct
             description: '{{ doc("shared_column_external_occpct") }}'
             tests:
@@ -215,7 +216,7 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: >
-                      external_occpct (occupancy percentage) should not be 0
+                      external_occpct (Occupancy %) should not be 0
           - name: external_propct
             description: '{{ doc("shared_column_external_propct") }}'
             tests:
@@ -227,7 +228,7 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: >
-                      external_propct (proration percentage) should not be 0
+                      external_propct (Proration %) should not be 0
           - name: external_rcnld
             description: '{{ doc("shared_column_external_rcnld") }}'
             tests:
@@ -239,7 +240,7 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: >
-                      external_rcnld (calculated full FMV) should not be 0
+                      external_rcnld (Full Market Value) should not be 0
           - name: gfact
             description: Grade factor
           - name: grade
--- a/dbt/models/iasworld/schema/iasworld.dweldat.yml
+++ b/dbt/models/iasworld/schema/iasworld.dweldat.yml
@@ -65,7 +65,7 @@ sources:
                       AND deactivat IS NULL
                       AND class NOT IN ('201', '213', '218', '219', '220', '221', '224', '225', '236', '240', '241', '290', '294', '297')
                   meta:
-                    description: attic (attic type) should not be null
+                    description: attic (Attic Type) should not be null
               - accepted_values:
                   name: iasworld_dweldat_attic_in_accepted_values
                   values: ['1', '2', '3', '4', '5']
@@ -73,7 +73,7 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: >
-                      attic (attic type) should be an integer between 1 and 5
+                      attic (Attic Type) should be an integer between 1 and 5
           - name: atticarea
             description: Attic area
           - name: atticval
@@ -98,7 +98,7 @@ sources:
                   additional_select_columns: *select-columns
                   config: *unique-conditions
                   meta:
-                    description: bsmt (basement type) should not be null
+                    description: bsmt (Basement Type) should not be null
               - accepted_values:
                   name: iasworld_dweldat_bsmt_in_accepted_values
                   values: ['1', '2', '3', '4', '5', '6', '7']
@@ -106,7 +106,7 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: >
-                      bsmt (basement type) should be an integer between 1 and 7
+                      bsmt (Basement Type) should be an integer between 1 and 7
           - name: bsmtarea
             description: Unfinished basement area
           - name: bsmtcar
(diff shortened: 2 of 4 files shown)


dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -74,7 +74,8 @@ sources:
                       AND deactivat IS NULL
                       AND calc_meth IS NOT NULL
                   meta:
-                    description: calc_meth should be 'E' if not null
+                    description: >
+                      calc_meth (Calculation Method) should be 'E' if not null
```
jeancochrane, line 78: Note that I've only updated the descriptions for these new calculated fields in `comdat` and `oby`, since they are identical to the ones in `dweldat` that are also fixed as of this PR. If you think we should do a full pass on _all_ fields in `comdat` and `oby`, and not just these new calculated fields, I'm happy to do so as well; this PR is just following the instructions of https://github.com/ccao-data/data-architecture/issues/354 somewhat narrowly.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -65,15 +65,15 @@ sources:
                       AND deactivat IS NULL
                       AND class NOT IN ('201', '213', '218', '219', '220', '221', '224', '225', '236', '240', '241', '290', '294', '297')
                   meta:
-                    description: attic (attic type) should not be null
+                    description: attic (Attic Type) should not be null
```
jeancochrane, line 68: Where a field is visible in iasWorld, we now match the iasWorld capitalization scheme in its description.

dbt/models/qc/schema.yml
```diff
@@ -310,319 +318,319 @@ models:
...
+                description: val41 (BOR FACT BLDG) should not be negative
               <<: *test-non-negative
       - name: val42
         tests:
           - accepted_range:
               name: qc_vw_neg_asmt_value_val42_gte_0
               meta:
                 table_name: asmt
-                description: val42 should not be negative
+                description: val42 (40. .:) should not be negative
```
jeancochrane, line 529: Some of these fields don't have verbose names in iasWorld; in this case, `val42` maps to a single dot (`.`). I've styled it as `40. .:` to more closely match the way the field appears in the UI, but I'm open to alternatives if this is confusing.

I wonder if the lack of a name is an indication that this field isn't used?

dbt/models/qc/schema.yml
```diff
@@ -131,7 +135,9 @@ models:
           meta:
             table_name: asmt
             category: relationships
-            description: FMV should not increase more than $500k YoY
+            description: >
+              cur_year_fmv (ASMT TOTAL) should not increase more than $500k
+              over pri_year_fmv
```
dfsnow, line 140: **suggestion (non-blocking):** Considering Brian's note about trying to make tests that are high signal, we might want to remove this one and others that look for % changes.

dbt/models/qc/schema.yml
```diff
@@ -310,319 +318,319 @@ models:
...
+                description: val16 (DEMOLITIONS ASSD) should not be negative
               <<: *test-non-negative
       - name: val17
         tests:
           - accepted_range:
               name: qc_vw_neg_asmt_value_val17_gte_0
               meta:
                 table_name: asmt
-                description: val17 should not be negative
+                description: val17 (STATE RPT OMITTED)  should not be negative
```
dfsnow, line 329:
```suggestion
                description: val17 (STATE RPT OMITTED) should not be negative
```

dbt/models/qc/schema.yml
dfsnow: I wish I knew. I assume it's unused but we should confirm with MV

dfsnow (approved):
Sweet! Thanks for handling this @jeancochrane.

dbt/models/qc/schema.yml
jeancochrane: Good thinking! I made this change in b57d1a1f538988d00c1c4a7c6a2e6e9ccb539920, and also got rid of the `qc.vw_incorrect_asmt_value` view and moved its remaining tests onto `asmt_all` and `pardat` (as we discussed offline).

One subtlety to this change is that `qc.vw_incorrect_asmt_value` was previously defining these tests on a rehydrated version of the missing-but-implied `asmt` table:

https://github.com/ccao-data/data-architecture/blob/e56518dd22b45d88d9803829541788856f62bbcf/aws-athena/views/qc-vw_incorrect_asmt_value.sql#L38-L51

In moving the remaining tests to `asmt_all`, I switched to use the correct filtering for `asmt_all` (filtering on `(procname, valclass, rolltype, deactivat)` instead of `(cur, deactivat)`) which resulted in slightly different failure counts for all tests except `iasworld_pardat_class_equals_luc:

| new test name | old failure count | new failure count | difference |
| --- | --- | --- | --- |
| iasworld_asmt_all_valasm1_non_ex_rr_gt_0 | 3497 | 3229 | -268 |
| iasworld_asmt_all_valasm2_impr_class_no_val | 957 | 352 | -605 |
| iasworld_asmt_all_valasm2_vact_class_has_val | 224 | 282 | +58 |
| iasworld_asmt_all_valasm3_non_ex_rr_gt_0 | 158 | 129 | -29 |
| iasworld_pardat_class_equals_luc | 2412 | 2412 | +0 |

Since most of the new failure counts are decreasing, and the only increasing one is going up by just a little bit, this seems OK to me; but let me know if you think my new filtering is incorrect, and we can troubleshoot.

jeancochrane:
I think this could use another look since [b57d1a1](https://github.com/ccao-data/data-architecture/pull/355/commits/b57d1a1f538988d00c1c4a7c6a2e6e9ccb539920) makes a pretty substantial change!

aws-athena/views/qc-vw_incorrect_asmt_value.sql
dfsnow, line 1: **praise:** Nice simplification!

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -189,10 +189,88 @@ sources:
...
+                        '100', '200', '239', '240', '241', '300', '400', '500',
+                        '535', '550', '637', '637A', '637B', '650', '651', '651A',
+                        '651B', '700', '700A', '700B', '742', '742A', '742B',
+                        '800', '800A', '800B', '850', '850A', '850B', '900',
+                        '999', 'EX', 'RR'
+                      )
+                  meta:
+                    table_name: asmt
+                    description: >
+                      valasm2 (ASMT BLDG) should not be 0 for non-vacant classes
```
dfsnow, line 263: **suggestion (non-blocking):** It might simplify the description and logic a little bit if this test were something like `valasm2 (ASMT BLDG) should be >= 0 for non-vacant classes`

dfsnow (approved):

dbt/models/iasworld/schema/iasworld.asmt_all.yml
jeancochrane: @dfsnow Currently the test is just checking that the value isn't 0, should we change the test along with the description so that it's checking for value > 0?

https://github.com/ccao-data/data-architecture/blob/b57d1a1f538988d00c1c4a7c6a2e6e9ccb539920/dbt/models/iasworld/schema/iasworld.asmt_all.yml#L241-L243

dfsnow: I think so. Easier to grok than the double negative.

jeancochrane: Cool, done in aabc1aa52eccdb5e506c5ab29da5f947a88873c7.

PR #360, opened 2024-03-25
Refactor `row_values_match_after_join` and `res_class_matches_pardat` for join flexibility

jeancochrane:
We would like to use the `row_values_match_after_join` and `res_class_matches_pardat` generics to implement https://github.com/ccao-data/ptaxsim/issues/31, but the generics as they exist now are too rigid to support the kinds of joins we need to do in those tests. In particular, neither test currently supports base models whose `parid` and `taxyr` columns have different names (e.g. `pin` or `year`), and `row_values_match_after_join` doesn't support left or outer joins.

This PR makes some tweaks to the two generics to support base models whose name schemes do not include `parid` and `taxyr`, and adds a `join_type` parameter to `row_values_match_after_join`. The changes are backwards-compatible in the case of `res_class_matches_pardat`, but breaking in the case of `row_values_match_after_join`; this is fine since the latter test is not yet in use anywhere in the codebase.

In the process of implementing these changes, I also beefed up the docs for these two generics (along with the `format_additional_select_columns` macro, which the generic docs reference) in order to make them easier to use by a dbt novice. I also implemented tests that use the new forms of these generics in order to test the changes, but I'm not including them here since I would like the implementer of https://github.com/ccao-data/ptaxsim/issues/31 to figure them out as an exercise.

### DIFF
--- a/dbt/tests/generic/test_res_class_matches_pardat.sql
+++ b/dbt/tests/generic/test_res_class_matches_pardat.sql
@@ -5,21 +5,21 @@
 --
 -- Optional parameters:
 --
---   * major_class_only (bool): Compare only the first digit of classes. When
---         set to False, compare the first three digits instead.
---   * parid_column_name (str): The name of the column on the base model that
---         corresponds to `parid`, in case the model uses a different name scheme.
---   * taxyr_column_name (str): The name of the column on the base model that
---         corresponds to `taxyr`, in case the model uses a different name scheme.
---   * additional_select_columns (dict): Standard parameter for selecting
---         additional columns from the base model for use in reporting failures.
---         `model.{column_name}` and `pardat.class` are already selected by
---         default and do not need to be included using this parameter.
---         See the `format_additional_select_columns` macro for more information.
---         Note that while this parameter can often be either a string or a dict,
---         a dict is required for this test and it must contain the `alias`
---         and `agg_func` parameters given that results are grouped by `parid`
---         and `taxyr`.
+-- * major_class_only (bool): Compare only the first digit of classes. When
+-- set to False, compare the first three digits instead.
+-- * parid_column_name (str): The name of the column on the base model that
+-- corresponds to `parid`, in case the model uses a different name scheme.
+-- * taxyr_column_name (str): The name of the column on the base model that
+-- corresponds to `taxyr`, in case the model uses a different name scheme.
+-- * additional_select_columns (dict): Standard parameter for selecting
+-- additional columns from the base model for use in reporting failures.
+-- `model.{column_name}` and `pardat.class` are already selected by
+-- default and do not need to be included using this parameter.
+-- See the `format_additional_select_columns` macro for more information.
+-- Note that while this parameter can often be either a string or a dict,
+-- a dict is required for this test and it must contain the `alias`
+-- and `agg_func` parameters given that results are grouped by `parid`
+-- and `taxyr`.
 --
 {% test res_class_matches_pardat(
     model,
@@ -90,8 +90,7 @@
         on filtered_model.{{ parid_column_name }} = pardat.parid
         and filtered_model.{{ taxyr_column_name }} = pardat.taxyr
     group by
-        filtered_model.{{ parid_column_name }},
-        filtered_model.{{ taxyr_column_name }}
+        filtered_model.{{ parid_column_name }}, filtered_model.{{ taxyr_column_name }}
     having
         sum(
             case
--- a/dbt/tests/generic/test_res_class_matches_pardat.sql
+++ b/dbt/tests/generic/test_res_class_matches_pardat.sql
@@ -3,17 +3,31 @@
 -- filters for residential parcels by anti-joining the model against `comdat`
 -- using parid and taxyr; as a result, it filters out mixed-use parcels as well.
 --
--- By default, the test will compare the first 3 digits of each set of classes;
--- if `major_class_only=true`, however, the test will compare the first digit
--- only.
+-- Optional parameters:
 --
--- Since the output is grouped by (parid, taxyr), additional columns that would
--- normally be selected via `additional_select_columns` need to be selected
--- with one of two aggregation methods: `select_columns_aggregated_with_array`
--- (which uses the `array_agg()` function to select the columns) or
--- `select_columns_aggregated_with_max` (which uses the `max()` function).
+--   * major_class_only (bool): Compare only the first digit of classes. When
+--         set to False, compare the first three digits instead.
+--   * parid_column_name (str): The name of the column on the base model that
+--         corresponds to `parid`, in case the model uses a different name scheme.
+--   * taxyr_column_name (str): The name of the column on the base model that
+--         corresponds to `taxyr`, in case the model uses a different name scheme.
+--   * additional_select_columns (dict): Standard parameter for selecting
+--         additional columns from the base model for use in reporting failures.
+--         `model.{column_name}` and `pardat.class` are already selected by
+--         default and do not need to be included using this parameter.
+--         See the `format_additional_select_columns` macro for more information.
+--         Note that while this parameter can often be either a string or a dict,
+--         a dict is required for this test and it must contain the `alias`
+--         and `agg_func` parameters given that results are grouped by `parid`
+--         and `taxyr`.
+--
 {% test res_class_matches_pardat(
-    model, column_name, major_class_only=false, additional_select_columns=[]
+    model,
+    column_name,
+    major_class_only=false,
+    parid_column_name="parid",
+    taxyr_column_name="taxyr",
+    additional_select_columns=[]
 ) %}
     {#-
         Process `additional_select_columns` to add the necessary prefix for
@@ -55,8 +69,8 @@
                     from {{ source("iasworld", "comdat") }}
                     where comdat.cur = 'Y' and comdat.deactivat is null
                 ) as comdat
-                on model.parid = comdat.parid
-                and model.taxyr = comdat.taxyr
+                on model.{{ parid_column_name }} = comdat.parid
+                and model.{{ taxyr_column_name }} = comdat.taxyr
             where comdat.parid is null
         )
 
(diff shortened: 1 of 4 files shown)


dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -3,17 +3,31 @@
...
--- with one of two aggregation methods: `select_columns_aggregated_with_array`
--- (which uses the `array_agg()` function to select the columns) or
--- `select_columns_aggregated_with_max` (which uses the `max()` function).
 {% test res_class_matches_pardat(
-    model, column_name, major_class_only=false, additional_select_columns=[]
+    model,
+    column_name,
+    major_class_only=false,
+    parid_column_name="parid",
+    taxyr_column_name="taxyr",
```
jeancochrane, line 29: Note that this join interface is less flexible than the one implemented on `row_values_match_after_join` (where the join is specified via a raw SQL string passed to a `join_condition` parameter), but I think that's acceptable for now since this test is more specific than `row_values_match_after_join` and hence can afford to be less flexible for the sake of a simpler interface.

dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -2,41 +2,85 @@
...
+--         by default and do not need to be included using this parameter.
+--         See the `format_additional_select_columns` macro for more information.
+--         Note that while this parameter can often be either a string or a dict,
+--         a dict is required for this test and it must contain the `alias`
+--         and `agg_func` parameters given that results are grouped by the
+--         columns specified in `group_by`.
+--
 {% test row_values_match_after_join(
     model,
-    column,
```
jeancochrane, line 7: For some reason, this test wasn't properly reading this parameter when it was called `column`, so I switched to `column_name` and it fixed the issue. `column_name` is the standard name for this parameter across the rest of our tests anyway.

dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -2,41 +2,85 @@
...
+    column_name,
     external_model,
-    external_column,
-    column_name="model_col",
-    external_column_name="external_model_col",
-    join_columns=[],
+    external_column_name,
```
jeancochrane, line 46: `external_column` gets updated to `external_column_name` to match the change to `column_name`.

dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -2,41 +2,85 @@
...
+    join_condition,
+    group_by,
+    column_alias="model_col",
+    external_column_alias="external_model_col",
```
jeancochrane, line 50: The variables corresponding to the previous names `column_name` and `external_column_name` now use `*_alias` instead of `*_name`, both to avoid collision with the other two variables that now end in `*_name` and because `*_alias` is a bit more explicit about what these variables are used for.

dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -2,41 +2,85 @@
...
-    column,
+    column_name,
     external_model,
-    external_column,
-    column_name="model_col",
-    external_column_name="external_model_col",
-    join_columns=[],
+    external_column_name,
+    join_condition,
+    group_by,
```
jeancochrane, line 48: Since `join_condition` can now contain an arbitrary number of conditions, the `GROUP BY` clause can no longer reuse the join columns for grouping, and so it has to be specified manually.

dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -2,41 +2,85 @@
...
     model,
-    column,
+    column_name,
     external_model,
-    external_column,
-    column_name="model_col",
-    external_column_name="external_model_col",
-    join_columns=[],
+    external_column_name,
+    join_condition,
```
jeancochrane, line 47: I considered using a data structure like a list of dicts to represent a more flexible set of join conditions, but ultimately it felt too restrictive for a test that I'm anticipating we'll want to use pretty widely in the future. Here's an example of the set of conditions that I had in mind while designing this:

```yaml
join_condition: |
  ON model.pin = external_model.parid
  AND model.year = external_model.taxyr
  AND external_model.rolltype != 'RR'
  AND external_model.procname = 'CCAOVALUE'
  AND external_model.deactivat IS NULL
  AND external_model.valclass IS NULL
```

It's possible we could use some combination of params and data structures to support joins like this, but allowing an arbitrary join condition seemed like the most flexible approach. I'm curious what you think!

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -3,17 +3,31 @@
 -- filters for residential parcels by anti-joining the model against `comdat`
 -- using parid and taxyr; as a result, it filters out mixed-use parcels as well.
 --
--- By default, the test will compare the first 3 digits of each set of classes;
--- if `major_class_only=true`, however, the test will compare the first digit
--- only.
+-- Optional parameters:
+--
+-- * major_class_only (bool): Compare only the first digit of classes. When
+-- set to False, compare the first three digits instead.
```
dfsnow, line 9: **issue (non-blocking):** We should make sure that these checks work for incentive (e.g. `597A`) and omitted classes (e.g. `OA2`).

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -3,17 +3,31 @@
...
+-- corresponds to `taxyr`, in case the model uses a different name scheme.
+-- * additional_select_columns (dict): Standard parameter for selecting
+-- additional columns from the base model for use in reporting failures.
+-- `model.{column_name}` and `pardat.class` are already selected by
+-- default and do not need to be included using this parameter.
+-- See the `format_additional_select_columns` macro for more information.
+-- Note that while this parameter can often be either a string or a dict,
+-- a dict is required for this test and it must contain the `alias`
+-- and `agg_func` parameters given that results are grouped by `parid`
+-- and `taxyr`.
```
dfsnow, line 22: **nitpick (non-blocking):** Can we indent these param lists such that all lines except the first start with 2 spaces? (for readability)

dbt/tests/generic/test_row_values_match_after_join.sql
dfsnow: I agree. It's a little haphazard, but the string join condition is the most flexible. Let's keep it.

dfsnow (approved):
@jeancochrane Nice work. Thanks for cleaning up my messy test macros. FWIW I'm not tied to this approach/macro if you think there's a better way we could do this, it was always a little hacky.

dbt/tests/generic/test_res_class_matches_pardat.sql
jeancochrane: Let me break my follow-up questions down by value of `major_class_only`:

1. When `major_class_only=False`, omitted classes should be fine as long as they're all 3 digits and they need to match exactly (i.e. `OA2` is not equivalent to `OA5`). Do incentive classes require a match in the letter suffix? If so, perhaps we need to compare four digits instead of three.
2. When `major_class_only=True`, do we want all omitted classes to match each other (i.e. `OA2` is equivalent to `OA5`), or do we need a special case for those? And is it fine to still compare the first digit of incentive classes?

dbt/tests/generic/test_res_class_matches_pardat.sql
jeancochrane: Annoyingly, removing the indentation is a rule that `sqlfmt` applies automatically. Let me see if I can override it.

jeancochrane: I went ahead and disabled the formatting for these docstrings in d8bed6c1f7a5de9c020a4f1cd651bf9c147ab489, which is the easiest way to allow for custom formatting given that [sqlfmt doesn't expose any config options other than line length](https://docs.sqlfmt.com/getting-started/configuring-sqlfmt#configuring-style). It works and isn't too invasive; the only downside is that we won't get automatic line-length formatting for docstrings, but I think that's an acceptable price to pay for readability of bulleted lists.

dbt/tests/generic/test_res_class_matches_pardat.sql
dfsnow: 1. I don't know enough about them yet to say definitively, but for now let's assume they need to match on all 3 characters.
2. Annoyingly, we'll probably have to build in logic for a special case, since the major class for the omitted is the _last_ digit.

jeancochrane: > Annoyingly, we'll probably have to build in logic for a special case, since the major class for the omitted is the last digit.

So to clarify, does this mean that e.g. a PIN with class `OA2` should have a pardat class of either `OA2` or `2xx`? I just spot checked the current failures for omitted classes, and all of the mismatched classes are `OA2` or `OA5` matching to either `EX` or `100` (no clear pattern I could discern from a cursory glance based on class alone). Is it possible that those are valid values for omitted classes?

dfsnow: The OA classes are sufficiently new that I actually don't know. I'll bring it up tomorrow with MV. Feel free to merge + fast-follow if you don't want this blocked.

dfsnow: So from what I understand, the `OA2`, `OA3`, and `OA5` codes should only exist in the various detail tables (`COMDAT`, `DWELDAT`, etc.). If they exist they can basically have _any_ class code in `pardat`, but should match `EX`, `100`, or their corresponding major class most of the time. Further, if something has an `OA` class code, it should have a corresponding `asmt.tot57` value in the same tax year, so we should add a test for that.

jeancochrane: Issue open here: https://github.com/ccao-data/data-architecture/issues/363

PR #369, opened 2024-04-01
Add class checks and other data validation prior to final db export

Damonamajor:
This PR creates 3 of the 4 dbt tests requested in the following [issue ](https://github.com/ccao-data/ptaxsim/issues/31). The fourth is on hold, due to the fact that there are known valuation discrepancies between iasWorld and PTaxSim. Thus, we may want to include a buffer on the matching function.  

I can't link the issue in the development sections since it's in the PTaxSim repository.

### DIFF
--- a/dbt/models/tax/schema.yml
+++ b/dbt/models/tax/schema.yml
@@ -23,7 +23,49 @@ sources:
 
       - name: pin
         description: '{{ doc("table_pin") }}'
-
+          columns:  
+            - name: tax_code_num
+              description: '{{ doc("shared_column_tax_code")}}'
+              tests:
+                - row_values_match_after_join:
+                  name: pin_tax_dist_matches_iasworld_legdat
+                  external_model: source('iasworld', 'legdat')
+                  external_column_name: taxdist
+                  column_alias: tax_code_num
+                  external_column_alias: taxdist
+                  group_by: year, pin
+                  additional_select_columns: 
+                    - column: pin AS parid
+                      agg_func: max
+                    - column: year AS taxyr
+                      agg_func: max
+                  join_condition:
+                    ON  model.pin = external_model.parid
+                    AND model.year = external_model.taxyr
+                    AND external_model.cur = 'Y'
+                    AND external_model.deactivat IS NULL
+            - name: class
+              description: '{{ doc("shared_column_char_class")}}'
+              tests:
+                - res_class_matches_pardat:
+                    name: tax_pin_class_equals_iasworld_pardat
+                    parid_column_name: pin
+                    taxyr_column_name: year
+                    additional_select_column: year
+                      agg_func: max
+                - relationships:
+                    name: tax_pin_all_class_in_ccao_class_dict
+                    to: source('ccao', 'class_dict')
+                    field: class_code
+                    additional_select_columns: &select-columns
+                      - year
+                    config:
+                      where:
+                        CAST(year AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                    meta:
+                      category: class_mismatch_or_issue
+                      description: class code must be valid
+    
       - name: pin_geometry_raw
         description: '{{ doc("table_pin_geometry_raw") }}'
 
--- a/dbt/models/tax/schema.yml
+++ b/dbt/models/tax/schema.yml
@@ -23,48 +23,52 @@ sources:
 
       - name: pin
         description: '{{ doc("table_pin") }}'
-          columns:  
-            - name: tax_code_num
-              description: '{{ doc("shared_column_tax_code")}}'
-              tests:
-                - row_values_match_after_join:
+        columns:  
+          - name: tax_code_num
+            description: '{{ doc("shared_column_tax_code")}}'
+            tests:
+              - row_values_match_after_join:
                   name: pin_tax_dist_matches_iasworld_legdat
                   external_model: source('iasworld', 'legdat')
                   external_column_name: taxdist
                   column_alias: tax_code_num
                   external_column_alias: taxdist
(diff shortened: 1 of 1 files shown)


dbt/models/tax/schema.yml
```diff
@@ -23,7 +23,53 @@ sources:
...
+              - res_class_matches_pardat:
+                  name: tax_pin_class_equals_iasworld_pardat
+                  parid_column_name: pin
+                  taxyr_column_name: year
+                  additional_select_columns:
+                    - agg_func: max
+                      column: year
+                  config:
+                    where:
+                      CAST(year AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
```
Damonamajor, line 59: @Removing these clauses increases the number of errors by a significant amount. @jeancochrane recommended leaving them in, because that's our operating practice, but it's up to you @dfsnow on if we want to test the historical data.

dbt/models/tax/schema.yml
```diff
@@ -23,7 +23,53 @@ sources:
...
+                    - year
+                    - pin
+                  join_condition:
+                    ON  model.pin = external_model.parid
+                    AND model.year = external_model.taxyr
+                    AND external_model.cur = 'Y'
+                    AND external_model.deactivat IS NULL
+                  config:
+                    where:
+                      CAST(year AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
```
jeancochrane, line 46: **[Nitpick, non-blocking]** For completeness' sake, let's add the `meta.description` attribute to this test so that we can properly output it for review if it starts failing:

```suggestion
                  config:
                    where:
                      CAST(year AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
                  meta:
                    description: tax_code_num should match legdat.taxdist
```

dbt/models/tax/schema.yml
jeancochrane, line 59: **[Nitpick, non-blocking]** Let's also add `meta.description` here:

```suggestion
                  config:
                    where:
                      CAST(year AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
                  meta:
                    description: at least one class should match pardat class
```

jeancochrane (approved):
Looking great! A question for @dfsnow: Do we want these tests to be output to the QC workbook if they fail? Currently they won't be, because A) they're not tagged with a `test_qc_*` tag and B) the output workbook isn't written to handle tests outside the `iasworld` schema anyway. If we do want to output these tests to the QC workbook, I can put up a follow-up PR to handle both of these problems; if not, perhaps we should discuss when and how we expect them to run, since they don't really strike me as unit tests either.

dbt/models/tax/schema.yml
```diff
@@ -45,8 +45,7 @@ sources:
                     where:
                       CAST(year AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
                   meta:
-                    category: class_mismatch_or_issue
-                    description: class code must be valid
+                    description: tax_code_num should match legdat.taxdist
```
dfsnow, line 48: So, we actually know and expect this test to fail for the time being, as the `taxdist` values in `legdat` are out of date. We don't really have a handler for the "this is failing now but needs to be fixed in the future" case. Perhaps we should make this into a warning rather than an explicit error. Thoughts @jeancochrane?

dbt/models/tax/schema.yml
dfsnow: We can keep the clauses in for now.

dfsnow:
> Looking great! A question for @dfsnow: Do we want these tests to be output to the QC workbook if they fail? Currently they won't be, because A) they're not tagged with a `test_qc_*` tag and B) the output workbook isn't written to handle tests outside the `iasworld` schema anyway. If we do want to output these tests to the QC workbook, I can put up a follow-up PR to handle both of these problems; if not, perhaps we should discuss when and how we expect them to run, since they don't really strike me as unit tests either.

No, these shouldn't go to the QC workbooks. For now, I'm just going to run them manually (via GHA) whenever I export the PTAXSIM DB.

dbt/models/tax/schema.yml
jeancochrane: That makes sense to me @dfsnow! Do we know which `taxdist` values are out of date, and when we expect them to be up to date? If not, I think a naive warning on the failing row count is a good start, but if we do know how many and when we expect to be fixed it could be nice to add some additional filtering logic so that it starts failing automatically if one of those assumptions is broken.

dfsnow: If I recall right all values after 2021 are not kept up-to-date and are supposed to be fixed by the Clerk once they come on board iasWorld. The tax codes don't change super frequently so many of them may still match fine. I would start with a naive warning for now.

Damonamajor: @dfsnow 
Do we want the warning if the errors are >0 or if they are > 'current value'?

Damonamajor: @dfsnow Also, as the query is currently constructed, there are mismatched values, but none after 2021. Below is a table with the years where errors are populated (after removing the year clause).

2006
2007
2008
2009
2010
2011
2012
2014
2017
2018
2019
2020
2021

dfsnow: Whoops, maybe I had that backwards, it could be that values **before** 2021 are not up-to-date. Either way, let's throw a warning for errors>0.

Damonamajor: Done

PR #370, opened 2024-04-02
Update `test_dbt_models` workflow to format and upload run results to S3

jeancochrane:
This PR updates the `test_dbt_models` workflow to format run results and upload them to S3 for analysis. In the process, it refactors the `format_dbt_test_failures` script into a more general form (now called `transform_dbt_test_results`) that outputs test result metadata to parquet files in addition to outputting the failures workbook.

## Schema

Here's the schema of the new generated parquet files:

* `metadata/test_run/run_id=*/*.parquet`: Metadata about the full test run, partitioned by run ID. Populates the `qc.test_run` table in Athena. Unique by `run_id`.
    * `run_id`: primary key string representing the unique ID of the test run (technically the [invocation ID](https://docs.getdbt.com/reference/dbt-jinja-functions/invocation_id) of the `dbt test` command)
    * `run_date`: string representing the UTC date that the test ran, in `YYYY-MM-DD` format
    * `elapsed_time`: decimal representing the total number of seconds that the tests took to execute
    * `run_by`: string representing the username of the person who ran the script, or `"CI"` on CI
    * `var_year_start`: string representing the value of `test_qc_year_start` for the test run in `YYYY-MM-DD` format
    * `var_year_end`: string representing the value of `test_qc_year_end` for the test run in `YYYY-MM-DD` format
    * `git_sha`: string representing the hash of the script's commit
    * `git_branch`: string representing the branch of the script's commit 
    * `git_author`: string representing the name and email address of the author of the script's commit

* `metadata/test_run_result/run_id=*/*.parquet`: Metadata about individual test results (i.e. pass, fail, number of failing rows, etc.) in this run, partitioned by run ID. Populates the `qc.test_run_result` table in Athena. Unique by `run_id`, `test_name`, and `township`.
    * `run_id`: composite primary key string representing the unique ID of the test run. Foreign key to `qc.test_run.run_id`.
    * `test_name`: composite primary key string representing the name of the test, e.g. `iasworld_sales_cur_in_accepted_values`
    * `township_code`: composite primary key string representing the township of the test results; can be null if the failing rows can't be mapped to townships, or if the test passed
    * `table_name`: string representing the name of the iasWorld table that the test is defined on, e.g. `pardat`
    * `category`: string representing the category of the test, e.g. `class_mismatch_or_issue`
    * `description`: string representing the long description of the test, e.g. `at least one class should match pardat class`
    * `status`: string enum representing the status of the test, one of `pass` or `fail`
    * `elapsed_time`: decimal representing the number of seconds that the test took to execute
    * `num_failing_rows`: integer representing the number of rows that failed the test, or 0 if the test passed

## Testing

See [here](https://github.com/ccao-data/data-architecture/actions/runs/8529144247/job/23364275892) for a successful workflow run, and check the `qc.test_run` and `qc.test_run_result` tables in Athena to browse the metadata for that run. An example of the types of queries we can run using this schema:

```
SELECT run_id, SUM(num_failing_rows)
FROM qc.test_run_result
WHERE status = 'fail' AND category = 'class_mismatch_or_issue'
GROUP BY run_id
```


.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+    township_code will always be None in the case of a passing test."""
+
+    def __init__(
+        self, township_code: typing.Optional[str], *args, **kwargs
+    ) -> None:
+        self.township_code = township_code
+        super().__init__(*args, **kwargs)
+
+
+class TestCategory:
```
jeancochrane, line 243: This class was adapted from the previous version of the `format_dbt_test_failures` script, where it was called [`FailedTestGroup`](https://github.com/ccao-data/data-architecture/blob/66c85bfa973d163bd29c4e41f08ecf58eafff08d/.github/scripts/format_dbt_test_failures.py#L95). The main difference between this version and the old version of the class is that, like `TestResult`, `TestCategory` can store results for passing tests in addition to failing tests.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+        of its TestResult objects."""
+        return (
+            Status.PASS
+            if all(
+                result.status == Status.PASS for result in self.test_results
+            )
+            else Status.FAIL
+        )
+
+    def add_to_workbook(self, workbook: openpyxl.Workbook) -> None:
```
jeancochrane, line 355: This function is adapted from the previous version of the `format_dbt_test_failures` script, where it was called [`add_sheet_to_workbook`](https://github.com/ccao-data/data-architecture/blob/66c85bfa973d163bd29c4e41f08ecf58eafff08d/.github/scripts/format_dbt_test_failures.py#L564). I've moved it to a method on the `TestCategory` class since it is essentially a function to add a `TestCategory` object to an existing workbook.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+        """Add a sheet of failed dbt tests to an openpyxl Workbook using data
+        from the TestCategory object. Note that we expect the workbook to be
+        initialized with write_only=True."""
+        # Only add this category to the workbook if it has any failing tests
+        if self.status == Status.PASS:
+            print(
+                f"Skipping add_to_workbook for category {self.category} since "
+                f"its status is '{Status.PASS.value}'"
+            )
+            return
```
jeancochrane, line 365: This is the only substantive difference between this method and the previous version of the function: Now that `TestResult` and `TestCategory` store info on passing tests as well as failing tests, we need to adjust the workbook logic to skip adding `TestCategory` objects that are passing.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+        # happen _before_ data is added. See here for details:
+        # https://stackoverflow.com/a/66159254
+        for cell_range in merged_cell_range:
+            sheet.merged_cells.ranges.add(cell_range)
+
+    @property
+    def debugging_fieldnames(self) -> typing.List[str]:
+        """Get a list of fieldnames (e.g. ["foo", "bar"]) for fields that
+        are used for debugging."""
+        return self._filter_for_existing_fieldnames(self._debugging_fieldnames)
```
jeancochrane, line 548: This property function and all of the ones that follow were copied over as-is from the old `FailedTestGroup` object.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+    ]:
+        table = pyarrow.Table.from_pylist(
+            [dataclasses.asdict(meta_obj) for meta_obj in metadata_list]
+        )
+        metadata_root_path = os.path.join(
+            output_directory, "metadata", tablename
+        )
+        pyarrow.parquet.write_to_dataset(
+            table, metadata_root_path, partition_cols
+        )
```
jeancochrane, line 725: I'm not 100% sure about partitioning by run ID. It seems like an obvious choice, but I suspect it's going to lead to small file sizes for each `test_run` and `test_run_result` object, which we know creates overhead in Athena queries. It would probably be more performant from a query perspective if each script run loaded and then saved a couple shared parquet files in S3, each one partitioned by year, but from an engineering perspective I have a strong preference for small immutable files over big mutable files. My thinking is that it's easier to move forward with small immutable files and then refactor to big mutable files if we find that latency due to query overhead is severe, but I'm open to other opinions.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+                category=test_category.category,
+                description=township_result.description,
+                township_code=township_result.township_code,
+                status=township_result.status.value,
+                elapsed_time=township_result.elapsed_time,
+                num_failing_rows=len(township_result.failing_rows),
+            )
+            for test_category in test_categories
+            for test_result in test_category.test_results
+            for township_result in test_result.split_by_township()
```
jeancochrane, line 821: In normal circumstances I would find a triple-nested list comprehension to be a little too hairy, but in this case it feels straightforward enough that I don't mind it -- curious if you agree!

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+        # Basic attrs for the test result that apply whether or not the test
+        # failed
+        base_result_kwargs = {
+            "name": test_name,
+            "table_name": tablename,
+            "description": test_description,
+            "elapsed_time": execution_time,
+        }
+
+        if status == Status.PASS.value:
```
jeancochrane, line 926: The previous version of this function, [`get_test_results_by_category`](https://github.com/ccao-data/data-architecture/blob/66c85bfa973d163bd29c4e41f08ecf58eafff08d/.github/scripts/format_dbt_test_failures.py#L402), only supported tests with status `fail`. With the addition of this block, we now support passing tests as well.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+                "in dbt run results"
+            )
+
+    # Now that we've accumulated all of the test results and they are grouped
+    # into categories, we no longer need the category key in the dict, so
+    # transform the output into a list
+    return list(tests_by_category.values())
+
+
+def get_category_from_node(node: typing.Dict) -> str:
```
jeancochrane, line 1017: `get_category_from_node` and `get_tablename_from_node` are untouched since the last version of this script.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+            output_directory, "metadata", tablename
+        )
+        pyarrow.parquet.write_to_dataset(
+            table, metadata_root_path, partition_cols
+        )
+        print(f"{tablename} metadata saved to {metadata_root_path}/")
+
+
+@dataclasses.dataclass
+class TestRunMetadata:
```
jeancochrane, line 730: The dataclasses `TestRunMetadata` and `TestRunResultMetadata` are new as of this version of the script, and are used to store and format the data that get output to parquet files for storage in S3.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1082 @@
...
+                elapsed_time=township_result.elapsed_time,
+                num_failing_rows=len(township_result.failing_rows),
+            )
+            for test_category in test_categories
+            for test_result in test_category.test_results
+            for township_result in test_result.split_by_township()
+        ]
+
+
+def get_test_cache_path(
```
jeancochrane, line 825: `get_test_cache_path`, `get_test_categories_from_file`, `save_test_categories_to_file`, and `get_test_categories_from_athena` are basically unchanged since the last version of this script; the only changes are to symbol names, and using the new `TestCategory` interface for (de)serialization.

.github/workflows/test_dbt_models.yaml
```diff
@@ -22,8 +22,65 @@ jobs:
...
-        run: dbt test --target "$TARGET"
+      - name: Install Python requirements
+        run: pip install -r ./.github/scripts/requirements.transform_dbt_test_results.txt
+        shell: bash
+
+      - name: Run tests
+        run: |
+          # dbt doesn't differentiate between test failures and errors, but we
+          # need to since we expect errors. Do this by capturing the output
+          # and checking its contents to look for errors
```
jeancochrane, line 33: This is annoying, but it works.

.github/workflows/test_dbt_models.yaml
```diff
@@ -22,8 +22,65 @@ jobs:
...
+          if output=$(dbt test --target "$TARGET" \
+                               --select \
+                                   iasworld_addn_card_not_null \
+                                   iasworld_asmt_all_class_matches_pardat_class \
```
jeancochrane, line 37: Note to self: Remove these lines (used to speed up testing) before merging, and replace with `tag:test_qc_*`.

.github/workflows/test_dbt_models.yaml
```diff
@@ -22,8 +22,63 @@ jobs:
...
-        run: dbt test --target "$TARGET"
+      - name: Install Python requirements
+        run: pip install -r ./.github/scripts/requirements.transform_dbt_test_results.txt
+        shell: bash
+
+      - name: Run tests
+        run: |
+          # dbt doesn't differentiate between test failures and errors, but we
+          # need to since we expect failures. Do this by capturing the output
+          # and checking its contents to look for errors
```
jeancochrane, line 33: This is annoying but it works, as evidenced by [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/8527517308/job/23359117365) that failed due to the QC schema not existing for my branch.

.github/scripts/transform_dbt_test_results.py
dfsnow: Whoops, didn't see this until scrolling down, but see my earlier comment: we're on the same page. We really don't need ultra-performant queries, and lots of small immutable files should be way easier to deal with than some big mutable ones.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1081 @@
...
+        )
+        print(f"{tablename} metadata saved to {metadata_root_path}/")
+
+
+@dataclasses.dataclass
+class TestRunMetadata:
+    """Metadata object storing information about a test run."""
+
+    run_id: str
+    run_date: str
```
dfsnow, line 736: **suggestion (non-blocking):** IMO it's useful to store basically as much metadata as possible here. Some ideas:

- Who/what instantiated the run (captured from the environment?)
- The git sha + author + info for the current dbt code
- The _total_ number of tests, failing tests, warning tests, and passing tests

Could be a separate PR though

.github/scripts/transform_dbt_test_results.py
dfsnow: I think it's clear enough what's happening that it's fine. Can't say I've ever seen a triple before though :D

.github/scripts/transform_dbt_test_results.py
dfsnow: Same question as above: what happens for `WARNING` tests?

.github/scripts/transform_dbt_test_results.py
```diff
@@ -0,0 +1,1081 @@
...
+#    ccao-athena-results-us-east-1 bucket
+#
+# Outputs three files:
+#
+#   1. `qc_test_failures_<date>.xlsx`: Excel workbook to share with other teams
+#   2. `metadata/test_run/run_id=*/*.parquet`: Metadata about this run,
+#       partitioned by run ID and prepped for upload to S3
+#   3. `metadata/test_run_result/run_id=*/*.parquet`: Metadata about test
+#      results (pass, fail, number of failing rows, etc.) in this run,
+#      partitioned by run ID and prepped for upload to S3
```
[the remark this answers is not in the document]
jeancochrane: Good call, first pass implemented in 95682a523a489a763cf682e45b775118e3297d30 and I'll test and refine next week.

.github/scripts/transform_dbt_test_results.py
jeancochrane: Good catch, I added support for warnings in d54120e6bf2b7e2020fbfeb887fc5d7841aac83. Now the ailing rows for a warning test get logged to `test_run_result` with a `WARN` status, but they don't get added to the workbook.

.github/scripts/transform_dbt_test_results.py
jeancochrane: Good thinking! I added bullets 1 and 2 in 382d2b858fda5a08ee326a088f594bb52484ba17 and am going to test them out shortly. I don't think bullet 3 is necessary since a given `test_run` can be joined to its `test_run_results` to aggregate the total number of tests of each status, but I'm open to pushback if you think there's a case for duplicating that info on the `test_run` table as well.

.github/scripts/transform_dbt_test_results.py
jeancochrane: Answered here: https://github.com/ccao-data/data-architecture/pull/370#discussion_r1554070863

jeancochrane:
This is ready for another look @dfsnow! The tables in Athena should also be up to date from [this latest run](https://github.com/ccao-data/data-architecture/actions/runs/8606692923/job/23585634814) if you'd like to browse the updated schema.

.github/scripts/transform_dbt_test_results.py
dfsnow: You're right, probably not needed. Let's table it for now.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -704,14 +728,41 @@ def main() -> None:
     workbook.save(workbook_filepath)
     print(f"Output workbook saved to {workbook_filepath}")
 
+    # Get run metadata from the environment
+    try:
+        run_by = os.environ["USER"]
```
dfsnow, line 733: **thought (non-blocking):** It might be nice to set this to `GITHUB_TRIGGERING_ACTOR` if it's set.

dfsnow (approved):
Looks great! Excited to see this come to fruition @jeancochrane.

.github/scripts/transform_dbt_test_results.py
jeancochrane: Excellent idea, done in [0e68fff](https://github.com/ccao-data/data-architecture/pull/370/commits/0e68fffd3238ae89c281716abe6303f98f59d9c0).

PR #373, opened 2024-04-10
Add selector for `test-dbt-models` workflow and set its indirect selection to "cautious"

jeancochrane:
This PR fixes a bug in the `test-dbt-models` workflow revealed by the [latest workflow run](https://github.com/ccao-data/data-architecture/actions/runs/8629528053/job/23653918137). The [indirect selection](https://docs.getdbt.com/reference/node-selection/test-selection-examples#indirect-selection) for our tests is set to "eager" by default, which means:

> By default, runs tests if any of the parent nodes are selected, regardless of whether all dependencies are met. This includes ANY tests that reference the selected nodes.

As a result, tests that are defined on a non-selected model will be run if they reference a selected model in an argument. This causes `test-dbt-models` to run `tax.pin` tests that reference `legdat` and `pardat` in their arguments:

https://github.com/ccao-data/data-architecture/blob/9e821802f2e964e4e6fff6fd04685d9137f4abf3/dbt/models/tax/schema.yml#L30-L64

This is a problem since the failure workbook code can only parse tests defined on models in the `iasworld` schema, so it raises an error when it tries to parse tests defined on `tax.pin`.

In order to fix this bug, this PR sets the indirect selection to "cautious" for this workflow, which means:

> Restricts tests to only those that exclusively reference selected nodes. Tests will only be executed if all the nodes they depend on are selected, which prevents tests from running if one or more of its parent nodes are unselected...

In the process, we define a [YAML selector](https://docs.getdbt.com/reference/node-selection/yaml-selectors) for this particular selection string, to make it easier to read and reuse.

Evidence that the workflow succeeds here: https://github.com/ccao-data/data-architecture/actions/runs/8634848786/job/23671719425

### DIFF
--- a/.github/workflows/test_dbt_models.yaml
+++ b/.github/workflows/test_dbt_models.yaml
@@ -31,10 +31,7 @@ jobs:
           # dbt doesn't differentiate between test failures and errors, but we
           # need to since we expect failures. Do this by capturing the output
           # and checking its contents to look for errors
-          if output=$(dbt test --target "$TARGET" \
-                               --select tag:test_qc_* \
-                               --exclude tag:test_qc_exclude_from_workbook \
-                               --store-failures); then
+          if output=$(dbt test --target "$TARGET" --selector qc_tests --store-failures); then
             echo "$output"
           else
             status_code="$?"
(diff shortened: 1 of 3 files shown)


dfsnow (approved):
Wow, dbt really has an option for just about everything. Continues to impress. We should build some other selectors for common workflows.

PR #375, opened 2024-04-10
 Fix missing names for state rep and state senate census districts

wrridgeway:
`spatial.census` has `NULL` names for these two census district types.

This PR also removes `ZZZ` geoid state congressional districts.


wrridgeway:
This isn't super important, but the `ZZZ` geoids are annoying.

dfsnow (approved):
Looks good to me! I'm moving to `ZZZ` district.

Issue #371, opened 2024-04-04
Small edits to dbt tests based on feedback


jeancochrane:
We got a new batch of feedback based on the use of our dbt tests in Rogers. Once these edits are done, we should run an export for West and send it over.

## All categories

- [x] Update `test_qc_year_start` to 2024 so that we don't report on any 2023 data
- [x] Make sure that `class` is selected via `additional_select_columns` on as many tests as possible, since it's useful for workbook readers
- [x] Move `class` from the red "Problematic fields" section of the workbook over to the blue "Unique identifier fields" section
- [x] Some issues are addressed at the beginning of the year; data may be out of date

## Incorrect values

- [x] `iasworld_dweldat_fixhalf_between_0_and_6`: half baths are pretty heterogenous, so either adjust this test to just make sure fixhalf is greater than 0 or design a heuristic such that it's a multiple of the number of units
    - [x] Same for `iasworld_dweldat_rmbed_sf_between_1_and_15`
    - [x] Same for `iasworld_dweldat_sfla_sf_between_1_and_10000`
- [x] `iasworld_dweldat_stories_in_accepted_values` should also allow 6.00, 7.00, 8.00, and 9.00
  - [x] Also update the `shared_column_char_type_resd` docstring to document these values
- [ ] Update `iasworld_oby_card_between_1_and_100` to exclude condos, and see if the results are accurate (i.e., make sure that there are no non-condo buildings where card >100 makes sense)
    * Clarify with Tia/Dan; I think this test is actually correct

## Missing values

- [ ] Move `iasworld_dweldat_card_proration_rate_not_null` from `dweldat.user24` to `dweldat.external_propct` and adjust its filters so that it only checks cards that have tieback PINs
    - Clarify with Tia/Dan; the rules here are complex
- [x] Remove `iasworld_dweldat_char_frpl_not_null` since a null can indicate no fireplaces
- [x] Adjust dweldat garage char not-null tests to filter for rows where user33 (garage size) is not 7 (NONE)
  - [x] `iasworld_dweldat_char_gar_att_not_null`
  - [x] `iasworld_dweldat_char_gar_area_not_null`
  - [x] `iasworld_dweldat_char_gar_cnst_not_null`
- [x] Add dweldat test to user34 (garage construction) to ensure it is null if user33 (garage size) is 7 (NONE)
[...]

PR #377, opened 2024-04-11
Small edits to dbt tests based on feedback

jeancochrane:
This PR makes a number of small edits to our dbt tests as described in https://github.com/ccao-data/data-architecture/issues/371. See that issue for a full list of the changes.

There are a few changes from the list that I haven't made yet. I'm curious to get your take:

* Update `iasworld_oby_card_between_1_and_100` to exclude condos, and see if the results are accurate (i.e., make sure that there are no non-condo buildings where card >100 makes sense)
    * These query results make me doubt that the failures are related to condos with large numbers of cards:

```sql
select oby.class, failures.*
from z_dev_jecochr_test_failure.iasworld_oby_card_between_1_and_100 failures
join iasworld.oby oby
    on oby.parid = failures.parid
    and oby.taxyr = failures.taxyr
    and oby.cur = 'Y'
    and oby.deactivat is null
```

* Move `iasworld_dweldat_card_proration_rate_not_null` from `dweldat.user24` to `dweldat.external_propct` and adjust its filters so that it only checks for prorated PINs (i.e. PINs with tiebacks)
    * I tried a few different ways of joining `dweldat` to `pardat` and filtering for only PINs with tiebacks but none of them seemed to produce a result set that made sense to me. I'm wondering if maybe my interpretation of the suggestion is incorrect; it seems a bit weird to be comparing a proration rate set on the dwelling level against a tieback set at the parcel level, which is making me think that perhaps I'm getting two different types of proration mixed up. Let me know if I'm misunderstanding the suggestion, and if not (meaning that the join is indeed just very complex) I can split this out into a separate issue and kick it over to you.   
* Adjust `iasworld_dweldat_char_ncu_between_0_and_5` to filter for 212s
    * These query results are making me think that this test is actually finding useful things:

```sql
select * from z_dev_jecochr_test_failure.iasworld_dweldat_char_ncu_between_0_and_5
```

### DIFF
--- a/.github/scripts/transform_dbt_test_results.py
+++ b/.github/scripts/transform_dbt_test_results.py
@@ -117,6 +117,7 @@ CUSTOM_TEST_NAMES = {
     "macro.athena.test_unique_combination_of_columns": "duplicate_records",
     "macro.dbt_utils.test_unique_combination_of_columns": "duplicate_records",
     "macro.athena.test_not_null": "missing_values",
+    "macro.athena.test_is_null": "missing_values",
     "macro.athena.test_res_class_matches_pardat": "class_mismatch_or_issue",
 }
 # Directory to store failed test caches
@@ -1061,6 +1062,8 @@ def get_test_categories(
             # source table in an attempt to rehydrate the missing columns.
             test_results_select = "select test_results.*"
             test_results_join = ""
+            # We need parid and taxyr at minimum in order to rehydrate any
+            # missing fields
             if PARID_FIELD in fieldnames and TAXYR_FIELD in fieldnames:
                 if TOWNSHIP_FIELD not in fieldnames:
                     test_results_select += f", leg.user1 AS {TOWNSHIP_FIELD}"
@@ -1073,17 +1076,23 @@ def get_test_categories(
                     """
                 if CLASS_FIELD not in fieldnames:
                     if CARD_FIELD in fieldnames:
-                        test_results_select += f", dwel.class AS {CLASS_FIELD}"
+                        # Figure out the right table to join to in order to
+                        # query the class
+                        card_join_table = "dweldat"
+                        if tablename in ["comdat", "land", "oby"]:
+                            card_join_table = tablename
+
+                        test_results_select += f", card.class AS {CLASS_FIELD}"
                         test_results_join += f"""
-                            left join iasworld.dweldat as dwel
-                                on dwel.{PARID_FIELD} =
+                            left join iasworld.{card_join_table} as card
+                                on card.{PARID_FIELD} =
                                     test_results.{PARID_FIELD}
-                                and dwel.{TAXYR_FIELD} =
+                                and card.{TAXYR_FIELD} =
                                     test_results.{TAXYR_FIELD}
-                                and dwel.{CARD_FIELD} =
+                                and card.{CARD_FIELD} =
                                     test_results.{CARD_FIELD}
-                                and dwel.cur = 'Y'
-                                and dwel.deactivat is null
+                                and card.cur = 'Y'
+                                and card.deactivat is null
                         """
                     else:
                         test_results_select += f", par.class AS {CLASS_FIELD}"
--- a/.github/scripts/transform_dbt_test_results.py
+++ b/.github/scripts/transform_dbt_test_results.py
@@ -104,6 +104,7 @@ TAXYR_FIELD = "taxyr"
 PARID_FIELD = "parid"
 CARD_FIELD = "card"
 TOWNSHIP_FIELD = "township_code"
+CLASS_FIELD = "class"
 WHO_FIELD = "who"
 WEN_FIELD = "wen"
 # Overrides for default display names for dbt tests
@@ -272,6 +273,7 @@ class TestCategory:
         TAXYR_FIELD,
         PARID_FIELD,
         CARD_FIELD,
+        CLASS_FIELD,
         TOWNSHIP_FIELD,
         WHO_FIELD,
         WEN_FIELD,
@@ -1052,24 +1054,54 @@ def get_test_categories(
             # SHOW COLUMNS often returns field names with trailing whitespace
             fieldnames = [row["field"].strip() for row in cursor]
 
-            test_results_query = f"select * from {test_results_relation_name}"
-            if (
-                PARID_FIELD in fieldnames
-                and TAXYR_FIELD in fieldnames
-                and TOWNSHIP_FIELD not in fieldnames
-            ):
-                # If parid and taxyr are present, try to retrieve the township
-                # code for every row. It's most efficient to do this via a
-                # join in the query rather than in a Python lookup since
-                # legdat has 43m rows
-                test_results_query = f"""
-                    select test_results.*, leg.user1 as {TOWNSHIP_FIELD}
-                    from {test_results_relation_name} as test_results
-                    left join iasworld.legdat as leg
-                        on leg.{PARID_FIELD} = test_results.{PARID_FIELD}
-                        and leg.{TAXYR_FIELD} = test_results.{TAXYR_FIELD}
-                """
+            # Construct the query to select the test results that were stored
+            # in Athena. We want to rehydrate a few fields like township_code
+            # and class without requiring that all tests select them, so
+            # check if those fields are missing and join against the correct
+            # source table in an attempt to rehydrate the missing columns.
+            test_results_select = "select test_results.*"
+            test_results_join = ""
+            if PARID_FIELD in fieldnames and TAXYR_FIELD in fieldnames:
+                if TOWNSHIP_FIELD not in fieldnames:
+                    test_results_select += f", leg.user1 AS {TOWNSHIP_FIELD}"
+                    test_results_join += f"""
+                        left join iasworld.legdat as leg
+                            on leg.{PARID_FIELD} = test_results.{PARID_FIELD}
+                            and leg.{TAXYR_FIELD} = test_results.{TAXYR_FIELD}
+                            and leg.cur = 'Y'
+                            and leg.deactivat is null
+                    """
+                if CLASS_FIELD not in fieldnames:
+                    if CARD_FIELD in fieldnames:
+                        test_results_select += f", dwel.class AS {CLASS_FIELD}"
+                        test_results_join += f"""
+                            left join iasworld.dweldat as dwel
+                                on dwel.{PARID_FIELD} =
+                                    test_results.{PARID_FIELD}
+                                and dwel.{TAXYR_FIELD} =
+                                    test_results.{TAXYR_FIELD}
+                                and dwel.{CARD_FIELD} =
+                                    test_results.{CARD_FIELD}
+                                and dwel.cur = 'Y'
+                                and dwel.deactivat is null
+                        """
+                    else:
+                        test_results_select += f", par.class AS {CLASS_FIELD}"
+                        test_results_join += f"""
+                            left join iasworld.pardat as par
+                                on par.{PARID_FIELD} =
+                                    test_results.{PARID_FIELD}
+                                and par.{TAXYR_FIELD} =
+                                    test_results.{TAXYR_FIELD}
+                                and par.cur = 'Y'
+                                and par.deactivat is null
+                        """
 
+            test_results_query = (
+                f"{test_results_select} "
+                f"from {test_results_relation_name} as test_results "
+                f"{test_results_join}"
+            )
             cursor.execute(test_results_query)
             query_results = cursor.fetchall()
             if len(query_results) == 0:
--- a/.github/scripts/transform_dbt_test_results.py
+++ b/.github/scripts/transform_dbt_test_results.py
@@ -1062,6 +1062,19 @@ def get_test_categories(
             # and class without requiring that all tests select them, so
             # check if those fields are missing and join against the correct
             # source table in an attempt to rehydrate the missing columns.
+            #
+            # The conditionals that follow are a bit ugly, but they reflect the
+            # fact that we need to rehydrate different columns from different
+            # source tables, and that those source tables can differ based on
+            # whether the base model is keyed by card or parcel. In the future,
+            # we might consider simplifying this by joining to a helper view
+            # with a universe of township and class codes split out by table,
+            # e.g.:
+            #
+            #    pin  | card | taxyr | township_code |  table   | class
+            #   ----- | ---- | ----- | ------------- | -------- | -----
+            #   12345 | 1    | 2024  | 10            | land     | 200
+            #   12345 | 1    | 2024  | 10            | dweldat  | 211
             test_results_select = "select test_results.*"
             test_results_join = ""
             # We need parid and taxyr at minimum in order to rehydrate any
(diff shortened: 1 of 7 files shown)


.github/scripts/transform_dbt_test_results.py
```diff
@@ -1052,24 +1055,62 @@ def get_test_categories(
...
+                        test_results_select += f", par.class AS {CLASS_FIELD}"
+                        test_results_join += f"""
+                            left join iasworld.pardat as par
+                                on par.{PARID_FIELD} =
+                                    test_results.{PARID_FIELD}
+                                and par.{TAXYR_FIELD} =
+                                    test_results.{TAXYR_FIELD}
+                                and par.cur = 'Y'
+                                and par.deactivat is null
+                        """
```
jeancochrane, line 1107: I wonder if there is a cleaner way to do this?

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -403,22 +403,39 @@ sources:
...
+                    - card
+                    - who
+                    - wen
+                    - user14
                   config: *unique-conditions
                   meta:
                     description: >
-                      fixhalf (Number of Half Baths) should be between 0 and 6
+                      fixhalf (Number of Half Baths) should be <= 5 times the
+                      number of units (user14)
```
jeancochrane, line 438: I spot-checked these results and the heuristic seemed to be performing reasonably well to me, but you may want to take a look at the spreadsheet output yourself as well.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -689,33 +706,24 @@ sources:
...
+                      WHEN user14 = '5' THEN 48
+                      ELSE 8
+                    END
+                  additional_select_columns: *select-columns-with-num-units
+                  config: *unique-conditions
                   meta:
                     description: >
-                      rmbed (Number of Bedrooms) should be between 1 and 24
+                      rmbed (Number of Bedrooms) should be <= 8 times the
+                      number of units (user14)
```
jeancochrane, line 726: Similar deal here.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -851,20 +859,63 @@ sources:
...
+                        '210', '211', '212', '234', '278', '295'
+                      )
                       AND cur = 'Y'
                       AND deactivat IS NULL
                   meta:
                     description: >
-                      sfla (Building Square Footage) should be between 1 and 10000
-                      for single-family properties (excl. 208 or 209)
+                      sfla (Building Square Footage) should match the class
+                      definition for class 2 regression properties
```
jeancochrane, line 918: Does this seem reasonable to you? It relies on an assumption that the square foot thresholds set in our [class definitions](https://github.com/ccao-data/wiki/blob/master/Data/Class-Definitions.pdf) are hard and fast rules, but I don't have full confidence in that.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -974,20 +1025,6 @@ sources:
                     description: user3 (Renovated) should be '0' or '1'
           - name: user4
             description: '{{ doc("shared_column_char_tp_dsgn") }}'
-            tests:
-              - not_null:
-                  name: iasworld_dweldat_char_tp_dsgn_not_null
```
jeancochrane, line 979: Our docs describe this field as deprecated, and it's no longer visible in iasWorld as far as I can tell, so Tia and I agreed to remove its tests.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1126,24 +1165,19 @@ sources:
                   meta:
                     description: >
                       user20 (No. of Commercial Units) should not be null
-          - name: user24
-            description: '{{ doc("shared_column_card_proration_rate") }}'
-            tests:
-              - not_null:
-                  name: iasworld_dweldat_card_proration_rate_not_null
```
jeancochrane, line 1133: `user24` is no longer the canonical proration rate column, so we remove this test. See the PR description for details on my (so far failed) attempts to recreate it on `external_propct`.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1126,24 +1165,19 @@ sources:
...
-                    description: user24 (Proration %) should not be null
+                    description: >
+                      user20 (No. of Commercial Units) should be an integer between 0 and 5
+          - name: user24
+            description: '{{ doc("shared_column_card_proration_rate") }}'
           - name: user30
             description: '{{ doc("shared_column_char_porch") }}'
             tests:
-              - not_null:
-                  name: iasworld_dweldat_char_porch_not_null
```
jeancochrane, line 1142: Apparently, `char_porch` (like a few other chars that follow) can be null if the dwelling does not have a porch. I'm not sure if I should be updating the docs in cases like these; happy to do so if you think it's useful to have information in the column docs on what nulls mean.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1126,24 +1165,19 @@ sources:
                   meta:
                     description: >
                       user20 (No. of Commercial Units) should not be null
-          - name: user24
-            description: '{{ doc("shared_column_card_proration_rate") }}'
-            tests:
-              - not_null:
-                  name: iasworld_dweldat_card_proration_rate_not_null
+              - accepted_values:
+                  name: iasworld_dweldat_char_ncu_between_0_and_5
```
jeancochrane, line 1169: This test is not new, but rather has been moved from an `expression_is_true` test that was defined on the model and translated into an `accepted_values` test that is defined on the column. I'm preferring to move tests from the model level to the column level where possible, since it allows the DAG to determine and display which tests are defined on a column in the docs; happy to discuss more if you disagree with that preference.

dbt/models/iasworld/schema/iasworld.land.yml
```diff
@@ -24,7 +24,7 @@ sources:
                   additional_select_columns:
                     - taxyr
                     - parid
-                    - card
+                    - lline
```
jeancochrane, line 27: Whoops, turns out that `card` is a vestigal attribute on `land`, and we should be using `lline` as part of the composite primary key instead.

dbt/models/default/schema/default.vw_pin_value.yml
```diff
@@ -69,12 +69,12 @@ models:
           column_name: certified_class
           config:
             where: CAST(year AS int) < {{ var('test_qc_year_start') }}
-            error_if: ">13" # as of 2024-03-15
+            error_if: ">15" # as of 2024-03-15
```
jeancochrane, line 72: A few of these tests started failing once `test_qc_year_start` got bumped to 2024. My assessment is that the `certified` values represent two new true failures in the underlying data, but the `board` values should actually be tested on the prior year, since board values are finalized much later than mailed/certified values. But correct me if I'm wrong!

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -364,14 +364,30 @@ sources:
...
+                    - card
+                    - who
+                    - wen
+                    - user14
                   config: *unique-conditions
                   meta:
                     description: >
-                      fixbath (Number of Full Baths) should be between 1 and 16
+                      fixbath (Number of Full Baths) should be <= 7 times the
+                      number of units (user14)
```
jeancochrane, line 390: This heuristic seemed reasonable based on my own testing, but you're welcome to double check it.

.github/scripts/transform_dbt_test_results.py
dfsnow: This is horrendous but also I love it. If it works it works. The only way I can think of that might be simpler is to create a "helper view" i.e. a view with the universe of township and class codes by table:

| PIN   | CARD | TAXYR | TOWNSHIP_CODE | TABLE   | CLASS |
|-------|------|-------|---------------|---------|-------|
| 12345 | 1    | 2024  | 10            | LAND    | 200   |
| 12345 | 1    | 2024  | 10            | DWELDAT | 211   |

Then filter the table and join it to get the town and class.

dbt/models/default/schema/default.vw_pin_value.yml
dfsnow: Yes I suppose that's right: there won't be any 2024 board values until next year, so it doesn't really make sense to test for them.

As an aside, we should change or remove the date on these static thresholds.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -364,14 +364,30 @@ sources:
...
+                  max_value: |
+                    CASE
+                      WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 7
+                      WHEN user14 = '1' THEN 14
+                      WHEN user14 = '2' THEN 21
+                      WHEN user14 = '3' THEN 28
+                      WHEN user14 = '4' THEN 35
+                      WHEN user14 = '5' THEN 42
+                      ELSE 7
+                    END
```
dfsnow, line 378: **praise:** Smart!

dbt/models/iasworld/schema/iasworld.dweldat.yml
dfsnow: I think this makes sense!

dbt/models/iasworld/schema/iasworld.dweldat.yml
dfsnow: This makes senses for now, as they _are_ technically supposed to be hard definitions. I commented the ones that need specific changes.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -851,20 +869,63 @@ sources:
...
+                  max_value: |
+                    CASE
+                      WHEN class = '202' THEN 999
+                      WHEN class = '203' THEN 1800
+                      WHEN class = '204' THEN 10000
+                      WHEN class = '205' THEN 2200
+                      WHEN class = '206' THEN 4999
+                      WHEN class = '207' THEN 2000
+                      WHEN class = '208' THEN 4999
+                      WHEN class = '209' THEN 20000
```
dfsnow, line 900: You'd be surprised how big some of these places get:

```suggestion
                      WHEN class = '209' THEN 50000
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -851,20 +869,63 @@ sources:
...
+                      WHEN class = '210' THEN 10000
+                      WHEN class = '211' THEN 20000
+                      WHEN class = '212' THEN 20000
```
dfsnow, line 903:
```suggestion
                      WHEN class = '211' THEN 40000
                      WHEN class = '212' THEN 40000
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -851,20 +869,63 @@ sources:
...
+                      WHEN class = '234' THEN 10000
+                      WHEN class = '278' THEN 3800
+                      WHEN class = '295' THEN 10000
```
dfsnow, line 906:
```suggestion
                      WHEN class = '295' THEN 20000
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
dfsnow: It's probably going to require some chats with Mirella to figure out how these new columns work, we didn't need to touch them for 2024 modeling but will going forward. I opened a new issue #392 to cover this.

dbt/models/iasworld/schema/iasworld.dweldat.yml
dfsnow: That makes sense to me!

dbt/models/iasworld/schema/iasworld.dweldat.yml
dfsnow: I always assumed that but never had it explicitly confirmed. Let's definitely update the docs.

dbt/models/iasworld/schema/iasworld.land.yml
dfsnow: We definitely should, as it's already correctly defined as the key in [the docs](https://ccao-data.github.io/data-architecture/#!/source/source.athena.iasworld.land). Must've missed this during a prior review!

.github/scripts/transform_dbt_test_results.py
jeancochrane: Very smart! I like the idea of a helper table a lot but I don't want to hold up the PR for it. I left a comment to this effect in 77305a584265f770a3b4df6be50d5b1277c03fa3, we can pick it up next time we find ourselves needing to adjust this logic.

jeancochrane:
Thanks for the detailed review @dfsnow! I left clarifying questions on two of your comments and made fixes based on the rest. This should be ready for another look.

Unrelated to your review, I also decided to split out `iasworld_dweldat_sfla_matches_class_definition` into individual tests based on class, which I think makes for a clearer workbook reading experience. Now the description states exactly what the square footage should be for the record's class. You can run just this new set of tests if you'd like to see what the output looks like:

```
dbt test --store-failures --select \
	iasworld_dweldat_sfla_between_1_and_999_for_class_202
	iasworld_dweldat_sfla_between_1000_and_1800_for_class_203
	iasworld_dweldat_sfla_between_1801_and_10000_for_class_204
	iasworld_dweldat_sfla_between_1_and_2200_for_class_205
	iasworld_dweldat_sfla_between_2201_and_4999_for_class_206
	iasworld_dweldat_sfla_between_1_and_2000_for_class_207
	iasworld_dweldat_sfla_between_3801_and_4999_for_class_208
	iasworld_dweldat_sfla_between_5000_and_50000_for_class_209
	iasworld_dweldat_sfla_between_1_and_10000_for_class_210
	iasworld_dweldat_sfla_between_1_and_40000_for_class_211
	iasworld_dweldat_sfla_between_1_and_20000_for_class_212
	iasworld_dweldat_sfla_between_1_and_10000_for_class_234
	iasworld_dweldat_sfla_between_2001_and_3800_for_class_278
	iasworld_dweldat_sfla_between_1_and_20000_for_class_295
```

dbt/models/default/schema/default.vw_pin_value.yml
dfsnow: Oh, the commented date on here for the error threshold is `# as of 2024-03-15` but it wasn't updated with the change. My suggestion was just to update it to `# as of 2024-04-16` or similar.

dbt/models/shared_columns.md
```diff
@@ -649,7 +650,8 @@ Possible values for this variable are:
 {% docs shared_column_char_hbath %}
 Number of half baths.
 
-Defined as bathrooms without a shower or bathtub
+Defined as bathrooms without a shower or bathtub. Note that either 0 or
```
dfsnow, line 653: **suggestion:** Let's make this same update for `char_porch`

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -916,31 +882,208 @@ sources:
...
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '278'
+                  meta:
+                    description: >
+                      sfla (Building Square Footage) should be between
+                      2001 and 3800 for class 278 cards
+              - accepted_range:
```
dfsnow, line 1072: **suggestion:** We may want to create similar tests for `yrblt`. Probably worth opening another issue so as not to hold this one up any longer.

dfsnow (approved):

dbt/models/default/schema/default.vw_pin_value.yml
jeancochrane: Gotcha, I removed them in 250050ec5bc2b88bff74e98a993a85c864554c78 since we can always inspect the git diff if we want to see when a threshold was set.

dbt/models/shared_columns.md
jeancochrane: `char_porch` has a bulleted list of values, so I just added it there:

https://github.com/ccao-data/data-architecture/blob/250050ec5bc2b88bff74e98a993a85c864554c78/dbt/models/shared_columns.md?plain=1#L682-L687

Does that seem good enough to you, or should we add a prose description instead?

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: I was thinking the same thing, issue open here: https://github.com/ccao-data/data-architecture/issues/398

PR #378, opened 2024-04-12
Add new test to make sure tieback pins don't contain hyphens

Damonamajor:
This test results in a failure with 3024 results.

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.pardat.yml
+++ b/dbt/models/iasworld/schema/iasworld.pardat.yml
@@ -261,11 +261,12 @@ sources:
             description: '{{ doc("shared_column_tieback_key_pin") }}'
             tests:
               - expression_is_true: 
-                  name: iasworld_pardat+tieback_does_not_contain_hyphen
+                  name: iasworld_pardat_tieback_does_not_contain_hyphen
                   expression: |
                     not like '%-%'
-                  meta:
-                    description: tieback should not have any hyphens                  
+                  additional_select_columns: 
+                    - parid
+                    - taxyr
           - name: tiebkcd
             description: Tieback code
           - name: tiebldgpct
--- a/dbt/models/iasworld/schema/iasworld.pardat.yml
+++ b/dbt/models/iasworld/schema/iasworld.pardat.yml
@@ -259,6 +259,11 @@ sources:
             description: '{{ doc("shared_column_year") }}'
           - name: tieback
             description: '{{ doc("shared_column_tieback_key_pin") }}'
+            tests:
+              - expression_is_true: 
+                  name: iasworld_pardat_does_not_contain_hyphen
+                  expression: |
+                    not like '%-%'
           - name: tiebkcd
             description: Tieback code
           - name: tiebldgpct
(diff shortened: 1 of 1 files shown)


Damonamajor:
@jeancochrane - Do we want a timeframe attached to this, or other configs that we used in the past?

dbt/models/iasworld/schema/iasworld.pardat.yml
```diff
@@ -259,6 +259,14 @@ sources:
             description: '{{ doc("shared_column_year") }}'
           - name: tieback
             description: '{{ doc("shared_column_tieback_key_pin") }}'
+            tests:
+              - expression_is_true: 
+                  name: iasworld_pardat_tieback_does_not_contain_hyphen
+                  expression: |
+                    not like '%-%'
```
jeancochrane, line 266: **[Nitpick, non-blocking]** The pipe character in this context indicates a multiline string, but since the string itself is short enough to fit on one line, there shouldn't be a need for it:

```suggestion
                  expression: not like '%-%'
```

dbt/models/iasworld/schema/iasworld.pardat.yml
```diff
@@ -259,6 +259,14 @@ sources:
...
+                  additional_select_columns: 
+                    - parid
+                    - taxyr
```
jeancochrane, line 269: **[Nitpick]** We can use a [YAML alias](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) here to select the same set of columns as the majority of the other tests in this file. Let's also add a `config` attribute and set it to the `*unique-conditions` alias so that we can apply the same filtering by date and active records that the other tests apply:

```suggestion
                  additional_select_columns: *select-columns
                  config: *unique-conditions
```

For reference, here are the definitions of the `&select-columns` and `&unique-conditions` anchors, so you can see what they're doing:

https://github.com/ccao-data/data-architecture/blob/14095e2a9bb15106ae95807f2e8302a34a3b7e45/dbt/models/iasworld/schema/iasworld.pardat.yml#L82-L86

https://github.com/ccao-data/data-architecture/blob/14095e2a9bb15106ae95807f2e8302a34a3b7e45/dbt/models/iasworld/schema/iasworld.pardat.yml#L106-L110

jeancochrane, line 269: **[Nitpick]** Let's add `meta` attributes to make sure this test gets displayed correctly in the output workbook. We'll need to set these two attributes under the `meta` key:

* `category: incorrect_values` (necessary because `expression_is_true` is very general and so cannot be automatically assigned to a specific test category by [the code in `transform_dbt_test_results.py`](https://github.com/ccao-data/data-architecture/blob/14095e2a9bb15106ae95807f2e8302a34a3b7e45/.github/scripts/transform_dbt_test_results.py#L1113-L1131))
* `description: tieback should not contain hyphens`

For reference, here's an example of a test that defines these two attributes, albeit with different values:

https://github.com/ccao-data/data-architecture/blob/14095e2a9bb15106ae95807f2e8302a34a3b7e45/dbt/models/iasworld/schema/iasworld.pardat.yml#L77-L96

jeancochrane (changes requested):
Looks great, I tested it out locally and the test runs as expected! Just a few small changes to set the required `meta` properties and implement date/active record filtering, and then this should be good to go.

jeancochrane (approved):
Great work here!

PR #389, opened 2024-04-17
Tweak dbt data docs sidebar

dfsnow:
I got sufficiently annoyed with the dbt docs sidebar that I took a minute to tweak it. Here we:

- Remove the useless "Group" tab
- Open the "Database" tab and "awsdatacatalog" DB by default


jeancochrane (approved):
Looks good, I'm down to test in prod!

Issue #328, opened 2024-02-29
Define a dbt test to ensure that exempt `pardat` parcels are also exempt in all related tables


jeancochrane:
From [discussion](https://github.com/ccao-data/data-architecture/pull/323#pullrequestreview-1909703397) on #323:

> Land actually can be exempt on a non-exempt parcel e.g. you can have a `EX` land code on a 211 pardat parcel. However, the reverse is not true (can't have an EX pardat parcel and a 200 land class).

Add one or more tests to confirm that parcels that are exempt in `pardat` are also exempt in all of the other tables that have `class` columns.

jeancochrane:
Once https://github.com/ccao-data/data-architecture/pull/390 lands, we can use it to test the following models:

* `addn`
* `asmt_all`
* `land`
* `oby`

PR #390, opened 2024-04-17
Tweak `res_class_matches_pardat` generic test to add pardat join kwargs

jeancochrane:
This PR makes a small modification to the `res_class_matches_pardat` generic test in order to support the test spec in https://github.com/ccao-data/data-architecture/issues/328.

We add two new optional kwargs to the generic interface:

* `join_type`: The type of join to use for joining the model to pardat (e.g. `inner` or `left`, defaults to `left`).
* `additional_pardat_filter`: A SQL string representing additional conditions to use when selecting records from pardat to join to the model (e.g. `class != 'EX' and class != 'RR'`). `cur = 'Y'` and `deactivat IS NULL` are already set, hence "additional" pardat filter.

The combination of these two new kwargs will allow us to filter a test for only those rows whose corresponding pardat class is `"EX"`. This is demonstrated in a dweldat test that can be generalized to `addn`, `asmt_all`, `oby`, and `land` in order to complete https://github.com/ccao-data/data-architecture/issues/328.

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.dweldat.yml
+++ b/dbt/models/iasworld/schema/iasworld.dweldat.yml
@@ -203,7 +203,7 @@ sources:
                   name: iasworld_dweldat_exempt_classes_match_pardat_class
                   additional_select_columns: *select-columns-for-pardat-join
                   join_type: inner
-                  additional_join_condition: AND class = 'EX'
+                  additional_pardat_filter: AND class = 'EX'
                   config:
                     where: |
                       CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
--- a/dbt/models/iasworld/schema/iasworld.dweldat.yml
+++ b/dbt/models/iasworld/schema/iasworld.dweldat.yml
@@ -185,7 +185,7 @@ sources:
                     description: class code must be valid
               - res_class_matches_pardat:
                   name: iasworld_dweldat_class_matches_pardat_class
-                  additional_select_columns:
+                  additional_select_columns: &select-columns-for-pardat-join
                     - column: card
                       agg_func: array_agg
                     - column: taxyr
@@ -199,6 +199,21 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: at least one class should match pardat class
+              - res_class_matches_pardat:
+                  name: iasworld_dweldat_exempt_classes_match_pardat_class
+                  additional_select_columns: *select-columns-for-pardat-join
+                  join_type: inner
+                  additional_join_condition: and class = 'EX'
+                  config:
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class NOT LIKE 'OA%'
+                  meta:
+                    description: >
+                      at least one class should be exempt or omitted if pardat
+                      is exempt
           - name: cline
             description: Condominium line
           - name: cndadjval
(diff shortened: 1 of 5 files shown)


dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -28,8 +28,8 @@
 --
 -- Optional parameters:
 --
---    * join_type(str): The type of join to use, e.g. "inner join" or
---    "left join". Defaults to "inner join".
+--    * join_type(str): The type of join to use, e.g. "inner" or "left".
+--    Defaults to "inner".
```
jeancochrane, line 32: I tweaked this test to match the interface used by `res_class_matches_join`, since on second thought it feels more intuitive for me for "join type" to refer to the type of join ("inner" or "outer") rather than the full join string itself ("inner join" or "outer join"). I'm definitely open to pushback if you disagree. There's no need to adjust any tests to use this new interface, since `join_type` isn't actually specified in any active tests right now.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -199,6 +199,21 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: at least one class should match pardat class
+              - res_class_matches_pardat:
+                  name: iasworld_dweldat_exempt_classes_match_pardat_class
+                  additional_select_columns: *select-columns-for-pardat-join
+                  join_type: inner
+                  additional_join_condition: and class = 'EX'
```
dfsnow, line 206: **mega-nitpick:** For the sake of consistency:

```suggestion
                  additional_join_condition: AND class = 'EX'
```

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -86,12 +95,16 @@
         -- Pardat should be unique by (parid, taxyr), so we can select the
         -- max rather than array_agg
         max(pardat.class) as pardat_class
-    from filtered_model
-    left join
+    from filtered_model {{ join_type }}
+    join
```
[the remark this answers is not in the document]
dfsnow: **suggestion:** You could throw a `-- fmt: off` on that specific line.

dbt/tests/generic/test_res_class_matches_pardat.sql
```diff
@@ -86,12 +95,16 @@
...
         (
             select *
             from {{ source("iasworld", "pardat") }}
-            where cur = 'Y' and deactivat is null
+            where
+                cur = 'Y' and deactivat is null
+                {% if additional_join_condition %}
+                    {{ additional_join_condition }}
+                {% endif %}
```
dfsnow, line 107: FWIW I interpreted `additional_join_conditions` to mean they would be placed within the join itself i.e. `ON {{ additional_join_conditions`. I don't _think_ that would change the query outcome, though it might change how the planner works. Probably fine either way.

dfsnow (approved):
Looks great! Thanks for the edits

dbt/tests/generic/test_res_class_matches_pardat.sql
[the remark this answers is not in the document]
jeancochrane: Of course! I thought I remembered there being an exclude tag for sqlfmt, but I couldn't remember it 🤦🏻 Fixed in f8e75ee0b0068f80886f17ca9de9cab4b77bd738.

dbt/tests/generic/test_res_class_matches_pardat.sql
jeancochrane: Ah, good point. Renamed to `additional_pardat_filter` in 9bf577e63dd08268e869f493b8c42cfaac63cd32 for clarity.

PR #391, opened 2024-04-17
Add dbt tests for sales view legacy filters

wrridgeway:
We need to be sure our legacy sales filters are doing what is expected of them. This can be handled through dbt tests for `default.vw_pin_sale`.


jeancochrane (approved):
Looks great!

PR #395, opened 2024-04-18
Define a dbt test to ensure that exempt pardat parcels are also exempt in all related tables

Damonamajor:
This produces three tests, one for asmt_all, oby, and land, each of which tests that all values in pardat which are exempt, are exempt in the tested tables. 
It fails in two occasions, land (14 failures) and oby (3), while passing in asmt_all.

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.asmt_all.yml
+++ b/dbt/models/iasworld/schema/iasworld.asmt_all.yml
@@ -37,6 +37,29 @@ sources:
                   meta:
                     category: class_mismatch_or_issue
                     description: class code must be valid
+              - row_values_match_after_join:
+                  name: iasworld_asmt_all_exempt_pardat_exempt
+                  external_model: source('iasworld', 'pardat')
+                  external_column_name: class
+                  column_alias: class
+                  external_column_alias: class
+                  group_by:
+                    - taxyr
+                    - parid                    
+                  join_condition:
+                    On model.parid = external_model.parid
+                    AND model.taxyr = external_model.taxyr
+                    AND external_model.cur = 'Y'
+                    AND external_model.deactivat IS NULL
+                  where:
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND rolltype != 'RR'
+                      AND deactivat IS NULL
+                      AND valclass IS NULL
+                      AND class IN ('EX')
+                  meta:
+                    category: class_mismatch_or_issue
+                    description: exempt values in land must be exempt
               - res_class_matches_pardat:
                   name: iasworld_asmt_all_class_matches_pardat_class
                   additional_select_columns:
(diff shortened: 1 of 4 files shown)


dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -57,6 +57,16 @@ sources:
                       AND valclass IS NULL
                   meta:
                     description: at least one class should match pardat class
+              - res_class_matches_pardat:
+                  name: iasworld_asmt_all_ex_matches_pardat_class_ex
+                  major_class_only: false
+                  join_type: inner
+                  additional_select_columns: *select-columns-aggregated
+                  additional_pardat_filter: AND class = 'EX'
```
Damonamajor, line 65: I took a look at each of the unique conditions. As far as I understood, none interacted with 'Ex', but that is something to keep in mind.

dbt/models/iasworld/schema/iasworld.land.yml
```diff
@@ -112,6 +112,16 @@ sources:
                   meta:
                     description: >
                       at least one major class (first digit) should match pardat class
+              - res_class_matches_pardat:
+                  name: iasworld_land_ex_matches_pardat_class_ex
+                  major_class_only: false
+                  join_type: inner
```
Damonamajor, line 118: Join counts with left are quite high. But, inner joins also make more sense conceptually.

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -57,6 +57,16 @@ sources:
                       AND valclass IS NULL
                   meta:
                     description: at least one class should match pardat class
+              - res_class_matches_pardat:
+                  name: iasworld_asmt_all_ex_matches_pardat_class_ex
+                  major_class_only: false
```
Damonamajor, line 62: major class true and false produced the same outcome.

dbt/models/iasworld/schema/iasworld.land.yml
jeancochrane: **[Praise]** Yup! We need the inner join here since otherwise the `class = 'EX'` filter on `pardat` will cause the majority of records in `land` to have no match in `pardat`, causing the test to compare the land class to `NULL`.

dbt/models/iasworld/schema/iasworld.asmt_all.yml
```diff
@@ -57,6 +57,16 @@ sources:
...
+                  join_type: inner
+                  additional_select_columns: *select-columns-aggregated
+                  additional_pardat_filter: AND class = 'EX'
+                  config: *unique-conditions
+                  meta:
+                    description: >
+                      Exempt parcels in pardat should be exempt in asmt all
```
jeancochrane, line 69: **[Nitpick]** The output workbook includes the tablename of the failing row as the first column, so I try to avoid repeating the tablename in the test description for the sake of brevity. I also try to start the description with the name of the field that is being tested so that a reader can scan the descriptions and quickly parse out which field failed a test. A better description here (and in the rest of the tests in this PR) would be something like:

```suggestion
                      class should be EX if pardat class is EX
```

jeancochrane (changes requested):
Looking great! Just a few small changes and this should be ready to go.

jeancochrane (approved):
Very nice!

PR #400, opened 2024-04-19
Add test for nbhd code with mismatched township

### LINKED ISSUE #275: Double-dipped neighborhood codes
In `vw_pin_universe`, at least one nbhd code is wandering around, hanging out in two townships. 

Specifically, when `tax_year` = 2023 and `neighborhood_code` = 77131, there are 24,639 rows associated with township code 77 (West Chicago)... and 463 rows with 76 (South Chicago). Weird to have a between-geography discrepancy within a tax year 🗺. 

This is also replicated in the Parcel Universe asset on Open Data. 

The goal of this issue is to investigate (and fix), per each tax year, where one 5-digit neighborhood code is associated with more than one township. 

Let's please help these poor lost PINs! Please:
- [ ] identify the scope and source of the issue
- [ ] plan to correct (if possible), ideally at the source so that corrections flow downstream. 
- [ ] make sure this doesn't affect, or is corrected before, we finish our 2024 modeling.
- [ ] if advisable, add to dbt tests (perhaps something like, check for a discrepancy between first two digits of a PIN's nbhd code and its township code).

Clues: 
- All of the PINs associated in this neighborhood and "South Chicago" are 299s with the same PIN10. Hopefully this entire issue is caused by just one building!
- May be related to divisions/consolidations. That's good because it means it probably won't affect our 2024 modeling...but even with that it still seems weird to have a between-geography discrepancy within one tax year.

Routing to @dfsnow to prioritize, assign, etc. as you deem fit

dfsnow:
This PR adds a iasWorld QC test for neighborhood codes that don't match their township, pursuant to the research in #275.

Closes #275.

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.pardat.yml
+++ b/dbt/models/iasworld/schema/iasworld.pardat.yml
@@ -167,6 +167,30 @@ sources:
             description: Jurisdictions (Municipalities) used as part of Folio number (BCA)
           - name: nbhd
             description: '{{ doc("shared_column_nbhd_code") }}'
+            tests:
+            - row_values_match_after_join:
+                name: iasworld_pardat_nbhd_matches_legdat_township
+                external_model: source('iasworld', 'legdat')
+                external_column_name: user1
+                column_alias: nbhd_code
+                external_column_alias: township_code
+                group_by:
+                  - parid
+                  - taxyr
+                join_condition:
+                  ON  substr(model.nbhd, 1, 2) != external_model.user1
+                  AND model.parid = external_model.parid
+                  AND model.taxyr = external_model.taxyr
+                  AND model.cur = 'Y'
+                  AND model.deactivat IS NULL
+                  AND external_model.cur = 'Y'
+                  AND external_model.deactivat IS NULL
+                config:
+                  where:
+                    CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                meta:
+                  category: relationships
+                  description: nbhd code first 2 digits should match legdat.user1 (township code)
           - name: nbhdie
             description: Neighborhood income valuation
           - name: nonfarminc
--- a/dbt/models/iasworld/schema/iasworld.pardat.yml
+++ b/dbt/models/iasworld/schema/iasworld.pardat.yml
@@ -183,13 +183,9 @@ sources:
                   ON  substr(model.nbhd, 1, 2) != external_model.user1
                   AND model.parid = external_model.parid
                   AND model.taxyr = external_model.taxyr
-                  AND model.cur = 'Y'
-                  AND model.deactivat IS NULL
                   AND external_model.cur = 'Y'
                   AND external_model.deactivat IS NULL
-                config:
-                  where:
-                    CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                config: *unique-conditions
                 meta:
                   category: relationships
                   description: nbhd code first 2 digits should match legdat.user1 (township code)
--- a/dbt/models/iasworld/schema/iasworld.pardat.yml
+++ b/dbt/models/iasworld/schema/iasworld.pardat.yml
@@ -174,9 +174,11 @@ sources:
                 external_column_name: user1
                 column_alias: nbhd_code
                 external_column_alias: township_code
-                group_by:
-                  - parid
-                  - taxyr
+                additional_select_columns:
+                  - model.taxyr
+                  - model.parid
+                  - model.who
+                  - model.wen
                 join_condition:
                   ON  substr(model.nbhd, 1, 2) != external_model.user1
                   AND model.parid = external_model.parid
--- a/dbt/tests/generic/test_row_values_match_after_join.sql
+++ b/dbt/tests/generic/test_row_values_match_after_join.sql
@@ -107,6 +107,8 @@
     group by {{ group_by_csv }}
     having
         sum(case when {{ external_model_col }} = {{ model_col }} then 1 else 0 end) = 0
+    {%- else %}
+    where {{ external_model_col }} != {{ model_col }}
     {%- endif %}
 
 {% endtest %}
--- a/dbt/tests/generic/test_row_values_match_after_join.sql
+++ b/dbt/tests/generic/test_row_values_match_after_join.sql
@@ -20,13 +20,13 @@
 --    This is not necessary in the case of USING, since USING does not
 --    refer to tablenames directly.
 --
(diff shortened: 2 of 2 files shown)


dbt/models/iasworld/schema/iasworld.pardat.yml
```diff
@@ -167,6 +167,32 @@ sources:
...
+                name: iasworld_pardat_nbhd_matches_legdat_township
+                external_model: source('iasworld', 'legdat')
+                external_column_name: user1
+                column_alias: nbhd_code
+                external_column_alias: township_code
+                additional_select_columns:
+                  - model.taxyr
+                  - model.parid
+                  - model.who
+                  - model.wen
```
dfsnow, line 181: I had to prefix these columns with `model.` since they exist in both the target model and the joined one. It's kind of messy. Maybe we could build some handling into the macro?

dbt/tests/generic/test_row_values_match_after_join.sql
dfsnow, line 1: All the changes in this test are related to making the `group_by` argument optional.

dbt/models/iasworld/schema/iasworld.pardat.yml
```diff
@@ -167,6 +167,32 @@ sources:
...
+                join_condition:
+                  ON  substr(model.nbhd, 1, 2) != external_model.user1
```
jeancochrane, line 183: **[Praise]** Nice, this is a clever way to adapt this test to your use case.

dbt/models/iasworld/schema/iasworld.pardat.yml
```diff
@@ -167,6 +167,32 @@ sources:
...
+                  AND model.parid = external_model.parid
+                  AND model.taxyr = external_model.taxyr
+                  AND model.cur = 'Y'
+                  AND model.deactivat IS NULL
```
jeancochrane, line 187: **[Nitpick, non-blocking]** It's a little bit arbitrary, but our pattern so far as been to include static filters like these in the `config.where` block rather than the `join_condition`, mostly because it helps us reuse `config.where` logic via the `unique-conditions` anchor. I'm open to being convinced that there's a good reason to do it here rather than in `config.where`, though.

dbt/models/iasworld/schema/iasworld.pardat.yml
jeancochrane: **[Suggestion, non-blocking]** Yup, I support updating the `row_values_match_after_join` to make this easier. Here's how we do it in the `res_class_matches_pardat` generic:

https://github.com/ccao-data/data-architecture/blob/e313087555fa3afebd91e16445640947175a1570/dbt/tests/generic/test_res_class_matches_pardat.sql#L50-L78

I'm fine with breaking this out into its own issue to avoid blocking this PR.

dbt/tests/generic/test_row_values_match_after_join.sql
```diff
@@ -87,18 +87,26 @@
...
+            , {{ additional_select_columns_csv }}
         {% endif %}
-        array_agg({{ model_col }}) as {{ column_alias }},
-        array_agg({{ external_model_col }}) as {{ external_column_alias }}
     from
         {{ external_model }} as external_model
     {{ join_type }} join -- fmt: off
         (select * from {{ model }}) as model {{ join_condition }}
+
+    {%- if group_by is defined %}
```
jeancochrane, line 106: **[Question, non-blocking]** If `group_by` is not defined, but the test doesn't implement a `join_condition` that specifically filters for mismatching rows (as the one in this PR does), then am I right that the test won't catch mismatching rows due to the `having` clause being excluded? That works, but it requires the caller to understand the implementation details of the generic to know that it won't automatically filter for mismatching rows, so I wonder if there's a way we could make the interface friendlier in that case.

jeancochrane (approved):
Looking great! A couple suggestions for interface improvements, but I'm fine punting them to later so we can bring this in and get the test out the door.

dbt/models/iasworld/schema/iasworld.pardat.yml
dfsnow: There actually is a reason here: conditionals in the `config: where:` clause can't be namespaced to a table since they're applied to the whole test result (which is an inner query). Since columns here share the same name across joined tables, we need to include the namespacing.

dbt/tests/generic/test_row_values_match_after_join.sql
dfsnow: Good catch! I think defaulting to a WHERE clause here solves this, see c4f473f.

dbt/models/iasworld/schema/iasworld.pardat.yml
dfsnow: Moved to a new issue! #405

dbt/models/iasworld/schema/iasworld.pardat.yml
jeancochrane: Hmm, I don't quite understand why namespacing is required in this case. I understand why any filters applied in the `join_condition` block need to be namespaced, but any filters in the `config.where` block are applied to the base model in a subquery prior to the join, so there shouldn't be any namespacing issues as far as filtering the base model is concerned. For example, running the test with this diff returns the same set of results as the current code in this PR:

```diff
--- a/dbt/models/iasworld/schema/iasworld.pardat.yml
+++ b/dbt/models/iasworld/schema/iasworld.pardat.yml
@@ -183,13 +183,9 @@ sources:
                   ON  substr(model.nbhd, 1, 2) != external_model.user1
                   AND model.parid = external_model.parid
                   AND model.taxyr = external_model.taxyr
-                  AND model.cur = 'Y'
-                  AND model.deactivat IS NULL
                   AND external_model.cur = 'Y'
                   AND external_model.deactivat IS NULL
-                config:
-                  where:
-                    CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                config: *unique-conditions
                 meta:
                   category: relationships
                   description: nbhd code first 2 digits should match legdat.user1 (township code)
```

But maybe I'm misunderstanding something about the way the join is constructed! I don't think it's a particularly big deal either way.

jeancochrane (approved):
Nice work! One remaining thought on the issue with namespacing the filters for the base model, but you should feel free to resolve that however you see fit and merge when ready.

dbt/models/iasworld/schema/iasworld.pardat.yml
dfsnow: Oh I see I see, you only wanted to move the `model.` conditions into the `config:` not both the `model.` and `external_model.` ones. That makes sense. Fixed in 24b97da!

PR #404, opened 2024-04-22
Adjust `mktadj` tests to reflect the switch to `external_occpct`and new `5B` code

### LINKED ISSUE #399: Adjust `mktadj` tests to reflect the switch to `external_occpct`
In 2023, the office started moving from `mktadj` to `external_occpct` on dweldat, comdat, and oby to indicate occupancy percentage. We have a few tests (particularly on `mktrsn`) that reference `mktadj` exclusively to indicate this field; we should adjust these tests so that they check `external_occpct` as well.

jeancochrane:
In 2023, we began migrating from the `mktadj` column to the `external_occpct` column on `dweldat`, `comdat`, and `oby`. Prior to 2023, only `mktadj` should be populated in cases of fractional occupancy; in 2023, either `mktadj` or `external_occpct` (but not both) can be populated, representing the time when the fields were in transition; and starting in 2024, only `external_occpct` should be populated.

At the same time, we recently added `"5B"` as a valid option for `chgrsn` and `mktrsn` to help differentiate between 1-year reductions issued by our office and the board of review.

This PR adjusts our tests for the `mktadj`, `mktrsn`, and `chgrsn` columns on `dweldat`, `comdat`, and `oby` to reflect these two changes.

Closes #399.

Run the tests to ensure they execute properly:

```
dbt test --select \
    iasworld_comdat_chgrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null \
    iasworld_comdat_external_occpct_not_null_when_chgrsn_eq_5_or_5b_and_mktadj_is_null \
    iasworld_comdat_mktadj_not_null_when_chgrsn_eq_5_or_5b_and_external_occpct_is_null \
    iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null \
    iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null \
    iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null \
    iasworld_oby_chgrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null \
    iasworld_oby_external_occpct_not_null_when_chgrsn_eq_5_or_5b_and_mktadj_is_null \
    iasworld_oby_mktadj_not_null_when_chgrsn_eq_5_or_5b_and_external_occpct_is_null
```

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.comdat.yml
+++ b/dbt/models/iasworld/schema/iasworld.comdat.yml
@@ -101,21 +101,30 @@ sources:
           - name: chgrsn
             description: '{{ doc("column_chgrsn") }}'
             tests:
-              - expression_is_true:
-                  name: iasworld_comdat_chgrsn_eq_5_mktadj_not_null
-                  expression: |
-                    != '5' OR mktadj IS NOT NULL
+              - accepted_values:
+                  name: iasworld_comdat_chgrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null
+                  values:
+                    - '5'
+                    - 5B
                   additional_select_columns:
                     - parid
                     - taxyr
                     - card
                     - who
                     - wen
+                    - external_occpct
                     - mktadj
-                  config: *unique-conditions
+                  config:
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND (external_occpct IS NOT NULL OR mktadj IS NOT NULL)
                   meta:
-                    category: incorrect_values
-                    description: chgrsn should only be 5 if mktadj is not null
+                    description: >
+                      chgrsn (Reason for Change) should be 5 or 5B
+                      if external_occpct (Occupancy % [current]) or
+                      mktadj (Occupancy % [deprecated]) is not null
           - name: class
             description: '{{ doc("shared_column_class") }}'
             tests:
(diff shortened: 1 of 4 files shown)


dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -101,21 +101,30 @@ sources:
           - name: chgrsn
             description: '{{ doc("column_chgrsn") }}'
             tests:
-              - expression_is_true:
-                  name: iasworld_comdat_chgrsn_eq_5_mktadj_not_null
-                  expression: |
-                    != '5' OR mktadj IS NOT NULL
+              - accepted_values:
```
jeancochrane, line 104: While the `expression_is_true` version of this test had a nice conciseness, I've always found it confusing to wrap my head around the double negative defined in the `expression`. In this PR, I've split each `expression_is_true` test out into three `accepted_values` tests, which are (in my opinion) more verbose but easier to understand.

dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -101,21 +101,30 @@ sources:
...
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND (external_occpct IS NOT NULL OR mktadj IS NOT NULL)
                   meta:
-                    category: incorrect_values
-                    description: chgrsn should only be 5 if mktadj is not null
+                    description: >
+                      chgrsn (Reason for Change) should be 5 or 5B
+                      if external_occpct (Occupancy % [current]) or
+                      mktadj (Occupancy % [deprecated]) is not null
```
jeancochrane, line 127: Note that this test logic is not as thorough as it could be; we've been told that the presence of `mktadj` or `external_occpct` should correspond to a specific timeline (see PR description), but so far these tests do not capture that time dimension. Do you think it's worth refactoring the tests to specify that `mktadj` should be the populated field pre-2023, and `external_occpct` should be the populated field post-2024?

dbt/models/iasworld/schema/iasworld.comdat.yml
dfsnow: **praise:** It was definitely confusing, this is much more readable.

dfsnow: **praise:** It was definitely confusing, this is much more readable.

dbt/models/iasworld/schema/iasworld.comdat.yml
dfsnow: No, if we can't correct historical data then there's not really a point to worrying about a time dimension for any tests IMO.

dfsnow (approved):
This looks great! Nice refactor @jeancochrane. I see we already have some failures 👀

dbt/models/iasworld/schema/iasworld.comdat.yml
jeancochrane: That makes sense @dfsnow. I wonder if perhaps we should just test `external_occpct` then, and leave `mktadj` out of it?

dfsnow: I think it's fine to test both, on the off chance that someone accidentally uses `mktadj`.

PR #406, opened 2024-04-22
Refactor pre-commit to use composite action

wagnerlmichael:
This PR refactors the `.github/workflows/pre-commit.yaml` to use the `action.yaml` in the actions repo. However, there are key difference that makes it so that it isn't a 1:1 replacement. I'm not sure how to incorporate these differences, mostly because I am not familiar with the exact use case. I could build these into the `action.yaml`, or perhaps try to get some sort of conditional logic going.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -13,4 +13,4 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Run pre-commit checks
-        uses: ccao-data/actions/pre-commit@create-composite-pre-commit-action
+        uses: ccao-data/actions/pre-commit@main
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -10,29 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Setup python
-        uses: actions/setup-python@v4
-        with:
-          python-version: 3.x
-
-      - name: Install pre-commit
-        run: python -m pip install pre-commit
-        shell: bash
-
-      - name: Freeze dependencies
-        run: python -m pip freeze --local
-        shell: bash
-
-      - name: Cache pre-commit environment
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cache/pre-commit
-            ~/.cache/R
-          key: pre-commit-3-${{ env.pythonLocation }}-${{ hashFiles('.pre-commit-config.yaml') }}
-
-      - name: Run pre-commit
-        run: pre-commit run --show-diff-on-failure --color=always --all-files
-        shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit.yaml
```diff
@@ -10,29 +10,7 @@ jobs:
...
-        with:
-          python-version: 3.x
-
-      - name: Install pre-commit
-        run: python -m pip install pre-commit
-        shell: bash
-
-      - name: Freeze dependencies
-        run: python -m pip freeze --local
-        shell: bash
```
wagnerlmichael, line 26: The current iteration of the `action.yaml` in the actions repo doesn't have this included. Should I include this there? I'm not sure how useful/generalizable this would be to other repos.

.github/workflows/pre-commit.yaml
```diff
@@ -10,29 +10,7 @@ jobs:
...
-
-      - name: Cache pre-commit environment
-        uses: actions/cache@v3
-        with:
-          path: |
-            ~/.cache/pre-commit
-            ~/.cache/R
-          key: pre-commit-3-${{ env.pythonLocation }}-${{ hashFiles('.pre-commit-config.yaml') }}
```
wagnerlmichael, line 34: These are 2 other difference that aren't included in the `action.yaml` file from the actions repo. Thoughts on editing the action file?

.github/workflows/pre-commit.yaml
jeancochrane: This looks like a holdover from [the now-deprecated official pre-commit action](https://github.com/pre-commit/action/blob/f7acafac0271bdd064cdfa1b13f17b4350e565ed/action.yml#L13-L14) that this workflow was originally based on. My guess is that the purpose of this line was to aid in debugging errors by printing the exact versions of Python dependencies that are installed alongside the `pre-commit` package. I say we continue to leave it out and reevaluate if and when we find ourselves debugging issues with the dependencies.

.github/workflows/pre-commit.yaml
jeancochrane: I don't think adding `env.pythonLocation` to the cache key will be useful unless we start publishing Python packages where we would like to run pre-commit against multiple supported versions of Python. I also don't expect that explicitly saving the R cache directory should be necessary, since pre-commit claims it [saves all environments in its cache](https://pre-commit.com/#managing-ci-caches), but I could be wrong about this; I say we check to see whether R hooks are being cached properly when we refactor a repo that includes R hooks to switch to the composite action (e.g. [`model-res-avm`](https://github.com/ccao-data/model-res-avm/blob/cb7cdc1b65ce6321f5a968ee979d6ad0eda215fb/.pre-commit-config.yaml#L4-L14)) and reevaluate then.

jeancochrane (approved):
Looks great!

jeancochrane:
Oh, one minor nitpick -- can we change the title of this PR to something more descriptive, like "Refactor pre-commit to use composite action" @wagnerlmichael?

PR #408, opened 2024-04-23
Add a test to dweldat to make sure yrblt matches the class code

Damonamajor:
This produces six tests where the following [table](https://github.com/ccao-data/wiki/blob/master/Data/Class-Definitions.pdf) identifies a date range.
The results of the tests are as follows:
iasworld_dweldat_yrblt_205_more_than_62_years_old - 36 failures
iasworld_dweldat_yrblt_206_more_than_62_years_old - 31 failures
iasworld_dweldat_yrblt_207_less_than_62_years_old - pass
iasworld_dweldat_yrblt_208_less_than_62_years_old - 1 failure
iasworld_dweldat_yrblt_278_less_than_62_years_old - 1 failure
iasworld_dweldat_yrblt_295_less_than_62_years_old - pass

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.dweldat.yml
+++ b/dbt/models/iasworld/schema/iasworld.dweldat.yml
@@ -1512,7 +1512,7 @@ sources:
               - accepted_range:
                   name: iasworld_dweldat_yrblt_between_1850_and_now
                   min_value: 1850
-                  max_value: cast(year(current_date) as integer)
+                  max_value: cast(year(current_date) as decimal)
                   additional_select_columns: *select-columns
                   config: *unique-conditions
                   meta:
--- a/dbt/models/iasworld/schema/iasworld.dweldat.yml
+++ b/dbt/models/iasworld/schema/iasworld.dweldat.yml
@@ -1517,6 +1517,90 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: yrblt should be between 1850 and now
+                  name: iasworld_dweldat_yrblt_205_more_than_62_years_old
+                  min_value: 1850
+                  max_value: cast(year(current_date) as decimal) - 62
+                  additional_select_columns: *select-columns
+                  config: 
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '205'
+                  meta:
+                    description: 205 should be > 62 years old
+                  name: iasworld_dweldat_yrblt_206_more_than_62_years_old
+                  min_value: 1850
+                  max_value: cast(year(current_date) as decimal) - 62
+                  additional_select_columns: *select-columns
+                  config: 
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '206'
+                  meta:
+                    description: 206 should be > 62 years old
+                  name: iasworld_dweldat_yrblt_207_less_than_62_years_old
+                  min_value: cast(year(current_date) as decimal) - 62
+                  max_value: cast(year(current_date) as decimal)
+                  additional_select_columns: *select-columns
+                  config: 
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '207'
+                  meta:
+                    description: 207 should be <= 62 years old
+                  name: iasworld_dweldat_yrblt_208_less_than_62_years_old
+                  min_value: cast(year(current_date) as decimal) - 62
+                  max_value: cast(year(current_date) as decimal)
+                  additional_select_columns: *select-columns
+                  config: 
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '208'
+                  meta:
+                    description: 208 should be <= 62 years old
+                  name: iasworld_dweldat_yrblt_210_more_than_62_years_old
+                  min_value: 1850
+                  max_value: cast(year(current_date) as decimal) - 62
+                  additional_select_columns: *select-columns
+                  config: 
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '210'
+                  meta:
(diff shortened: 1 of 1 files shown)


dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1517,6 +1517,97 @@ sources:
...
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '205'
+                  meta:
+                    description: 205s should be > 62 years old
+              - accepted_range:
+                  name: iasworld_dweldat_yrblt_206_more_than_62_years_old
+                  min_value: 1850
```
Damonamajor, line 1535: We have this in a previous test. Do we want to switch it to 0, so that they function separately?

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1517,6 +1517,97 @@ sources:
...
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '206'
+                  meta:
+                    description: 206s should be > 62 years old
+              - accepted_range:
+                  name: iasworld_dweldat_yrblt_207_less_than_62_years_old
+                  min_value: cast(year(current_date) as decimal) - 62
```
Damonamajor, line 1548: Both min and max are coded as 62. By the class definitions, it would seem as though the minimum should be 62 (up to 62 years old) and when it is max, it should be 63 (over 62 years old). I don't know how accurately these transitions occur, and trying to accurately code these may cause unexpected spikes.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1517,6 +1517,97 @@ sources:
...
+                  meta:
+                    description: 207s should be <= 62 years old
+              - accepted_range:
+                  name: iasworld_dweldat_yrblt_208_less_than_62_years_old
+                  min_value: cast(year(current_date) as decimal) - 62
+                  max_value: cast(year(current_date) as decimal)
+                  additional_select_columns: *select-columns
+                  config: 
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
```
Damonamajor, line 1566: Is there a way to do *select_conditions AND class = '208.
I took a look through the examples you mentioned, and it was always written out completely, but just wanted to ping this question.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1517,6 +1517,97 @@ sources:
                   config: *unique-conditions
                   meta:
                     description: yrblt should be between 1850 and now
+              - accepted_range:
+                  name: iasworld_dweldat_yrblt_205_more_than_62_years_old
+                  min_value: 1850
+                  max_value: cast(year(current_date) as decimal) - 62
```
Damonamajor, line 1523: I used as decimal because that's how the >1850 accepted range is coded, but it doesn't seem like the most intuitive for dates.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1517,6 +1517,97 @@ sources:
...
+                  additional_select_columns: *select-columns
+                  config: 
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '205'
+                  meta:
+                    description: 205s should be > 62 years old
```
jeancochrane, line 1532: **[Nitpick, non-blocking]** Let's bring these descriptions in line with our typical description style, where fieldnames are used to indicate the columns in the workbook that explain the failure:

```suggestion
                    description: yrblt should be > 62 years old when class is 205
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: **[Nitpick]** Same as above, I think we should get rid `min_year` since 206s don't have a maximum age:

```suggestion
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: **[Suggestion, non-blocking]** This is a good question, I'm not sure which of the two ranges should include dwellings that are exactly 62 years old. I've escalated this question, but for now I would interpret the class definition language to indicate that "up to 62 years of age" includes 62-year-old dwellings while "over 62 years of age" excludes them.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1517,6 +1517,97 @@ sources:
...
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class = '206'
+                  meta:
+                    description: 206s should be > 62 years old
+              - accepted_range:
+                  name: iasworld_dweldat_yrblt_207_less_than_62_years_old
+                  min_value: cast(year(current_date) as decimal) - 62
+                  max_value: cast(year(current_date) as decimal)
```
jeancochrane, line 1549: **[Suggestion]** In a similar vein, our test for [overall age ranges of dwellings](https://github.com/ccao-data/data-architecture/blob/66442f5a33153e33de914421967181dc53839c54/dbt/models/iasworld/schema/iasworld.dweldat.yml#L1561-L1568) already checks to make sure that `yrblt` is less than the current year, so we can remove this range to avoid duplicate failures:

```suggestion
```

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Good question! Sadly there is not a way to extend YAML aliases like that, so we have to fall back to duplicating all of the logic in cases where we want to extend it.

jeancochrane (changes requested):
Looking great! Just a few small tweaks and this should be good to go.

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Hm, yeah -- I would go with an integer here if it produces the same behavior, even though the other test does it differently.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1546,7 +1546,6 @@ sources:
               - accepted_range:
                   name: iasworld_dweldat_yrblt_207_less_than_62_years_old
                   min_value: cast(year(current_date) as decimal) - 62
-                  max_value: cast(year(current_date) as decimal)
```
Damonamajor: @jeancochrane 
You didn't explicitly state to drop the max year, so I just wanted to flag this, as I did it in the updated push.

jeancochrane (approved):
Nice work!

PR #409, opened 2024-04-24
Use jinja templating to reduce duplicated operations in `reporting.res_report_summary`

wrridgeway:
`reporting.res_report_summary` performs a lot of repetitive operations across a number of CTEs. Creating a jinja template for these operations will save us code and better ensure accuracy.

### DIFF
--- a/aws-athena/ctas/reporting-res_report_summary.sql
+++ b/aws-athena/ctas/reporting-res_report_summary.sql
@@ -190,28 +190,28 @@ class_modes AS (
 -- By township, assessment_stage, and property group
 values_town_groups AS (
     {{ summarize_res_report_summary(
-        True, True, 'assessment_stage, triad, township_code, year, property_group'
+        'Town', True, 'assessment_stage, triad, township_code, year, property_group'
         ) }}
 ),
 
 -- By township and assessment stage
 values_town_no_groups AS (
     {{ summarize_res_report_summary(
-        True, False, 'assessment_stage, triad, township_code, year'
+        'Town', False, 'assessment_stage, triad, township_code, year'
         ) }}
 ),
 
 -- By neighborhood, assessment_stage, and property group
 values_nbhd_groups AS (
     {{ summarize_res_report_summary(
-        False, True, 'assessment_stage, triad, townnbhd, year, property_group'
+        'TownNBHD', True, 'assessment_stage, triad, townnbhd, year, property_group'
         ) }}
 ),
 
 -- By neighborhood and assessment stage
 values_nbhd_no_groups AS (
     {{ summarize_res_report_summary(
-        False, False, 'assessment_stage, triad, townnbhd, year'
+        'TownNBHD', False, 'assessment_stage, triad, townnbhd, year'
         ) }}
 ),
 
--- a/aws-athena/ctas/reporting-res_report_summary.sql
+++ b/aws-athena/ctas/reporting-res_report_summary.sql
@@ -188,32 +188,18 @@ class_modes AS (
 
 -- Here we aggregate stats on AV and characteristics for each reporting group
 -- By township, assessment_stage, and property group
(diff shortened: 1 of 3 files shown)


wrridgeway:
@dfsnow or @jeancochrane is this the best way to implement this? As in: should I have these macros stored in one script, with a name that pertains to the only script they should be used in?

aws-athena/ctas/reporting-res_report_summary.sql
```diff
@@ -169,176 +169,63 @@ class_modes AS (
...
         ) AS group_townnbhd_mode,
         FIRST_VALUE(class) OVER (
             PARTITION BY assessment_stage, township_code, year
-            ORDER BY group_townnbhd_count DESC
+            ORDER BY group_townnbhd_count DESC, class DESC
         ) AS town_mode,
         FIRST_VALUE(class) OVER (
             PARTITION BY assessment_stage, townnbhd, year
-            ORDER BY group_townnbhd_count DESC
+            ORDER BY group_townnbhd_count DESC, class DESC
```
wrridgeway, line 184: Without this these columns were not deterministic before 😬

jeancochrane (commented):
This looks like the right approach to me! Feel free to re-request my review once you're done with testing and I can take a more detailed look.

wrridgeway:
## Testing

This query tests if there are any groupings that don't join between the old and new table, as well as whether there are any discrepancies in the number of pins per group or the sale price stats.

```
select
    *
from "reporting".res_report_summary as old
full outer join "z_ci_401-use-jinja-templating-to-reduce-duplicated-operations-in-reportingres-report-summary_reporting".res_report_summary as new
    on old.geography_id = new.geography_id
    and old.property_group = new.property_group
    and old.year = new.year
    and old.assessment_stage = new.assessment_stage
where old.pin_n != new.pin_n
    or old.pin_n is null
    or new.pin_n is null
    or old.sale_min != new.sale_min
```

Nothing is returned, which means we're good here.

dbt/macros/res_report_summary.sql
```diff
@@ -0,0 +1,53 @@
+-- Macros that summarize FMVs and sales by different groupings specifically for
+-- reporting.res_report_summary.
+{% macro res_report_summarize_values(vals, geo_type, prop_group) %}
```
jeancochrane, line 3: **[Nitpick, non-blocking]** I find `vals` and `sales` to be a bit confusing as parameter names, and I don't think I would have immediately known what they were for without reading the source code of the `res_report_summarize_values` macro. A clearer parameter name might be something like `from`, to indicate that the argument will be used as the source table for the summary query, but this is more of a personal preference so I don't feel strongly about it.

dbt/macros/res_report_summary.sql
```diff
@@ -0,0 +1,53 @@
...
+    from {{ vals }}
+    group by
+        assessment_stage,
+        triad,
+        {% if geo_type == "Town" %} township_code,
+        {% elif geo_type == "TownNBHD" %} townnbhd,
+        {% endif %}
+        {% if prop_group %} year, property_group
+        {% elif not prop_group %} year
+        {% endif %}
```
jeancochrane, line 29: **[Nitpick, non-blocking]** A slight simplification, if it appeals to you:

```suggestion
        year
        {%- if prop_group %}, property_group{%endif %}
```

jeancochrane (approved):
Nice work here!

dbt/macros/res_report_summary.sql
wrridgeway: I like it.

dbt/macros/res_report_summary.sql
wrridgeway: I had written this oddly in order to try to make it explicit that that expect values should only be True or False, but i don't think it adds much.

PR #413, opened 2024-04-25
Reporting views for municipalities

wrridgeway:
> In reporting, we currently have vw_assessment_roll and vw_top_5 for townships. I'd like to create two similar separate views for municipalities.
> ### Use case
> 
> - This is frequently requested data for Assessor Leadership; let's institutionalize it.
> 
> - I also want to have an auto-updating version of a [municipal dashboard](https://www.cookcountyassessor.com/node/1887075).
> 
> 
> ### Complication: Assessment cycles
> 
> the assessment system works on a township cycle, not a municipal cycle, so some municipalities might have some PINs in one assessment stage and other PINs in a different stage.
> ### Possible solution
> 
> I _think_ the best way around this is to add an additional column on the assessment_roll version indicating, for each municipality, for the year & stage, the percentage of its PINs that have passed through that year & stage.
> 
> The ideal behavior of this "percent complete" column would be if the percent indicator had "memory." For example:
> 
> - if a multi-township muni like Chicago had fully passed through the 2023 certified stage but only 50% of its PINs had reached the 2023 bor stage, then the percentage indicator would say 100% when year = 2023 and stage = assessor certified, and 50% when year = 2023 and stage = bor.
> 
> - and, because in this scenario 2022 is complete, then for all 3 stages, the percent complete would be 100%.

### DIFF
--- a/aws-athena/views/reporting-vw_assessment_roll_muni.sql
+++ b/aws-athena/views/reporting-vw_assessment_roll_muni.sql
@@ -22,8 +22,9 @@ SELECT
     munis.major_class AS class,
     COUNT(*) AS n,
     pin_counts.total_n,
-    COUNT(*) / pin_counts.total_n AS stage_portion,
-    AVG(CAST(munis.reassessment_year AS INT)) AS reassessment_portion,
+    CAST(COUNT(*) AS DOUBLE)
+    / CAST(pin_counts.total_n AS DOUBLE) AS stage_portion,
+    AVG(CAST(munis.reassessment_year AS DOUBLE)) AS reassessment_portion,
     SUM(vpvl.bldg) AS bldg_sum,
     CAST(APPROX_PERCENTILE(vpvl.bldg, 0.5) AS INT) AS bldg_median,
     SUM(vpvl.land) AS land_sum,
--- a/aws-athena/views/reporting-vw_assessment_roll_muni.sql
+++ b/aws-athena/views/reporting-vw_assessment_roll_muni.sql
@@ -34,7 +34,8 @@ Starting in 2020 a small number of PINs are present in iasworld.asmt_all for
 one or two but not all three stages of assessment when we would expect all three
 stages to be present for said PINs. This is also a data error, but is NOT
 addressed in this view and leads to a few instances where stage_portion ends up
-being less than 1 when it should equal 1. */
+being less than 1 when it should equal 1. 16-07-219-029-1032 missing a mailed
+value but having CCAO and BOR certified values in 2021 is an example. */
 trimmed_town_class AS (
     SELECT vptc.*
     FROM {{ ref('reporting.vw_pin_township_class') }} AS vptc
--- a/aws-athena/views/reporting-vw_assessment_roll_muni.sql
+++ b/aws-athena/views/reporting-vw_assessment_roll_muni.sql
@@ -34,8 +34,7 @@ Starting in 2020 a small number of PINs are present in iasworld.asmt_all for
 one or two but not all three stages of assessment when we would expect all three
 stages to be present for said PINs. This is also a data error, but is NOT
 addressed in this view and leads to a few instances where stage_portion ends up
-being less than 1 when it should equal 1.
-*/
+being less than 1 when it should equal 1. */
 trimmed_town_class AS (
     SELECT vptc.*
     FROM {{ ref('reporting.vw_pin_township_class') }} AS vptc
--- a/aws-athena/views/reporting-vw_assessment_roll_muni.sql
+++ b/aws-athena/views/reporting-vw_assessment_roll_muni.sql
@@ -20,7 +20,7 @@ LEFT JOIN {{ ref('reporting.vw_pin_township_class') }} AS townships
     ON values_by_year.pin = townships.pin
     AND values_by_year.year = townships.year
 LEFT JOIN {{ ref('location.tax') }} AS tax
-    ON SUSBTR(values_by_year.pin, 1, 10) = tax.pin10
+    ON SUBSTR(values_by_year.pin, 1, 10) = tax.pin10
     AND values_by_year.year = tax.year
 WHERE townships.township_name IS NOT NULL
 GROUP BY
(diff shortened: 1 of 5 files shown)


aws-athena/views/reporting-vw_assessment_roll_muni.sql
```diff
@@ -0,0 +1,114 @@
...
+derive the numerator and denominator from for stage_portion are the same. These
+differences are inherent in the tables that feed these views (iasworld.pardat
+and iasworld.asmt_all, respectively) and are data errors - not emblematic of
+what portion of a municipality has actually progressed through an assessment
+stage.
+
+It does NOT remove PINs from the most recent year of
+reporting.vw_pin_township_class since we expect these differences based on how
+iasworld.asmt_all is populated through the year as the assessment cycle
+progresses.
```
dfsnow, line 31: **question:** Won't this not quite work when we're just into a new year i.e. if we roll over into 2025 with a new `taxyr` in `pardat` (which feeds `vw_pin_township_class`), there will still be stage records missing from the Board in `vw_pin_value_long`?

aws-athena/views/reporting-vw_assessment_roll_muni.sql
```diff
@@ -0,0 +1,114 @@
...
+
+Starting in 2020 a small number of PINs are present in iasworld.asmt_all for
+one or two but not all three stages of assessment when we would expect all three
+stages to be present for said PINs. This is also a data error, but is NOT
+addressed in this view and leads to a few instances where stage_portion ends up
+being less than 1 when it should equal 1. 16-07-219-029-1032 missing a mailed
+value but having CCAO and BOR certified values in 2021 is an example. */
```
dfsnow, line 38: 🙃 Gah. Probably not worth making a view-level fix for this if it's only a few PINs right? Perhaps we could add some dbt tests? "Each PIN in the prior tax year has three stage/procname values"

aws-athena/views/reporting-vw_assessment_roll_muni.sql
```diff
@@ -0,0 +1,114 @@
...
+pin_counts AS (
+    SELECT
+        vptc.municipality_name,
+        vptc.major_class,
+        vptc.year,
+        stages.stage,
+        COUNT(*) AS total_n
+    FROM trimmed_town_class AS vptc
+    CROSS JOIN stages
+    WHERE vptc.municipality_name IS NOT NULL
```
dfsnow, line 70: **thought:** We should maybe add a group for unincorporated areas.

aws-athena/views/reporting-vw_assessment_roll_muni.sql
```diff
@@ -0,0 +1,114 @@
...
+-- each municipality that has progressed through an assessment stage by class.
+SELECT
+    pin_counts.year,
+    pin_counts.stage,
+    pin_counts.municipality_name,
+    munis.major_class AS class,
+    SUM(CAST(vpvl.pin IS NOT NULL AS INT)) AS n,
+    pin_counts.total_n,
+    SUM(CAST(vpvl.pin IS NOT NULL AS DOUBLE))
+    / CAST(pin_counts.total_n AS DOUBLE) AS stage_portion,
```
dfsnow, line 88: suggestion: I find these column names slightly ambiguous. Maybe instead:

- n --> num_pin_w_value
- total_n --> num_pin_total_in_group
- stage_portion --> pct_pin_w_value_in_group

dfsnow, line 88: **praise:** These are extremely cool and useful columns!

dbt/models/reporting/docs.md
```diff
@@ -30,6 +30,15 @@ assessment stage, and year. Feeds public reporting assets.
 **Primary Key**: `year`, `township_name`, `class`, `stage`
 {% enddocs %}
 
+# vw_assessment_muni
```
dfsnow, line 33:
```suggestion
# vw_assessment_roll_muni
```

dfsnow (commented):
This looks great @wrridgeway! Will definitely be useful and a good baseline for building out a similar view for all geographies. I think I followed all the logic in here.

This seems sufficiently important that it's probably worthwhile to get another pair of eyes on it. @jeancochrane can you also take a look?

aws-athena/views/reporting-vw_assessment_roll_muni.sql
wrridgeway: I think we'll need to discuss this in person. This part is complicated, but we're already in a situation where we don't have a complete set of PINs in `vw_pin_value_long` for 2024 and the view is working as intended.

aws-athena/views/reporting-vw_assessment_roll_muni.sql
wrridgeway: I've reached out to Mirella. dbt test is a good idea. If we were to try to address it in the view, it'd be very tricky dynamically identifying which pins are allowed to have less than three rows and which aren't.

aws-athena/views/reporting-vw_assessment_roll_muni.sql
wrridgeway: I considered this, but leaned towards no since unincorporated isn't a cohesive political unit. Definitely still an open question.

aws-athena/views/reporting-vw_assessment_roll_muni.sql
wrridgeway: I'm cool with this.

jeancochrane (approved):
Very nice work!

aws-athena/views/reporting-vw_assessment_roll_muni.sql
ccao-jardine: Good suggestion to have an unincorporated group. Let's do it!

First, helps to have a ready answer for questions like "How many parcels are in unincorporated Cook." Second, it helps with QC and transparency: adding up, for example, PIN and AV totals of municipal and unincorporated should roughly equal the Countywide totals.

Issue #376, opened 2024-04-11
Add dbt testing framework for seeds


dfsnow:
> What do we think about adding some tests to this model? In particular, I'm thinking about how the join conditions in this PR depend on the assumption that the combination of `township_code_coverage` values for final models in a given year cover all triads in that year's township. I'm not 100% confident that our current strategy for testing changed and added models would work for seeds -- we know that models will get tested if the queries that comprise them change, but I'm not sure if we can expect the same behavior for seeds whose data change -- but if we could get it working I think it would be a helpful safeguard for future maintainers who need to modify the `model.final_model.csv` seed file.
>
> Just an idea for the future, we can do this investigation in a follow-up issue!
>
_Originally posted by @jeancochrane in https://github.com/ccao-data/data-architecture/pull/365#discussion_r1559679508_

Damonamajor:
1. Define a simple test on a seed.
2. Modify the seed data and see if the dbt test command that we run on CI, will run the tests on the modified seed.
3. else
              echo "Running tests on modified/new resources only"
              dbt test -s state:modified state:new --exclude "tag:test_qc*" --defer --state "target"
  This is hopefully going to run for modified seeds, but we're not sure.

PR #421, opened 2024-04-30
Add initial LoA seed

### LINKED ISSUE #420: Add LoA dbt seed
We should add a dbt seed containing the historical level of assessment of every CCAO class. This should be a "long" format, with a row per year and class (for easy joining). This would also let us close other issues like https://github.com/ccao-data/ptaxsim/issues/31 and https://github.com/ccao-data/ptaxsim/issues/22.

dfsnow:
This PR adds a Level of Assessment (LoA) dbt seed. We'll need to update it yearly with new LoAs and classes, pursuant to any legislation or internal changes. Closes #420.


dfsnow:
FYI @ccao-jardine

dbt/seeds/ccao/docs.md
```diff
@@ -8,6 +8,19 @@ regressions and reporting classes.
...
 
+# loa
+
+{% docs seed_loa %}
+Table containing the Level of Assessment (LoA) for each minor property class
+for each year. LoAs change over time due to legislation, changes to Assessor
+class codes, etc.
+
+Do *not* use this table as a definitive list of existent class codes per year.
+Please use `ccao.class_dict` instead.
```
jeancochrane, line 19: **[Question, non-blocking]** `ccao.class_dict` doesn't yet have historical class codes, correct?

dbt/seeds/ccao/schema.yml
```diff
@@ -34,6 +34,24 @@ seeds:
...
+    config:
+      column_types:
+        year: string
+        class_code: string
+
+    columns:
+      - name: year
+        description: '{{ doc("shared_column_year") }}'
+      - name: class_code
+        description: '{{ doc("shared_column_class") }}'
```
jeancochrane, line 48: **[Thought, non-blocking]** I wonder if we should file an issue to test this view to make sure that the class codes are all in `ccao.class_dict`?

dbt/seeds/ccao/ccao.loa.csv
```diff
@@ -0,0 +1,4945 @@
+year,class_code,loa
+2000,0,0
+2000,100,0.22
```
jeancochrane, line 3: **[Question, non-blocking]** It seems like `ccao.class_dict` records LoA as the percentage value rather than decimal value (i.e. `22` instead of `0.22`). Is it a problem that they're different, or are we intending to move toward decimal as our standard?

jeancochrane (approved):
I can't confirm that the LoAs are correct, but the code itself looks great!

dbt/seeds/ccao/docs.md
dfsnow: Not yet, and it may not for awhile, but IMO it's fine for the moment.

dbt/seeds/ccao/schema.yml
dfsnow: There are a _bunch_ of classes in the LoA table that unfortunately aren't in the class dict. Most of them don't have any instances. I'll defer fixing/adding tests until #376.

dbt/seeds/ccao/ccao.loa.csv
dfsnow: Good point! I dropped the LoA from `class_dict` in 56ca182

PR #422, opened 2024-05-01
Refactor ratio stats Glue job to dbt python model

dfsnow:
This PR migrates the `ratio_stats` Glue job to a dbt Python model that uses Athena's Spark integration as the Python runtime. There are some tradeoffs here:

## Pros

- Vastly simplifies the Glue job code (no need for polling, context setup, etc)
- Integrates the `ratio_stats` table with the DAG, meaning it will get rebuild on upstream changes AND it works with dbt profiles (so changes can first be built in CI buckets)
- Very easy to add further Python models now

## Cons

- Needs external packages. We'll have to setup zip files of packages in order to use third-party stuff.
- Cannot source from views. Athena Spark pulls from the _files_ rather than SQL, so it doesn't work with views at all.
- Very new integration. Support was only added in the most recent version of `dbt-athena`.
- Hard to debug
- Ugly linting

IMO the pros are worth it here, but it's probably worth it to merge this and keep it up as a test for a month or so before integrating it more deeply.


.gitignore
```diff
@@ -16,3 +16,4 @@ logs/
 # Ignore most CSVs, except those that are used as dbt seeds
 *.csv
 !dbt/seeds/**/*.csv
+venv/
```
dfsnow, line 19: I've been using a root level venv to prototype python scripts and figured it would be good to ignore it.

.github/workflows/build_daily_dbt_models.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     # Every day at 8am (2pm UTC-5)
-    - cron: '0 13 * * *'
+    - cron: '0 12 * * *'
```
dfsnow, line 7: The Python model takes about an hour to build. We could likely speed it up significantly by refactoring it to use spark explicitly. For now, bumping back the daily build by an hour ensure it's ready to go each morning.

aws-athena/ctas/reporting-ratio_stats_input.sql
dfsnow, line 1: So [Athena Spark](https://docs.aws.amazon.com/athena/latest/ug/notebooks-spark.html#notebooks-spark-considerations-and-limitations) is _not_ able to pull from views. I don't think this is a huge hindrance for our use case, since most python models would be upstream from views.

However, in this case, it requires making the ratio stats input view into a table. The table will be rebuilt every morning as part of the DAG.

dbt/dbt_project.yml
```diff
@@ -1,4 +1,4 @@
-name: 'athena'
+name: 'ccao_data_athena'
```
dfsnow, line 1: Per the bug [that Jean filed](https://github.com/dbt-athena/dbt-athena/issues/607), `dbt-athena` can't have the _project_ named Athena or the name spacing gets messed up. Annoyingly the project name can't contain dashes, so prefixing it with `ccao_data_` for now.

dbt/models/reporting/reporting.ratio_stats.py
dfsnow, line 1: This is the actual refactored Glue job. Athena/Spark abstracts away all the polling code, client setup, and location setup.

dbt/models/reporting/reporting.ratio_stats.py
```diff
@@ -1,125 +1,17 @@
...
-
-# Athena query ----
-SQL_QUERY = "SELECT * FROM reporting.vw_ratio_stats;"
-
-
-# Run run_query_get_result to get file like object ----
-r_file_object = run_query_get_result(
-    athena_client, s3_bucket, SQL_QUERY, athena_db, s3_output, s3_prefix
+# type: ignore
+# pylint: skip-file
```
dfsnow, line 2: Annoyingly, our linters and pre-commit checks absolutely hate this setup. There are a two problems:

1. `sc` is undefined in the script (it's added when you submit the job).
2. The imports come necessarily _after_ `addPyFile()`, which loads the packages from S3.

dbt/models/reporting/reporting.ratio_stats.py
```diff
@@ -1,125 +1,17 @@
...
+sc.addPyFile(  # noqa: F821
+    "s3://ccao-dbt-athena-dev-us-east-1/packages/spark-packages.zip"
```
dfsnow, line 4: This is the path to the zip file of Python packages. I figure for our simple package manager, we can create a single zip file of all packages for all jobs, with each of the environments (`dev`, `ci`, `prod`) getting its own file.

We would need to then find some way to dynamically set this path based on the active environment, rather than hardcoding it.

dbt/models/reporting/reporting.ratio_stats.py
```diff
@@ -249,7 +141,13 @@ def ccao_prb(fmv, sale_price):
...
 
 def report_summarise(df, geography_id, geography_type):
-    # Aggregates data and calculates summary statistics for given groupings
+    """
+    Aggregates data and calculates summary statistics for given groupings
+    """
+
+    # Convert the Spark input dataframes to Pandas for compatibility
+    # with assesspy functions
+    df = df.toPandas()
```
dfsnow, line 150: This line is absolutely _critical_ to making this work. It converts the spark table returned by `dbt-athena` (in this case all of `reporting.ratio_stats_input`) to a pandas data frame. IMO, we should refactor this to use pyspark since it will probably be much faster.

dbt/models/reporting/reporting.ratio_stats.py
```diff
@@ -276,7 +174,6 @@ def report_summarise(df, geography_id, geography_type):
                 "mki": ccao_mki(fmv=x["fmv"], sale_price=x["sale_price"]),
                 "prd": ccao_prd(fmv=x["fmv"], sale_price=x["sale_price"]),
                 "prb": ccao_prb(fmv=x["fmv"], sale_price=x["sale_price"]),
-                "detect_chasing": detect_chasing(ratio=x["ratio"]),
```
dfsnow, line 279: I dropped the `detect_chasing()` stuff because:

a) It's not actually super effective at detecting sales chasing.
b) It was causing extremely difficult to debug errors in the spark job.

We can work to add it back later if it's truly needed. FYI @ccao-jardine.

dbt/models/reporting/schema.yml
```diff
@@ -1,10 +1,16 @@
...
+      - daily
       - load_auto
     tables:
       - name: ratio_stats
         description: '{{ doc("table_ratio_stats") }}'
+        config:
+          packages:
+            - "assesspy==1.1.*"
+            - "numpy==1.26.*"
+            - "pandas==2.*"
```
dfsnow, line 13: This is the "dbt way" to define package dependencies. I'm not yet clear on exactly how we'd use it in the package tooling. My first pass idea:

1. Scrape this out of the dbt metadata
2. Construct a requirements.txt file
3. Install the packages to a virtual env
4. Zip up the installed packages then upload

dbt/profiles.yml
```diff
@@ -11,6 +11,7 @@ athena:
       schema: z_static_unused-dbt-stub-database
       # "database" here corresponds to a Glue data catalog
       database: awsdatacatalog
+      spark_work_group: Spark_primary
```
dfsnow, line 14: This is the only setting needed to actually enable the spark integration. I'll make a separate PR to capture the new workgroup, permissions, and IAM roles.

dbt/requirements.txt
```diff
@@ -1 +1 @@
-dbt-athena-community==1.7.1
+dbt-athena-community==1.7.2
```
dfsnow, line 1: `1.7.2` is the version of `dbt-athena` that has the Python model integration.

dbt/models/reporting/reporting.ratio_stats.py
dfsnow, line 1: @wrridgeway It looks like these functions are still cutting out the top and bottom 5% of ratios. I can't quite remember where we landed on this, but I thought we were shifting to kicking out sales using the new flags.

dbt/models/reporting/docs.md
```diff
@@ -10,6 +10,15 @@ reporting Tableau dashboards.
 `property_group`
 {% enddocs %}
 
+# ratio_stats_input
+
+{% docs table_ratio_stats_input %}
+View to feed the Python dbt job that creates the `reporting.ratio_stats` table.
```
jeancochrane, line 16: **[Nitpick, non-blocking]** Technically not a view anymore, right?

```suggestion
Table to feed the Python dbt job that creates the `reporting.ratio_stats` table.
```

dbt/models/reporting/schema.yml
jeancochrane: **[Praise]** I think that's a good place to start! It's nice that dbt already has a syntax for defining requirements.

dbt/models/reporting/reporting.ratio_stats.py
jeancochrane: **[Question, non-blocking]** Unfortunately this is pretty common when 1) writing code for environments with implicit context and B) writing setup code that needs to run prior to import. I don't think there's an obviously better way to do it, although we might consider switching to a different linter that supports block-level ignores (unlike flake8).

That being said, can you provide some context on why pylint is turned off? Is that just for your editor integration?

.github/workflows/build_daily_dbt_models.yaml
jeancochrane: **[Question, non-blocking]** It seems like [the current run](https://github.com/ccao-data/data-architecture/actions/runs/8925881909/job/24515607821?pr=422) has taken 2.5 hours (so far) to build models, does that interfere at all with this schedule?

jeancochrane (approved):
Really nice! I agree that the pros seem to outweigh the cons, and that we should wait a little bit to gather experience with this approach before refactoring more code to use Python models.

I'm a bit concerned that the build is taking so long to complete and also seems to be blocking logs from loading (although that could just be something with my network). Once it completes and we're feeling good about the output, I'm down for us to merge.

dbt/models/reporting/reporting.ratio_stats.py
dfsnow: Ah, it's because it's also mad about `sc` not being defined. Pylint in this case is kind of extraneous since flake8 is running already in precommit.

dfsnow:
@jeancochrane Hit another small roadblock on this after a bunch of permissions debugging. Seems like spark can't write tables with a dash in the name. Not an issue in prod, but doesn't work with our kebab-case name spacing schema. Do you think it's worth changing the macro to force everything to snake_case?

jeancochrane:
> @jeancochrane Hit another small roadblock on this after a bunch of permissions debugging. Seems like spark can't write tables with a dash in the name. Not an issue in prod, but doesn't work with our kebab-case name spacing schema. Do you think it's worth changing the macro to force everything to snake_case?

@dfsnow It's a little bit annoying to have to change our naming scheme, but there's not actually a huge difference between the hyphenated vs. underscored versions so I don't mind. Were you able to find documentation confirming this limitation with Athena PySpark? I couldn't find anything so I'm wondering if the root cause is actually an error in the underlying plugin implementation (e.g. not using backticks to escape special characters).

dfsnow:
> > @jeancochrane Hit another small roadblock on this after a bunch of permissions debugging. Seems like spark can't write tables with a dash in the name. Not an issue in prod, but doesn't work with our kebab-case name spacing schema. Do you think it's worth changing the macro to force everything to snake_case?
> 
> @dfsnow It's a little bit annoying to have to change our naming scheme, but there's not actually a huge difference between the hyphenated vs. underscored versions so I don't mind. Were you able to find documentation confirming this limitation with Athena PySpark? I couldn't find anything so I'm wondering if the root cause is actually an error in the underlying plugin implementation (e.g. not using backticks to escape special characters).

Sadly, [there are official docs](https://spark.apache.org/docs/latest/sql-ref-identifier.html#regular-identifier) supporting this. I also forked `dbt-athena` and tested various backtick configurations with the fork, none of them worked. I think basically only the `z_ci_` DBs ever even use hyphens, so it's not a major change. But yes, annoying.

dfsnow:
@jeancochrane @wrridgeway This could probably use one more look before merging. I'll post the diff between the current `reporting.ratio_stats` and the new version tomorrow.

dfsnow:
A quick diff shows that the old `ratio_stats` and new one are basically identical but for the following differences:

- Bootstrapping is not deterministic and therefore all `_ci` columns have different values
- The `mki` column has different values due to https://github.com/ccao-data/assesspy/pull/19
- `detect_chasing` is removed in the new version (replaced with `False`), due to reasons I outlined earlier. It's not a reliable measurement of chasing and it doesn't work in Spark for some reason
- The float columns have a slightly different precision between versions. The new version records _more_ digits

All that said, I think this is ready to merge. I've kept it as close as I can to the original `ratio_stats` table. FYI @ccao-jardine

aws-athena/ctas/reporting-ratio_stats_input.sql
```diff
@@ -1,5 +1,10 @@
 -- View containing ratios by pin, intended to feed the
```
wrridgeway, line 1:
```suggestion
-- Table containing ratios by pin, intended to feed the
```

wrridgeway:
Assuming testing comes back fine, looks good to me, though this all begs the question: is there a point to having the input table exist outside the python script to generate stats any longer if both processes are handled by dbt on a daily basis and the input table serves no function outside feeding the script? It probably didn't need to exist in the first place. Just a thought.

dfsnow:
> Assuming testing comes back fine, looks good to me, though this all begs the question: is there a point to having the input table exist rather than just being a pull in the python script to generate stats any longer if both processes are handled by dbt on a daily basis and the input table serves no function outside feeding the script? It probably didn't need to exist in the first place. Just a thought.

Probably not tbh, we can likely ditch the input table. However, I'm going to punt that to a follow-up issue since I want to get this merged.

dbt/models/reporting/reporting.ratio_stats.py
```diff
@@ -335,13 +230,45 @@ def report_summarise(df, geography_id, geography_type):
...
+        + "property_group: string, assessment_stage: string, "
+        + "geography_id: string, sale_year: bigint, sale_n: bigint, "
+        + "median_ratio: double, median_ratio_ci: string, cod: double, "
+        + "cod_ci: string, cod_n: bigint, prd: double, prd_ci: string, "
+        + "prd_n: bigint, prb: double, prb_ci: string, prb_n: bigint, "
+        + "mki: double, mki_n: bigint, detect_chasing: boolean, "
+        + "ratio_met: boolean, cod_met: boolean, prd_met: boolean, "
+        + "prb_met: boolean, mki_met: boolean, vertical_equity_met: boolean, "
+        + "within_20_pct: bigint, within_10_pct: bigint, within_05_pct: bigint"
+    )
```
jeancochrane, line 270: **[Suggestion, non-blocking]** Since this is essentially just a comma-and-space-separated string of key-value pairs, it might make future development more convenient to define it as a dict and then serialize it to this format with a oneliner like so:

```
schema_dict = { ... }
schema = ", ".join(f"{key}: {val}" for key, val in schema_dict.items())
```

But I leave it up to you to decide which data structure you find more convenient!

jeancochrane (approved):
Awesome! I can't really speak to the accuracy of the underlying table, but the model definition portion of the PR looks good to go.

dbt/models/reporting/reporting.ratio_stats.py
dfsnow: I don't expect that we'll persist this schema definition. It will probably only live until Nicole gets back and can confirm that Tableau is still working fine.

Issue #148, opened 2023-10-02
Add README


dfsnow:
We should add a root-level README to the repo to serve as a directory to other data documentation efforts. I would point at:

- The dbt documentation site
- The Data Department GitHub wiki
- The dbt README

Additionally, we should check the licenses of the software used in this repo, then choose an appropriate license. Our other repos almost exclusively use AGPL-3, but MIT might be appropriate here.

PR #423, opened 2024-05-02
Create issue template for new dbt tests

wagnerlmichael:
This PR adds an issue template for new DBT tests.


.github/ISSUE_TEMPLATE/new-dbt-test.md
```diff
@@ -0,0 +1,49 @@
+# New DBT Test
```
jeancochrane, line 1: **[Nitpick]** As you can see from [the official website](https://docs.getdbt.com/), the official style is actually all lowercase for dbt:

```suggestion
# New dbt test
```

.github/ISSUE_TEMPLATE/new-dbt-test.md
```diff
@@ -0,0 +1,49 @@
...
+        combination_of_columns:
+            - pin
+            - year
+    ```
+
+   - Reference code implementation [here](https://github.com/ccao-data/data-architecture/blob/66ad8159bcb3d96dcdc62b7355f8fbce64affc78/dbt/models/default/schema/default.vw_pin_universe.yml#L248-L252).
+
+## Running Tests
+
+Execute the tests against development or production environments using the following commands:
```
jeancochrane, line 38: **[Suggestion, non-blocking]** I think we should decide whether we want folks to run development or production tests; otherwise, I expect readers will think that they should run both. I would lean in the direction of recommending that they just run development tests.

.github/ISSUE_TEMPLATE/new-dbt-test.md
```diff
@@ -0,0 +1,49 @@
...
+
+Run a test against development models:
```
jeancochrane, line 40: **[Nitpick, non-blocking]** Let's make it a bit clearer that the test author needs to update the name of the test when running this command:

```suggestion
Run your test against development models. Make sure to change the name of the test that is passed to the `--select` flag below (`default_vw_pin_universe_unique_by_14_digit_pin_and_year`) to match the name(s) of the test(s) you want to run:
```

.github/ISSUE_TEMPLATE/new-dbt-test.md
```diff
@@ -0,0 +1,49 @@
+# New DBT Test
+
+## Issue Summary
+
+_Provide a brief description of the test and its motivation._
+
+## Prerequisites
+
+Before proceeding, ensure you have a basic understanding of DBT tests. The [official DBT documentation on data tests](https://docs.getdbt.com/docs/build/data-tests) is a great resource.
```
jeancochrane, line 9: **[Suggestion, non-blocking]** I think it might be worthwhile to make our expectations clearer:

```suggestion
Before proceeding, make sure you've read the [official dbt documentation on data tests](https://docs.getdbt.com/docs/build/data-tests) and [our internal dbt documentation](https://github.com/ccao-data/data-architecture/blob/master/dbt/README.md) so that you know how dbt models and tests work. In particular, you should understand:

* The difference between a generic test and a test that is defined on a model in a `schema.yml` file
* How generic tests are written and extended
* How tests defined on a model in a `schema.yml` file are written and run

If you don't feel like you have a strong understanding of all of these items, ask a senior data scientist for clarification before starting.
```

.github/ISSUE_TEMPLATE/new-dbt-test.md
```diff
@@ -0,0 +1,49 @@
...
+   - [ ] Include the new test in the `schema.yaml` file under the `dbt/models/` directory for the specific data model.
+   - Example: Here's how the test is implemented for the `default.vw_pin_universe` view:
+
+    ```yaml
+    - unique_combination_of_columns:
+        name: default_vw_pin_universe_unique_by_14_digit_pin_and_year
+        combination_of_columns:
+            - pin
+            - year
+    ```
```
jeancochrane, line 32: **[Suggestion]** There are a couple pieces of key information that are missing here:

* QC tests need either the test or the model to be tagged with a tag that starts with `test_qc_*` (usually `test_qc_iasworld` for an iasworld test) in order to run as part of our QC test suite
* QC tests require a few extra properties in order to be properly displayed in the output workbook:
    * `meta.description` (required): Short description of the test
    * `meta.category` (optional): Category for the test, if the test is not configured with a default category in the [`transform_dbt_test_results`](https://github.com/ccao-data/data-architecture/blob/master/.github/scripts/transform_dbt_test_results.py) script
    * `meta.table_name` (optional): The iasWorld table name that the test operates on, if one cannot be automatically determined from the test

This is a lot of information, and I'm not sure whether it's better documented here in the issue template or elsewhere in our dbt documentation.

.github/ISSUE_TEMPLATE/new-dbt-test.md
```diff
@@ -0,0 +1,49 @@
...
+
+   - Reference code implementation [here](https://github.com/ccao-data/data-architecture/blob/66ad8159bcb3d96dcdc62b7355f8fbce64affc78/dbt/models/default/schema/default.vw_pin_universe.yml#L248-L252).
```
jeancochrane, line 34: **[Nitpick]** I would actually prefer we link to an iasworld test, since tests defined on our view are much more likely to be unit tests and so require a different process. Any test on `iasworld.dweldat` should be fine.

.github/ISSUE_TEMPLATE/new-dbt-test.md
```diff
@@ -0,0 +1,49 @@
...
+## Checklist
+
+Our workflow focuses on creating reusable generic test templates in the `dbt/tests/generic/` directory, which are invoked in the `schema.yaml` file for a given data model. Follow these steps to add a new test:
+
+1. **Check for Existing Tests:** Ensure the test or a similar functionality does not already exist among our generic templates. Consider modifying an existing template if it closely aligns with the required functionality.
+
+2. **Create Test Template:**
+   - [ ] Add a new test template in the `dbt/tests/generic/` directory.
+   - Example: View an existing DBT test template [here](https://github.com/ccao-data/data-architecture/blob/master/dbt/tests/generic/test_unique_combination_of_columns.sql).
+   - _We use [Jinja](https://jinja.palletsprojects.com/en/3.1.x/templates/) to parameterize the generic test templates._
```
jeancochrane, line 20: **[Nitpick]** I think we need some kind of conditional branching here to indicate the different steps the reader should take based on which of the following situations is true:

* A generic test exists that meets the reader's needs
* A generic test exists that _almost_ meets the reader's needs, but must be extended to support their use case
* No generic test exists that meets the reader's needs

Currently these docs imply that the third bullet is the most important step, but I actually think the third bullet should only be a last resort, because writing a new generic requires advanced understanding of dbt.

jeancochrane (changes requested):
This is a good start! I'm worried that we've gotten ahead of ourselves by attempting to write an issue template before drafting detailed dbt test docs, though; as you can tell from my comments, there's a lot of missing context, but I don't think an issue template is the appropriate way to provide that context. My suggestion would be for us to pause this effort, but up a separate PR to beef up the testing documentation to explain how our test suites work in detail, and then come back to this issue template once we can link out to our docs for necessary context. But I'm curious what @dfsnow thinks!

.github/ISSUE_TEMPLATE/new-dbt-test.md
wagnerlmichael: Ah right, because this issue could still be invoked to add an existing generic test to the `schema.yaml` for a given data model.

dfsnow:
> This is a good start! I'm worried that we've gotten ahead of ourselves by attempting to write an issue template before drafting detailed dbt test docs, though; as you can tell from my comments, there's a lot of missing context, but I don't think an issue template is the appropriate way to provide that context. My suggestion would be for us to pause this effort, but up a separate PR to beef up the testing documentation to explain how our test suites work in detail, and then come back to this issue template once we can link out to our docs for necessary context. But I'm curious what @dfsnow thinks!

I agree with this, better to document our various testing patterns in the dbt README. Could also be built on out #148.

.github/ISSUE_TEMPLATE/new-dbt-test.md
wagnerlmichael: I added a link out to the docs. You'll have to let me know what the right balance is.

dfsnow:
@wagnerlmichael What's the deal with this PR? Can we close it out or does it still need some work?

wagnerlmichael:
I believe it still needs some work. I think I forgot to pin Jean after I made some changes. Let me take a look

jeancochrane:
We backburnered this effort and it turns out we don't add dbt tests often enough for an issue template to be worth maintaining, in my opinion. We can reopen this in the future if we decide it's worth spending time on.

PR #431, opened 2024-05-06
Refactor naming macro to use snake_case

dfsnow:
This PR alters the `kebab_slugify()` dbt macro. It removes support for `kebab-case` and make slugify return strict `snake_case`. This is to support #422, which require snake_case table names.


dbt/macros/generate_schema_name.sql
```diff
@@ -27,9 +27,9 @@
     {%- set default_schema = target.schema -%}
 
     {%- if target.name == "dev" -%}
-        {%- set schema_prefix = "z_dev_" ~ env_var_func("USER") ~ "_" -%}
+        {%- set schema_prefix = "z_dev_" ~ slugify(env_var_func("USER")) ~ "_" -%}
```
dfsnow, line 30: Usernames can actually contain hyphens, so we need to slugify this now.

dbt/macros/slugify.sql
```diff
@@ -1,13 +1,13 @@
...
-{% macro kebab_slugify(string) %}
+-- Variation of dbt_utils.slugify macro using strict snake_case
+{% macro slugify(string) %}
     {#- Lower case the string -#}
     {% set string = string | lower %}
 
-    {#- Replace spaces, slashes, and underscores with hyphens -#}
-    {% set string = modules.re.sub("[ _/]+", "-", string) %}
+    {#- Replace spaces, slashes, and hyphens with underscores -#}
+    {% set string = modules.re.sub("[ /-]+", "_", string) %}
```
dfsnow, line 7: Note that the order of this regex is now super important. The hyphen _must_ be the last character.

jeancochrane (approved):
Looks good! Note that the `cleanup-dbt-models` workflow will no longer work for open PRs once this is merged, so we'll either need to rename those resources or clean them up manually sometime between the time this PR is merged and the time those PRs are ready to be cleaned up.

dbt/macros/slugify.sql
jeancochrane: **[Suggestion, non-blocking]** Sounds tricky! Might actually be worth preserving as a comment in the code?

Issue #417, opened 2024-04-26
Add workflow to deploy Python dependencies to S3


jeancochrane:
Once #422 is merged, we need to build out a simple package management and deployment system to manage third-party packages used by Spark. My idea on this:

> This is the "dbt way" to define package dependencies. I'm not yet clear on exactly how we'd use it in the package tooling. My first pass idea:
>
> 1. Scrape this out of the dbt metadata
> 2. Construct a requirements.txt file
> 3. Install the packages to a virtual env
> 4. Zip up the installed packages then upload

_Originally posted by @dfsnow in https://github.com/ccao-data/data-architecture/pull/422#discussion_r1587832066_

PR #435, opened 2024-05-08
Add script and update `build-and-test-dbt` workflow to push dbt Python dependencies to S3

jeancochrane:
This PR adds a new script `deploy_dbt_model_dependencies.sh` that downloads dependencies for dbt Python models, bundles them into zip archives, and pushes them to S3. It also updates the `build-and-test-dbt` workflow to use the new script, including the following features:

* If a dbt cache exists, only deploy packages for models that are new or have changed
* If a caller has dispatched the workflow for a specific set of models, only deploy packages for those models
* In all cases, only deploy package bundles that are new or have changed
* Deploy a separate package bundle for each model
* Deploy package bundles to separate S3 locations for dev, CI, and prod, and isolate dev/CI environments by user or branch

Note that this PR leaves out of scope the additional step to refactor the existing `reporting.ratio_stats` model to use this new dependency management system. That step will be handled in a follow-up issue (https://github.com/ccao-data/data-architecture/issues/439).

Closes https://github.com/ccao-data/data-architecture/issues/417.

## Testing

Evidence of `build-and-test-dbt` pushing packages for changed models to S3 on `synchronize` event: [Link](https://github.com/ccao-data/data-architecture/actions/runs/9021296669/job/24788308838#step:7:102)

Evidence of `build-and-test-dbt` skipping dependencies that have not changed on `synchronize` event: [Link](https://github.com/ccao-data/data-architecture/actions/runs/9021342661/job/24788460993#step:7:53)

Evidence of `build-and-test-dbt` skipping models that have no Python dependencies on `workflow_dispatch` event: [Link](https://github.com/ccao-data/data-architecture/actions/runs/9022831999/job/24793287938#step:7:55)

Evidence of `cleanup-dbt-resources` removing packages from S3 on `closed` event: [Link](https://github.com/ccao-data/data-architecture/actions/runs/9021452274/job/24788819738#step:5:35)


jeancochrane:
Closing this to test the cleanup step.

dbt/macros/get_s3_dependency_dir.sql
```diff
@@ -0,0 +1,48 @@
+-- Get the S3 location that is used to store Python model dependencies
+-- for the current target
```
jeancochrane, line 2: Ideally we would just add a new attribute `s3_dependency_dir` to our profiles defined in `profiles.yml`, and then we could have full control over the location where dependencies get stored for each target. That doesn't appear to be possible without extending the `dbt-athena` plugin itself to support a new profile variable, so this macro represents a workaround that allows us to compute a dependency dir based on the `s3_data_dir` variables set in our profiles. 

I think we'll probably need a Python version of this macro as well, but I'm leaving that for myself to figure out later when I tackle https://github.com/ccao-data/data-architecture/issues/439.

dbt/macros/get_s3_dependency_dir.sql
```diff
@@ -0,0 +1,48 @@
...
+                return(
+                    raise_error_func(
+                        "HEAD_REF env var must be set when target is 'ci'"
+                    )
+                )
+            }}
+        {% endif %}
+        {% set dir_suffix = "/" ~ head_ref %}
+    {% endif %}
+    {% set s3_dependency_dir = target.s3_data_dir ~ "packages" ~ dir_suffix %}
```
jeancochrane, line 40: One downside of this approach is that it stores packages _inside_ the data dir, when ideally they would be stored alongside the data dir. For clarity, here are the three paths for the three different targets we support:

* `dev`
  * data dir:  `s3://ccao-dbt-athena-dev-us-east-1/data/`
  * dependency dir: `s3://ccao-dbt-athena-dev-us-east-1/data/packages/<username>/`
* `ci`
  * data dir: `s3://ccao-dbt-athena-ci-us-east-1/data/`
  * dependency dir: `s3://ccao-dbt-athena-ci-us-east-1/data/packages/<branch_slug>/`
* `prod`
  * data dir: `s3://ccao-athena-ctas-us-east-1/`
  * dependency dir: `s3://ccao-athena-ctas-us-east-1/packages/`

I think `prod` is probably the most reasonable of the three. They all feel acceptable right now, but not perfect. What do you think?

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -57,6 +57,40 @@ jobs:
...
+              model_names=$(\
+                dbt list \
+                  --quiet -t "$TARGET" -s state:modified state:new \
+                  --state "$STATE_DIR" --resource-types model \
+                  --output json --output-keys name | \
+                jq -r '.name' | \
+                tr '\n' ' '
+              )
+              if [ -z "$model_names" ]; then
+                echo "No modified/new resources found, skipping dependency deploy"
```
jeancochrane, line 77: See [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/9024036185/job/24797106700?pr=435#step:7:53) for an example of this conditional branch in action.

dbt/macros/tests/test_generate_schema_name.sql
```diff
@@ -5,13 +5,6 @@
     {% do test_generate_schema_name_raises_for_default_schema_name() %}
 {% endmacro %}
 
-{% macro mock_env_var(var_name) %}
-    {% if var_name == "USER" %} {{ return("test-user") }}
-    {% elif var_name == "HEAD_REF" %} {{ return("testuser/feature-branch-1") }}
-    {% else %} {{ return("") }}
-    {% endif %}
-{% endmacro %}
```
jeancochrane, line 13: This mock hasn't been deleted, I just moved it to the `utils` module below since I'm now reusing it outside the context of the `test_generate_schema_name` module.

dbt/macros/get_s3_dependency_dir.sql
dfsnow: **suggestion (non-blocking):** Well, if we're following our existing schema for the dev, ci, prod split, something like this might be better:

* `dev`
  * data dir:  `s3://ccao-dbt-athena-dev-us-east-1/data/`
  * dependency dir: `s3://ccao-dbt-athena-dev-us-east-1/packages/<username>/`
* `ci`
  * data dir: `s3://ccao-dbt-athena-ci-us-east-1/data/`
  * dependency dir: `s3://ccao-dbt-athena-ci-us-east-1/packages/<branch_slug>/`
* `prod`
  * data dir: `s3://ccao-athena-ctas-us-east-1/`
  * dependency dir: `s3://ccao-athena-packages-us-east-1/`

.github/scripts/deploy_dbt_model_dependencies.sh
dfsnow, line 1: **praise:** This is some pretty wild bash. Appreciate the clear comments!

dfsnow (approved):
This is awesome @jeancochrane. Just a minor issue and a suggestion and I think this is set to go.

dbt/macros/get_s3_dependency_dir.sql
jeancochrane: @dfsnow I agree, that pattern would be cleaner, but [per my comment above](https://github.com/ccao-data/data-architecture/pull/435/files#r1596033934) we aren't actually able to configure the dependency directory as a dedicated profile variable, so it needs to be derivable from other variables like `s3_data_dir`. This is why I proposed the organization sketched out in this PR, as imperfect as it is; if you have suggestions for a better one that is derivable from existing profile variables I would be grateful!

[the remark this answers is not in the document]
jeancochrane: Good call, I fixed handling for all-space variables and added tests in 9dde41d5cd69f3067085ee2457e64fd8206cf9e8!

jeancochrane: I like that idea! Done in 02cc83a3ee89b3bda571f489d9f04a125ab069b3.

jeancochrane:
@dfsnow Want to give this one more look before I merge?

.github/scripts/cleanup_dbt_resources.sh
```diff
@@ -51,5 +51,12 @@ echo "$schemas"
 
 echo "$schemas" | xargs -i bash -c 'delete_database {}'
 
+s3_dependency_dir=$(dbt run-operation print_s3_dependency_dir --quiet \
+    --target "$1"
+)
+
+echo "Deleting Python model requirements at $s3_dependency_dir"
+aws s3 rm "$s3_dependency_dir" --recursive
```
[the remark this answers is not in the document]
dfsnow: Hmmm alright, as long as there's no way for that call to ever target anything other than the target env buckets then I think it's okay. I can dig a little more on this when we do our permissions cleanup.

dbt/models/default/schema/default.vw_pin_address.yml
```diff
@@ -1,6 +1,11 @@
 models:
   - name: default.vw_pin_address
     description: '{{ doc("view_vw_pin_address") }}'
+    config:
+      packages:
+        - "assesspy==1.1.*"
+        - "numpy==1.26.*"
+        - "pandas==2.*"
```
dfsnow, line 8: **question (blocking):** Why are these here? Just for testing?

dfsnow (approved):

dbt/models/default/schema/default.vw_pin_address.yml
jeancochrane: Sorry about that, this was one of the things that prompted me to un-request your review after realizing I had requested it prematurely! I'll just DM you next time that happens instead of messing around with the GitHub review request settings. Fixed in [a5f5262](https://github.com/ccao-data/data-architecture/pull/435/commits/a5f5262cc49cff8db3659e31b1cab979588b44dc).

.github/scripts/cleanup_dbt_resources.sh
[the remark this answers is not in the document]
jeancochrane: Sounds good! I think we're also slightly more protected by the switch to variables for defining `s3_dependency_dir` in [02cc83a](https://github.com/ccao-data/data-architecture/pull/435/commits/02cc83a3ee89b3bda571f489d9f04a125ab069b3), since now our S3 paths for dependencies don't overlap with our paths for data.

PR #437, opened 2024-05-09
Bump role duration for daily builds

dfsnow:
Our [daily dbt build](https://github.com/ccao-data/data-architecture/actions/runs/9017116315/job/24780765574) failed this morning with a role timeout. I'm fairly certain we just need to bump the role duration passed during the setup stage of the workflow.


jeancochrane (approved):
Let's give it a try! If we need to bump it past 4 hours we'll have to bump the maximum session duration for the IAM role in the AWS console, but that should be easy if it comes to it.

Issue #433, opened 2024-05-07
More small tweaks to dbt tests based on Valuations feedback


jeancochrane:
- [ ] Update the `config.where` expression in `iasworld_dweldat_char_attic_fnsh_not_null` to filter for only records where `attic` is not `3` ("None"), and adjust the description in `meta.description` to add "when attic is not 3 (None)"
- [ ] Update the `config.where` expression in `iasworld_dweldat_char_bsmt_fin_not_null` to filter for only records where `bsmt` is not `2` ("Slab"), and adjust the description in `meta.description` to add "when bsmt is not 2 (Slab)"
- [ ] Update the `config.where` expression in `iasworld_dweldat_char_ncu_not_null` to filter for only records where `class` is 212, adjust the description in `meta.description` to add "when class is 212", and change the `name` to `iasworld_dweldat_char_ncu_not_null_class_212`
- [ ] Add a new test `iasworld_dweldat_char_ncu_null_or_0_class_non_212` similar to `iasworld_dweldat_char_ncu_not_null_class_212` but with a filter for `class != '212'` and with allowed values being `0` or null

PR #441, opened 2024-05-10
Tweak `dweldat` and `pardat.tieback` tests based on feedback

jeancochrane:
This PR makes a few tweaks to `dweldat` and `pardat.tieback` tests based on feedback from Valuations:

- Updates not-null test for `dweldat` attic finish to filter out cases where `attic` is `3: None`
- Updates not-null test for `dweldat` basement finish to filter out cases where `bsmt` is `2: Slab`
- Updates `dweldat` tests for number of commercial units to only allow nulls or 0 when class is not `212`, and disallow these values when class is `212`
- Excludes `pardat` test to ensure hyphens are not in tieback from workbook, since Valuations can't act on it

Closes https://github.com/ccao-data/data-architecture/issues/433.


dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1263,10 +1263,23 @@ sources:
...
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class NOT IN ('201', '213', '218', '219', '220', '221', '224', '225', '236', '240', '241', '290', '294', '297')
+                      AND attic != '3'
                   meta:
-                    description: user6 (Attic Finish) should not be null
+                    description: >
+                      user6 (Attic Finish) should not be null unless attic is 3 (None)
```
jeancochrane, line 1282: It just occurred to me that I don't know if the reverse is true (should `user6` _always_ be null if attic is 3?) but I will check with Valuations. They're busy right now though, so I don't think the result of that discussion should block this PR.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1295,10 +1308,23 @@ sources:
...
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class NOT IN ('201', '213', '218', '219', '220', '221', '224', '225', '236', '240', '241', '290', '294', '297')
+                      AND bsmt != '2'
                   meta:
-                    description: user12 (Basement Finished) should not be null
+                    description: >
+                      user12 (Basement Finished) should not be null unless bsmt is 2 (Slab)
```
jeancochrane, line 1327: Same deal here with the reverse test.

dbt/models/iasworld/schema/iasworld.dweldat.yml
```diff
@@ -1376,12 +1402,42 @@ sources:
...
+                      when class is 212
+              - is_null:
+                  name: iasworld_dweldat_char_ncu_null_when_class_is_not_212_and_char_ncu_not_0
+                  additional_select_columns: *select-columns-with-class
+                  config:
+                    where: |
+                      CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
+                      AND cur = 'Y'
+                      AND deactivat IS NULL
+                      AND class != '212'
```
jeancochrane, line 1435: These test failures include a lot of exempt dwellings (class `EX`). I wonder if this class restriction needs a little bit more polish?

dbt/models/iasworld/schema/iasworld.pardat.yml
```diff
@@ -282,14 +282,18 @@ sources:
...
                   expression: not like '%-%'
                   additional_select_columns: *select-columns
                   config: *unique-conditions
                   meta:
                     category: incorrect_values
                     description: tieback should not contain hyphens
+                  tags:
+                    # Valuations can't act on these failures since hyphens are
+                    # not distinguishable in the iasWorld UI
+                    - test_qc_exclude_from_workbook
```
jeancochrane, line 296: I'm thinking about emailing Mirella to ask if we can act on these ourselves, since there are <1k of them. Does that sound OK to you?

dbt/models/iasworld/schema/iasworld.dweldat.yml
dfsnow: Almost certainly yes, but let's check with them to be sure.

dbt/models/iasworld/schema/iasworld.dweldat.yml
dfsnow: Almost certainly these are thing that _were_ 212s and are now exempt. I would exclude exempt (`EX`) and railroad (`RR`) from this test.

dbt/models/iasworld/schema/iasworld.pardat.yml
dfsnow: Yes absolutely, let's suggest that we provide a mass upload with the current and then corrected versions.

dfsnow (approved):

dbt/models/iasworld/schema/iasworld.dweldat.yml
jeancochrane: Makes sense, I excluded non-regression classes and exempt/RR dwellings in 5fc90ec80f649475291fa47fec80fa543678a208!

dbt/models/iasworld/schema/iasworld.pardat.yml
jeancochrane: Done! I also wonder if perhaps we should update our consuming code to strip the hyphens -- as far as I can tell we don't do that yet.

PR #444, opened 2024-05-13
Sort out renv for aws-s3, update dependencies

wrridgeway:


wrridgeway:
Was able to fully `renv::restore()` using these package versions.

PR #447, opened 2024-05-14
Add dbt testing framework for seeds

Damonamajor:
File changes: 

This produces a simple test which identifies if all LOA values are less than .38. This is the largest value in the ccao.loa.csv file, and passes as could be expected.

Does testing seeds work as expected?

To figure this out, we created a temporary target file and changed one of the values in ccao.loa.csv. We ran` dbt build -s state:modified state:new --exclude "tag:test_qc*" --state "target-old"`. When we did not change any values, the seed test was not initiated. We then modified ccao.loa.csv with a number which we suspected would create an error. When this file was modified, the test was run, and an error was produced.


jeancochrane (approved):
Awesome! We can see from these workflow run logs that the test executed and passed as expected: https://github.com/ccao-data/data-architecture/actions/runs/9083267108/job/24961597577?pr=447#step:8:47

PR #459, opened 2024-05-20
Add initial SoT design doc

### LINKED ISSUE #458: Create reporting "source-of-truth" design doc
The Data Department gets asked to fill many ad-hoc reporting requests for various aggregate stats, usually partitioned by geography. This has resulted in a buildup of ad-hoc reporting views and tables over time. These ad-hoc views are difficult to debug and trust, because each of them is complicated in a slightly different way.

Rather than continue to create such ad-hoc views, we should consolidate our reporting efforts into a few larger views that cover the majority of CCAO reporting use cases. These new views will cover different groupings, geographies, and valuation stages, and will use a consistent methodology across time. They will serve as the new "source-of-truth" for both internal and external CCAO reporting.

This issue covers the scoping and design docs process for this effort.

dfsnow:
This PR adds a draft design document outlining the motivation, requirements, and design of reporting "source-of-truth" tables. We can use this PR to get everyone on the same page and possibly solicit feedback from other departments.

Closes #458.

### DIFF
--- a/docs/design-docs/reporting-sot.md
+++ b/docs/design-docs/reporting-sot.md
@@ -1,6 +1,6 @@
-# Design doc: Reporting source-of-truth (SoT) views
+# Design doc: Reporting source-of-truth (SoT) tables
 
-This doc describes the design of a new set of reporting SQL views that aim
+This doc describes the design of a new set of reporting SQL tables that aim
 to make the CCAO's reporting more consistent, durable, debuggable, and useful.
 
 ## Motivation
@@ -12,8 +12,8 @@ are difficult to debug and trust, because each of them is complicated in a
 slightly different way.
 
 Rather than continue to create such ad-hoc views, we should consolidate our
-reporting efforts into a few larger views that cover the majority of CCAO
-reporting needs. These new views will cover different groupings, geographies,
+reporting efforts into a few larger tables that cover the majority of CCAO
+reporting needs. These new tables will cover different groupings, geographies,
 and valuation stages, and will use a consistent methodology and schema.
 They will serve as the new "source-of-truth" for both internal and external
 CCAO reporting.
@@ -38,7 +38,7 @@ However, this constant change creates two problems:
   needs, as the available assets are often changing.
 
 Now that the pace of institutional change has slowed a bit, it's possible to
-create a more permanent set of reporting views that would help solve these
+create a more permanent set of reporting tables that would help solve these
 problems.
 
 #### Requirements
@@ -49,7 +49,7 @@ We should be able to:
   disappearing in a few years
 - Store the data in a persistent and/or fully reproducible way, such that
   there is a near-zero chance of data loss
-- Build long-term applications and extracts on top of these views (e.g. in
+- Build long-term applications and extracts on top of these tables (e.g. in
   Open Data or Tableau)
 
 ### Consistent methods over time
@@ -63,7 +63,7 @@ published ratio statistics across years is very difficult, if not impossible.
 Even aggregate statistics like counts can change over time based on what is
 or is not included in the underlying data.
 
-The goal of the new views would to be apply a single set of consistent filters,
+The goal of the new tables would to be apply a single set of consistent filters,
 methods, and SOPs, such that the resulting statistics are comparable across
 time.
 
@@ -86,7 +86,7 @@ We should be able to:
 The vast majority of Data Department reporting needs/asks are some version of,
 "Fetch an aggregate statistic by group and geography." Currently, these needs
 are met on an ad-hoc basis by individual assets, but we can likely create
-larger views that cover *most* possible geographies, groups, and aggregate
+larger tables that cover *most* possible geographies, groups, and aggregate
 statistics.
 
 #### Requirements
@@ -109,13 +109,13 @@ the time. As a result, it's difficult to know if they are perfectly accurate,
 given the necessary complexity of their underlying SQL. Their complexity and
 lack of shared schema also means they're difficult to work on for newcomers.
 
-Consolidating reporting into larger views would let us focus our debugging and
+Consolidating reporting into larger tables would let us focus our debugging and
 documentation efforts into just a few areas, rather than a dispersed set of
 assets.
 
 #### Requirements
 
-New reporting views should:
+New reporting tables should:
 
 - Have strict tests for:
   - Completeness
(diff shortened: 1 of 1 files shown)


docs/design-docs/reporting-sot.md
```diff
@@ -0,0 +1,533 @@
...
+problems.
+
+#### Requirements
+
+We should be able to:
+
+- Reference columns or tables by name without fear of them changing or
+  disappearing in a few years
+- Store the data in a persistent and/or fully reproducible way, such that
+  there is a near-zero chance of data loss
```
dfsnow, line 51: This becomes much more complex if the tables need to be immutable, see open questions section below.

docs/design-docs/reporting-sot.md
```diff
@@ -0,0 +1,538 @@
...
+- `tax_fire_protection_district`
+- `tax_community_college_district`
+- `tax_sanitation_district`
+- `tax_special_service_area`
+- `tax_tif_district`
+- `central_business_district`
+
+#### Enumerated group_type
+
+The `group_type` column above should span the following values.
```
dfsnow, line 201: @ccao-jardine @wrridgeway Any other groups you all can think of?

docs/design-docs/reporting-sot.md
```diff
@@ -0,0 +1,538 @@
...
+
+- `no_group` (geography breakout only)
+- `class` (minor class, 3-digit code)
+- `major_class` (first digit of class code, excluding letter)
+- `modeling_group` (SF, MF, CONDO, LAND, NON-RES)
+- `res_comm` (residential or commercial)
+- `av_quintile`
+
+### Tables
```
dfsnow, line 210: @wrridgeway @ccao-jardine I tried my best to catalog and enumerate all the possible/common reporting columns we use. Let me know if we can nix some or if there are any missing.

docs/design-docs/reporting-sot.md
```diff
@@ -0,0 +1,538 @@
...
+AND geography_id = '2'
+AND geography_year = '2023'
+AND group_type = 'class'
+AND group_id IN ('202', '203', '204')
+AND stage = 'mailed'
+```
+
+## Open Questions
+
+### Mutability
```
dfsnow, line 512: This is probably the biggest outstanding question. @ccao-jardine what do you think? Is this a requirement?

docs/design-docs/reporting-sot.md
```diff
@@ -0,0 +1,538 @@
...
+- Stages contain their representative sets, i.e. `mailed`, `certified`, and
+  `board` stages exist for each year regardless of whether or not they've
+  been finalized
+- Values fall within expected ranges
+  - Counts are always greater than or equal to 0
+  - FMVs and AVs are constrained by
+- No null, inf, or NaN values are present (where appropriate, some nulls are
+  expected e.g. not all PINs are covered by some geography types)
+- Tables are unique by primary key
+- Tests are run daily (as part of the `build-daily-dbt-models` workflow)
```
jeancochrane, line 477: **[Thought, non-blocking]** Nice tests! Just flagging that we don't currently have a system for running non-iasWorld data tests on a set schedule, so this will likely involve coming up with a new tag and then updating `build-daily-dbt-models` to run tests with that tag -- I think that's what this section is suggesting too, but I just want to make it explicit.

jeancochrane (approved):
Really nice, I'm excited for this!

docs/design-docs/reporting-sot.md
wrridgeway: No, this covers them.

docs/design-docs/reporting-sot.md
wrridgeway: I think this suffices. We can always revisit this if necessary.

wrridgeway:
The only comment I have to make is that we will still some additional views in reporting for tableau, but we can use these tables as the basis for most of those views. Things like `reporting.vw_top_5` won't be able to draw from these tables unless I've misunderstood them, and sometimes @ccao-jardine is looking specifically for wide vs long tables.

wrridgeway (approved):
The only comment I have to make is that we will still some additional views in reporting for tableau, but we can use these tables as the basis for most of those views. Things like reporting.vw_top_5 won't be able to draw from these tables unless I've misunderstood them, and sometimes @ccao-jardine is looking specifically for wide vs long tables.

docs/design-docs/reporting-sot.md
```diff
@@ -0,0 +1,538 @@
...
+- `group_type (string)` - The sub-group for aggregation within each geometry
+  e.g. class, modeling group, etc.
+- `group_id (string)` - The sub-group identifier e.g. for class it would be
+  `202`, `203`, etc.
+- `(optional) stage_name (string)` - Name of the assessment stage used for
+  grouping, one of `mailed`, `certified`, or `board`
+- `(optional) stage_id (string)` - Numeric identifier of the assessment
+  stage, one of `1` (mailed), `2` (certified), `3` (board)
+
+#### Enumerated geography_type
```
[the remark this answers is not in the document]
ccao-jardine: This is great! My only note is a complication about `ward_num`; because we have ward nums for two municipalities (Chicago and Evanston), historically, I've asked for the municipality name to be appended with `ward_num` in the `geo_id `(for example, "Chicago 1").

docs/design-docs/reporting-sot.md
dfsnow: @ccao-jardine I do want to get your input on this before we merge the design doc, as your answer here would seriously change how we build this project.

docs/design-docs/reporting-sot.md
ccao-jardine: In addition to things like municipality containing all county municipalities, let's avoid confirmation bias and be sure to test the opposite: that there are no municipalities that are not in the list of county municipalities, or wards that are not real wards (e.g., Evanston ward 30). I suspect we already have tests for this but just signal-boosting!

docs/design-docs/reporting-sot.md
ccao-jardine: Let's start with a mutable version and go from there!

ccao-jardine (commented):
Nice work. I've not yet gotten feedback on the utility of our pilot version of these kinds of tables, but this is an exciting high value project.

PR #455, opened 2024-05-16
Restructure and cleanup repository

dfsnow:
This PR simplifies, restructures, and cleans up our data architecture repository. Don't be fooled by the absolutely heinous diff, this is mostly just moving things around. I recommend reviewing this PR by just [switching branches](https://github.com/ccao-data/data-architecture/tree/dfsnow/restructure-directories) and browsing the new structure. The main changes are:

- Remove the `aws-athena` directory and all symlinks to that directory (replacing with hard copies instead). This simplifies dbt model setup considerably. I've also updated the documentation to reflect the change.
- Add a top-level README to signpost to other documentation.
- Remove all RPIE-related models and assets. The data will still exist in S3 if it's needed.
- Update various miscellaneous files like .gitignores and pre-commit configs
- Remove other deprecated files, such as .gitkeep files and our old data catalog
- Rename the `aws-s3` directory to just `etl`

I'll call out other changes with PR comments. Note that merging this may rebuild the entire dbt DAG and will likely create substantial merge conflicts.

Also closes #148.

PR #461, opened 2024-05-20
Refactor Python model dependencies to use global package repository

jeancochrane:
In the course of discussion on #453, we [decided](https://github.com/ccao-data/data-architecture/pull/453#pullrequestreview-2063739974) to simplify the design of the Python model dependency deployment system:

> Rather than a per-model, per-environment zip of packages, we could create a global directory of packages by version (probably in `s3://ccao-athena-dependencies-us-east-1`). [...] In this setup, people would specify the specific package version they want for a model in that model's schema file. We'd then have a CI job collect and dedupe those versions into a single list. We can then `pip install --no-deps` each of the packages, zip the results, and name the file by the package name and version. Finally, we'd `aws s3 sync` the CI package dir with the bucket.

This PR refactors `deploy_dbt_model_dependencies.sh` and the `reporting.ratio_stats` Python model to implement this new design.


.github/scripts/deploy_dbt_model_dependencies.sh
```diff
@@ -1,184 +1,155 @@
...
 # Determine the remote location where bundled dependencies will be deployed
-s3_dependency_dir=$(dbt run-operation print_s3_dependency_dir --quiet \
-    --target "$target")
-
-# Parse optional positional arguments representing a restricted list of
-# models for which to upload dependencies
-declare -a specified_models
-if [ $# -gt 0 ]; then
-    specified_models=("$@")
-fi
```
jeancochrane, line 33: Since packages are now configured in a global repository rather than in environments isolated by model and dbt target, it felt like `specified_models` and its positional args were adding unnecessary complexity, so I went ahead and removed it. We may want to reintroduce this logic to speed up build times if the universe of packages we use for our scripts ends up growing by a lot, but in the meantime I think reduced maintenance complexity is more important than a slight reduction in build times. Let me know if you disagree!

dbt/dbt_project.yml
```diff
@@ -66,8 +66,6 @@ models:
       +schema: proximity
     reporting:
       +schema: reporting
-    rpie:
-      +schema: rpie
```
jeancochrane, line 70: The presence of this config was throwing sporadic errors given that `rpie` models were all removed in https://github.com/ccao-data/data-architecture/pull/455.

dbt/models/reporting/reporting.ratio_stats.py
```diff
@@ -1,234 +1,238 @@
 # pylint: skip-file
```
jeancochrane, line 1: The diff for this file is large, but the actual changes are very minor; mostly, we undo the indentation we added in https://github.com/ccao-data/data-architecture/pull/453 by pulling the imports and function definitions back out of the `main()` file. Ignoring whitespace should help with reviewing this file again.

dbt/README.md
```diff
@@ -206,12 +206,10 @@ dbt build --select default.* --resource-types model seed
...
-> To make these dependencies available to your model, run
-> `../.github/scripts/deploy_dbt_model_dependencies.sh <your_model_name>`
-> in the context of the `dbt/` directory before you run `dbt build`.
-> You can also run the script with no arguments to make dependencies available
-> for all Python models in the project, but this will take a longer time to
-> complete and will use extra storage space on S3.
+> To make these dependencies available to your model, run the
+> `build-and-test-dbt` workflow on your branch to deploy any Python dependencies
+> that you've added to [the `config.packages` attribute](https://docs.getdbt.com/docs/build/python-models#configuring-packages)
+> on your model.
```
jeancochrane, line 212: Juniors don't have permissions to modify the `ccao-athena-dependencies-us-east-1` bucket, so running the workflow will be the easiest way to update dependencies. I wonder if it would be better to just give them permissions to modify the bucket, given that it's easy to recreate its contents?

dfsnow (approved):
This looks great @jeancochrane. Sorry to make you redo some work. Confident this will be much simpler in the long-run.

dbt/README.md
jeancochrane: Cool! I'll leave it as-is for now then, but we can keep an eye on it and adjust their permissions if it proves to be easier for them to run the script directly.

dbt/dbt_project.yml
```diff
@@ -52,8 +52,6 @@ models:
     +materialized: view
     +write_compression: zstd
     +format: parquet
-    ccao:
-      +schema: ccao
```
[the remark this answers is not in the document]
jeancochrane: As with the `rpie` schema, if we leave it in, we get a warning that no models are configured to use the schema:

```
16:26:41  [WARNING]: Configuration paths exist in your dbt_project.yml file which do not apply to any resources.
There are 1 unused configuration paths:
- models.ccao_data_athena.ccao
```

It doesn't seem to affect dbt's behavior either way, but I figured we might as well silence the warning.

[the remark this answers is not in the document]
dfsnow: As long as it doesn't affect the creation of the `ccao` seeds!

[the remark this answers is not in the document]
jeancochrane: Seeds should be unaffected since they're configured under a separate `seeds` attribute:

https://github.com/ccao-data/data-architecture/blob/aebf2752056da89b9d47bd338450739d80ce47b8/dbt/dbt_project.yml#L78-L81

PR #464, opened 2024-05-21
Add `expression_is_false` dbt test macro

wrridgeway:
Our current `expression_is_true` dbt tests can lead to some weird backwards engineering trying to figure out how to invert logic given how it employs `NOT` before the `WHERE` clause it tests. Having an alternative to avoid these situations would be helpful.

### DIFF
--- a/dbt/tests/generic/test_expression_is_false.sql
+++ b/dbt/tests/generic/test_expression_is_false.sql
@@ -1,0 +1,27 @@
+-- Filter for rows where a given `expression` is true.
+--
+-- Adapted from our own `expression_is_true`. Supports an optional
+-- `additional_select_columns` option representing an array of columns to select
+-- for failing rows in addition to `column_name`. If no
+-- `additional_select_columns` array is provided, defaults to selecting the
+-- column represented by `column_name`; if `column_name` is also missing, falls
+-- back to selecting 1 for failing rows.
+{% test expression_is_false(
+    model, column_name, expression, additional_select_columns=[]
+) %}
+    {%- set select_columns_csv = format_additional_select_columns(
+        additional_select_columns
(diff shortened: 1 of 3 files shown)


wrridgeway:
This is ready to go, but these test names feel backwards to me. Probably not important.

dbt/tests/generic/test_expression_is_false.sql
```diff
@@ -0,0 +1,27 @@
...
+            {%- set columns_csv = columns_csv ~ ", " ~ select_columns_csv -%}
+        {%- endif -%}
+    {%- elif select_columns_csv -%} {%- set columns_csv = select_columns_csv -%}
+    {%- else -%} {%- set columns_csv = "1 AS fail" -%}
+    {%- endif -%}
+
+    select {{ columns_csv }}
+    from {{ model }}
+    where ({{ column_name }} {{ expression }})
+{% endtest %}
```
jeancochrane, line 27: **[Thought, non-blocking]** Looks like everything here is identical with `expression_is_true` except for the lack of a `not` operator on line 26. Perhaps a sign that we could factor out the shared logic into a third macro that `expression_is_true` and `expression_is_false` both inherit from? Not required by any means, but something to think about!

jeancochrane (approved):
Nice! Real quick before we merge this in, could we add some simple docs for `expression_is_false` to the [generic tests README](https://github.com/ccao-data/data-architecture/tree/master/dbt/tests/generic#readme)?

dbt/tests/generic/test_expression_is_false.sql
wrridgeway: I'm not opposed to that at all.

PR #471, opened 2024-05-22
Ingest affordability risk index ari for modeling

Damonamajor:
Billy and I are running into some conceptual issues regarding the ARI - DCI indexes. Originally, the query was built off IHS. This made sense since they were both geographic indexes for housing rankings. But that join is based on 2010 census data and specific view “census_2010”. Billy tried playing with removing that view, but deemed it to be necessary. Because of this, it’s been housed in the ‘other’ bucket. Do we want to keep it there, or shift it to the ‘spatial’ bucket? Billy was wishy-washy on the need to create a census_2020. Is that a desire?

And, I’ve been reading through the docs, and peeping through the different aws buckets. It appears we have a lot of single-year databases (walkability, economy, etc.). Do we want ARI- DCI to be joined in that structure (based off a single year), or the expectation that we will download updated datasets on a regular basis? 

Related to this, when we last touched on it, we discussed the data being joined based on year of download. When I mentioned it to Billy / Nicole, there was hesitancy, and they suggested based on the year that the data was constructed. This would be 2023 for ARI and 2024 for DCI. If this is something we construct a single time, this would be easy to forward fill. But, if we have some expectation to update it annually (or at odd intervals), the sql join may be a bit wonkier. I've been playing with a query where the goal is to identify the maximum year of upload, which is less than the tract year, but it wouldn't seem useful to implement until we have a multi-year structure in place. For the moment, a join based on >= year would make sense since it's only downloaded once. This seems to be our "standard" approach for things where there isn't a full expectation of yearly download (census). 

I also modified the script to directly query the url rather than the structure I had before of identifying the xlsx file on the page, since the script / processes seemed too ambiguous for that. 

I included etl/scripts-ccao-data-warehouse-us-east-1/export/export-geojson.R, because IHS was included in it, but then ran into issues where the function read_geoparquet_sf was deprecated. Billy said you are aware of it, but I expect these changes just to be reverted.


etl/scripts-ccao-data-raw-us-east-1/housing/housing-ari_index.py
```diff
@@ -0,0 +1,81 @@
...
+            "(KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.3"
+        )
+    }
+    response = requests.get(base_url, headers=headers)
+    response.raise_for_status()
+
+    soup = BeautifulSoup(response.text, "html.parser")
+    links = soup.find_all("a", href=True)
+
+    xlsx_links = [
```
Damonamajor, line 23: This pulls an xlsx link rather than a specified sone, so if the string changes, the file will still work.

etl/scripts-ccao-data-warehouse-us-east-1/housing/housing-ari_index.py
```diff
@@ -0,0 +1,43 @@
...
+s3 = boto3.client("s3")
+temp_file = tempfile.NamedTemporaryFile(suffix=".xlsx", delete=False)
+
+load_dotenv("etl/.Renviron")
+AWS_S3_RAW_BUCKET = os.getenv("AWS_S3_RAW_BUCKET")[5:]
+file_key = os.path.join("housing", "ari_index", "2023-ARI.xlsx")
+
+s3.download_file(AWS_S3_RAW_BUCKET, file_key, temp_file.name)
+
+# Use pandas to read the Excel file, skipping the first two rows
```
Damonamajor, line 19: First two rows of the file are bogus.

etl/scripts-ccao-data-warehouse-us-east-1/housing/housing-ari_index.py
```diff
@@ -0,0 +1,43 @@
...
+data = pd.read_excel(temp_file.name, skiprows=2, engine="openpyxl")
+data = data[["Census Tract", "Total ARI Score"]]
```
Damonamajor, line 21: Keeps two of the three columns.

dbt/models/shared_columns.md
```diff
@@ -1050,6 +1050,14 @@ DePaul Institute of Housing Studies quarterly index.
 Averaged up to the yearly level. Unit of observation is Census PUMAs
 {% enddocs %}
 
+## ari_index
+
+{% docs shared_column_ihs_avg_year_index %}
```
wrridgeway, line 1055:
```suggestion
{% docs shared_column_ari %}
```

etl/scripts-ccao-data-warehouse-us-east-1/housing/housing-ari_index.py
```diff
@@ -0,0 +1,48 @@
...
+load_dotenv("etl/.Renviron")
+AWS_S3_RAW_BUCKET = os.getenv("AWS_S3_RAW_BUCKET")[5:]
+file_key = os.path.join("housing", "ari_index", "2023-ARI.xlsx")
+
+s3.download_file(AWS_S3_RAW_BUCKET, file_key, temp_file.name)
+
+# Use pandas to read the Excel file, skipping the first two rows
+data = pd.read_excel(temp_file.name, skiprows=2, engine="openpyxl")
+current_year = datetime.now().year
+data["year"] = current_year
```
Damonamajor, line 23: We don't know if/how often the file will be updated. The creation of the year was to upload to allow it to forward-fill.

dbt/models/model/model.vw_pin_shared_input.sql
```diff
@@ -66,6 +66,16 @@ housing_index AS (
     GROUP BY puma.pin10, ihs.year
 ),
 
+affordability_risk_index AS (
+    SELECT
+        tract.pin10,
+        ari.year
+    FROM {{ source('other', 'ari_index') }} AS ari
+    LEFT JOIN {{ ref('location.tract_2020') }} AS tract
```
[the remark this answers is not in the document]
wrridgeway: This isn't actually a table, so perhaps we should look into building location.census_2020 or figuring out a new solution.

jeancochrane:
@wrridgeway I think the cache permissions warning is a red herring, since it's just a warning. [This](https://github.com/ccao-data/data-architecture/actions/runs/9275562861/job/25521246660#step:4:437) is the line that I think is the culprit:

```
etl/scripts-ccao-data-raw-us-east-1/housing/housing-ari.py:72: error: Value of type "Optional[str]" is not indexable  [index]
```

The issue there is that `os.environ.get()` can return `None` if the env var isn't set, in which case attempting to slice the indeces `[5:]` would raise a `TypeError`.

wrridgeway:
Thanks, I should have realized. I was addressing that same issue elsewhere.

etl/scripts-ccao-data-warehouse-us-east-1/housing/housing-ari.py
```diff
@@ -0,0 +1,40 @@
...
+)
+data["geoid"] = data["geoid"].astype(str)
+temp_file.close()
+
+AWS_S3_WAREHOUSE_BUCKET = os.getenv("AWS_S3_WAREHOUSE_BUCKET")
+
+# Upload the Parquet file to S3
+data.to_parquet(
+    os.path.join(
+        AWS_S3_WAREHOUSE_BUCKET,  # type: ignore
```
wrridgeway, line 34: Not sure what to do about this. Is it worth coding it differently to avoid the mypy error?  We _want_ the script to fail if `os.getenv("AWS_S3_WAREHOUSE_BUCKET")` returns `None`, don't we?

jeancochrane: @wrridgeway In that case I would recommend retrieving the value with `os.environ["AWS_S3_WAREHOUSE_BUCKET"]` instead, which will raise a `KeyError` if the variable is not set!

dbt/models/model/model.vw_pin_shared_input.sql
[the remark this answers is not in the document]
Damonamajor: @dfsnow @wrridgeway I did some more digging and the 2023 census tracts are a 1:1 match. Would location.census_2020 a table worth building?

dbt/models/shared_columns.md
```diff
@@ -1050,6 +1050,14 @@ DePaul Institute of Housing Studies quarterly index.
 Averaged up to the yearly level. Unit of observation is Census PUMAs
 {% enddocs %}
 
+## ari
+
+{% docs shared_column_ari %}
+Illinois Housing Development Authority's Affordability Risk Index.
+
+Data may not be udpated regularly, so year refers to year of download
```
dfsnow, line 1058: **issue:** Should be year of construction, I misspoke.

etl/scripts-ccao-data-warehouse-us-east-1/export/export-geojson.R
dfsnow, line 1: **issue (blocking):** We probably _don't_ actually need to update this script. At this point I don't even think it's used. You can definitely revert the changes here.

etl/scripts-ccao-data-warehouse-us-east-1/housing/housing-ari.py
```diff
@@ -0,0 +1,43 @@
...
```
dfsnow, line 37: I would implement Jean's suggestion below and also move the loading of the environment variables to the top of the script after the imports.

dfsnow (commented):
First to address your questions:

> Billy and I are running into some conceptual issues regarding the ARI - DCI indexes. Originally, the query was built off IHS. This made sense since they were both geographic indexes for housing rankings. But that join is based on 2010 census data and specific view “census_2010”. Billy tried playing with removing that view, but deemed it to be necessary. Because of this, it’s been housed in the ‘other’ bucket. Do we want to keep it there, or shift it to the ‘spatial’ bucket? Billy was wishy-washy on the need to create a census_2020. Is that a desire?

I don't _love_ the kludgy nature of an additional, dedicated 2010 Census view, but I also don't want to block this PR. Let's keep it for now and we can factor it out later if needed. No need for a dedicated 2020 view.

> And, I’ve been reading through the docs, and peeping through the different aws buckets. It appears we have a lot of single-year databases (walkability, economy, etc.). Do we want ARI- DCI to be joined in that structure (based off a single year), or the expectation that we will download updated datasets on a regular basis?

It looks to me like both ARI and DCI have yearly updates, in which case we should ingest the each year. The single-year datasets are only for times when we expect to only receive the data once.

> Related to this, when we last touched on it, we discussed the data being joined based on year of download. When I mentioned it to Billy / Nicole, there was hesitancy, and they suggested based on the year that the data was constructed. This would be 2023 for ARI and 2024 for DCI. If this is something we construct a single time, this would be easy to forward fill. But, if we have some expectation to update it annually (or at odd intervals), the sql join may be a bit wonkier. I've been playing with a query where the goal is to identify the maximum year of upload, which is less than the tract year, but it wouldn't seem useful to implement until we have a multi-year structure in place. For the moment, a join based on >= year would make sense since it's only downloaded once. This seems to be our "standard" approach for things where there isn't a full expectation of yearly download (census).

For now, let's treat this as a single-year dataset but plan to fill it in with subsequent years once they're available. We should definitely join by year of construction, not by download (though in this case they should be the same). Question: is there any way to get historical data for these two datasets?

> I also modified the script to directly query the url rather than the structure I had before of identifying the xlsx file on the page, since the script / processes seemed too ambiguous for that.

Great! That makes sense to me.

> I included etl/scripts-ccao-data-warehouse-us-east-1/export/export-geojson.R, because IHS was included in it, but then ran into issues where the function read_geoparquet_sf was deprecated. Billy said you are aware of it, but I expect these changes just to be reverted.

Hmmm, yes we do need to address the deprecated geoparquet functions. I'll put Billy on that in a separate issue (#492).

## Review

This mostly looks really good! It should be set to go with some super minor changes.

dfsnow (approved):
Alright @Damonamajor, this looks set to merge!

PR #472, opened 2024-05-22
Ingest distressed communities index dci for modeling

Damonamajor:
One of the linting errors is that I don't have write privileges' to the necessary bucket, which I'm not going to be able to change. The other is for the following line. As I understand it, because we are using an .Renviron, it's not exactly the same, so we cut out the first 5 digits. This was what Billy recommended rather than creating a new .env file.


etl/scripts-ccao-data-raw-us-east-1/housing/housing-dci_index.py
```diff
@@ -0,0 +1,55 @@
...
+    s3 = boto3.client("s3")
+    # Create a buffer
+    csv_buffer = StringIO()
+    # Write DataFrame to buffer
+    df.to_csv(csv_buffer, index=False)
+    # Upload buffer content to S3
+    s3.put_object(Bucket=bucket, Key=file_name, Body=csv_buffer.getvalue())
+
+
+# URL of the file you want to load
```
Damonamajor, line 43: the noqa is necessary because concatenating it within {} doesn't allow the query to work for some reason.

etl/scripts-ccao-data-raw-us-east-1/housing/housing-dci_index.py
```diff
@@ -0,0 +1,55 @@
...
+file_url = "https://eig.org/dci-maps-2023/data/1cd12716-de4a-4ef6-884b-af6e1066b581.csv"  # noqa: E501
+
+# Load the data
+df = load_csv_from_url(file_url)
+
+load_dotenv("etl/.Renviron")
+AWS_S3_RAW_BUCKET = os.getenv("AWS_S3_RAW_BUCKET")[5:]
```
Damonamajor, line 50: This removes the first five digits of the bucket allowing us to download the correct environment. It also causes linter to fail.

etl/scripts-ccao-data-warehouse-us-east-1/housing/housing-dci_index.py
```diff
@@ -0,0 +1,58 @@
...
+file_key = os.path.join("housing", "dci_index", "dci_index.csv")
+
+s3.download_file(AWS_S3_RAW_BUCKET, file_key, temp_file.name)
+
+# Use pandas to read the Excel file, skipping the first two rows
+data = pd.read_csv(temp_file.name)
+
+data = data[data["County"] == "Cook County, Illinois"]
+
+data = data[
```
Damonamajor, line 25: These are the columns relevant to distressed scores. It's probably more than what we want, but I left them all in for the time being.

wrridgeway:
@Damonamajor I've got one minor request for changes. Once that's done, can we do a row count comparison for `vw_pin_shared` between master and dev branches and a check to see how much missingness there is in the dci column for the two input views and if matches with what we expect?

Damonamajor:
@wrridgeway 
The join that you recommended works, but becomes even wonkier. I have it up and running, but because of how the census data gets uploaded, the process intended to forward fill, ends up back filling. But, it should even out if we keep the data up. This is because we have the most recent data for DCI (2024). Because we then join to the max value of census (2023) we will keep data coded for that year. As soon as the census data gets updated, to 2024, the max would update, data; 2023 would become NULL and data for 2024 would be included in the correct year. To make things slightly more complicated, our most recent column for 'census_data_year' is 2023, but for 'census_acs5_data_year' it is 2022. As far as I know, there's not really a difference between them, but it is something to note.

dbt/models/model/model.vw_pin_shared_input.sql
```diff
@@ -71,15 +71,42 @@ housing_index AS (
...
+        ON dci.geoid = zcta.census_zcta_geoid
+        AND CASE
+            WHEN
+                dci.year
+                > (SELECT MAX(year) FROM {{ ref('location.census') }})
+                THEN (SELECT MAX(year) FROM {{ ref('location.census') }})
+            ELSE dci.year
+        END
+        = zcta.year
+),
```
wrridgeway, line 91: We gotta clean up the tract/zcta switching here.

wrridgeway:
I think the join is working as anticipated; I only see data for 2023. It's joining onto the most recent census zcta data we have, which is 2023 for now. Once that becomes 2024 it will join onto that and then stay there.  Is this the best way to join the data on? Probably not - but I don't know if this is data we anticipate continuing to gather, and it doesn't feel like data we should back fill?

Ultimately, I'm leaning towards including another column when we ingest this that makes it clear it uses 2022 census data and that we should then join on 2022 geoids and forward fill only. 2024 is when the data was made available, but the data isn't describing 2024. Thoughts @dfsnow ?

dfsnow:
> I think the join is working as anticipated; I only see data for 2023. It's joining onto the most recent census zcta data we have, which is 2023 for now. Once that becomes 2024 it will join onto that and then stay there. Is this the best way to join the data on? Probably not - but I don't know if this is data we anticipate continuing to gather, and it doesn't feel like data we should back fill?
> 
> Ultimately, I'm leaning towards including another column when we ingest this that makes it clear it uses 2022 census data and that we should join on 2022 geoids and forward fill only. 2024 is when the data was made available, but the data isn't describing 2024. Thoughts @dfsnow ?

@wrridgeway @Damonamajor Sorry I missed this. I would do as @wrridgeway suggests and forward fill from 2022.

Damonamajor:
> > I think the join is working as anticipated; I only see data for 2023. It's joining onto the most recent census zcta data we have, which is 2023 for now. Once that becomes 2024 it will join onto that and then stay there. Is this the best way to join the data on? Probably not - but I don't know if this is data we anticipate continuing to gather, and it doesn't feel like data we should back fill?
> > Ultimately, I'm leaning towards including another column when we ingest this that makes it clear it uses 2022 census data and that we should join on 2022 geoids and forward fill only. 2024 is when the data was made available, but the data isn't describing 2024. Thoughts @dfsnow ?
> 
> @wrridgeway @Damonamajor Sorry I missed this. I would do as @wrridgeway suggests and forward fill from 2022.


@dfsnow So to clarify, we want 1 column for year, which represents the year of underlying census data and one column which represents the year of download (2024) or their data construction (2023)?

dfsnow:
> @dfsnow So to clarify, we want 1 column for year, which represents the year of underlying census data and one column which represents the year of download (2024) or their data construction (2023)?

One column (`year`) for the underlying census data and another for year of construction, similar to the `_data_year` column in the `proximity` tables.

wrridgeway:
Where are we on this PR? Ready for review again?

dbt/models/model/model.vw_pin_shared_input.sql
```diff
@@ -79,15 +79,65 @@ housing_index AS (
     GROUP BY puma.pin10, ihs.year
 ),
```
Damonamajor, line 81: @wrridgeway 
A summary of the join:
Essentially, we want a join which accounts for the lagged variable (the census data for the most recent version is from 2021, but we want it to join with 2024 data). Then we want to allow for new data to be input, but we don't know if the data will be input in sequential years. This query identifies the largest value which is less than the year of the census year, and keeps that.

wrridgeway:
Alright, let's just assume ari and dci are static snapshots for now that won't get updated.
- year and data year columns seem to be reversed, so let's fix that
- lets just join the data to the shared input view on dci/ari.data_year >= uni.year and geoid

if we ever update the data we can cross that bridge when we get to it.

Damonamajor:
@wrridgeway 
This should be good. Can you upload to S3 and clean out the previous uploaded versions?

wrridgeway (approved):

PR #474, opened 2024-05-23
Fix broken categorization in `transform_dbt_test_results.py` and remove `dbt_utils` dependency

jeancochrane:
I noticed today that the categorization of tests performed by `transform_dbt_test_results.py` broke after we changed the name of our dbt project from `athena` to `ccao_data_athena` in https://github.com/ccao-data/data-architecture/pull/422, causing most of our tests to get incorrectly categorized as `miscellaneous`. This PR fixes that bug and removes the coupling of `transform_dbt_test_results.py` to the dbt project name so that we can change the name of the project without breaking the script in the future.

I also searched the repo to double-check that there aren't any more hidden references to the old project name hanging around, and as far as I can tell this is the last one.

Note that this PR also removes the dependency on `dbt_utils`, since I noticed that the last remaining use of a `dbt_utils` macro could be refactored to use one of our macros. We've been following a de facto pattern of vendoring `dbt_utils` macros since we need them to support our `additional_select_columns` interface, and this change makes that pattern official.

I tested this by rerunning `transform_dbt_test_results.py` and confirming that the test categorization works as expected again. I'm happy to share the output if you would like to double-check.

### DIFF
--- a/.github/scripts/transform_dbt_test_results.py
+++ b/.github/scripts/transform_dbt_test_results.py
@@ -108,16 +108,13 @@ WEN_FIELD = "wen"
 # Mapping that defines category names that should be reported for tests
 # based on their generics
 TEST_CATEGORIES = {
-    "macro.athena.test_accepted_range": "incorrect_values",
-    "macro.dbt_utils.test_accepted_range": "incorrect_values",
-    "macro.athena.test_accepted_values": "incorrect_values",
-    "macro.athena.test_not_accepted_values": "incorrect_values",
-    "macro.dbt_utils.test_not_accepted_values": "incorrect_values",
-    "macro.athena.test_unique_combination_of_columns": "duplicate_records",
-    "macro.dbt_utils.test_unique_combination_of_columns": "duplicate_records",
-    "macro.athena.test_not_null": "missing_values",
-    "macro.athena.test_is_null": "missing_values",
-    "macro.athena.test_res_class_matches_pardat": "class_mismatch_or_issue",
+    "test_accepted_range": "incorrect_values",
+    "test_accepted_values": "incorrect_values",
+    "test_not_accepted_values": "incorrect_values",
+    "test_unique_combination_of_columns": "duplicate_records",
+    "test_not_null": "missing_values",
+    "test_is_null": "missing_values",
+    "test_res_class_matches_pardat": "class_mismatch_or_issue",
 }
 # Fallback for tests whose category we can't determine from either the
 # test name, the `meta.category` attribute, or the TEST_CATEGORIES mapping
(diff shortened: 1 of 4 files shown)


.github/scripts/transform_dbt_test_results.py
```diff
@@ -108,16 +108,13 @@
...
-    "macro.athena.test_not_null": "missing_values",
-    "macro.athena.test_is_null": "missing_values",
-    "macro.athena.test_res_class_matches_pardat": "class_mismatch_or_issue",
+    "test_accepted_range": "incorrect_values",
+    "test_accepted_values": "incorrect_values",
+    "test_not_accepted_values": "incorrect_values",
+    "test_unique_combination_of_columns": "duplicate_records",
+    "test_not_null": "missing_values",
+    "test_is_null": "missing_values",
+    "test_res_class_matches_pardat": "class_mismatch_or_issue",
```
jeancochrane, line 117: This list remains the same, I've just removed the `macro.<project_name>.*` prefix from each key and deduplicated the keys that previously had a `dbt_utils` version.

dbt/tests/generic/test_sequential_values.sql
```diff
@@ -8,7 +8,7 @@
     additional_select_columns=[]
 ) %}
 
-    {%- set previous_column_name = "previous_" ~ dbt_utils.slugify(column_name) -%}
+    {%- set previous_column_name = "previous_" ~ slugify(column_name) -%}
```
jeancochrane, line 11: Our version of slugify is basically identical to the `dbt_utils` version; the only differences are that we convert backslashes to underscores and we don't prepend underscores if the string starts with a number. Neither of those differences should matter in this case since our column names don't contain backslashes or start with numbers.

For reference, here's our version of the macro:

https://github.com/ccao-data/data-architecture/blob/6efd15a7b874cd8aa403bbd02c704398344e1020/dbt/macros/slugify.sql#L1-L13

And[ here's the dbt utils version](https://github.com/dbt-labs/dbt-utils/blob/fc98e4e30cd299ddd67c43902f2a5efa9511edfb/macros/jinja_helpers/slugify.sql#L1-L14):

```sql
{% macro slugify(string) %}


{#- Lower case the string -#}
{% set string = string | lower %}
{#- Replace spaces and dashes with underscores -#}
{% set string = modules.re.sub('[ -]+', '_', string) %}
{#- Only take letters, numbers, and underscores -#}
{% set string = modules.re.sub('[^a-z0-9_]+', '', string) %}
{#- Prepends "_" if string begins with a number -#}
{% set string = modules.re.sub('^[0-9]', '_' + string[0], string) %}


{{ return(string) }}


{% endmacro %}
```

dbt/package-lock.yml
dfsnow, line 1: **question (non-blocking):** If we end up needing other macros from `dbt_utils`, are we going to just vendor the specific macro or will we re-add the dep?

dfsnow (approved):
Looks good! Sorry I didn't catch this on the Python model PR.

dbt/package-lock.yml
jeancochrane: I think the right move is to vendor the macro, unless two specific conditions arise: A macro that is 1) particularly complex and 2) does not require the `additional_select_columns` parameter that we use for all of our generic tests. In that case, it might make sense to reintroduce the dependency; in the meantime, every macro we've needed from `dbt_utils` so far has either been simple enough that it warrants vendoring or does require `additional_select_columns`.

PR #475, opened 2024-05-24
Rename `tests:` yaml key to `data_tests:`

dfsnow:
Looks like we missed a few `tests:` keys during the upgrade to dbt 1.8.0. This renames them to the new `data_tests:` key and gets rid of the deprecation warning:

```
[WARNING]: Deprecated functionality
The `tests` config has been renamed to `data_tests`. Please see
https://docs.getdbt.com/docs/build/data-tests#new-data_tests-syntax for more
information.
```


dbt/dbt_project.yml
```diff
@@ -37,14 +36,6 @@ vars:
...
   test_qc_year_end: 2030
 
-  # Locations in S3 where Python model dependencies are stored. Since these
-  # locations vary by target, it would be nice if we could define them as target
-  # attributes in profiles.yml, but we can't add new profiles.yml attributes
-  # without extending dbt-athena so we define the locations here instead
-  s3_dependency_dir_dev: s3://ccao-dbt-athena-dev-us-east-1/dependencies
-  s3_dependency_dir_ci: s3://ccao-dbt-athena-ci-us-east-1/dependencies
-  s3_dependency_dir_prod: s3://ccao-athena-dependencies-us-east-1
-
```
dfsnow, line 47: Also dropping these now that the package location is global and hard-coded in the deps script.

dbt/dbt_project.yml
```diff
@@ -69,7 +60,7 @@ models:
       +schema: sale
     qc:
       +schema: qc
-tests:
+data_tests:
```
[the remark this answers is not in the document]
dfsnow: It works! It's the [new syntax as of 1.8](https://docs.getdbt.com/docs/build/data-tests#new-data_tests-syntax). I ran it locally with `--store-failures` and everything checks out. [Here's](https://github.com/ccao-data/data-architecture/actions/runs/9353780075/job/25745084901) a successful workflow run.

dbt/macros/get_s3_dependency_dir.sql
dfsnow, line 1: Also dropping the helper here (and associated tests) since I don't think it's needed anymore, but correct me if I am wrong/if we want to keep it around!

jeancochrane (approved):
Good to go, thanks for answering my questions!

PR #477, opened 2024-05-28
Improve the dbt documenation for ccaocorner lot

Damonamajor:


dfsnow:
We can close this since it got dealt with in #536.

Issue #456, opened 2024-05-16
Reconfigure sqlfluff linter and pre-commit hooks


dfsnow:
It's been awhile since we've touched the SQLFluff/sqlfmt configuration we use for CI/CD. SQLFluff has been updated a lot in the last year or so. We should take a quick look at our configuration and make sure it's still doing everything we want/expect.

Specifically, I don't think the SQLFluff linting in the `macros/` or `tests/` directories is working quite right. It doesn't lint locally for me within macro blocks. Perhaps it's time for us to retry the dbt templater.

dfsnow:
Note that SQLFluff just did a compatibility release for dbt 1.8: https://github.com/sqlfluff/sqlfluff/releases/tag/3.0.7

jeancochrane:
I did a bunch of research on this today, including test driving the dbt templater and trying out a few different sets of configs, and unfortunately my conclusion is that sqlfluff does not support linting macro blocks at all. Since it's not supported using either the dbt templater or the jinjat templater, and since the dbt templater is slightly slower, I think we should stick to the jinja templater for now.

It's not really documented anywhere, but it appears that the sqlfluff policy is to not lint macros because they are only fragments and hence may not be valid SQL. Macros do not get linted in isolation, but instead are linted when they are called by a non-macro SQL statement (like in a model), at which point the templater uses the macro to generate a full SQL statement and then validate it. This policy is particularly apparent when using the dbt templater, which explicitly skips macro files -- here's a sample log line when running `sqlfluff lint` with the dbt templater enabled:

```
WARNING    Skipped file /home/jecochr/data-architecture/dbt/macros/log_results.sql because it is a macro
```

[This SO answer](https://stackoverflow.com/questions/76917011/sqlfluff-skips-linting-macro-files) suggests using the dbt templater for models and the jinja templater for macros, but that doesn't work because the jinja templater will skip anything inside a `{% macro %}` block. (It will lint anything in a macro file but outside a `{% macro %}` block, however, so I do agree that it is probably a better choice for macros.)

I couldn't find any discussion on this particular issue, but I did find some discussion and planned work around [linting unexecuted portions of jinja templates](https://github.com/sqlfluff/sqlfluff/issues/4061). It's possible this may eventually lead to support for linting macros in isolation, but it's hard to tell based on the current discussion, which is very weedsy. The fact that there's no discussion outside of the SO answer I linked leaves me less than confident that I've exhausted all options, however, so it's possible I've just missed something.

[...]

PR #478, opened 2024-05-28
Upgrade sqlfluff to 3.0.7

jeancochrane:
This PR brings us up to date with the latest sqlfluff release, and ignores a few lines for a new rule ([ST03](https://docs.sqlfluff.com/en/stable/rules.html#rule-ST03)) that is incorrectly failing in cases where we define a CTE so that we can pass its name into a dbt macro.

Closes https://github.com/ccao-data/data-architecture/issues/456.


wrridgeway (approved):
Looks good @jeancochrane, thanks for keeping this updated.

Issue #460, opened 2024-05-20
Explore using `dbt-athena`'s high availability option


dfsnow:
Recently, we've merged some PRs that required rebuilding our full dbt project in Athena. A full rebuild deletes the existing tables from Athena *before the queries to replace the tables are finished*. The means that the tables are unavailable for the duration of the rebuild.

This is typically fine, as we can simply merge such rebuild PRs after work hours. However, as more people, applications, and process come to depend on the Athena warehouse, such downtime may become untenable.

The dbt Athena adapter has a built-in solution for this in the form of [high availability tables](https://github.com/dbt-athena/dbt-athena?tab=readme-ov-file#highly-available-table-ha). We should experiment with the capability to see if it works as promised. If it does, it should reduce our downtime and/or the consequences of a failed rebuild PR.

jeancochrane:
Looks like this wasn't quite ready for production yet. Some failures we noticed after merging:

> Looks like dbt is too dumb to clean up old, non-HA stuff i.f.f. the tables have partitions.

This means we have to delete the old prod data after switching over to the new scheme. Not very ergonomic, but not necessarily a deal killer.

> dbt doesn't know/check if the stored data underlying a seed has been deleted, it just checks the state manifest to see if the seed data has changed. That means if you delete the seed data in S3, all the queries downstream of the seed will still run (to the extent they even can with a 0 row table).

We need to test if this behavior is specific to seeds, or to all models. Then, see if we can do anything about it.

> I'm not sure exactly how the HA works, but it doesn't seem to be materializing the Parquet files in the prod bucket. Weirdly, there are delete markers at the paths to the tables. It seems like all the tables that solely use Athena/SQL work fine, but building with Spark which depends on the explicit existence of the files in S3 seems to fail.

Here's the error:

```
File "<stdin>", line 387, in <module>
  File "<stdin>", line 239, in model
  File "<stdin>", line 384, in <lambda>
  File "<stdin>", line 292, in ref
  File "<stdin>", line 378, in get_spark_df
  File "/opt/amazon/spark/python/lib/pyspark.zip/pyspark/sql/session.py", line 700, in table
    return DataFrame(self._jsparkSession.table(tableName), self._sc, self._jconf)
  File "/opt/amazon/spark/python/lib/py4j-0.10.9.3-src.zip/py4j/java_gateway.py", line 1321, in __call__
    return_value = get_return_value(
[...]

PR #481, opened 2024-05-30
Switch dev and staging dbt targets to dedicated staging Spark workgroup

jeancochrane:
I noticed during the process of investigating problems with https://github.com/ccao-data/data-architecture/issues/460 that our dbt targets all share the same workgroup for Athena PySpark. This isn't ideal from a permissions perspective, since it means that  dev and staging targets can access a role that has CRUD permissions over prod tables and views. This PR changes our profiles so that our dev and staging targets point to a dedicated staging Spark workgroup, `primary-spark-staging`, that only has CRUD permissions over dev/staging tables and views.

The PR also sneaks in one minor unrelated change: Recently as I've been debugging our builds more extensively, I've been finding myself wishing there was a way that we could manually dispatch a run that rebuilds the entire DAG. As far as I can tell, there is no built-in [selector](https://docs.getdbt.com/reference/node-selection/syntax) pattern to do this, but I could be wrong. In any case, this PR also adds a feature whereby a single asterisk (`*`) in the `models` input will indicate to the `build-and-test-dbt` workflow that it should rebuild the entire DAG.

### DIFF
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -66,8 +66,12 @@ jobs:
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
-              echo "Running build on manually selected resources"
-              dbt build -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR" --resource-types model seed
+              echo "Running build on manually selected resources: $SELECTED_RESOURCES"
+              if [ "$SELECTED_RESOURCES" == "*" ]; then
+                dbt build -t "$TARGET" --resource-types model seed
+              else
+                dbt build -t "$TARGET" -s $SELECTED_RESOURCES --defer --state "$STATE_DIR" --resource-types model seed
+              fi
             else
               echo "Running build on modified/new resources only"
               dbt build -t "$TARGET" -s state:modified state:new --defer --state "$STATE_DIR" --resource-types model seed
@@ -78,6 +82,8 @@ jobs:
           fi
         working-directory: ${{ env.PROJECT_DIR }}
         shell: bash
+        env:
+          SELECTED_RESOURCES: ${{ inputs.models }}
 
       - name: Test models
         run: |
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -66,12 +66,8 @@ jobs:
         run: |
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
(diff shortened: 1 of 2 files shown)


dbt/profiles.yml
```diff
@@ -22,7 +22,7 @@ athena:
       region_name: us-east-1
       schema: z_static_unused_dbt_stub_database
       database: awsdatacatalog
-      spark_work_group: primary-spark
+      spark_work_group: primary-spark-staging
```
jeancochrane, line 25: See [here](https://github.com/ccao-data/data-architecture/actions/runs/9306870644/job/25616957485#step:8:55) for a workflow run demonstrating that Python models still work using this workgroup.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -66,8 +66,12 @@ jobs:
...
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
-              echo "Running build on manually selected resources"
-              dbt build -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR" --resource-types model seed
+              echo "Running build on manually selected resources: $SELECTED_RESOURCES"
+              if [ "$SELECTED_RESOURCES" == "*" ]; then
+                dbt build -t "$TARGET" --resource-types model seed
+              else
+                dbt build -t "$TARGET" -s "$SELECTED_RESOURCES" --defer --state "$STATE_DIR" --resource-types model seed
+              fi
```
jeancochrane, line 74: See [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/9320458427/job/25657288808#step:8:41) for an example of this conditional working when `models == "*"` (canceled after confirming that 87 models had been selected, for the sake of time) and see [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/9320501875/job/25657425019#step:8:41) for a regression test demonstrating that it still works when `models == "default.vw_pin_sale default.vw_pin_address"`.

One semi-related thought I had while developing this: [I accidentally ran this workflow on `master` with `models == "*"`](https://github.com/ccao-data/data-architecture/actions/runs/9320431951) and was lucky enough to catch my mistake and cancel it before causing mass downtime. Should we gate the `*` operator so that it can't run when the target is `prod`?

dfsnow: [Supposedly](https://docs.getdbt.com/reference/node-selection/syntax#specifying-resources) just using `ccao_data_athena` as the selector will build everything. I don't think it will be necessary to gate once we merge #462.

dfsnow: **suggestion (blocking):** And to clarify, I don't see the need for a dedicated `*` selector when this is the case, so let's remove it unless I'm missing something.

jeancochrane: Makes sense, reverted in 76a88d1838a96642897832f0ca02cea57c8f12ae.

dfsnow (approved):
Good to go!

PR #486, opened 2024-06-03
Test out high availability on `model.final_model`

jeancochrane:
This PR updates the `model.final_model` table to use [high availability](https://github.com/dbt-athena/dbt-athena?tab=readme-ov-file#highly-available-table-ha). By limiting high availability to one model, we should make it easier and safer to debug the problems we encountered after merging https://github.com/ccao-data/data-architecture/pull/462.


dfsnow (approved):
This looks good to go as soon as @ccao-jardine gives the go ahead.

PR #487, opened 2024-06-03
Configure high availability for `location` schema

jeancochrane:
Testing out more high availability tables following a successful experiment in https://github.com/ccao-data/data-architecture/pull/486.

We also bump the number of threads used by dbt for issuing queries to 10.


dfsnow (approved):
Great job 👍

PR #490, opened 2024-06-03
Revert all HA changes

dfsnow:
This PR reverts all the changes from #488, #487, and #486 that we made for testing dbt-athena high availability.


jeancochrane (approved):
Good to go!

PR #491, opened 2024-06-04
[Do not merge] Re-enable high availability tables

jeancochrane:
This PR acts as a reversion of https://github.com/ccao-data/data-architecture/pull/490, mostly for the debugging purposes. If we can get to a place where we're very confident in a fix, I'll plan to merge this into the main branch; otherwise, I'll close this once we have a root cause analysis of the failures we're seeing on prod when deploying high availability tables.

### DIFF
--- a/dbt/dbt_project.yml
+++ b/dbt/dbt_project.yml
@@ -43,6 +43,7 @@ models:
     +materialized: view
     +write_compression: zstd
     +format: parquet
+    +ha: true
     census:
       +schema: census
     default:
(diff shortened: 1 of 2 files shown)


jeancochrane:
I wasn't able to reproduce the prod behavior in spite of trying a bunch of changes to the staging configs. My next step is to do an after hours debugging session where I recreate the prod DAG locally using a `dbt --log-level debug run --target prod` call to see if I can find any clues in the debug logs.

PR #494, opened 2024-06-04
Add empty `packages.yml` as cache key

### LINKED ISSUE #489: Update dbt cache key
The Python setup/install is [currently missing its cache](https://github.com/ccao-data/data-architecture/actions/runs/9358554102/job/25760611085#step:3:266) because we deleted the `packages.yml` file (which it hashes) in a previous PR.

dfsnow:
Quick PR to close #489. GitHub Actions isn't properly caching dbt packages because it expects the `packages.yml` file to exist as a key. Since we aren't currently using any packages, this file does not exist, and the hash function used for the key returns an empty string. Just replacing this with an empty file should fix the problem.

### DIFF
--- a/dbt/packages.yml
+++ b/dbt/packages.yml
@@ -1,0 +1,3 @@
+# This file is hashed by our GitHub workflows for use as a key in the dbt deps
+# cache. The file is empty since we currently have no deps, but we need it
+# here so that we don't break the caching step.


dbt/packages.yml
jeancochrane, line 1: **[Suggestion, non-blocking]** How about a brief comment in this file explaining why it exists so we don't accidentally remove it in the future?

```yaml
# This file is hashed by our GitHub workflows for use as a key in the dbt deps
# cache. The file is empty since we currently have no deps, but we need it
# here so that we don't break the caching step.
```

jeancochrane (approved):
Thanks for picking this up!

dbt/packages.yml
dfsnow: Done dae700c!

PR #498, opened 2024-06-07
Bump number of threads used by all dbt targets

dfsnow:
The dbt [threads](https://docs.getdbt.com/docs/running-a-dbt-project/using-threads) argument specifies the maximum number of models that can be simultaneously built without breaking the dependency graph. We had it set to 5 for reasons I can't quite remember, but Athena can handle many more simultaneous queries (DDL statements, in this case). This PR bumps the maximum number of threads to 16. This should result in slightly faster full DAG builds.


jeancochrane (approved):
Let's do it!

PR #503, opened 2024-06-12
Sales val migration q2 2024 ias world spec

wagnerlmichael:
This PR updates `vw_pin_sale` and `vw_ias_salesval_upload` with the new column structure from the [sales val migration](https://github.com/ccao-data/model-sales-val/pull/130/files). We also edit schema and shared_columns files accordingly.

Note: I will need to change the refs back to the prod table at some point after we do the migration.

I am getting a sql fluff error on the pre-commit check that I'm not entirely sure is working correctly, curious about your thoughts @jeancochrane. 

Eyeballing these queries confirm that things are not totally out-of-whack:

```sql
SELECT sv_is_outlier, COUNT(*)
FROM "z_ci_sales_val_migration_q2_2024_ias_design_doc_default"."vw_pin_sale"
WHERE sv_run_id is not null
GROUP BY sv_is_outlier
ORDER BY COUNT(*) DESC;
```

```sql
SELECT sv_outlier_reason1, COUNT(*)
FROM "z_ci_sales_val_migration_q2_2024_ias_design_doc_default"."vw_pin_sale"
WHERE sv_run_id is not null
GROUP BY sv_outlier_reason1
ORDER BY COUNT(*) DESC;
```

```sql
SELECT sv_outlier_reason2, COUNT(*)
FROM "z_ci_sales_val_migration_q2_2024_ias_design_doc_default"."vw_pin_sale"
WHERE sv_run_id is not null
GROUP BY sv_outlier_reason2
ORDER BY COUNT(*) DESC;
```

### DIFF
--- a/dbt/models/default/default.vw_pin_sale.sql
+++ b/dbt/models/default/default.vw_pin_sale.sql
@@ -248,7 +248,7 @@ sales_val AS (
         sf.sv_outlier_reason3,
         sf.run_id AS sv_run_id,
         sf.version AS sv_version
-    FROM 
+    FROM
         {{ source('sale', 'flag') }}
             AS sf
     INNER JOIN max_version_flag AS mv
--- a/dbt/models/default/default.vw_pin_sale.sql
+++ b/dbt/models/default/default.vw_pin_sale.sql
@@ -243,7 +243,9 @@ sales_val AS (
         sf.sv_is_outlier,
         sf.sv_is_ptax_outlier,
         sf.sv_is_heuristic_outlier,
-        sf.sv_outlier_type,
+        sf.sv_outlier_reason1,
+        sf.sv_outlier_reason2,
+        sf.sv_outlier_reason3,
         sf.run_id AS sv_run_id,
         sf.version AS sv_version
     FROM {{ source('sale', 'flag') }} AS sf
@@ -325,7 +327,9 @@ SELECT
     sales_val.sv_is_outlier,
     sales_val.sv_is_ptax_outlier,
     sales_val.sv_is_heuristic_outlier,
-    sales_val.sv_outlier_type,
+    sales_val.sv_outlier_reason1,
+    sales_val.sv_outlier_reason2,
+    sales_val.sv_outlier_reason3,
     sales_val.sv_run_id,
     sales_val.sv_version
 FROM unique_sales
--- a/dbt/models/default/default.vw_pin_sale.sql
+++ b/dbt/models/default/default.vw_pin_sale.sql
@@ -233,8 +233,7 @@ max_version_flag AS (
     SELECT
         meta_sale_document_num,
         MAX(version) AS max_version
-    FROM
-        "z_ci_0002-update-outlier-column-structure-w-iasworld-2024-update_sale"."new_prod_data"
+    FROM {{ source('sale', 'flag') }}
     GROUP BY meta_sale_document_num
 ),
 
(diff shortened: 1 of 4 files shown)


jeancochrane:
Taking a look now! Two things before I dig in:

> I am getting a sql fluff error on the pre-commit check that I'm not entirely sure is working correctly, curious about your thoughts

[This looks reasonable to me](https://github.com/ccao-data/data-architecture/actions/runs/9503582101/job/26194321086?pr=503#step:3:31518), I think the issue is just that sqlfluff doesn't think the quotes around `new_prod_data` are necessary because it doesn't contain special characters. If you run `pre-commit run --all-files` then sqlfluff should fix it for you, but let me know if that doesn't work and I can help troubleshoot.

Also: What do you think is up with these sales that are outliers with no reasons?

```sql
SELECT sv_is_outlier, sv_outlier_reason1, sv_outlier_reason2, sv_outlier_reason3, COUNT(*)
FROM "z_ci_sales_val_migration_q2_2024_ias_design_doc_default"."vw_pin_sale"
WHERE sv_run_id IS NOT NULL
AND sv_is_outlier = TRUE
AND sv_outlier_reason1 IS NULL
AND sv_outlier_reason2 IS NULL
AND sv_outlier_reason3 IS NULL
GROUP BY 1, 2, 3, 4
ORDER BY COUNT(*) DESC;
```

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -233,7 +233,8 @@ max_version_flag AS (
     SELECT
         meta_sale_document_num,
         MAX(version) AS max_version
-    FROM {{ source('sale', 'flag') }}
+    FROM
+        "z_ci_0002-update-outlier-column-structure-w-iasworld-2024-update_sale"."new_prod_data"
```
jeancochrane, line 237: **[Nitpick]** Note to revert these `FROM` clauses before merging.

dbt/models/sale/schema.yml
```diff
@@ -25,8 +25,12 @@ sources:
...
           - name: sv_is_outlier
             description: '{{ doc("shared_column_sv_is_outlier") }}'
-          - name: sv_outlier_type
-            description: '{{ doc("shared_column_sv_outlier_type") }}'
+          - name: sv_outlier_reason1
+            description: '{{ doc("shared_column_sv_outlier_reason1") }}'
+          - name: sv_outlier_reason1
+            description: '{{ doc("shared_column_sv_outlier_reason2") }}'
+          - name: sv_outlier_reason1
+            description: '{{ doc("shared_column_sv_outlier_reason3") }}'
```
jeancochrane, line 33: **[Nitpick, non-blocking]** Instead of documenting each numbered reason separately, let's just have them share one doc block:

```suggestion
          - name: sv_outlier_reason1
            description: '{{ doc("shared_column_sv_outlier_reason") }}'
          - name: sv_outlier_reason1
            description: '{{ doc("shared_column_sv_outlier_reason") }}'
          - name: sv_outlier_reason1
            description: '{{ doc("shared_column_sv_outlier_reason") }}'
```

dbt/models/shared_columns.md
```diff
@@ -1300,9 +1301,27 @@ with `sv_is_ptax_outlier` (using OR logic).
...
 
-## sv_outlier_type
+## sv_outlier_reason1
+
+{% docs shared_column_sv_outlier_reason1 %}
+Heuristic or model used to flag an outlier.
+
+See the [model-sales-val](https://github.com/ccao-data/model-sales-val) repo
+for a list of possible flags.
+{% enddocs %}
```
jeancochrane, line 1311: **[Nitpick, non-blocking]** Same here, let's consolidate these into one `shared_column_sv_outlier_reason` (i.e. just rename `shared_column_sv_outlier_type` to `shared_column_sv_outlier_reason`).

jeancochrane (approved):
Looking great! See my question above about one particular curious combination of reasons, but otherwise this is looking good to go.

PR #482, opened 2024-05-31
Add a proximity to stadium feature

Damonamajor:
The built in proximity query that we had been using (proximity to cemetery, etc) kept producing the following error: 

> AMBIGUOUS_NAME: line 89:25: Column 'xy.x_3435' is ambiguous. You may need to manually clean the data at location 's3://ccao-dbt-athena-dev-us-east-1/results/tables/47cac154-5955-4ed2-873e-55a3d15ea8ab' before retrying. Athena will not delete data in your account.

My assumption is because the test is querying itself. I tested renaming the columns as well as removing the pins which we are trying to join to (NOT IN), but the error remained the same. Because of that, I built a SQL query on it's own which should work just fine. I did a decent amount of QCing with it. Let me know if I should go back to work with the "standard" structure.

PR #504, opened 2024-06-12
Add spatial.stadium.csv

Damonamajor:
Quick fix to add a seed for spatial.stadium to help with [distance to stadiums](https://github.com/ccao-data/data-architecture/pull/482)


dfsnow (changes requested):
@Damonamajor This needs to be a CSV file to work as a seed, not a .xlsx. You'd also need to change much of the code in #482 to reference and document the new seed, as you would a dbt model.

PR #507, opened 2024-06-13
Expand lines in `new-dbt-model` issue template for friendlier formatting

jeancochrane:
I noticed while showing off updates to the `new-dbt-model` issue template that the line breaks that we use in the Markdown file to make editing easier get reproduced in the final issue, where they are not automatically expanded out into full lines. This makes the final issue look a bit weird and hard to read. See [a draft issue](https://github.com/ccao-data/data-architecture/issues/new?template=new-dbt-model.md&title=Add+a+new+dbt+model) for an example (select the Preview tab to see how the issue gets rendered).

This PR expands out the lines in that issue template, which makes editing slightly more annoying but significantly improves the reading experience of the final issue. It also aligns the formatting of `new-dbt-model` with our other issue template, [`data-onboarding`](https://github.com/ccao-data/data-architecture/issues/new?template=data-onboarding.md).

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-dbt-model.md
+++ b/.github/ISSUE_TEMPLATE/new-dbt-model.md
@@ -14,54 +14,32 @@ _(Brief description of the task here.)_
 
 ## Model attributes
 
-* **Name**: _(What should the model be called? See [Model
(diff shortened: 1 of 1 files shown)


dfsnow (approved):
Woops! Good catch.

PR #509, opened 2024-06-13
Add MV to `reporting.vw_pin_value_long`

wrridgeway:
Quick PR to add market value columns from `asmt_all` to the long version of `vw_pin_value`.

### DIFF
--- a/dbt/models/default/schema/default.vw_pin_condo_char.yml
+++ b/dbt/models/default/schema/default.vw_pin_condo_char.yml
@@ -104,7 +104,7 @@ models:
             (is_common_area OR is_parking_space)
             AND (char_bedrooms IS NOT NULL OR char_full_baths IS NOT NULL)
           config:
-            error_if: ">48205" # as of 2024-06-03
+            error_if: ">48206" # as of 2024-06-14
       # TODO: Non-liveable unit heuristics
       # TODO: Row count matches PARDAT for condo classes
       # TODO: Non-unit and unit parcel count add up to total parcels per
--- a/dbt/models/reporting/docs.md
+++ b/dbt/models/reporting/docs.md
@@ -82,9 +82,6 @@ views.
 {% docs view_vw_pin_value_long %}
 Assessed and market values by PIN and year, for each assessment stage.
 
-- Market value (`_mv`) columns accurately reflect incentives, statute,
-  levels of assessment, building splits, etc.
-
 The assessment stages are:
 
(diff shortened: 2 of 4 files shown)


dbt/models/reporting/docs.md
```diff
@@ -80,7 +80,10 @@ views.
 # vw_pin_value_long
 
 {% docs view_vw_pin_value_long %}
-Assessed values by PIN and year, for each assessment stage.
+Assessed and market values by PIN and year, for each assessment stage.
+
+- Market value (`_mv`) columns accurately reflect incentives, statute,
+  levels of assessment, building splits, etc.
```
dfsnow, line 86: **issue:** This is an *assumption* and is listed in the other table as such. It should be listed as an assumption here as well.

dfsnow, line 86: **issue:** This is an *assumption* and is listed in the other table as such. It should be listed as an assumption here as well.

dbt/models/reporting/schema.yml
```diff
@@ -226,6 +226,13 @@ models:
           config:
             where: CAST(year AS int) < {{ var('test_qc_year_start') }}
             error_if: ">21" # as of 2024-04-01
+      - not_null:
+          name: reporting_vw_pin_value_long_tot_mv_not_null
+          column_name: tot_mv
+          config:
+            where: |
+              CAST(year AS int) < {{ var('test_qc_year_start') }} AND
+              year >= '2021'
```
dfsnow, line 235: **question (non-blocking):** We have the test for things >= 2021 but the values constructed for >= 2020, why the difference?

wrridgeway: don't we only have board values in 2020?

I can update this to address that.

dbt/models/default/schema/default.vw_pin_condo_char.yml
```diff
@@ -104,7 +104,7 @@ models:
             (is_common_area OR is_parking_space)
             AND (char_bedrooms IS NOT NULL OR char_full_baths IS NOT NULL)
           config:
-            error_if: ">48205" # as of 2024-06-03
+            error_if: ">48206" # as of 2024-06-14
```
wrridgeway, line 107: This feels like it's going to be an ongoing issue, now. I'll take another look next week...

dfsnow: See #515.

dfsnow (approved):

PR #510, opened 2024-06-13
Add a proximity to stadium feature v2

Damonamajor:
This provides an update to the original stadium feature, utilizing a seed (rather than manually constructed locations) and the dist_to_nearest_geography macro.


dbt/seeds/spatial/spatial.stadium.csv
```diff
@@ -0,0 +1,8 @@
+name,lon,lat,x_3435,y_3435,year,date_opened,year
+Soldier Field,-87.6165891,41.862322,1179603,1893122,2022,1924
```
Damonamajor, line 2: year is a required column in the dist_pin_to_geography macro. Having it as 2024 was not resulting in any joins.

dbt/models/proximity/proximity.dist_pin_to_stadium.sql
```diff
@@ -1,4 +1,4 @@
--- CTAS to create a table of distance to the stadiums for each PIN
+-- CTAS to create a table of distance to the nearest Metra stop for each PIN
```
dfsnow, line 1: **issue (blocking):**

```suggestion
-- CTAS to create a table of distance to the nearest stadium each PIN
```

dbt/seeds/spatial/spatial.stadium.csv
dfsnow, line 1: **issue (blocking):** The tables in the `spatial` database all follow a specific, standard format. They have the following columns:

- Some kind of ID column
- Metadata related to the ID (name, other info)
- `geometry` (binary 4326 lat/lon geometry)
- `geometry_3435` (binary 3435 geometry in feet)
- `year` (string, partition key)

We want to replicate this format using the dbt seed, but it would be pretty unwieldy since you'd need to save raw geometry binary data in the seed file. Instead let's what we did for the `model.final_model` seed with this case:

1. Rename this seed to `spatial.stadium_raw`.
2. Create a separate dbt model `spatial.stadium` that uses the seed as an input and results in the columns above. This would involve using the Athena geometry functions to properly construct the geometry from points (as you've done in the `dist_pin_to_stadium` model above.
3. Update downstream references to `spatial.stadium` to point to the new table and geometry columns.

dbt/seeds/spatial/spatial.stadium.csv
dfsnow: That's because the macro looks for that year of parcel data, and 2024 doesn't yet exist. You can actually just merge `date_opened` and `year`, keeping the open year for each stadium (in a column called `year`). That should fill appropriately.

dfsnow:
@Damonamajor This is also missing the other dbt model updates necessary to bring this feature all the way through the pipeline (e.g. the ones present in #512).

Damonamajor:
> @Damonamajor This is also missing the other dbt model updates necessary to bring this feature all the way through the pipeline (e.g. the ones present in #512).

These were pulled in through [482](https://github.com/ccao-data/data-architecture/pull/482).

dfsnow (approved):
Looks great @Damonamajor! Excited to have this added.

PR #514, opened 2024-06-14
Bump nonlivable chars test threshold

jeancochrane:
The test `default_vw_pin_condo_char_nonlivable_no_chars` occupies a bit of an awkward middle ground between a unit test and a data test. I've opened an issue to refactor it such that failures are more actionable (https://github.com/ccao-data/data-architecture/issues/515), but in the meantime, it's causing the full DAG build to fail ([example here](https://github.com/ccao-data/data-architecture/actions/runs/9507765142/job/26207844499)) so this PR bumps the threshold as a temporary workaround.

Evidence that this new threshold causes the test to warn instead of error in [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/9518288879/job/26238850535?pr=514#step:9:44).

### DIFF
--- a/dbt/models/default/schema/default.vw_pin_condo_char.yml
+++ b/dbt/models/default/schema/default.vw_pin_condo_char.yml
@@ -104,7 +104,7 @@ models:
             (is_common_area OR is_parking_space)
             AND (char_bedrooms IS NOT NULL OR char_full_baths IS NOT NULL)
           config:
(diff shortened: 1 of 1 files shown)


dfsnow (approved):
Thanks @jeancochrane. I'll kick off a full DAG rebuild tonight just to get the cache updated. I think this can be safely merged before that.

PR #520, opened 2024-06-21
Update dbt-core to 1.8.3

jeancochrane:
This PR updates `dbt-core` to [the latest release](https://github.com/dbt-labs/dbt-core/releases/tag/v1.8.3), which contains some minor security fixes to the docs JS dependencies.

I don't think anything in the changelog should affect us, but to be safe I ran a quick `build-and-test-model` run on `default.vw_card_res_char` and confirmed that things still generally work as expected ([see output here](https://github.com/ccao-data/data-architecture/actions/runs/9616775127/job/26527052111)).

### DIFF
--- a/dbt/requirements.txt
+++ b/dbt/requirements.txt
@@ -1,2 +1,2 @@
 dbt-athena-community==1.8.2
-dbt-core==1.8.1
+dbt-core==1.8.3


dfsnow (approved):
Thanks!

Issue #518, opened 2024-06-17
Spike using local Dagster for EL and orchestration


dfsnow:
While doing some research into potential software for this budget season, I came across [dagster](https://docs.dagster.io/getting-started/quickstart). It's an orchestration platform similar to Airflow that uses a declarative Python model similar to dbt. It integrates with dbt natively and also shares dbt's open-core licensing. IMO, this could be a great potential fit for replacing our ad-hoc Python/R ETL scripts. If we ever get the time, I think it would be worthwhile to do a local spike of dagster using the CLI.

PR #522, opened 2024-06-27
Add new model features to capture parcel shape

dfsnow:
This PR updates the parcel file ETL to add 6 new features for modeling. These features attempt to capture the complexity of the parcel shape, based on the idea that more "regular" shaped parcels are more desirable and are thus worth more, they were mostly taken from [this paper](https://eres.architexturez.net/system/files/P_20230206101430_7873.pdf). The 6 features are:

1. The standard deviation of the interior angles of the parcel polygon
2. The standard deviation of the parcel polygon edge length
3. The standard deviation of the parcel vertices' distance to the parcel centroid
4. The total number of vertices in the parcel polygon
5. The ratio of the length of the sides of the minimum spanning rotated rectangle
6. The ratio of the areas of the minimum spanning rotated rectangle and the parcel polygon itself

I haven't actually run this script yet to replace any cleaned up prod parcel files, but I'll do so once this is merged.

I'll make a follow-up PR to add the output features to dbt and all the necessary views/docs.

### DIFF
--- a/etl/etl.Rproj
+++ b/etl/etl.Rproj
@@ -12,6 +12,6 @@ Encoding: UTF-8
 RnwWeave: Sweave
 LaTeX: pdfLaTeX
 
-StripTrailingWhitespace: Yes
 AutoAppendNewline: Yes
+StripTrailingWhitespace: Yes
 LineEndingConversion: Posix
--- a/etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
@@ -153,7 +153,7 @@ process_parcel_file <- function(s3_bucket_uri,
     tictoc::toc()
 
     # Read attribute data and get unique attributes by PIN10
-    tictoc::tic(paste("Joined and wrote parquet for:", file_year))
+    tictoc::tic(paste("Merged boundary and attribute data:", file_year))
     attr_df <- read_parquet(local_attr_file) %>%
       mutate(pin10 = str_sub(pin, 1, 10)) %>%
       group_by(pin10) %>%
@@ -178,9 +178,11 @@ process_parcel_file <- function(s3_bucket_uri,
         town_code, year
       ) %>%
       distinct(pin10, .keep_all = TRUE)
+    tictoc::toc()
 
     # If centroids are missing from join (invalid geom, empty, etc.)
     # fill them in with centroid of the full multipolygon
+    tictoc::tic(paste("Repaired centroids:", file_year))
     if (any(
       is.na(spatial_df_merged$lon) |
         any(is.na(spatial_df_merged$x_3435))
@@ -202,7 +204,9 @@ process_parcel_file <- function(s3_bucket_uri,
         filter(!is.na(lon) & !is.na(x_3435)) %>%
         bind_rows(spatial_df_missing)
     }
+    tictoc::toc()
 
+    tictoc::tic(paste("Calculated shape features:", file_year))
     # Convert to a data.table of the vertices of each polygon. Using data.table
     # here because it's much faster
     spatial_mat_coords <- spatial_df_merged %>%
@@ -295,7 +299,9 @@ process_parcel_file <- function(s3_bucket_uri,
       ),
       by = "L2"
     ]
+    tictoc::toc()
 
+    tictoc::tic(paste("Finalized and wrote to file:", file_year))
     # Merge all calculated features back into the main dataframe
     # then sort by year, town code, and PIN10 for better compression
     spatial_df_final <- spatial_df_merged %>%
--- a/etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
@@ -280,8 +280,11 @@ process_parcel_file <- function(s3_bucket_uri,
         dist_to_centroid = sqrt(
           (X - mean(X)) ^ 2 +
             (Y - mean(Y)) ^ 2
-        )
+        ),
+        # Recalculate the angle after removing straight angles
+        angle = calculate_angles(as.matrix(.SD))
       ),
+      .SDcols = c("X", "Y"),
       by = c("L1", "L2", "L3")
     ]
 
--- a/etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
@@ -48,6 +48,25 @@ save_local_parcel_files <- function(year, spatial_uri, attr_uri) {
 }
 
 
+# Function to calculate the interior angles of a polygon given its X and Y
+# coordinates using the directions of the vectors between each pair of points
+# See: https://stackoverflow.com/questions/12083480/finding-internal-angles-of-polygon
+# Good example parcel for angle calc: 1418307019
+calculate_angles <- function(points) {
+  vectors <- diff(rbind(points, points[2, ])) * -1
+
+  cross_product <- vectors[-nrow(vectors), 1] * vectors[-1, 2] -
+    vectors[-1, 1] * vectors[-nrow(vectors), 2]
+  dot_product <- vectors[-nrow(vectors), 1] * vectors[-1, 1] +
+    vectors[-nrow(vectors), 2] * vectors[-1, 2]
+
+  angle <- pi + atan2(cross_product, dot_product)
+  angles <- c(NA_real_, angle * 180 / pi)
+
+  return(angles)
+}
+
+
 # Load local parcel file, clean, extract centroids, and write to partitioned
 # dataset on S3
 process_parcel_file <- function(s3_bucket_uri,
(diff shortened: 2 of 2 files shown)


etl/etl.Rproj
```diff
@@ -12,6 +12,6 @@ Encoding: UTF-8
 RnwWeave: Sweave
 LaTeX: pdfLaTeX
 
-StripTrailingWhitespace: Yes
 AutoAppendNewline: Yes
+StripTrailingWhitespace: Yes
```
dfsnow, line 16: This was auto-formatted by RStudio, not an actual change.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -174,29 +194,137 @@ process_parcel_file <- function(s3_bucket_uri,
...
+        # to the last row (again, same reason as above)
+        shp_parcel_interior_angle_sd = sd(tail(angle, -1), na.rm = TRUE),
+        shp_parcel_centroid_dist_ft_sd = sd(tail(dist_to_centroid, -1))
+      ),
+      by = "L3"
+    ]
+
+    # Collapse the edge-level data back to the parcel level, this time keeping
+    # the 180 degree vertices, since excluding them would drop some edges with
+    # non-zero length
```
jeancochrane, line 247: **[Question, non-blocking]** Would it be slightly more accurate to sum the length of any edges that are joined with angle ~180° when executing operations based on edge length, since the additional vertex is most likely a data error cutting a single edge in two? I don't feel super strongly about this (especially if the universe of 180° vertices is small) but I figure it's worth checking!

jeancochrane (approved):
Very nice work! I don't feel super qualified to evaluate the math or `data.tables` code in this PR, so let's get Billy's thoughts on this too, but overall it makes sense to me. There are a few linter warnings that you can see if you navigate to the PR file browser, might be nice to clean those up before merging but not required.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -173,28 +210,159 @@ process_parcel_file <- function(s3_bucket_uri,
...
+      as.data.table()
+
+    # Use the coordinate vertices to calculate additional features about
+    # the shape of the polygon
+    spatial_mat_calc <- spatial_mat_coords[
+      ,
+      # Calculate the interior angle between each pair of points
+      angle := calculate_angles(as.matrix(.SD)),
+      .SDcols = c("X", "Y"),
+      by = c("L1", "L2", "L3")
```
wrridgeway, line 233: I'm having a hard time parsing what L1, L2, and L3 are?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -122,40 +148,67 @@ process_parcel_file <- function(s3_bucket_uri,
       # For each PIN10, keep the centroid of the largest polygon
       group_by(pin10) %>%
       arrange(desc(area)) %>%
-      summarize(across(c(lon, lat, x_3435, y_3435), first)) %>%
+      summarize(across(c(lon, lat, x_3435, y_3435), dplyr::first)) %>%
```
dfsnow, line 151: Namespacing here just to prevent conflicts with `data.table`, which also has a `first()` function.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -122,40 +148,67 @@ process_parcel_file <- function(s3_bucket_uri,
...
       left_join(spatial_df_centroids, by = "pin10") %>%
       mutate(
         has_attributes = !is.na(town_code),
         geometry = st_transform(geometry, 4326),
-        geometry_3435 = st_transform(geometry, 3435)
+        area = units::drop_units(st_area(geometry))
       ) %>%
+      # Drop anything with an area less than 1 sq meter, as long as it has other
+      # polygons for the same PIN10 that are larger
+      filter(!(area <= 1 & n() >= 2 & !all(area <= 1)), .by = "pin10") %>%
```
dfsnow, line 177: There were ~200 parcels that had "sliver" polygons that were messing up the feature results. Here we drop these slivers i.f.f. they belong to the same PIN10 as a larger polygon. This cleans up the feature results significantly.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -122,40 +148,67 @@ process_parcel_file <- function(s3_bucket_uri,
...
       select(
         pin10, tax_code, nbhd_code, has_attributes,
-        lon, lat, x_3435, y_3435, geometry, geometry_3435,
+        lon, lat, x_3435, y_3435, geometry,
         town_code, year
-      ) %>%
-      distinct(pin10, .keep_all = TRUE)
```
dfsnow, line 154: This was actually a mistake/bug, so I'm glad I reviewed the rest of the parcel cleaning code. This `distinct()` drops an arbitrary polygon in cases where there are multiple polygons per PIN10, definitely not what we want. It wouldn't have actually impacted any model results or other processes, since the centroids (used for modeling) are (correctly) calculated separate from this block of code.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -122,40 +148,67 @@ process_parcel_file <- function(s3_bucket_uri,
...
+          across(geometry, st_union),
+          .by = "pin10"
+        ),
+      spatial_df_merged %>%
+        filter(!(n() >= 2), .by = pin10)
+    ) %>%
+      st_cast("MULTIPOLYGON") %>%
+      mutate(geometry_3435 = st_transform(geometry, 3435)) %>%
+      relocate(geometry_3435, .after = geometry)
+    tictoc::toc()
```
dfsnow, line 203: Here we aggregate to the PIN10 level using a geometry union instead of `distinct()`. This also improves the feature results quite a bit, as they now include all polygons.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -173,28 +226,159 @@ process_parcel_file <- function(s3_bucket_uri,
...
+
+    tictoc::tic(paste("Calculated shape features:", file_year))
+    # Simplify the planar geometry to reduce noise in shape-based features
+    # calculated below. Convert to a data.table of the vertices of each polygon.
+    # Using data.table here because it's much faster than dplyr
+    spatial_mat_coords <- spatial_df_merged %>%
+      st_set_geometry("geometry_3435") %>%
+      st_simplify(dTolerance = 2, preserveTopology = TRUE) %>%
+      st_cast("MULTIPOLYGON") %>%
```
dfsnow, line 238: Added a simplification step with a low, low tolerance (2 feet) to remove points that are basically overlapping.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
dfsnow: They are one of [the outputs](https://r-spatial.github.io/sf/reference/st_coordinates.html#value) from `st_coordinates()`. Basically, L3 corresponds to the PIN10 group while L1 and L2 can represent different sub-polygons of a multi-polygon.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -173,28 +226,162 @@ process_parcel_file <- function(s3_bucket_uri,
...
+          data.table::between(angle, 0, 1) |
+            data.table::between(angle, 359, 360)
+        ) | data.table::between(angle, 179, 181)
+      )
+    ][
+      ,
+      num_non_straight_angle := sum(!straight_angle, na.rm = TRUE),
+      by = c("L1", "L2", "L3")
+    ][
+      !straight_angle | num_non_straight_angle < 3 | is.na(angle),
```
dfsnow, line 269: Per @jeancochrane's prior comment, this calculates the interior vertex angle and removes straight angles _before_ calculating the rest of the features (edge length, distance to centroid, etc.). This yields way more accurate feature results, nearly all the PINs with a high angle SD are wacky things like this:

f71be41a-18b1-4129-b980-a84ec9997a92

dfsnow:
@jeancochrane @wrridgeway I'm requesting re-review on this as I just made some pretty significant changes. Mainly:

- Fixed an error that was dropping arbitrary parts of multi-polygon parcels
- Added a simplification step to remove overlapping points
- Added a filtering step to remove polygon slivers and lines
- Improved the feature calculation by first removing vertices with straight angles

All these changes should not only give us more accurate shape features, they should also give us a more accurate/cleaner parcel shapefile.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -173,28 +226,162 @@ process_parcel_file <- function(s3_bucket_uri,
...
+      st_drop_geometry() %>%
+      st_set_geometry("geometry_3435") %>%
+      mutate(parcel_area = st_area(geometry_3435)) %>%
+      st_minimum_rotated_rectangle() %>%
+      mutate(
+        shp_parcel_mrr_area_ratio = units::drop_units(
+          st_area(geometry_3435) / parcel_area
+        ),
+        # Sometimes the ratio here is very slightly less than 1, even though
+        # that should be impossible. Seems to an artifact of some malformed
```
jeancochrane, line 319: **[Question, non-blocking]** Not a huge deal, but the last sentence is a fragment and I'm wondering what you think might be malformed here?

jeancochrane (approved):
Nice work!

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
dfsnow: It seems to only be the result when calculating the ratio for these super long, skinny triangles, though I'm really not sure why. The minimum rotated rectangle is in red below. It definitely _should_ have an area larger than the triangle, but it is in fact smaller by 0.00657 ft^2. I don't think it's worth debugging further, given the magnitude of the error. Will update my comment to this effect.

2754c101-1d09-48ad-9c59-4bd5f8bfe59c

dfsnow:
@jeancochrane @wrridgeway Okay, last time requesting review on this. I had to make a bunch of edits to get the new features to even run (mostly in 16fcb07). The major changes are:

- Run an `st_make_valid()` call _before_ filtering out invalid parcels. This keeps some parcels that were getting dropped before, resulting in a larger parcel count per year than previously (see below).
- Simplify the geometry processing/transformations so that we're not transforming back and forth between 3435 and 4326. This is not only faster, it also preserves some parcels which are made invalid by the transformation.
- Add a check at the end of the script to ensure that the final count of unique PIN10s is very close to the initial count from the raw parcel file. They don't match exact because some parcels have empty geometries and are dropped.

There's a temporary `spatial.parcel_new` table in Athena if you all want to poke around it. Here's the yearly row counts between the existing table and the new one.

```sql
SELECT *
FROM (
    SELECT year, COUNT(*) AS cnt_old
    FROM spatial.parcel
    GROUP BY year
)
LEFT JOIN (
    SELECT year, COUNT(*) AS cnt_new
    FROM spatial.parcel_new
    GROUP BY year
) USING (year)
ORDER BY year DESC
```

| year | cnt_old | cnt_new | diff |
|------|---------|---------|------|
| 2023 | 1417110 | 1417110 | 0    |
| 2022 | 1416347 | 1416420 | 73   |
| 2021 | 1415410 | 1416038 | 628  |
| 2020 | 1415664 | 1415729 | 65   |
| 2019 | 1415193 | 1415409 | 216  |
| 2018 | 1414731 | 1414794 | 63   |
| 2017 | 1414896 | 1414954 | 58   |
| 2016 | 1414449 | 1414506 | 57   |
| 2015 | 1414155 | 1414227 | 72   |
| 2014 | 1413788 | 1413862 | 74   |
| 2013 | 1413759 | 1413832 | 73   |
| 2012 | 1413732 | 1413809 | 77   |
| 2011 | 1413698 | 1413775 | 77   |
| 2010 | 1413699 | 1413777 | 78   |
| 2009 | 1413740 | 1413816 | 76   |
| 2008 | 1413206 | 1413280 | 74   |
| 2007 | 1410605 | 1410677 | 72   |
| 2006 | 1408738 | 1408808 | 70   |
| 2005 | 1405047 | 1405112 | 65   |
| 2004 | 1401506 | 1401570 | 64   |
| 2003 | 1398092 | 1398094 | 2    |
| 2002 | 1395387 | 1395386 | -1   |
| 2001 | 1393174 | 1393174 | 0    |
| 2000 | 1391620 | 1391620 | 0    |

And here's an example of the type of invalid parcel we were previously dropping:

Screenshot 2024-07-08 at 12 04 28 PM

After this is merged we should re-run all the dbt models downstream of `spatial.parcel`, just to be safe.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -173,28 +222,181 @@ process_parcel_file <- function(s3_bucket_uri,
...
+      st_drop_geometry() %>%
+      summarize(n = n_distinct(pin10)) %>%
+      pull(n)
+    num_parcel_final <- spatial_df_final %>%
+      st_drop_geometry() %>%
+      summarize(n = n_distinct(pin10)) %>%
+      pull(n)
+    if (abs(num_parcel_raw - num_parcel_final) > 10) {
+      stop("More than 10 PINs dropped during geometry processing!")
+    }
```
jeancochrane, line 381: **[Thought, non-blocking]** Good thinking adding this test! It makes me wonder more broadly about how we should define QC tests on non-dbt transformation scripts like this one. Maybe running tests like this at the end of the script is the best solution? Or maybe we split tests out into a function defined in a different module, and then import and call that function at the end of this script for better separation of concerns? Either way, I support the current solution in the immediate term; this is just a thought to keep in mind if we find ourselves adding more QC tests like this one to non-dbt transformations.

jeancochrane (approved):
Really impressive and subtle work in this PR. Thanks for the attention to detail!

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
dfsnow: Let's leave it for now, since this script run rarely enough that I don't think investing lots of time into them is a good idea. However, this is definitely something to think about when we look at #518.

wrridgeway:
Rad work dude. I didn't get around to checking the math, but everything else looked just great.

Issue #524, opened 2024-07-05
Update `transform_dbt_test_results.py` script to optionally filter by township


jeancochrane:
A common workflow for working with the `transform_dbt_test_results.py` script is to output a workbook representing test failures in a specific township. This ends up requiring a fair amount of manual intervention to the workbook itself, since we need to A) filter each sheet by township and then B) hide any columns that are not used by failing tests in that township.

To improve this workflow, we should add a `--township` flag to the script so that we can automate the operation to filter results by township. It would probably also be worthwhile to refactor the script to use [click](https://click.palletsprojects.com/en/8.1.x/) for command line argument parsing, since now we'll have three arguments with defaults and it would be more ergonomic to change them to optional keyword arguments.

PR #527, opened 2024-07-08
Update `transform_dbt_test_results.py` script to allow filtering by township

jeancochrane:
This PR updates the `.github/scripts/transform_dbt_test_results.py` script to automate a common workflow: Filtering results by township. As of this PR, callers can now pass in a `--township` option with one or more numeric codes indicating a township that should be used to filter results. If the option is missing, the script will query for all townships, including rows with null townships (which usually indicates an invalid PIN).

As part of this change, we also refactor the script to use the built-in argparse library for more convenient CLI UX. This means that the arguments that were now positional arguments have been refactored to be options named `--run-results`, `--manifest`, and `--output-dir`, respectively. This should have the nice side effect of making it easier for CLI callers to modify only a selection of arguments, since changing any optional positional argument requires also specifying the positional arguments that precede it. We also rearrange docstrings to make use of argparse's built-in documentation using the `--help` flag.

I tested the new township filter by running dbt tests and then outputting three versions of the workbook: One unfiltered, and two filtered for different townships. I'm happy to share these over OneDrive if it would be useful for a reviewer to QC them.

```
dbt test --selector qc_tests --store-failures
python3 ../.github/scripts/transform_dbt_test_results.py --output-dir qc_test_results_2024-07-09_unfiltered
python3 ../.github/scripts/transform_dbt_test_results.py --township 70 --output-dir qc_test_results_2024-07-09_hyde_park
python3 ../.github/scripts/transform_dbt_test_results.py --township 70 72 --output-dir qc_test_results_2024-07-09_hyde_park_lake
```

Closes https://github.com/ccao-data/data-architecture/issues/524.


.github/scripts/transform_dbt_test_results.py
```diff
@@ -4,64 +4,10 @@
 # other teams for review and correction, along with metadata parquet files
 # that can be uploaded to S3 for long-term result tracking.
 #
-# This script assumes that it is being run in sequence after a call to
-# `dbt test --store-failures`, since it depends on two files created by
-# that operation (target/run_results.json and target/manifest.json).
-# It also requires Python dependencies be installed from
-# requirements.transform_dbt_test_results.txt.
```
jeancochrane, line 11: All of this text was preserved, just moved into the CLI documentation for more idiomatic access.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -676,28 +621,119 @@ def _filter_for_existing_field_indexes(
...
+
+Outputs three files to the directory specified by the `--output-dir` flag:
+
+  1. `qc_test_failures_<date>.xlsx`: Excel workbook to share with other teams
+  2. `metadata/test_run/run_year=YYYY/*.parquet`: Metadata about this run, partitioned by year of run and prepped for upload to S3
+  3. `metadata/test_run_result/run_year=YYYY/*.parquet`: Metadata about test results (pass, fail, number of failing rows, etc.) in this run,
+     partitioned by year of run and prepped for upload to S3
+
+Each sheet in the output workbook represents a category of test, e.g. "valid_range" or "not_null"; each row in a sheet represents a row in
+a database that failed a test, with enough metadata that a reader can figure out what conditions caused the test to fail and investigate the root cause."""  # noqa: E501
```
jeancochrane, line 651: The lines here are slightly longer than the typical flake8 limit of 80 characters for better terminal UX.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -676,28 +621,119 @@ def _filter_for_existing_field_indexes(
...
+        help=(
+            "The directory to which output artifacts should be written; "
+            "if the directory does not exist, it will be created. Defaults to "
+            "'./qc_test_results_<date>/'."
+        ),
+    )
+    parser.add_argument(
+        "--township",
+        required=False,
+        nargs="*",
```
jeancochrane, line 715: Setting this option instructs argparse to accept one or more values for this option and combine them into a tuple.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -1129,36 +1171,57 @@ def get_test_categories(
...
-            )
+                if townships:
+                    # Missing rows are most likely due to the township filter,
+                    # so print a warning and skip
+                    print(msg + ", possibly due to the township filter")
+                    test_result = TestResult(
+                        status=Status.PASS,
+                        failing_rows=[],
+                        **base_result_kwargs,
+                    )
```
jeancochrane, line 1202: This block only gets executed if a test failed, but none of its failing rows are in the township(s) specified by the `--township` option. In this case, we override the results to mimic the case in which the tests passed.

Note that there is technically a chance that this block could execute if a `--township` filter is present but there is some other unexpected problem with the results metadata. In this case, I expect that some piece of logic further down the call stack will raise a different error, and my hope is that the "possibly" modifier in the warning log line will be enough to flag for an end user that a problem occurred around the time of this block's execution,

dfsnow (approved):
Looks good @jeancochrane!

PR #530, opened 2024-07-10
Move nonlivable condos with characteristics test to qc database

wrridgeway:


jeancochrane (approved):
Nice work! I like this a lot better than the test.

PR #532, opened 2024-07-11
Add a dbt test: check AHSAP AVs

wrridgeway:
We get requests on a fairly regular basis for AHSAP parcels that have had had large swings in values between stages. This PR standardizes that request into a reproducible report.


dbt/models/qc/schema.yml
```diff
@@ -1,4 +1,7 @@
 models:
+  - name: qc.vw_change_in_ahsap_values
+    description: '{{ doc("view_vw_change_in_ahsap_values") }}'
+
```
wrridgeway, line 4: Not sure if this should get tagged with anything in particular, it's kind of its own thing

jeancochrane: **[Praise]** I think it's fine without a tag! I'm thinking that the tags will be useful in cases where we want to output multiple reports at once (e.g. on town close) but I don't think we need them in one-off cases like these.

jeancochrane (approved):
Very nice!

PR #533, opened 2024-07-11
Add QC test for exempt parcels that have AV

wrridgeway:


dbt/models/reporting/schema.yml
```diff
@@ -235,6 +235,15 @@ models:
               (year >= '2021' OR (
               year = '2020' AND stage_name = 'BOR CERTIFIED')
               )
+      - is_null:
+          name: reporting_vw_pin_value_long_exempt_parcels_values
+          column_name: pin
+          config:
+            tags:
+              - test_qc_iasworld
```
wrridgeway, line 243: Not sure iasworld is the correct tag, but we're really only worried about the underlying data here.

wrridgeway:
Thanks Jean, much appreciate all the input.

wrridgeway:
@jeancochrane what does automated joining to pardat, etc. for township involve? If I alias pin to parid and year to taxyr, is that enough? This view doesn't have township itself.

dbt/models/reporting/schema.yml
```diff
@@ -235,6 +235,24 @@ models:
...
+            class = 'EX' AND CAST(year AS int) = {{ var('test_qc_year_start') }}
+          additional_select_columns:
+            - {column: pin, alias: parid}
+            - class
+            - stage_name
+            - {column: year, alias: taxyr}
+          meta:
+            table_name: asmt_all
+            description: exempt parcels should not have AVs greater than 1
+            TEST_CATEGORIES: test_accepted_range
```
jeancochrane, line 255: **[Nitpick, non-blocking]** I don't think we need to override the category in this case, since `accepted_range` already has a good default in `TEST_CATEGORIES`:

https://github.com/ccao-data/data-architecture/blob/4b1b86885058fc2fae5e623afe82b2f6ed856cf4/.github/scripts/transform_dbt_test_results.py#L56

But if we do decide we need to override the category, we should use the `category` attribute instead of `TEST_CATEGORIES` -- `TEST_CATEGORIES` is the name of the default variable dict in the script, but it reads from `meta.category` for overrides:

https://github.com/ccao-data/data-architecture/blob/4b1b86885058fc2fae5e623afe82b2f6ed856cf4/.github/scripts/transform_dbt_test_results.py#L1252-L1256

```suggestion
            category: test_accepted_range
```

dbt/models/reporting/schema.yml
```diff
@@ -235,6 +235,24 @@ models:
...
               )
+      - accepted_range:
+          name: reporting_vw_pin_value_long_exempt_parcels_values
+          column_name: tot
+          config:
+            tags:
+              - test_qc_iasworld
+          max_value: 1
+          where: |
+            class = 'EX' AND CAST(year AS int) = {{ var('test_qc_year_start') }}
```
jeancochrane, line 246: **[Nitpick, non-blocking]** I think this is reasonable and will probably work, but we should probably align it with our other test filters by setting a maximum for the year:

https://github.com/ccao-data/data-architecture/blob/4b1b86885058fc2fae5e623afe82b2f6ed856cf4/dbt/models/iasworld/schema/iasworld.oby.yml#L40

```suggestion
          where: |
            class = 'EX' AND CAST(year AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
```

I honestly can't quite remember why we designed it this way... perhaps there are rows that have a taxyr incorrectly set to the future, or we're being defensive in case a rollover happens before January 1st? Dan will have more context!

jeancochrane (approved):
Awesome work! A couple tiny nits but nothing serious.

dbt/models/reporting/schema.yml
wrridgeway: i'll reopen this if we decide the time window needs to be adjusted. my thought was that reporting on data that can't be fixed would just be a distraction.

PR #534, opened 2024-07-11
Add configs and script for generating QC reports

jeancochrane:
This PR adds a new script `dbt/scripts/export_models.py` that can be used to export models to Excel files. The PR uses this as a basis for building a config system for generating QC reports, starting with the `Negative ASMT values` report that Valuations requests on town close.

Note that the PR does not add extensive documentation for defining QC reports; I think we should hold off on that until the system is more stable.

You can test this PR by checking out the branch and running the `Negative ASMT values` report:

```
python3 dbt/scripts/export_models.py --select tag:qc_report_town_close
```

This will output a report to `export/output/Negative ASMT values.xlsx` that you can inspect.


dbt/dbt_project.yml
```diff
@@ -36,6 +36,9 @@ vars:
   # frames for testing
   test_qc_year_end: 2030
 
+  # Township code to use for filtering town close QC reports
+  qc_report_town_code: "71"  # Lake Township
```
jeancochrane, line 40: I figure we can use the `qc_report_*` prefix to mark symbols that are used for reports, and if we wind up with multiple reports that need different town codes we can modify this variable to make it more specific.

Note that setting a variable isn't the most ergonomic way of accomplishing this goal (filtering reports by town) but I think it's the fastest path forward given dbt's design. The upside is that dbt exposes a built-in interface for variable overrides that we can use during report generation (via the `--vars` flag); the downside is that since the value is baked into the view, we have to rebuild any views that rely on this variable if we want to override it during export. See the script below for some details about how we handle this using the `--vars` and `--rebuild` flags. This is a little bit annoying, but I think it should be fine assuming A) none of our QC views take an obscene amount of time to execute and B) we continue following the historical pattern of generating QC reports once per town on town close.

An alternative might be to keep the underlying view unfiltered by township, and implement a `--township` option on the script that performs the filtering via query (similar to the way the `transform_dbt_test_results.py` script works). This would simplify the underlying view and make it more flexible, but in so doing would move more complexity into the export script and would tightly couple the export script to the notion of township filters, which may not apply to all exports that we end up running.  I feel somewhat agnostic on this direction but happy to move towards it if you feel strongly! I'll note that this alternative is probably more closely aligned with a potential future where reports are self-service and run frequently on demand, since in that future we might open ourselves up to race conditions if we're rebuilding models every time a use requests a report filtered by a township.

dbt/models/qc/schema.yml
```diff
@@ -609,6 +609,14 @@ models:
                 description: valasm3 (ASMT TOTAL) should not be negative
               <<: *test-non-negative
 
+  - name: qc.vw_report_neg_asmt_value
+    description: '{{ doc("view_vw_report_neg_asmt_value") }}'
+    config:
+      tags:
+        - qc_report_town_close
+    meta:
+      export_name: Negative ASMT values
```
jeancochrane, line 618: This config attribute will be used by the script to determine the name of the report file.

dbt/scripts/export_models.py
```diff
@@ -0,0 +1,212 @@
+# Export dbt models to Excel files.
```
jeancochrane, line 1: Note that this script is defined in a new `dbt/scripts/` subdirectory, rather than in `.github/scripts/` where `transform_dbt_test_results.py` is defined. I'm going to put up a PR soon to move `transform_dbt_test_results.py` into this directory since I've found that the location has caused confusion for folks who want to run it manually, which ends up happening fairly often.

dbt/scripts/export_models.py
```diff
@@ -0,0 +1,212 @@
...
+
+        writer_kwargs = (
+            {"mode": "a", "if_sheet_exists": "overlay"}
+            if template_exists
+            else {}
+        )
+        with pd.ExcelWriter(
+            output_path, engine="openpyxl", **writer_kwargs
+        ) as writer:
+            sheet_name = "Sheet1"
```
jeancochrane, line 185: In the future we might want to consider supporting a config value to override sheet names, but for now it's easiest to use the default value.

jeancochrane:
Oops, I forgot to add a requirements file before requesting review! Done in 1f11436cb37c41c0cb9cdcf4601899f1c3ce923b.

dfsnow (approved):
Looks great @jeancochrane! This will be very handy when sending people data, even stuff not related to QC work.

dbt/models/qc/docs.md
```diff
@@ -29,11 +29,15 @@ Pulled from the following Inquire queries:
 # vw_neg_asmt_value
 
 {% docs view_vw_neg_asmt_value %}
-Test that all `ASMT.val*` columns are not negative.
+Pull `ASMT.val*` columns to support tests confirming they are not negative.
```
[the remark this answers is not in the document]
jeancochrane: Yup, that's right. A little bit confusing, but it's necessary in order to meet Tia's request to keep the QC test workbooks exactly the same while we work on these exports. I'm hoping that as we start building out more QC infra we can push folks toward deduplication.

dbt/dbt_project.yml
jeancochrane: I added a `--where` filter in bce0bfe27f8c2d57d674d7a6b1744341639835bf and I think I like it better! Let me know what you think, happy to modify the design if you have a better idea.

jeancochrane:
This should be ready for another look @dfsnow!

dbt/export/templates/qc.vw_report_neg_asmt_value.xlsx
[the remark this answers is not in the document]
dfsnow: No I think it's fine! Let's keep it in.

dfsnow (approved):
Looks good @jeancochrane!

Issue #280, opened 2024-01-03
Improve the dbt documenation for `ccao.corner_lot`


wrridgeway:
Add a more in-depth description of the corner lots algorithm to the table description (`dbt\models\ccao\docs.md`) and corner lot indicator column description (`dbt\models\shared_columns.md`).

dfsnow:
Pinging @Damonamajor on this. Good thing to knock out if you ever have down time.

dfsnow:
Closed by #536.

Issue #274, opened 2023-12-22
Update corner lot data output and script locations


wrridgeway:
Currently our corner lot script lives in the aws-s3/raw `data-architecture` folder, though it doesn't collect and upload any raw data. It should probably be moved to the warehouse folder, and should also be renamed from a `spatial` prefix to `ccao` since it doesn't upload any spatial data.  I've included some rough code to help upload the data in its current form to the warehouse.

```
library(arrow)
library(DBI)
library(noctua)
library(readr)
source("utils.R")

# This script uploads corner lots to the S3 warehouse
AWS_S3_WAREHOUSE_BUCKET <- Sys.getenv("AWS_S3_WAREHOUSE_BUCKET")
output_bucket <- file.path(AWS_S3_WAREHOUSE_BUCKET, "ccao", "other", "corner_lot")

# Optionally enable query caching
noctua_options(cache_size = 10)

# Establish connection
AWS_ATHENA_CONN_NOCTUA <- dbConnect(noctua::athena())

# Grab townships for pin10 from iasWorld
townships <- dbGetQuery(
  conn = AWS_ATHENA_CONN_NOCTUA, "
  SELECT distinct
    substr(parid, 1, 10) AS pin10,
    user1 as township
  FROM iasworld.legdat
  WHERE user1 IS NOT NULL
    AND cur = 'Y'
    AND deactivat IS NULL
  ")

read_delim("Corners Final.csv") %>%
  summarise(
    is_corner_lot = as.logical(max(fnl_rsl, na.rm = TRUE)),
    .by = pin10
    ) %>%
  left_join(townships, by = "pin10") %>%
  group_by(township) %>%
  write_partitions_to_s3(output_bucket, is_spatial = FALSE, overwrite = TRUE)

```

wrridgeway:
Additionally:
- We should move from using one year's parcel shapefile to gathering all distinct pin10s across time
- Address non-contiguous parcels explicitly rather than while uploading

PR #536, opened 2024-07-11
Update and simplify corner lots generation

### LINKED ISSUE #1: Create corner lot indicator
## Overview

Assumption: corner lots tend to be more valuable than non-corner lots due to their size, increased accessibility (for commercial), and prominent location.

Given this assumption, we should add a corner lot indicator variable to all AVMs, as well as to commercial valuation spreadsheets/methods. We can impute a corner lot variable using a variety of GIS methods, including:

- Examining the nodal complexity of the street network
- Touching parcel boundaries at specific angles
- Lot size compared to neighbors

We should use a combination of methods to output a single indicator (0 or 1) for each PIN in the county. N at Mansueto and J at CSDS will likely have good ideas on this.

## Task

Create an R script that takes property locations (lat, lon) or shapes (parcel file) as an input, then outputs a corner lot indicator (1 or 0) for each PIN. Start with a prototype script using local data. The final script should pull raw data from S3 and write back to S3. It should run as a glue job. The script should be saved to `was-glue/jobs/location-corners.R`. Use the branch attached to this issue to save your work and submit it for review.

@miaoqi, figuring out how to identify these lots algorithmically will be a large part of this task. There is existing literature in the space in various geography and spatial data science journals. I can also put you in touch with folks at UChicago who can provide help if needed. Let's aim to have a running script by Feb 15th.

Some R libraries you'll probably need to get started:

- `sf`, for all spatial operations
- `igraph`, for any graph manipulation i.e. street networks
- `osmdata`, for OpenStreetMap data i.e. roads

dfsnow:
This PR vastly simplifies the code and logic we use to detect corner parcels. It keeps the same general algorithm, but makes much heavier use of vectorized functions in order to speed things up and simplify the logic. The result is significantly faster (all parcels for a year run in ~30 minutes) and uses much less memory (about 2% of the original).

Closes #1, #274, and #280.


dbt/models/default/default.vw_pin_universe.sql
```diff
@@ -50,7 +50,7 @@ SELECT
     sp.y_3435,
 
     -- Corner lot indicator
-    lot.is_corner_lot AS ccao_is_corner_lot,
+    COALESCE(lot.is_corner_lot, FALSE) AS ccao_is_corner_lot,
```
dfsnow, line 53: The `COALESCE` here is needed because the `spatial.corner` table now _only_ contains corners, i.e. it does not contain all PIN10 records. So, anything not in `spatial.corner` is assumed to be not-a-corner.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
dfsnow, line 1: It's much easier to review the scripts separately rather than try to parse the diff here, so I'm glad git decided they're different scripts.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
```diff
@@ -0,0 +1,237 @@
...
+  distinct(year) %>%
+  collect() %>%
+  pull(year)
+
+# Drop years before 2014, since OSM data for roads is spotty prior to that
+parcel_years <- parcel_years[parcel_years >= 2014]
+
+# Iterate over the years and townships, saving the results to a temporary file
+# after each iteration
+for (iter_year in parcel_years) {
```
dfsnow, line 31: This new script runs by year and township, meaning we'll get corners for all years after and including 2014. It takes about 30 minutes per year to run.

dfsnow:
Once _all_ the corners are built for all years the dbt builds will succeed, but that may take a day or two, and the code should be fine for review in the meantime.

etl/renv.lock
dfsnow, line 1: Managed to remove a ton of dependencies and simplify things quite a bit, hence the LoC decrease here.

dbt/models/spatial/docs.md
```diff
@@ -106,6 +106,39 @@ Cook County coordinated care area boundaries.
...
+   rectangle to determine the parcel's orientation and length. These values
+   are used in the next step to draw a cross on the parcel.
+2. Draw a cross that cuts the parcel in four and extends out slightly
+   beyond the minimum rectangle of the parcel. We will look for parcels and
+   streets that intersect this cross in order to determine neighbors. Use the
+   bearing and length calculated in the previous step to draw the cross (i.e
+   the long side of the cross is parallel to the minimum bounding rectangle).
+3. Find the arms of each cross which intersect neighboring parcels
+   which _also_ touch the cross-originating PIN. The idea here is that cross
+   cross arms that intersect neighboring (touching) parcels are likely to be
```
jeancochrane, line 127: **[Nitpick, non-blocking]** Tiny dupe here:

```suggestion
   which _also_ touch the cross-originating PIN. The idea here is that
   cross arms that intersect neighboring (touching) parcels are likely to be
```

dbt/models/spatial/docs.md
```diff
@@ -106,6 +106,39 @@ Cook County coordinated care area boundaries.
...
+1. Find the minimum rectangle that bounds the parcel, then use that
+   rectangle to determine the parcel's orientation and length. These values
+   are used in the next step to draw a cross on the parcel.
+2. Draw a cross that cuts the parcel in four and extends out slightly
+   beyond the minimum rectangle of the parcel. We will look for parcels and
+   streets that intersect this cross in order to determine neighbors. Use the
+   bearing and length calculated in the previous step to draw the cross (i.e
+   the long side of the cross is parallel to the minimum bounding rectangle).
+3. Find the arms of each cross which intersect neighboring parcels
+   which _also_ touch the cross-originating PIN. The idea here is that cross
```
jeancochrane, line 126: **[Question, non-blocking]** I don't totally understand what we get out of the condition that arms must touch the cross-originating PIN in order to be removed -- shouldn't we expect all cross arms to touch the cross-originating PIN, since that's where they originate from?

dbt/models/spatial/docs.md
```diff
@@ -106,6 +106,39 @@ Cook County coordinated care area boundaries.
 **Geometry:** `MULTIPOLYGON`
 {% enddocs %}
 
+# corner
+
+{% docs table_corner %}
+CCAO corner lot indicator. Determined algorithmically by unobstructed access to
+perpendicular streets.
+
+The algorithm is:
```
jeancochrane, line 115: **[Praise]** These docs are great!

etl/renv.lock
jeancochrane: **[Question, non-blocking]** Nice! Am I right that this lockfile covers all the R scripts in the `etl/` directory, or is it just for corner lots? If it's the former, I wonder if at some point in the future it'd be worth building in a little bit more dependency isolation for different scripts? Definitely not high priority and not something we should do soon, just curious what you think as the person who most recently touched these dependencies.

dbt/models/spatial/schema.yml
```diff
@@ -39,6 +39,31 @@ sources:
       - name: coordinated_care
         description: '{{ doc("table_coordinated_care") }}'
 
+      - name: corner
+        description: '{{ doc("table_corner") }}'
+        columns:
+          - name: pin10
+            description: '{{ doc("shared_column_pin10") }}'
```
jeancochrane, line 46: **[Question, non-blocking]** I would expect us to also have a `year` column on this table, are we missing docs for that or is it excluded for a reason?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
```diff
@@ -0,0 +1,236 @@
...
+    town_parcels_buf <- parcels %>%
+      filter(
+        between(lon, town_bbox$xmin, town_bbox$xmax) &
+          between(lat, town_bbox$ymin, town_bbox$ymax)
+      ) %>%
+      st_buffer(dist = units::set_units(2, "m"))
+
+    # Fetch the OSM street network for the township, removing any OSM way types
+    # that are not main roads
+    osm_streets <- opq(bbox = town_bbox) %>%
```
jeancochrane, line 74: **[Question, non-blocking]** I'm guessing you made this choice intentionally, but is it faster to issue a separate Overpass query for every bounding box vs. once for the whole county? Or is the difference negligible?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
```diff
@@ -0,0 +1,236 @@
...
+      mutate(
+        length = st_distance(geometry, lead(geometry), by_element = TRUE) +
+          units::set_units(40, "m"),
+        bearing = c(
+          lwgeom::st_geod_azimuth(geometry),
+          units::set_units(NA, "radians")
+        ),
+        bearing = units::set_units(bearing, "degrees")
+      ) %>%
+      filter(row_number() %% 5 != 0) %>%
```
jeancochrane, line 105: **[Question, non-blocking]** What's going on here?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
```diff
@@ -0,0 +1,236 @@
...
+    # Draw the lines that connect the centroids to the endpoints of the cross
+    draw_line <- function(r) st_linestring(t(matrix(unlist(r), 2, 2)))
+    cross$geometry <- st_sfc(sapply(
+      seq_len(nrow(cross)),
+      function(i) draw_line(cross[i, ]),
+      simplify = FALSE
+    ))
+
+    # Compute the length and aspect ratio of the cross. The aspect ratio will be
+    # useful for filtering out parcels that are dramatically longer than wide
```
jeancochrane, line 136: **[Question, non-blocking]** Am I right that we're no longer computing or using aspect ratio here?

jeancochrane (approved):
Looks fantastic! The new code is super clear and I'm excited about the performance gains.

One quick question related to deployment: Do we need to clean up the old version of the corner lots data as part of the deployment process for this PR, or is that already done?

dbt/models/spatial/docs.md
dfsnow: This is confusingly worded, but it's not the _arms_ that need to touch the cross-originating PINs, but the _neighboring PINs._ I clarified the language in 6c7a4c4f.

dbt/models/spatial/docs.md
dfsnow: Fixed in 6c7a4c4f!

dbt/models/spatial/schema.yml
dfsnow: Fixed in 6c7a4c4f!

etl/renv.lock
dfsnow: It is indeed the former. We should definitely address it at some point, but not a high priority. Perhaps when we get to #518.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
dfsnow: It indeed seems to be slightly faster to query by township than for the whole county.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
dfsnow: Added a clarifying comment in 6c7a4c4f!

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-ccao-corner.R
dfsnow: Correct! Removed in 6c7a4c4f65b1c729a20575d19633d9c1bc00adaa.

PR #544, opened 2024-07-15
 Adjust exempt parcels with non-zero AV test

wrridgeway:
Update https://github.com/ccao-data/data-architecture/pull/533 with [additional suggestion](https://github.com/ccao-data/data-architecture/pull/533#discussion_r1678306192).


jeancochrane (approved):
Thanks for this cleanup!

PR #546, opened 2024-07-16
Persist row-level test results in `transform_dbt_test_results` script

jeancochrane:
This PR updates the `.github/scripts/transform_dbt_test_results.py` script to output row-level test failure data to a parquet dataset with the name `test_run_failing_row`. This is intended to supplement the `test_run` dataset (which saves metadata about a specific run) and the `test_run_result` dataset (which saves aggregate statistics for tests at the township level).

In so doing, we also update `test_run_result` to include an nullable `column_name` attribute that matches the new `test_run_failing_row` dataset. This attribute has proven useful for the purposes of analysis, and saving it explicitly should be more robust than the current workflow where we extract the column name from the test name (which follows the pattern `iasworld_<tablename>_<column_name>_description` in cases where a column name exists for the test).

Sample output is available for review in Athena under the table name `qc.test_run_failing_row`.

### DIFF
--- a/.github/scripts/transform_dbt_test_results.py
+++ b/.github/scripts/transform_dbt_test_results.py
@@ -1160,17 +1160,14 @@ def get_test_categories(
     run_results.json file dict and a manifest.json file dict) and an optional
     township filter, generates a list of TestCategory objects storing the
     results of the tests."""
-    conn = pyathena.connect(
+    # Define a cursor with unload=True to output query results as parquet.
+    # This is particularly important when selecting aggregated columns,
+    # which are deserialized incorrectly by regular cursors
+    cursor = pyathena.connect(
         s3_staging_dir=AWS_ATHENA_S3_STAGING_DIR,
         region_name="us-east-1",
-    )
-    # Define a regular cursor for simple queries
-    dict_cursor = conn.cursor(pyathena.cursor.DictCursor)
-    # Define a cursor with unload=True which we'll use to get more detailed
-    # column information for more complex queries. This is particularly
-    # important when requesting aggregated columns, which otherwise are
-    # deserialized incorrectly by the DictCursor
-    arrow_cursor = conn.cursor(pyathena.arrow.cursor.ArrowCursor, unload=True)
+        cursor_class=pyathena.arrow.cursor.ArrowCursor,
+    ).cursor(unload=True)
 
     tests_by_category: typing.Dict[str, TestCategory] = {}
 
@@ -1226,9 +1223,9 @@ def get_test_categories(
             relation_name_unquoted = test_results_relation_name.replace(
                 '"', "`"
             )
-            dict_cursor.execute(f"show columns in {relation_name_unquoted}")
+            cursor.execute(f"show columns in {relation_name_unquoted}")
             # SHOW COLUMNS often returns field names with trailing whitespace
-            fieldnames = [row["field"].strip() for row in dict_cursor]
+            fieldnames = [row[0].strip() for row in cursor]
 
             # Construct the query to select the test results that were stored
             # in Athena. We want to rehydrate a few fields like township_code
@@ -1331,8 +1328,8 @@ def get_test_categories(
             )
             # Use the cursor with unload=True in this query, since otherwise
             # aggregated columns can be deserialized incorrectly
-            arrow_cursor.execute(test_results_query)
-            query_results = arrow_cursor.as_arrow().to_pylist()
+            cursor.execute(test_results_query)
+            query_results = cursor.as_arrow().to_pylist()
             if len(query_results) == 0:
                 msg = (
                     f"Test {test_name} has status {status!r} but no failing "
--- a/.github/scripts/transform_dbt_test_results.py
+++ b/.github/scripts/transform_dbt_test_results.py
@@ -27,6 +27,7 @@ import openpyxl.utils
 import pyarrow as pa
 import pyarrow.parquet
 import pyathena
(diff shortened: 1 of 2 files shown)


.github/scripts/transform_dbt_test_results.py
```diff
@@ -94,10 +96,11 @@ def __init__(
         """
         The failing_rows list should be formatted like the rows
         returned by a csv.DictReader or a DictCursor, i.e. a list of
-        dicts mapping `{column_name: row_value}`.
+        dicts mapping `{field_name: row_value}`.
```
jeancochrane, line 99: This tweak is intended to avoid a misunderstanding now that `column_name` is a parameter of the function.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -207,14 +213,14 @@ class TestCategory:
     results for output to a workbook and saving them to a cache."""
 
     # Names of fields that are used for debugging
-    _debugging_fieldnames = [TEST_NAME_FIELD, DOCS_URL_FIELD]
+    possible_debugging_fieldnames = [TEST_NAME_FIELD, DOCS_URL_FIELD]
```
jeancochrane, line 216: Now that we're referencing these attributes outside the context of the class (see below), they shouldn't be marked as private anymore.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -882,15 +901,20 @@ def create(
             git_author=git_author,
         )
 
+    def to_dict(self) -> typing.Dict:
+        return dataclasses.asdict(self)
```
jeancochrane, line 905: We need to extend `dataclasses.asdict()` with extra serialization steps for `TestRunFailingRowMetadata` (see below) so this stub function is intended to unify the interface across `Metadata` classes. I considered a class mixin to reduce some of the duplication between `TestRunMetadata` and `TestRunResultMetadata`, both of which implement this function in the same way, but I figure that unless/until we extend the `Metadata` interface further it's probably not worth the extra complexity that a mixin class incurs.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         self,
         name: str,
         table_name: str,
+        column_name: typing.Optional[str],
```
jeancochrane, line 90: Note that the `column_name` attribute of a dbt manifest node can be null in cases where the test was defined on a model rather than on a column (e.g. table uniqueness tests).

.github/scripts/transform_dbt_test_results.py
```diff
@@ -929,6 +954,104 @@ def create_list(
...
+    description: str
+    # Fields that identify the failing row. Some of these can occasionally
+    # be arrays for tests that query multiple rows (e.g. uniqueness tests)
+    # so for consistency we set them to always be arrays, even when there
+    # is only one value
+    parid: typing.Optional[str]
+    taxyr: typing.Optional[str]
+    card: typing.Optional[typing.List[int]]
+    lline: typing.Optional[typing.List[int]]
+    class_: typing.Optional[typing.List[str]]
```
jeancochrane, line 983: I don't love that we need to compromise by making these columns arrays, since I find array columns to be annoying to deal with, but it strikes me as the best way to reconcile the two different approaches to these columns.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -929,6 +954,104 @@ def create_list(
...
+
+    def to_dict(self) -> typing.Dict:
+        output_data = dataclasses.asdict(self)
+        # Serialize the "class" attribute to a more human-friendly name
+        output_data["class"] = output_data.pop("class_")
+        # Dump the problematic fields to string, since parquet can't handle
+        # the notion of an untyped JSON object
+        output_data["problematic_fields"] = json.dumps(
+            output_data["problematic_fields"]
+        )
```
jeancochrane, line 1052: I also don't love that this field needs to be serialized to a string, but as far as I know parquet doesn't support untyped JSON columns.

.github/scripts/transform_dbt_test_results.py
```diff
@@ -1037,9 +1163,14 @@ def get_test_categories(
...
-        cursor_class=pyathena.cursor.DictCursor,
     )
-    cursor = conn.cursor()
+    # Define a regular cursor for simple queries
+    dict_cursor = conn.cursor(pyathena.cursor.DictCursor)
+    # Define a cursor with unload=True which we'll use to get more detailed
+    # column information for more complex queries. This is particularly
+    # important when requesting aggregated columns, which otherwise are
+    # deserialized incorrectly by the DictCursor
+    arrow_cursor = conn.cursor(pyathena.arrow.cursor.ArrowCursor, unload=True)
```
dfsnow, line 1173: **question (non-blocking):** Why not just use the `arrow_cursor` for everything?

dfsnow (approved):
Nice work @jeancochrane! This is super clean and well done. Appreciate the clear comments.

.github/scripts/transform_dbt_test_results.py
jeancochrane: Good point! This was due to a slight preference I have for the `DictCursor` interface but I don't think that's worth the extra complexity of maintaining two cursors. I switched everything over to `ArrowCursor` in 94eb40115cb25ba310077d3f79f34167c6d23cd5 and confirmed it works the same by running on a subset of tests.

Issue #537, opened 2024-07-11
Strip out old references to class dict that predate the seed


jeancochrane:
In https://github.com/ccao-data/data-architecture/pull/397 we changed up the definition of `ccao.class_dict` to be controlled via a [seed](https://github.com/ccao-data/data-architecture/blob/master/dbt/seeds/ccao/ccao.class_dict.csv), but we didn't strip out old references to the prior definition that was based on a [warehouse ETL script](https://github.com/ccao-data/data-architecture/blob/a77b78d335f577397694d8f8476118776454f138/etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-dictionary.R#L18). There are a few cleanup steps here:

- [x] Remove [the `ccao-dictionary` script](https://github.com/ccao-data/data-architecture/blob/master/etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-dictionary.R) since it seems to no longer be used
- [x] Remove [the old `class_dict` source definition](https://github.com/ccao-data/data-architecture/blob/a77b78d335f577397694d8f8476118776454f138/dbt/models/ccao/schema.yml#L6-L10)
- [x] Change remaining references from `source("ccao", "class_dict")` to `ref("ccao.class_dict")` in a few dbt tests
- [x] Do a search of the code for `class_dict` and make sure these instructions haven't missed anything
- [x] Remove the raw data from `s3://ccao-data-warehouse-us-east-1/ccao/dictionary/class_dict/`
    - [x] Perhaps delete `aprval.rscd` as well? I'm less sure what's going on with that table

wrridgeway:
> Perhaps delete aprval.rscd as well? I'm less sure what's going on with that table

Would it make more sense to turn this into a seed? I think it got put into s3 so that our glue crawler would turn it into an athena table but we never bothered to document it or script its generation.

jeancochrane:
@wrridgeway I think that's a great idea!

PR #547, opened 2024-07-17
 Strip out old references to class dict that predate the seed

wrridgeway:
See https://github.com/ccao-data/data-architecture/issues/537


jeancochrane (approved):
Thanks for handling this cleanup!

PR #550, opened 2024-07-17
Fix hard-coded asset names in qc.vw_change_in_ahsap_values.sql

wrridgeway:
Let some standard sql view/table names slip through into this view.


dfsnow (approved):
Nice @wrridgeway, thanks for catching this!

Issue #541, opened 2024-07-15
Detatch municipal assessment progress from `reporting.vw_assessment_roll_muni`


wrridgeway:
`reporting.vw_assessment_roll_muni` contains the progress that has been made in re-assessing individual municipalities. This metric can be useful elsewhere and should be its own view that can be joined to municipal-level reporting rather than contained in any one view.

PR #551, opened 2024-07-18
Detatch municipal assessment progress from `reporting.vw_assessment_roll_muni`

wrridgeway:
`pct_pin_w_value_in_group` in `reporting.vw_assessment_roll_muni` is so useful! It lets us know whether we can reasonably report on the tax base in a given municipality yet, or whether, because of the way the assessment schedule divides up some municipalities, we shouldn't yet. 

It's so useful that in addition to [detaching it](https://github.com/ccao-data/data-architecture/issues/541) and making it its own view, I'd like two more changes:

1. Let's expand the geo types and geo ids. Specifically, let's now include Chicago Ward, and to include all school districts that are property tax agencies. To accommodate this requires adding geo type and geo id columns. 
2. Currently `pct_pin_w_value_in_group` is unique per municipality, year, stage, and major class, as originally specified. I believe that for our new use case, it only needs to be unique per geo id, year, and stage -- removing the major class. 

The above changes do not affect any existing public reporting assets.


wrridgeway:
Can't really run any tests to compare old and new since `assessment_roll_muni` is no longer grouped by class.

ccao-jardine:
> Can't really run any tests to compare old and new since `assessment_roll_muni` is no longer grouped by class.

I'm sorry @wrridgeway, I think was terrifically unclear! I'd like to keep vw_assessment_roll_muni _itself_ grouped by class, so that we can report at the class level within a muni and stage. 

But the municipal assessment tracker that we're _detaching_, i.e. the column  `pct_pin_w_value_in_group` does _not_ need to be grouped by class; i.e., it's unique to the year, stage, and geo_id. The goal of this tracker is to provide a number that we can use to guide whether to report on that municipality as a whole. 

The vision is that by joining the tracker to vw_assessment_roll_muni on year, stage, and geo_id, users can use the tracker to gauge whether there's enough assessment progress to keep looking in vw_assessment_roll_muni. If there is, then the class-level data would be there.

dbt/models/reporting/reporting.vw_assessment_progress.sql
```diff
@@ -0,0 +1,172 @@
...
+    {{ assessment_progress_pin_count(
+            from = 'expanded_geos2',
+            geo_type = 'Unified School District',
+            column_name = 'school_unified_district_name'
+            ) }}
+)
+
+-- Calculate the portion of each geography that has progressed through an
+-- assessment stage by class.
+SELECT
```
dfsnow, line 149: **thought (non-blocking):** This is a pretty wild query. It took me a bit of running the CTEs to understand exactly what's happening. Now that we've moved fully onto dbt, it might be worth decomposing the CTEs into individual models or ephemeral views (once they're ready). The goal is to enable folks without your domain knowledge to work on/debug parts of these larger views.

An alternative to decomposing: even though the commenting here is really good, it might be clearer/even more helpful to add ASCII table representations of the data at each step/CTE.

dbt/models/reporting/schema.yml
```diff
@@ -98,6 +98,28 @@ models:
           - property_group
           - assessment_stage
 
+  - name: reporting.vw_assessment_progress
```
dfsnow, line 101: **suggestion (non-blocking):** It might be worth materializing this view by adding the dbt `daily` tag, as it queries ~50GB of data for each run.

dfsnow (approved):

dbt/models/reporting/reporting.vw_assessment_progress.sql
wrridgeway: Yeah, I agree this query is a little rough on the brain. Between it's complexity and the amount of data it queries, I'll try to sketch out an alternative this week.

PR #554, opened 2024-07-25
[1/3] Migrate Will's town close QC reports to dbt models

jeancochrane:
## Overview

This PR is the first of three following up on https://github.com/ccao-data/data-architecture/pull/534 by adding a bunch of models and Excel templates to recreate the QC reports that Will sends out on town close.

This PR migrates the following reports that all share common logic:

* 0 land value
* 0 value
* Vacant class, bldg value
* Improved class, no bldg value
* Class does not equal LUC
* 500k increase, 1m decrease

The subsequent PRs finishing the migration are https://github.com/ccao-data/data-architecture/pull/556 and https://github.com/ccao-data/data-architecture/pull/557. I've tried to group PRs according to reports that share common logic, with the third PR being a grab bag of miscellaneous reports.

We make a few other small changes to support this effort:
* Existing QC report views (`qc.vw_change_in_ahsap_values` and `qc.vw_report_town_close_neg_asmt_value`) get refactored to remove hardcoded dates, in line with other QC models
* The `dbt/scripts/export_models.py` script is extended to support reading a `config.meta.export_format.columns` attribute from the schema config, which is used to set formatting options for the output Excel workbooks that can't be set in the template (such as right-aligning text data columns)

## Testing

To recreate 2024 Jefferson township reports for the purposes of testing, run the following command in the `dbt/` subdirectory:

```
python3 scripts/export_models.py --select tag:qc_report_town_close --where "township_code = '71' and taxyr = '2024'" --rebuild
```

The reports will be output to the `dbt/export/output/` directory.


dbt/models/qc/schema.yml
```diff
@@ -612,14 +585,6 @@ models:
                 description: valasm3 (ASMT TOTAL) should not be negative
               <<: *test-non-negative
 
-  - name: qc.vw_report_neg_asmt_value
-    description: '{{ doc("view_vw_report_neg_asmt_value") }}'
-    config:
-      tags:
-        - qc_report_town_close
-    meta:
-      export_name: Negative ASMT values
```
jeancochrane, line 621: This model got renamed to `qc.vw_report_town_close_neg_asmt_value`.

dbt/models/qc/schema.yml
```diff
@@ -868,3 +860,71 @@ models:
...
+    config:
+      tags:
+        - qc_report_town_close
+    meta:
+      export_name: 500k increase, 1m decrease
+      export_format:
+        columns:
+          - name: '% Change'
+            index: M
+            horizontal_align: right
```
jeancochrane, line 930: Per the PR body, this is the new syntax for specifying formatting that should be applied to the export workbook at runtime. This is necessary because (as far as I can tell) you can't set horizontal alignment in an Excel template before you populate it with data using openpyxl.

dbt/scripts/export_models.py
```diff
@@ -198,10 +199,34 @@ def main():
                 ),
             )
             table.tableStyleInfo = TableStyleInfo(
-                name="TableStyleMedium7", showRowStripes=True
+                name="TableStyleMedium11", showRowStripes=True
```
jeancochrane, line 202: I felt like the look of `TableStyleMedium11` more closely matched Will's reports than `TableStyleMedium7`, but I don't really think it matters very much!

dbt/models/qc/qc.vw_iasworld_asmt_all_with_prior_year_values.sql
```diff
@@ -0,0 +1,52 @@
+SELECT
+    legdat.parid,
+    asmt.taxyr,
+    legdat.user1 AS township_code,
+    pardat.nbhd,
+    pardat.class,
+    pardat.luc,
+    owndat.own1,
+    aprval.reascd,
```
jeancochrane, line 9: The raw `reascd` here is just a number, but it would be nice to make them more human-readable, e.g. `1: Assessor Correction`. Is the correct way to do this just to join to the `ccao.aprval_reascd` table?

dbt/scripts/export_models.py
```diff
@@ -198,10 +200,47 @@ def main():
...
             )
             sheet.add_table(table)
 
+            # If a parid column exists, format it explicitly as a number to
+            # avoid Excel converting it to scientific notation when a user
+            # edits the cell
```
dfsnow, line 209: **issue (blocking):** Be careful with this, I'm pretty sure Excel strips the leading zeros off of PINs when they're converted to numbers. There's a format (General > Custom > type 14 zeros) that doesn't strip leading zeros or convert to scientific notation, but I'm not sure how to use it in openpyxl.

.flake8
```diff
@@ -0,0 +1,6 @@
+[flake8]
+ignore=
+    # Whitespace before ':', incompatible with black
+    E203,
+    # Line break before binary operator, incompatible with black
+    W503
```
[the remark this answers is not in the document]
jeancochrane: Good catch! Ruff seems to make this config unnecessary so I removed it in 0e41f3e31e5312bb1ec4eb1becdbfaae994e9fa8.

dbt/export/templates/assessed_and_market_values.xlsx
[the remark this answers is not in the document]
jeancochrane: Thanks for flagging this! Those errors occur whenever a query returns no results, which causes the filterable table that we define on the sheet to throw an error. I've just been ignoring the errors since we don't send over empty workbooks anyway, but I realize now that it's a pretty confusing developer experience, so I resolved the error in ac37578693f0c16da0c35e6a0769f9fdd076b885 by tweaking the script logic to only add a table to each sheet if its result set is non-empty.

dbt/models/qc/qc.vw_iasworld_asmt_all_with_prior_year_values.sql
wrridgeway: Should be just fine that way.

dbt/models/qc/docs.md
```diff
@@ -117,19 +138,213 @@ In contrast to `qc.vw_neg_asmt_value`, this view directly performs filtering
 for negative values.
 {% enddocs %}
 
-# vw_sale_mydec_null_values
+# vw_report_town_close_0_land_value
```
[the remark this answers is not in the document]
dfsnow: Schema files should be alphanumerically ordered. The docs files are kind of a mixed bag (some are broken up by section), but are usually alphanumeric also.

dbt/export/templates/assessed_and_market_values.xlsx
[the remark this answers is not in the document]
dfsnow: Makes sense! Thanks for the fix.

dbt/models/qc/docs.md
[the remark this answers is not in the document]
jeancochrane: Sounds good, I also reordered the schema file in 10e033f8e583b78a70fc7a519002498a19b0739f. Sorry for the annoying diff! I didn't change anything and checked with a `dbt compile` that the graph still compiles correctly, which should hopefully rule out any obvious copy/paste errors.

dfsnow (approved):
Nice, this looks great @jeancochrane!

PR #556, opened 2024-07-26
[2/3] Migrate Will's town close QC reports to dbt models

jeancochrane:
This PR follows up on https://github.com/ccao-data/data-architecture/pull/554 adding more of Will's QC reports to our dbt config. See that PR for background on this effort.

This PR migrates the following reports that all share common logic:

* Card code 5s possibly not at 100% (COMDAT)
* Card code 5s possibly not at 100% (DWELDAT)
* Card code 5s possibly not at 100% (OBY)

PR #557, opened 2024-07-26
[3/3] Migrate Will's town close QC reports to dbt models

jeancochrane:
This PR follows up on https://github.com/ccao-data/data-architecture/pull/554 and https://github.com/ccao-data/data-architecture/pull/556 adding more of Will's QC reports to our dbt config. See #554 for background on this effort.

This PR migrates the following reports that **do not** share common logic:

* Res parcels not set to Cost Approach
* OVRRCNLD discrepancy checks
* Bldg parcel class mismatch
* 289s
* Res edits
* Res multicodes


dbt/models/qc/docs.md
```diff
@@ -27,16 +27,6 @@ Pulled from the following Inquire queries:
...
 
-# vw_incorrect_val_method
-
-{% docs view_vw_incorrect_val_method %}
-Test that no residential parcels are using a cost approach.
-
-Pulled from the following Inquire queries:
-
-- `FP Checklist - Res parcels not set to Cost Approach` (#1578) - Will C.
-{% enddocs %}
```
jeancochrane, line 38: This view has been refactored into `qc.vw_report_town_close_res_parcels_not_set_to_cost_approach`.

dbt/models/qc/qc.vw_iasworld_sales_latest_sale.sql
```diff
@@ -0,0 +1,21 @@
+SELECT
+    parid,
+    instruno,
+    price,
+    saledt,
+    DATE_FORMAT(
+        DATE_PARSE(saledt, '%Y-%m-%d %H:%i:%S.%f'), '%c/%e/%Y'
+    ) AS saledt_fmt
```
jeancochrane, line 8: A number of different QC reports reformat iasworld dates in a similar fashion. I thought about factoring out the logic into a dbt macro, or trying to figure out how to get Excel to recognize the strings as dates in order to format them natively, but this ended up feeling to me like the least complex approach.

dbt/models/qc/qc.vw_report_town_close_res_edit.sql
```diff
@@ -0,0 +1,193 @@
...
+),
+
+num_dwellings AS (
+    SELECT
+        parid,
+        taxyr,
+        COUNT(*) AS num_dwellings
+    FROM first_five_dwellings
+    GROUP BY parid, taxyr
+)
```
jeancochrane, line 78: Note that this is counting the number of dwellings _in the first five dwellings_, and not necessarily the number of dwellings in the parcel. I think this is probably fine, but it's a distinction to keep in mind, and it may make sense for me to double check with Will.

dbt/models/qc/schema.yml
```diff
@@ -2,33 +2,6 @@ models:
...
-                - township_code
-                - who
-                - wen
-              config:
-                where: |
-                  CAST(taxyr AS int) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
-                  AND class LIKE '2%'
-              meta:
-                table_name: aprval
-                description: revcode should be '1' for 200-class parcels
```
jeancochrane, line 30: This test got moved to the new `qc.vw_report_town_close_res_parcels_not_set_to_cost_approach` model.

dbt/models/qc/schema.yml
```diff
@@ -916,6 +892,64 @@ models:
     meta:
       export_name: Class does not equal LUC
 
+  - name: qc.vw_report_town_close_res_parcels_not_set_to_cost_approach
```
jeancochrane, line 895: I'm realizing now that this name might be clearer as `qc.vw_report_town_close_res_parcels_set_to_cost_approach` (as opposed to `not_set_to_cost_approach`), since that more accurately describes what is being queried for... what do you think?

dbt/models/qc/qc.vw_report_town_close_res_edit.sql
```diff
@@ -0,0 +1,193 @@
...
+    dweldat.sfla_{{ idx }},
+    dweldat.stories_{{ idx }},
+    dweldat.rmbed_{{ idx }},
+    dweldat.fixbath_{{ idx }},
+    dweldat.fixhalf_{{ idx }},
+    dweldat.bsmt_{{ idx }},
+    dweldat.bsmt_fin_{{ idx }},
+    dweldat.external_propct_{{ idx }},
+    dweldat.alt_cdu_{{ idx }}{% if not loop.last %},{% endif %}
+{% endfor %}
```
jeancochrane, line 157: It would be nice to recode these values a la `ccao::vars_recode()`, is there a way to do that in Athena? Or perhaps in Python as part of the export script?

dbt/models/qc/schema.yml
```diff
@@ -916,6 +892,64 @@ models:
...
+    description: '{{ doc("view_vw_report_town_close_res_parcels_not_set_to_cost_approach") }}'
+    config:
+      tags:
+        - qc_report_town_close
```
dfsnow, line 899: **issue:** Won't these tests never get run, since we're querying for `test_qc_*` tags when running the daily build?

dfsnow (commented):
@jeancochrane Thanks for handling this conversion, I can see it's a ton of work!

And thanks for doing the extra work of splitting these up for me. It helped make review much easier.

Overall this is looking great! My only concern is that our testing apparatus is getting a bit unwieldy. I think you already plan to do this, but we should write up a doc defining what tests run when, what triggers them, the purpose of each group of tests, etc. So far I can think of:

- QC reports (these)
- QC tests (output to the test failure workbook)
- Data tests run separately as part of the dbt CI that capture bad joins and things like that
- (Soon) unit tests to test our in-SQL logic

dbt/models/qc/schema.yml
jeancochrane: See my comment above for an explanation of why "res parcels *not* set to cost approach" is actually correct 😅

dbt/models/qc/docs.md
```diff
@@ -215,6 +213,61 @@ Pulled from the following Inquire queries:
...
+
+Pulled from the following Inquire queries:
+
+- `FP Checklist - Bldg and parcel class mismatch` (#2161) - Will C.
+{% enddocs %}
+
+# vw_report_town_close_289s
+
+{% docs view_vw_report_town_close_289s %}
+Check for rows in the  `OBY` table where the class is 289.
```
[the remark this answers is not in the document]
jeancochrane: Oops, good catch! Fixed in b9ac1baa648f2b9e619c48442a22eb541dac5f79 along with another stray double space.

dbt/models/qc/qc.vw_report_town_close_bldg_parcel_class_mismatch.sql
```diff
@@ -0,0 +1,71 @@
...
+    pardat.class AS parcel_class,
+    {% for tablename in comparison_tables %}
+        ARRAY_JOIN({{ tablename }}.classes, ', ')
+            AS {{ tablename }}_classes
+
+        {% if not loop.last %}
+
+            ,
+
+        {% endif %}
```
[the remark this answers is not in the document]
jeancochrane: sqlfluff was really annoying about fixing this but I finally got it to work in a slightly different form in b9ac1baa648f2b9e619c48442a22eb541dac5f79.

dbt/models/qc/qc.vw_report_town_close_ovrrcnlds_to_review.sql
```diff
@@ -0,0 +1,113 @@
...
+        dweldat.ovrrcnld,
+        dweldat.external_calc_rcnld - dweldat.ovrrcnld AS value_difference,
+        dweldat.mktrsn AS card_code,
+        dweldat.user16 AS alt_cdu,
+        dweldat.who,
+        DATE_FORMAT(
+            DATE_PARSE(dweldat.wen, '%Y-%m-%d %H:%i:%S.%f'), '%c/%e/%Y %H:%i'
+        ) AS wen
+    FROM {{ source('iasworld', 'dweldat') }} AS dweldat
+    WHERE dweldat.cur = 'Y' AND dweldat.deactivat IS NULL
```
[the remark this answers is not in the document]
jeancochrane: Oops, thanks! Fixed along with the three other occurrences below in b9ac1baa648f2b9e619c48442a22eb541dac5f79.

dbt/models/qc/qc.vw_report_town_close_ovrrcnlds_to_review.sql
```diff
@@ -0,0 +1,113 @@
...
+        oby.ovrrcnld,
+        oby.external_calc_rcnld - oby.ovrrcnld AS value_difference,
+        oby.chgrsn AS card_code,
+        oby.user16 AS alt_cdu,
+        oby.who,
+        DATE_FORMAT(
+            DATE_PARSE(oby.wen, '%Y-%m-%d %H:%i:%S.%f'), '%c/%e/%Y %H:%i'
+        ) AS wen
+    FROM {{ source('iasworld', 'oby') }} AS oby
+    WHERE oby.cur = 'Y' AND oby.deactivat IS NULL
```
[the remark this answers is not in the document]
jeancochrane: Fixed in Oops, good catch! Fixed in b9ac1baa648f2b9e619c48442a22eb541dac5f79.

dbt/models/qc/schema.yml
jeancochrane: Good question! Technically `qc_vw_incorrect_val_method_revcode_eq_1` would still get run because I moved the tag over to the test from the model. Still, I think that was a bit unnecessarily defensive of me to do -- thinking about it again, I realize now that we'll probably want any more tests we add to this model to be run as part of the QC test suite -- so I moved it back from the test to the model in b9ac1baa648f2b9e619c48442a22eb541dac5f79.

dbt/models/qc/qc.vw_report_town_close_res_edit.sql
jeancochrane: Makes sense, I switched to counting all cards in `dweldat` in 83ddeb8fdf2a5472f49bb0f152f56ddf0d58d61b and spot-checked the counts between the raw query and the view output.

jeancochrane:
@dfsnow This whole stack of PRs should be ready for another round of review! I've done my best to address all of your comments, but feel free to tag me on any of them that may have slipped through the cracks.

dbt/models/qc/schema.yml
dfsnow: Ah, I had a tired moment and thought we were running the town close QC reports daily along with the rest of the `test_qc_*` tagged stuff. Now I remember we just call them on demand from export script.

dbt/models/qc/docs.md
```diff
@@ -128,61 +147,32 @@ Pulled from the following Inquire queries:
...
 
-# vw_report_town_close_ovrrcnlds_to_review
+# vw_report_town_close_500k_increase_1m_decrease
 
-{% docs view_vw_report_town_close_ovrrcnlds_to_review %}
-Check for rows in the `DWELDAT`, `COMDAT`, and `OBY` tables
-where the calculated net market value does not match the override net market
-value.
+{% docs view_vw_report_town_close_500k_increase_1m_decrease %}
+Check for rows in the  `ASMT` table where the total market value has increased
```
dfsnow, line 166:
```suggestion
Check for rows in the `ASMT` table where the total market value has increased
```

dbt/models/qc/schema.yml
dfsnow, line 1: **question (non-blocking):** This is mostly just re-ordering things right? It's hard to tell via the huge diff.

Also, AFAIK all of these are currently excluded from both the test failure workbooks AND the QC reports right? Do you think we could in incorporate them easily into either? It's a shame to leave them sitting around.

dfsnow (approved):

dbt/models/qc/schema.yml
jeancochrane: No worries, it's confusing! I'm hoping the docs will help clear some of this stuff up.

dbt/models/qc/docs.md
jeancochrane: Oops, good catch! Fixed along with another stray whitespace in da946cc145e0b5c1f82d0dcb5075c957110734d4.

dbt/models/qc/schema.yml
jeancochrane: > This is mostly just re-ordering things right? It's hard to tell via the huge diff.

Sorry about that diff! In hindsight I think I should have done a fresh pre-review given how much changed between review rounds. Yup, the only thing that should have changed was the order of the items.

> Also, AFAIK all of these are currently excluded from both the test failure workbooks AND the QC reports right? Do you think we could in incorporate them easily into either? It's a shame to leave them sitting around.

Hmm, which tests/models are you referring to? Anything with the `test_qc_exclude_from_workbook` tag will be excluded from the test workbook (but not the QC reports). Everything else should be included, which covers many of the tests in this file.

jeancochrane:
@dfsnow I'm going to go ahead and merge even though you have an outstanding question about exclusion; I think I answered it, but if there's in fact an error that I'm missing I can fix it in a follow-up PR!

Issue #558, opened 2024-07-28
Investigate duplicates in `model.vw_card_res_input`


wrridgeway:
I had @Damonamajor add dbt tests for our modeling views to detect non-unique rows. There are currently 19 pin/card/year combos in `model.vw_card_res_input` with more than 1 row. This shouldn't be the case. My preliminary investigation suggests the nearest neighbor joins in the script for the view are creating the dupes:

```
FROM forward_fill AS f1
LEFT JOIN (
    SELECT *
    FROM forward_fill
    WHERE NOT ind_pin_is_multicard
        AND nearest_neighbor_1_dist_ft <= 500
) AS nn1
    ON f1.nearest_neighbor_1_pin10 = nn1.meta_pin10
    AND f1.meta_year = nn1.meta_year
LEFT JOIN (
    SELECT *
    FROM forward_fill
    WHERE NOT ind_pin_is_multicard
        AND nearest_neighbor_2_dist_ft <= 500
) AS nn2
    ON f1.nearest_neighbor_1_pin10 = nn2.meta_pin10
    AND f1.meta_year = nn2.meta_year
```

wrridgeway:
@dfsnow let me know if this is a lark

wrridgeway:
lol:

meta_pin | meta_class | meta_year | nearest_neighbor_1_pin10
-- | -- | -- | --
14201150381001 | 208 | 2017 | 1420115028
14201150280000 | 206 | 2017 | 1420115038
14201150381002 | 208 | 2017 | 1420115028

got some 208s that aren't unique by pin10

dfsnow:
Hmmmm, I'm guessing it's a join issue. Can you trace it and maybe walk @Damonamajor through how to do so.

wrridgeway:
I believe it's an issue with not all res parcels being unique by pin10 and doing the NN join on pin10. Take `1432413072` for example:

```
select meta_pin, meta_year, meta_class
from model.vw_card_res_input
where meta_pin10 = '1432413072' and year = '2017'
```

meta_pin | meta_year | meta_class
-- | -- | --
14324130721003 | 2017 | 278
14324130721002 | 2017 | 278
14324130721001 | 2017 | 278

PR #560, opened 2024-07-29
Investigate duplicates in `model.vw_card_res_input`

wrridgeway:
See https://github.com/ccao-data/data-architecture/issues/558

Long story short, `model.vw_card_res_input` is not currently unique by pin10, card, and year due to nearest neighbor joins. These joins introduce duplicates because some non-condo res parcels have 14-digit pins rather than the 10 we assume.

### DIFF
--- a/dbt/models/model/model.vw_card_res_input.sql
+++ b/dbt/models/model/model.vw_card_res_input.sql
@@ -654,6 +654,7 @@ LEFT JOIN (
     SELECT *
     FROM forward_fill
     WHERE NOT ind_pin_is_multicard
+        AND SUBSTR(meta_pin, 11, 4) = '0000'
         AND nearest_neighbor_1_dist_ft <= 500
 ) AS nn1
     ON f1.nearest_neighbor_1_pin10 = nn1.meta_pin10
@@ -662,6 +663,7 @@ LEFT JOIN (
     SELECT *
     FROM forward_fill
     WHERE NOT ind_pin_is_multicard
+        AND SUBSTR(meta_pin, 11, 4) = '0000'
         AND nearest_neighbor_2_dist_ft <= 500
 ) AS nn2
     ON f1.nearest_neighbor_1_pin10 = nn2.meta_pin10
--- a/dbt/models/model/model.vw_card_res_input.sql
+++ b/dbt/models/model/model.vw_card_res_input.sql
@@ -654,7 +654,11 @@ LEFT JOIN (
     SELECT *
     FROM forward_fill
     WHERE NOT ind_pin_is_multicard
-        -- Some res parcels are not unique by pin10
+        /* Unfortunately, some res parcels are not unique by pin10, card, and
+        year. This complicates one of our core assumptions about the res parcel
+        universe.
+        See https://github.com/ccao-data/data-architecture/issues/558 for more
+        information. */
         AND SUBSTR(meta_pin, 11, 4) = '0000'
         AND nearest_neighbor_1_dist_ft <= 500
 ) AS nn1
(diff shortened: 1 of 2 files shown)


wrridgeway:
I _believe_ this is the most straightforward way to handle this issue. I don't think it's a problem with how we calculate nearest pins, so I didn't touch the macro for that - it's just a problem with us needing to assume res pins are unique by pin10 for this type of filling when they're not. Thankfully this issue seems like it was rectified in the data after 2017. Removing the config clause for the `unique_combination_of_columns` test shows ensures the view is now unique.

```
with old AS (
    select year, count(*) as old
    from "model".vw_card_res_input group by year
),

new as (
    select year, count(*) as new
    from "z_ci_558_investigate_duplicates_in_modelvw_card_res_input_model".vw_card_res_input
    group by year
)

select
    old.year,
    old.old,
    new.new,
    old.old - new.new as diff
from old
left join new
    on old.year = new.year
```

year | old | new | diff
-- | -- | -- | --
2024 | 1099572 | 1099572 | 0
2023 | 1098999 | 1098999 | 0
2022 | 1098520 | 1098520 | 0
2021 | 1098460 | 1098460 | 0
2020 | 1098310 | 1098310 | 0
2019 | 1098203 | 1098203 | 0
2018 | 1098026 | 1098026 | 0
2017 | 1097869 | 1097866 | 3
2016 | 1097330 | 1097322 | 8
2015 | 1097308 | 1097305 | 3
2014 | 1096888 | 1096885 | 3
2013 | 1097208 | 1097200 | 8
2012 | 1097928 | 1097914 | 14
2011 | 1098162 | 1098154 | 8
2010 | 1098741 | 1098733 | 8
2009 | 1098606 | 1098598 | 8
2008 | 1099090 | 1099082 | 8
2007 | 1097785 | 1097777 | 8
2006 | 1095396 | 1095388 | 8
2005 | 1092271 | 1092263 | 8
2004 | 1088624 | 1088616 | 8
2003 | 1084651 | 1084643 | 8
2002 | 1076780 | 1076769 | 11
2001 | 1072950 | 1072950 | 0
2000 | 1069983 | 1069983 | 0
1999 | 1057865 | 1057865 | 0

dbt/models/model/model.vw_card_res_input.sql
```diff
@@ -654,6 +654,8 @@ LEFT JOIN (
     SELECT *
     FROM forward_fill
     WHERE NOT ind_pin_is_multicard
+        -- Some res parcels are not unique by pin10
+        AND SUBSTR(meta_pin, 11, 4) = '0000'
```
dfsnow, line 658: **question (non-blocking):** Isn't there a (slim) chance that this will result in NULL values on the filled characteristic/final view if _both_ nearest neighbor PINs?

dfsnow, line 658: **todo:** Can you add a little more context to the comment about why this is needed? Maybe just link to this issue.

dfsnow (approved):
Good lookin out @wrridgeway. @Damonamajor, if we haven't already, let's add some additional tests to the rest of the `default` and `model` views and see if we can catch any more stuff like this.

dbt/models/model/model.vw_card_res_input.sql
wrridgeway: Yes, there is. The other solutions I considered
1. making the two select statements for the NN joins exclude pin14 and use distinct
2. excluding pins from the `nearest_pin_neighbors` macro that are in our res classes and have 14 digit pins as possible neighbors

Those two solutions seemed pretty intense for the level of risk, but i could look into them further.

dfsnow: Hmmm, alright. Let's stick with what you've done for now.

PR #564, opened 2024-08-02
Switch to ruff Python setup + bump Actions versions

dfsnow:
This PR swaps ruff for older Python tooling (flake8, black, isort). Similar to https://github.com/ccao-data/assesspy/pull/20. No actual content or settings change except for 1 line, since ruff is a drop-in replacement for all these tools.

It also bumps the versions of some of all our first-party GitHub Actions.


.github/scripts/transform_dbt_test_results.py
```diff
@@ -626,8 +626,7 @@ def _filter_for_existing_field_indexes(
         return tuple(
             openpyxl.utils.get_column_letter(
                 # openpyxl is 1-indexed while the index() method is 0-indexed
-                existing_fieldnames.index(field)
-                + 1
+                existing_fieldnames.index(field) + 1
```
dfsnow, line 629: This is the only line that actually got corrected by ruff.

pyproject.toml
dfsnow, line 1: Just consolidating the config files for sqlfluff and ruff into this single file. No changes to the actual settings.

dfsnow, line 1: Just consolidating the config files for sqlfluff and ruff into this single file. No changes to the actual settings.

dfsnow, line 1: Just consolidating the config files for sqlfluff and ruff into this single file. No changes to the actual settings.

.github/workflows/lint.yaml
```diff
@@ -17,24 +17,27 @@ jobs:
...
-        uses: github/super-linter@v4
+        uses: github/super-linter@v6
         env:
           VALIDATE_ALL_CODEBASE: false
           DEFAULT_BRANCH: master
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           VALIDATE_MARKDOWN: false
           # These tools are handled by pre-commit, so we don't need SuperLinter
           # to run them as well
+          VALIDATE_CHECKOV: false
```
dfsnow, line 35: These got introduced with the version bump. One of them (CHECKOV) throws some interesting new warnings [about our IaC setup](https://github.com/ccao-data/data-architecture/actions/runs/10210408742/job/28250004495#step:4:229). Might be worth pursuing, fixing, and then re-enabling this linter. What do you think?

jeancochrane: Interesting, thanks for flagging this! On first glance, the [top-level permissions error](https://github.com/ccao-data/data-architecture/actions/runs/10210332289/job/28249802654#step:4:251) seems reasonable enough and I would support fixing it, but I'm finding myself struggling to understand the [workflow_dispatch inputs check](https://github.com/ccao-data/data-architecture/actions/runs/10210332289/job/28249802654#step:4:238). I'm down to file an issue to get a handle on these checks and enable the linter!

jeancochrane (approved):
Nice work, thank you!

PR #566, opened 2024-08-06
Implement socrata updating through publisher API

wrridgeway:
This PR contains:
- A python script that can be triggered through a github workflow to update a user determined asset on socrata through the socrata publisher api
  - user can decide which years the asset is updated for, whether it should be chunked by township, and whether or not it should be overwritten
  - the publisher api crashes if we try to push too much data at once, so uploads are handled by pulling chunked data into memory from athena and then pushed to socrata
- Addition of open data related metadata to athena views that feed socrata assets
  - Currently the asset id for `default.vw_pin_universe` is *not* accurate - it's set to update the api test asset I created.

My guess is the workflow probably won't work as intended since I haven't been able to test it.

### Improvements to consider once script is stable:
- Allow township chunking without year chunking

Changes to make in a subsequent PR
---
- Remove limit on sql queries used for testing
- Replace test version of "Parcel Universe" asset id


socrata/socrata_upload.py
```diff
@@ -0,0 +1,174 @@
...
+app_token = os.getenv("SOCRATA_APP_TOKEN")
+auth = (os.getenv("SOCRATA_USERNAME"), os.getenv("SOCRATA_PASSWORD"))
+cursor = connect(
+    s3_staging_dir=os.getenv("AWS_ATHENA_S3_STAGING_DIR") + "/",
+    region_name=os.getenv("AWS_REGION"),
+    cursor_class=PandasCursor,
+).cursor(unload=True)
+
+
+def build_query(athena_asset, row_identifiers, years=None):
```
wrridgeway, line 19: @dfsnow is this something as user should pass to the function, or something we could automatically grab from dbt

.github/workflows/socrata_upload.yaml
```diff
@@ -0,0 +1,73 @@
...
+      # Set these env vars as secrets so they get masked in the GitHub
+      # Actions logs
+      - run: pip install -r socrata/requirements.txt
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+      - name: Upload to Socrata
+        env:
```
wrridgeway, line 65: I could not get this to work using `secrets`

.github/workflows/socrata_upload.yaml
```diff
@@ -0,0 +1,77 @@
+name: update_socrata_asset
+
+on:
+  workflow_dispatch:
```
wrridgeway, line 4: We'll want to add a cron component as well with some different default values, but I just wanted to get this preliminary build finished up.

socrata/socrata_upload.py
```diff
@@ -0,0 +1,337 @@
...
+        )
+
+    elif method == "post":
+        response = requests.post(
+            url=url,
+            data=cursor.execute(sql_query, query_conditionals)
+            .as_pandas()
+            .to_json(orient="records"),
+            auth=auth,
+        )
```
jeancochrane, line 179: **[Question, non-blocking]** I wonder if we should call [`raise_for_status()`](https://3.python-requests.org/user/quickstart/#response-status-codes) on these responses so that we can surface errors? More broadly, what's our desired behavior in case an upload fails partway through a job?

socrata/socrata_upload.py
```diff
@@ -0,0 +1,337 @@
...
+                    }
+                elif flag == "years":
+                    upload_args = {
+                        "asset_id": asset_id,
+                        "sql_query": sql_query,
+                        "overwrite": overwrite,
+                        "year": item,
+                    }
+                response = upload("post", **upload_args)
+                print(response.content)
```
jeancochrane, line 326: **[Suggestion, non-blocking]** I wonder if we could consolidate these two conditional branches if we treated `overwrite` as something that only matters for determining whether subsequent requests after the first request should be `PUT` or `POST`? Then we'd only need one block that looks something like this:

```python
for item in groups:
    if flag == "both":
        upload_args = {
            "asset_id": "asset_id",
            "sql_query": "sql_query",
            "overwrite": "overwrite",
            "year": item[0],
            "township": item[1],
        }
    elif flag == "years":
        upload_args = {
            "asset_id": asset_id,
            "sql_query": sql_query,
            "overwrite": overwrite,
            "year": item,
        }
    if count == 0 and overwrite:
        response = upload("put", **upload_args)
    else:
        response = upload("post", **upload_args)
    print(response.content)
    count = count + 1
```

socrata/socrata_upload.py
```diff
@@ -0,0 +1,337 @@
...
+                    upload_args = {
+                        "asset_id": asset_id,
+                        "sql_query": sql_query,
+                        "overwrite": overwrite,
+                        "year": item,
+                    }
+                if count == 0:
+                    response = upload("put", **upload_args)
+                else:
+                    response = upload("post", **upload_args)
```
jeancochrane, line 304: **[Question, non-blocking]** This is somewhat counterintuitive to me, since that my understanding of the distinction between `PUT` vs. `POST` is that `POST` is typically used to create a resource and `PUT` is used to update it (see [this SO question](https://stackoverflow.com/q/630453) for links to the RFC definitions). Wouldn't we expect to `POST` on the first request, and `PUT` thereafter?

socrata/socrata_upload.py
```diff
@@ -0,0 +1,337 @@
...
+    toc = time.perf_counter()
+    print(f"Total upload in {toc - tic:0.4f} seconds")
+
+
+socrata_upload(
+    socrata_asset=os.getenv("SOCRATA_ASSET"),
+    overwrite=os.getenv("OVERWRITE"),
+    years=os.getenv("YEARS"),
+    by_township=os.getenv("BY_TOWNSHIP"),
+)
```
jeancochrane, line 337: **[Thought, non-blocking]** One high-level thought that I have for you about the design of the functions in this module: Currently, there are three possible filter conditions (no filters; all filters; or year _but not_ township filter) and each one of the `build_query`, `generate_groups`, and `upload` functions are responsible for implementing a different solution for each condition. This works fine, but it makes the functions hard to read since each one has its own set of complex conditional branches, and if we ever needed to alter those conditions or add a new one I think we would find ourselves having to basically rewrite the whole module to accommodate the new or changed condition.

I think there's potential for a simpler design that consolidates all of the responsibility for evaluating the filters in the `generate_groups` function, which could return a list of groups that could then be operated on by each of the `build_query` and `upload` functions agnostic of the exact contents of the filters themselves. This is a bit speculative, so I would be happy to take a stab at a follow-up PR to demonstrate a refactor if you're interested but not sure what I mean. Let me know and we can hash it out! But I don't think it's necessary to get too deep into it while you're still trying to just get the functionality up and running.

socrata/socrata_upload.py
```diff
@@ -0,0 +1,337 @@
...
+
+    if not years:
+        query = (
+            "SELECT\n"
+            + row_identifier
+            + "\n"
+            + ",\n".join(columns["column"])
+            + "\nFROM "
+            + athena_asset
+            + "\nLIMIT 1000"
```
jeancochrane, line 97: **[Question, non-blocking]** I'm guessing these limits are for testing and we're going to eventually strip them out, is that right?

jeancochrane (approved):
Awesome job here! Seems like it works, and I have no major concerns about it. However, I think there are a bunch of small opportunities for aligning the file structure with our existing dbt patterns, and also for cleaning up the Python script itself to make it more maintainable. I would put the priority of aligning this effort with our dbt setup to be higher than perfecting the Python script, but ultimately I think both workstreams are up to you. Let me know if you want to talk through any of these suggestions!

.github/workflows/socrata_upload.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - Property Tax-Exempt Parcels
+        default: Parcel Universe
+        required: true
+      overwrite:
+        type: boolean
+        description: Overwrite or update socrata asset
+        required: true
+      years:
+        type: string
+        description: Years to update or overwrite
```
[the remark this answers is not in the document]
wrridgeway: Yes, is that something I should note?

.github/workflows/socrata_upload.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - Assessed Values
+        - Appeals
+        - Parcel Addresses
+        - Parcel Proximity
+        - Property Tax-Exempt Parcels
+        default: Parcel Universe
+        required: true
+      overwrite:
+        type: boolean
+        description: Overwrite or update socrata asset
```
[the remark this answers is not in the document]
wrridgeway: True for overwrite : False for update. I'll add a note.

.github/workflows/socrata_upload.yaml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      # contents:read and id-token:write permissions are needed to interact
+      # with GitHub's OIDC Token endpoint so that we can authenticate with AWS
+      contents: read
+      id-token: write
+      # While packages:write is usually not required for workflows, it is
+      # required in order to allow the reusable called workflow to push to
+      # GitHub Container Registry
+      packages: write
```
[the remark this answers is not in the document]
wrridgeway: I'll remove this package.

socrata/socrata_upload.py
wrridgeway: Yep, I'll add a note to the PR. I also need to replace the test asset ID for parcel universe.

socrata/socrata_upload.py
wrridgeway: For whatever reason, here's [how socrata does it](https://dev.socrata.com/publishers/soda-producer/soda-producer-basics).

socrata/socrata_upload.py
```diff
@@ -0,0 +1,337 @@
...
+        )
+
+    return query
+
+
+def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
+    """
+    Function to perform the upload to Socrata. `puts` or `posts` depending on
+    user's choice to overwrite existing data.
```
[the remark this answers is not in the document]
wrridgeway: There's definitely overlap, but having these two different params allows the function to print "overwriting" for subsequent chunks even when using `post` after an initial `put` overwrites the open data asset.

socrata/socrata_upload.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+def generate_groups(athena_asset, years=None, by_township=False):
+    """
+    Helper function to determine what groups need to be iterated over for
+    upload.
+    """
+
+    if not years and by_township:
+        raise ValueError("Cannot set 'by_township' when 'years' is None")
```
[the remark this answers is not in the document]
wrridgeway: I suppose there isn't. I can't actually remember why I wrote this code to so adamantly preclude township chunking without year chunking.

socrata/socrata_upload.py
wrridgeway: Perfect.

PR #567, opened 2024-08-07
Update docs.md

wrridgeway:
pin and year are not primary keys for `default.vw_pin_sale`. `doc_no` has to be combined with `pin` since multisales make `doc_no` non-unique.


jeancochrane (approved):
Oops, good catch!

Issue #539, opened 2024-07-12
Improve documentation for QC tests


jeancochrane:
Now that QC tests are being developed more widely among the team, it's become clear that we're lacking good docs explaining how and when we run QC tests. We also need to document the emerging distinction between QC tests (errors) and QC reports (warnings).

Documentation checklist:

- [x] Difference between the different types of tests we have
    - [x] QC tests
    - [x] QC reports
    - [x] Unit tests
- [x] How and when we run QC tests
- [x] How and when we run QC reports
- [x] Writing a new QC test
    - [x] Syntax for metadata
         - [x] Exclusion tag  
- [x] Writing a new QC report
    - [x] Syntax for `export_*` configs in schema files

PR #568, opened 2024-08-08
Improve documentation for QC products

jeancochrane:
This PR updates `dbt/README.md` to improve our documentation for our QC tests and reports. In the process, we also take the opportunity to move the `transform_dbt_test_results.py` script from its old location at `.github/scripts/transform_dbt_test_results.py` to a new home at `dbt/scripts/transform_dbt_test_results.py`. This consolidates the location of the script alongside our other script `dbt/scripts/export_models.py`, and reflects the new reality that we run this script by hand more often than we thought we would when we initially wrote it.

Closes https://github.com/ccao-data/data-architecture/issues/539.


.github/workflows/test_dbt_models.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
 
       - name: Extract test results
         run: |
-          python3 ../.github/scripts/transform_dbt_test_results.py \
+          python3 scripts/transform_dbt_test_results.py \
```
jeancochrane, line 55: While the `pip install` command above provides the `dbt/` subdir as a prefix to this command, we don't need to do that here since the `working-directory` is already set to `dbt` on line 57 below.

README.md
```diff
@@ -10,8 +10,10 @@ modeling, reporting, and data integrity work.
   Documentation for all CCAO data lakehouse tables and views
 - [:nut_and_bolt: dbt README](/dbt/README.md) - How to develop CCAO data
   infrastructure using dbt
-- [:test_tube: dbt Test Documentation](/dbt/tests/generic/README.md) -
-  Definitions for CCAO generic dbt data tests
+- [:test_tube: dbt QC Tests and Reports](dbt/README.md#-how-to-add-and-run-qc-tests-and-reports) -
+  How to add and run QC tests and reports using dbt
```
jeancochrane, line 14: I figure that QC testing and reporting is an important enough feature of this repo that it's worth linking to directly from the root README.

Maybe this is a personal preference thing, but the 🧪 emoji makes me think the link will be discussing QC products (as opposed to dbt generics), so I moved it from the generic docs to the QC  products docs and switched the generic docs to the 📝 emoji.

dbt/README.md
```diff
@@ -8,10 +8,10 @@ This directory stores the configuration for building our data catalog using
 ### In this document
 
 * [🖼️ Background: What does the data catalog do?](#%EF%B8%8F-background-what-does-the-data-catalog-do)
-* [🔨 How to rebuild models using GitHub Actions](#-how-to-rebuild-models-using-github-actions)
```
jeancochrane, line 11: It feels more natural to me to flow from "Background" > "Developing" > "Add model" > "Rebuild model" rather than "Background" > "Rebuild model" > "Developing" > "Add model", so I went ahead and reordered the sections. Happy to revert if you have a strong preference for the previous outline of the doc.

dbt/README.md
```diff
@@ -8,10 +8,10 @@ This directory stores the configuration for building our data catalog using
...
 * [💻 How to develop the catalog](#-how-to-develop-the-catalog)
 * [➕ How to add a new model](#-how-to-add-a-new-model)
-* [📝 How to add and run tests](#-how-to-add-and-run-tests)
+* [🔨 How to rebuild models using GitHub Actions](#-how-to-rebuild-models-using-github-actions)
+* [🧪 How to add and run QC tests and reports](#-how-to-add-and-run-qc-tests-and-reports)
```
jeancochrane, line 14: Again, for the sake of consistency and my own expectation as a reader I switched to using the 🧪 emoji here for QC tests and reports.

dbt/README.md
```diff
@@ -44,8 +44,7 @@ use to predict property values.
 
 ### 2. Define integrity checks that specify the correct format for our data
 
-The tests defined in the `schema.yml` files in the `models/` directory,
-alongside some one-off tests defined in the [`tests/`](./tests/) directory,
```
jeancochrane, line 48: We don't actually have any tests in the `tests/` directory anymore, so there's no use documenting them.

dbt/README.md
```diff
@@ -222,11 +199,8 @@ You should almost never have to manually build tables and views in our
 production environment, since this repository is configured to automatically
 deploy production models using GitHub Actions for continuous integration.
 However, in the rare case that you need to manually build models in production,
-use the `--target` option:
-
-```
-dbt build --target prod --resource-types model seed
-```
```
jeancochrane, line 229: Folks should be using the GitHub workflow to rebuild prod models, not running `dbt build --target prod` in the command line like a chaos monkey.

dbt/README.md
```diff
@@ -494,63 +468,141 @@ We use the following pattern to determine where to define each column descriptio
...
+passed directly to `dbt build`. Model names _must include the database schema name_. Some possible inputs include:
 
-1. **Data tests** check that our assumptions about our raw data are correct
+- `default.vw_pin_sale` - Rebuild a single view
+- `default.vw_pin_sale default.vw_pin_universe` - Rebuild two views at once
+- `+default.vw_pin_history` - Rebuild a view and all its upstream dependencies
+- `location.*` - Rebuild all views under the `location` schema
+- `path:models` - Rebuild the full DAG (:warning: takes a long time!)
+
+For more possible inputs using dbt node selection, see the [documentation site](https://docs.getdbt.com/reference/node-selection/syntax#examples).
```
jeancochrane, line 493: This section shouldn't have changed at all, I just moved it from a different location.

dbt/README.md
```diff
@@ -494,63 +468,141 @@ We use the following pattern to determine where to define each column descriptio
...
+
+## 🧪 How to add and run QC tests and reports
```
jeancochrane, line 495: The expanded QC test/report section is getting big enough that I'm starting to wonder if it's worth moving it to its own README in a place like `dbt/models/qc/README.md`. I'm not going to do that here for the sake of the diff, but happy to do so in a follow-up PR if you think it's worth it.

dbt/README.md
```diff
@@ -494,63 +468,141 @@ We use the following pattern to determine where to define each column descriptio
...
+  data problems
+* Parquet files representing metadata tables that can be uploaded to S3 for aggregate
+  analysis
+
+**Non-QC tests** are much less common than QC tests in our test suite. We don't currently
+have a great process for handling them, so there is not much to document yet. If you
+are being asked to add a test that appears to be a non-QC test, meaning that it
+does not confirm our assumptions about iasWorld data, double check with the person who
+assigned the test to you and ask them when and how the test should be run so that its
+attributes can be set accordingly.
```
jeancochrane, line 551: I wonder if it's even worth mentioning this distinction, given that we don't really support non-QC tests anymore? And perhaps it would be more accurate to describe the distinction here as being between iasWorld QC tests and non-iasWorld QC tests? I'm a little unsure, curious what you think!

dbt/README.md
```diff
@@ -625,12 +669,160 @@ do so, you have two options:
...
+model during and after export:
+
+* **Tagging**: [Model tags](https://docs.getdbt.com/reference/resource-configs/tags) are not
+  required for QC reports, but they are helpful in cases where we need to export more than one
+  report at a time. For example, Valuations typically requests all of the town close QC
+  reports at the same time, so we tag each model with the `qc_report_town_close` tag such that
+  we can select them all at once when running the `export_models` script using
+  `--select tag:qc_report_town_close`. For consistency, prefer tags that start with the `qc_*`
+  prefix, but beware not to use the `test_qc_*` prefix, which is instead used for [QC
+  tests](#adding-qc-tests).
```
jeancochrane, line 766: The distinction between these two tag formats strikes me as a little bit confusing... Any ideas for a more intuitive tagging scheme?

dbt/README.md
```diff
@@ -625,12 +669,160 @@ do so, you have two options:
...
+    * **`meta.export_name`**: The base name that the script will use for the output file, not
+      including the file extension. The script will output the file to
+      `dbt/export/output/{meta.export_name}.xlsx`.
+    * **`meta.export_template`**: A path to an Excel template to populate with data, relative
+      to the [`dbt/export/templates/`
+      directory](https://github.com/ccao-data/data-architecture/tree/master/dbt/export/templates).
+      The script will read this template from `dbt/export/templates/{meta.export_template}`.
+      In contrast to `meta.export_name`, this option _does_ expect you to include the file
+      extension in the template name, to allow you more flexibility over the format of the
+      template files you use. Templates are useful if you want to apply custom headers,
```
jeancochrane, line 782: I wonder if it would be simpler if we just standardized both configs as requiring or not requiring file extensions? It doesn't feel super urgent to me since I doubt most reports will use both of these options anyway, but I'm curious what you think.

dbt/README.md
```diff
@@ -494,63 +468,142 @@ We use the following pattern to determine where to define each column descriptio
...
 
-### Data tests vs. unit tests
+- `default.vw_pin_sale` - Rebuild a single view
+- `default.vw_pin_sale default.vw_pin_universe` - Rebuild two views at once
+- `+default.vw_pin_history` - Rebuild a view and all its upstream dependencies
+- `location.*` - Rebuild all views under the `location` schema
+- `path:models` - Rebuild the full DAG (:warning: takes a long time!)
 
-There are two types of tests that we might consider for a model:
+For more possible inputs using dbt node selection, see the [documentation site](https://docs.getdbt.com/reference/node-selection/syntax#examples).
```
jeancochrane, line 493: This section shouldn't have changed at all, I just moved it from a different location.

dbt/README.md
dfsnow: AFAIK the non-QC tests are only run in the context of dbt builds. I think that's useful and worth keeping around, as there may be non-unit, non-QC tests that we want to perform related to table relations between stuff not in iasWorld. 

I agree this is pretty confusing though. Building off my earlier thought, what if renamed things as:

- **QC tests** --> **Data Integrity (DI) tests**
- **Non-QC tests** --> **Integration tests**
- **Unit tests** (unchanged)
- **QC reports** (unchanged)

dbt/README.md
```diff
@@ -494,63 +468,142 @@ We use the following pattern to determine where to define each column descriptio
...
+assigned the test to you and ask them when and how the test should be run so that its
+attributes can be set accordingly.
+
+#### Running QC tests
+
+There are two instances when QC tests typically run:
+
+1. Once per day by the [`test-dbt-models` GitHub
+   workflow](https://github.com/ccao-data/data-architecture/actions/workflows/test_dbt_models.yaml),
+   which pushes parquet output to S3 in order to support our analysis of test failures over time
```
dfsnow, line 559:
```suggestion
   which pushes Parquet output to S3 in order to support our analysis of test failures over time
```

dbt/README.md
```diff
@@ -494,63 +468,142 @@ We use the following pattern to determine where to define each column descriptio
...
+2. On demand by a Data team member whenever Tia, Tom, or another Valuations staff member requests
+   a copy of the Excel workbook for a township, usually right before the town closes
+
+Since the first instance is a scheduled job that requires no intervention, the following
+steps describe how to respond to a request from Tia, Tom, or other Valuations staff for
+a fresh copy of the test failure output before town closing.
```
dfsnow, line 565: **todo (blocking):** Let's just use "Valuations staff member" instead of names.

dbt/README.md
```diff
@@ -494,63 +468,142 @@ We use the following pattern to determine where to define each column descriptio
...
-and their output is interpreted by the [`transform_dbt_test_results`
-script](https://github.com/ccao-data/data-architecture/blob/master/.github/scripts/transform_dbt_test_results.py).
-This script reads the metadata for a test run and outputs an Excel
-workbook with detailed information on each failure to aid in resolving
-any data problems that the tests reveal.
+Typically, Valuations staff will ask for test output for a specific township. We'll refer to the
+[township code](https://github.com/ccao-data/wiki/blob/master/Data/Townships.md) for this township
+using the bash variable `$TOWNSHIP_CODE`.
+
+First, run the tests using dbt and the [QC test
```
dfsnow, line 571:
```suggestion
First, run the tests locally using dbt and the [QC test
```

dbt/README.md
```diff
@@ -571,7 +624,7 @@ the tag `test_qc_iasworld`
     * `category` (optional): A workbook category for the test, required if
       a category is not defined for the test's generic in the `TEST_CATEGORIES`
       constant in the [`transform_dbt_test_results`
-      script](https://github.com/ccao-data/data-architecture/blob/master/.github/scripts/transform_dbt_test_results.py)
+      script](https://github.com/ccao-data/data-architecture/blob/master/dbt/scripts/transform_dbt_test_results.py)
```
dfsnow, line 627: **nitpick (non-blocking):** A super minor thing, but IMO relative URLs are better for things inside the repo, just in case the path/owner ever changes. It also lets you navigate the web UI correctly while in a branch other than master.

dbt/README.md
dfsnow: Following up from above, how about something like:

- QC tests --> Data Integrity (DI) tests = `test_di...`
- Non-QC tests --> Integration tests = `test_int...`
- Unit tests (unchanged) = `test_unit...`
- QC reports (unchanged) = `report_qc...`

With the expectation that we may extend the report tag to encompass other types of exports and reporting.

dbt/README.md
dfsnow: What's the context in which we use/will use template formats other than `.xlsx`?

dfsnow (commented):
**praise:** Great work @jeancochrane! This is definitely needed since our testings situation seems to grow by the day. Very clear docs and examples.

dbt/README.md
jeancochrane: Great call, I switched all of the links to code in the README to use relative paths in 7d20c44b3e20d39367c3f026fd1d350c645ecda5.

dbt/README.md
jeancochrane: Makes sense, I switched up both references in 2571113f66dd2df01dffec0599df7d79dff89bcd.

dbt/README.md
jeancochrane: Hm, good point -- I can't think of one. I switched things up in d42e97edd1d9e8a65a4bb18e57779173c79a5894 so that `meta.export_template` doesn't expect a file extension, either.

dbt/README.md
```diff
@@ -494,63 +468,142 @@ We use the following pattern to determine where to define each column descriptio
...
+tests](https://docs.getdbt.com/docs/build/data-tests#singular-data-tests).
 
-There are two types of data tests that we support:
+There are two subtypes of data tests that we support:
 
 1. **QC tests** confirm our assumptions about iasWorld data and are run at
    scheduled intervals to confirm that iasWorld data meets spec
 2. **Non-QC tests** confirm all other assumptions about data sources outside
    of iasWorld, and are run in an ad hoc fashion depending on the needs of
    the transformations that sit on top of the raw data
```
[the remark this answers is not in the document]
jeancochrane: I switched to a new naming scheme in [db596f4](https://github.com/ccao-data/data-architecture/pull/568/commits/db596f4e857a68c5713794d2d976254657942f88) that resolves this suggestion!

dbt/README.md
jeancochrane: I switched to a slightly different naming scheme in [db596f4](https://github.com/ccao-data/data-architecture/pull/568/commits/db596f4e857a68c5713794d2d976254657942f88) that we talked about yesterday, namely consolidating "QC tests" and "non-QC tests" under a unified "data tests" category. Currently working on a follow-up https://github.com/ccao-data/data-architecture/pull/571 that will handle renaming the tags, since that touches a lot of code.

jeancochrane:
@dfsnow This should be ready for another look while I work on refactoring the iasWorld data test tags in https://github.com/ccao-data/data-architecture/pull/571!

dbt/README.md
```diff
@@ -523,18 +530,24 @@ tests](https://docs.getdbt.com/best-practices/writing-custom-generic-tests),
...
-transformed for analysis using the [`transform_dbt_test_results`
-script](https://github.com/ccao-data/data-architecture/blob/master/dbt/scripts/transform_dbt_test_results.py).
+Currently our primary use of data tests is to check assumptions about iasWorld data.
+We refer to this set of tests as "iasWorld data tests", and we've built a system
+for running and interpreting them that we will explain in the sections to follow.
+Other types of data tests do exist, and we primarily run them via automated
+GitHub workflows during CI when models change. However, we anticipate that in
+the future we will likely build out similar infrastructure for running and
+interpreting non-iasWorld data tests to accompany the infrastructure we have
+built for iasWorld data tests.
```
dfsnow, line 540: Woops, I just made a comment on #571 asking to add _when_ we run tests which tests and you've basically already done that here, so nevermind.

dbt/README.md
```diff
@@ -523,18 +530,24 @@ tests](https://docs.getdbt.com/best-practices/writing-custom-generic-tests),
...
-   scheduled intervals to confirm that iasWorld data meets spec
-2. **Non-QC tests** confirm all other assumptions about data sources outside
-   of iasWorld, and are run in an ad hoc fashion depending on the needs of
-   the transformations that sit on top of the raw data
-
-**QC tests** can be run using the [`dbt test`
-command](https://docs.getdbt.com/reference/commands/test) and their output can be
-transformed for analysis using the [`transform_dbt_test_results`
-script](https://github.com/ccao-data/data-architecture/blob/master/dbt/scripts/transform_dbt_test_results.py).
+Currently our primary use of data tests is to check assumptions about iasWorld data.
```
dfsnow, line 533:
```suggestion
Currently, our primary use of data tests is to check assumptions about iasWorld data.
```

dfsnow (approved):
Looks great!

PR #571, opened 2024-08-14
Rename tag for iasWorld data tests and refactor script to run tests

jeancochrane:
This PR implements a two changes to simplify our iasWorld data test suite:

1. Consolidates a `data_test_iasworld` tag from three different `test_qc_*` tags, and renames associated tags and variables to follow this pattern
2. Renames the `transform_dbt_test_results` script to `run_iasworld_data_tests`, and refactors it to run tests in addition to transforming the output

Arguably these are two different changes that merit two different PRs, but I consolidated them into one since the tag refactor was so simple that it didn't seem worth the extra complexity of a third PR. However, this results in a large number of edited files, most of which are small changes for renaming tags and variables for existing models and tests. Let me know if you'd strongly prefer me to split this into two PRs and I can take some time to do so.

In addition to testing a full run of `run_iasworld_data_tests`, I tested the following conditions to make sure that the new logic in the script works:

<details>
<summary>Run against just iasworld_addn_class_in_ccao_class_dict where the test fails</summary>
<br>
<pre>
$ python3 scripts/run_iasworld_data_tests.py
Running tests
> dbt test --select iasworld_addn_class_in_ccao_class_dict --store-failures
19:20:39  Running with dbt=1.8.3
19:20:40  Registered adapter: athena=1.8.2
19:20:41  Found 103 models, 9 seeds, 437 data tests, 138 sources, 10 exposures, 501 macros
19:20:41
19:20:48  Concurrency: 16 threads (target='dev')
19:20:48
19:20:48  1 of 1 START test iasworld_addn_class_in_ccao_class_dict ....................... [RUN]
19:20:55  1 of 1 FAIL 7 iasworld_addn_class_in_ccao_class_dict ........................... [FAIL 7 in 7.23s]
19:20:55
19:20:55  Finished running 1 test in 0 hours 0 minutes and 14.22 seconds (14.22s).
19:20:56
19:20:56  Completed with 1 error and 0 warnings:
19:20:56
19:20:56  Failure in test iasworld_addn_class_in_ccao_class_dict (models/iasworld/schema/iasworld.addn.yml)
19:20:56    Got 7 results, configured to fail if != 0
19:20:56
19:20:56    compiled code at target/compiled/ccao_data_athena/models/iasworld/schema/iasworld.addn.yml/iasworld_addn_class_in_ccao_class_dict.sql
19:20:56
19:20:56    See test failures:
  ----------------------------------------------------------------------------------------------------
  select * from "awsdatacatalog"."z_dev_jecochr_test_failure"."iasworld_addn_class_in_ccao_class_dict"
  ----------------------------------------------------------------------------------------------------
19:20:56
19:20:56  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
Loading test results from Athena
Querying failed rows from "awsdatacatalog"."z_dev_jecochr_test_failure"."iasworld_addn_class_in_ccao_class_dict"
Saving test results to the cache at test_cache/run_4ce66331a4330163df3b84cbc23b2018_manifest_7cbfdf61df79bf9f1478fc4a88cf4421.json
Generating the output workbook
Adding sheet for class_mismatch_or_issue
Output workbook saved to iasworld_test_results_2024-08-14/iasworld_test_failures_2024-08-14.xlsx
test_run metadata saved to iasworld_test_results_2024-08-14/metadata/test_run/
test_run_result metadata saved to iasworld_test_results_2024-08-14/metadata/test_run_result/
test_run_failing_row metadata saved to iasworld_test_results_2024-08-14/metadata/test_run_failing_row/
</pre>
</details>

<details>
<summary>Run against just iasworld_addn_class_in_ccao_class_dict where the test has a syntax error</summary>
<br>
<pre>
$ python3 scripts/run_iasworld_data_tests.py
Running tests
> dbt test --select iasworld_addn_class_in_ccao_class_dict --store-failures
19:15:37  Running with dbt=1.8.3
19:15:38  Registered adapter: athena=1.8.2
19:15:39  Found 103 models, 9 seeds, 437 data tests, 138 sources, 10 exposures, 501 macros
19:15:39
19:15:46  Concurrency: 16 threads (target='dev')
19:15:46
19:15:46  1 of 1 START test iasworld_addn_class_in_ccao_class_dict ....................... [RUN]
19:15:50  1 of 1 ERROR iasworld_addn_class_in_ccao_class_dict ............................ [ERROR in 4.05s]
19:15:50
19:15:50  Finished running 1 test in 0 hours 0 minutes and 11.66 seconds (11.66s).
19:15:50
19:15:50  Completed with 1 error and 0 warnings:
19:15:50
19:15:50    Runtime Error in test iasworld_addn_class_in_ccao_class_dict (models/iasworld/schema/iasworld.addn.yml)
  COLUMN_NOT_FOUND: line 17:121: Column 'jafskldfjlksdjf' cannot be resolved or requester is not authorized to access requested resources. You may need to manually clean the data at location 's3://ccao-dbt-athena-dev-us-east-1/results/tables/da866396-3eb1-4ec0-ac1c-363d087cb923' before retrying. Athena will not delete data in your account.
19:15:50
19:15:50  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
Traceback (most recent call last):
  File "/home/jecochr/code/data-architecture/dbt/scripts/run_iasworld_data_tests.py", line 1494, in <module>
    main()
  File "/home/jecochr/code/data-architecture/dbt/scripts/run_iasworld_data_tests.py", line 779, in main
    raise ValueError("Quitting due to error in dbt test run")
ValueError: Quitting due to error in dbt test run
</pre>
</details>

<details>
<summary>Run against just iasworld_addn_class_in_ccao_class_dict where the dbt CLI command has a syntax error</summary>
<br>
<pre>
$ python3 scripts/run_iasworld_data_tests.py
Running tests
> dbt test --selectasdf iasworld_addn_class_in_ccao_class_dict --store-failures
Traceback (most recent call last):
  File "/home/jecochr/code/data-architecture/dbt/scripts/run_iasworld_data_tests.py", line 1494, in <module>
    main()
  File "/home/jecochr/code/data-architecture/dbt/scripts/run_iasworld_data_tests.py", line 771, in main
    raise dbt_test_result.exception
dbt.cli.exceptions.DbtUsageException: No such option: --selectasdf
</pre>
</details>


.github/workflows/build_and_test_dbt.yaml
```diff
@@ -84,14 +84,14 @@ jobs:
           if [[ $MODIFIED_RESOURCES_ONLY == 'true' ]]; then
             if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
               echo "Running tests on manually selected resources"
-              dbt test -t "$TARGET" -s ${{ inputs.models }} --exclude "tag:test_qc*" --defer --state "$STATE_DIR"
+              dbt test -t "$TARGET" -s ${{ inputs.models }} --exclude "tag:data_test_iasworld" --defer --state "$STATE_DIR"
```
jeancochrane, line 87: Note that I decided on `data_test_iasworld` instead of `test_data_iasworld` because I doubt we'll need to tag our unit tests in any specific way such that we want `test_*` to be a pattern that encompasses `test_data_*` and `test_unit_*`. In the absence of that need, `data_test_*` feels clearer to me, although that's a subjective choice. I don't feel super strongly about and would be happy to switch up the tag if you have a strong opinion.

.github/workflows/test_dbt_models.yaml
```diff
@@ -23,37 +23,13 @@ jobs:
...
-              # failed due to an error, so print the output and fail the
-              # pipeline
-              echo "$output"
-              exit "$status_code"
-            else
-              # The tests must have failed rather than errored out, so
-              # print the output but don't raise an error
-              echo "$output"
-            fi
-          fi
```
jeancochrane, line 50: Now that we're running tests in the `run_iasworld_data_tests` script rather than requiring the script caller to know how to run tests separately, we can reassign ownership over this error logic to the script from the workflow. Ultimately I think this results in simpler code, since the Python utilities for running dbt commands have a clearer set of attributes for distinguishing test failures from test errors (which we'll see shortly).

dbt/models/qc/qc.vw_nonlivable_condos_with_chars.sql
```diff
@@ -20,6 +20,3 @@ LEFT JOIN {{ source('spatial', 'township') }} AS towns
 WHERE
     (condos.is_common_area OR condos.is_parking_space)
     AND (condos.char_bedrooms IS NOT NULL OR condos.char_full_baths IS NOT NULL)
-    AND CAST(
-        condos.year AS INT
-    ) BETWEEN {{ var('test_qc_year_start') }} AND {{ var('test_qc_year_end') }}
```
jeancochrane, line 25: This is an unrelated problem I decided to fix while changing references to `test_qc_year_*` variables: Previously this QC report view was relying on hardcoded values for the year filter, which could only change if the view were rerun. To bring it in line with our other QC reports, I'm removing the year filter and requiring us to apply the filter when calling the `scripts/export_models.py` script. Heads up to @wrridgeway who is the de facto owner of this view.

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -693,33 +700,13 @@ def main() -> None:
...
-    parser.add_argument(
-        "--manifest",
-        required=False,
-        default=os.path.join("target", "manifest.json"),
-        help=(
-            "The local path to a manifest.json file with the compiled dbt "
-            "DAG, generated by all dbt CLI commands (including `dbt test`). "
-            "Defaults to './target/manifest.json'."
-        ),
-    )
```
jeancochrane, line 715: Since the script is now responsible for running tests, we no longer need to expose config params to point to the location of the run results and manifest files.

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -1087,9 +1114,15 @@ def get_test_cache_path(
...
-    The `run_results_filepath` and `manifest_filepath` are used to generated
-    a hash key that uniquely defines the cache key for a given test run."""
+    """Return the path to the cache where test results are stored, or an
+    empty string if no cache exists yet. The `run_results_filepath` and
+    `manifest_filepath` are used to generated a hash key that uniquely defines
+    the cache key for a given test run."""
+    if not os.path.isfile(run_results_filepath) or not os.path.isfile(
+        manifest_filepath
+    ):
+        return ""
```
jeancochrane, line 1124: I'm not super confident in this interface, although I don't know if getting it exactly right is high priority. One alternative would be to raise an error if one or both of these files is missing, since we should only ever call this function under one of two conditions:

1. The caller wants to read from the cache for a previous run and has instructed the script to do so by passing in the `--use-cached` param, in which case we the script is already configured to error if the cache does not exist
2. The script wants to write to the cache for the most recent run, in which case it will have always already completed a dbt run in the immediate past

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -733,35 +720,73 @@ def main() -> None:
...
+        if dbt_test_result.exception is not None:
+            raise dbt_test_result.exception
+
+        if any(
+            result.status == TestStatus.Error
+            for result in getattr(dbt_test_result.result, "results", [])
+        ):
+            # No need to report the exception, since the dbt process
+            # will have printed it already
+            raise ValueError("Quitting due to error in dbt test run")
```
jeancochrane, line 778: This block replaces the old `Run tests` step in the `test-dbt-models` workflow, including the error handling.

.github/workflows/build_and_test_dbt.yaml
dfsnow: That works for me!

.github/workflows/test_dbt_models.yaml
dfsnow: **question (non-blocking):** Is the output still shown in the runner? I assume `dbt.invoke()` shows that output but I don't see a test run.

dbt/scripts/run_iasworld_data_tests.py
dfsnow: Yeah this is way way better. <3 Python

dbt/scripts/run_iasworld_data_tests.py
dfsnow: I think that's fine since it's just for local testing purposes!

dbt/selectors.yml
```diff
@@ -1,15 +1,15 @@
 selectors:
-  - name: qc_tests
-    description: Selector for running QC tests on iasWorld tables
+  - name: data_test_iasworld
```
dfsnow, line 2: **nitpick (non-blocking):** I find it pretty confusing to have the selector named the same thing as the tag. Maybe we can prefix it with something?

dfsnow (commented):
So just so I'm clear, we'll now have the following nomenclature:

- Data tests which encompass:
  - `data_test_iasworld` - iasWorld specific tests that run every morning after sqoop
  - Other data tests that run during dbt builds? Test aren't actually tagged though
- Unit tests (TBD), which run on dbt builds
- QC reports, built from `qc_report*` tags

Is that right? IMO it would be helpful to have in the README under each type _when_ we expect them to run.

Otherwise, this looks great! Thanks for tidying things up.

dbt/selectors.yml
jeancochrane: Great point, I renamed the selector to `select_data_test_iasworld` in b61a761a39614c77d8c853e5b0dc7ef1b7de27e0.

.github/workflows/test_dbt_models.yaml
jeancochrane: This was a good reminder for me to test a workflow run, which I hadn't done yet (I'd only tested locally) 😅 The `dbt.invoke()` call does indeed print logs in real time, which you can see here: https://github.com/ccao-data/data-architecture/actions/runs/10407263825/job/28822148281#step:5:30

jeancochrane: See here for a successful workflow run: https://github.com/ccao-data/data-architecture/actions/runs/10407263825/job/28822628961

jeancochrane:
@dfsnow Based on https://github.com/ccao-data/data-architecture/pull/568#discussion_r1717620739, it seems like you're actually satisfied with the docs explaining when each type of test gets run, is that right? If not, I'm happy to take a stab at clarifying things -- otherwise, let me know if this looks good to merge!

dfsnow:
> @dfsnow Based on [#568 (comment)](https://github.com/ccao-data/data-architecture/pull/568#discussion_r1717620739), it seems like you're actually satisfied with the docs explaining when each type of test gets run, is that right? If not, I'm happy to take a stab at clarifying things -- otherwise, let me know if this looks good to merge!

Whoops I missed this. Yes, looks good! Let's merge it.

dfsnow (approved):

PR #573, opened 2024-08-18
Add a proximity to stadium feature

Damonamajor:
This updates the proximity to stadium feature. It does four things -

1. Changes spatial.stadium connection to spatial.stadium_raw from source to reference. 
2. Makes sure that stadium years are coded in 2000 or after, properly allowing them to join to spatial.parcel
3. Makes sure that spatial.stadium has a full set of observations whenever new stadiums are added. For example, SeetGeek stadium is constructed in 2006. Thus, the stadiums which are built before that also have observations in 2006.
4. Adds documentation for points 2 and 3


dbt/seeds/spatial/spatial.stadium_raw.csv
dfsnow, line 1: **issue (blocking):** We should actually be able to keep the years in the original seed, then use the `spatial.stadium` model to "expand" the seed such that there is one row per year per stadium.

dbt/models/spatial/spatial.stadium.sql
```diff
@@ -2,9 +2,29 @@
...
-SELECT
-    stadium.name,
-    ST_ASBINARY(ST_POINT(stadium.lon, stadium.lat)) AS geometry,
-    ST_ASBINARY(ST_POINT(stadium.x_3435, stadium.y_3435)) AS geometry_3435,
-    CAST(stadium.year AS VARCHAR) AS year
-FROM {{ ref('spatial.stadium_raw') }} AS stadium
+WITH distinct_years AS (
+    SELECT DISTINCT CAST(year AS INTEGER) AS year
+    FROM {{ ref('spatial.stadium_raw') }}
+),
```
dfsnow, line 8: **issue (blocking):** This will only create records for years which exist in the original seed data, in the current case 2000, 2006, and 2017. You should amend this to create a record for all years between the min year present in the seed data and the max year of parcel data. See this [SO answer](https://stackoverflow.com/a/34994136) for a possible way to do this.

dfsnow (commented):
Good work catching the little bug that requires a full set of years to create proximity records! This is a step in the right direction toward a fix, but it still needs a couple fixes.

dbt/models/spatial/spatial.stadium.sql
```diff
@@ -2,9 +2,30 @@
...
+)
+
+SELECT DISTINCT
+    stadium_years.name,
+    CAST(stadium_years.year AS VARCHAR) AS year,
+    ST_ASBINARY(ST_POINT(stadium_years.lon, stadium_years.lat)) AS geometry,
+    ST_ASBINARY(ST_POINT(stadium_years.x_3435, stadium_years.y_3435))
+        AS geometry_3435
+FROM stadium_years
+ORDER BY year, name;
```
dfsnow, line 31: **nitpick:** Try not to include semi-colons in your queries, as sometimes they break the composability of queries in dbt.

```suggestion
ORDER BY year, name
```

dbt/seeds/spatial/docs.md
```diff
@@ -2,7 +2,9 @@
 
 {% docs seed_stadium_raw %}
 A list of stadiums and their latitude and longitude, manually constructed
-by CCAO staff
+by CCAO staff.
+All stadiums are coded to an earliest year of 2000 to ensure that the
+macro dist_to_nearest_geometry works properly.
```
dfsnow, line 7: **issue:** Gotta update this text still.

dfsnow (approved):
The query here looks right now. Once the minor nitpicks from the comments are fixed this is good to go.

Issue #209, opened 2023-11-01
Add GitHub Action to check dbt YAML key sort order


dfsnow:
Now that our dbt documentation is largely finished, it's critical that we keep it organized and up-to-date. To ease in organization and maintenance of the the documentation files, I propose we add a quick GitHub Actions workflow that checks the sorting of certain dbt arrays and markdown headers.

Specifically, we should check the that all array keys in the `columns:` and `tests:` items are _sorted alphabetically_. These items are located in each database's `schema.yml` file or the specific table `.yml` file for databases broken out by table.

We should also check that each `docs.md`, `columns.md`, and `shared_columnd.md` file has alphabetically sorted markdown headings.

## Tasks

- [x] Create a Python script to check the sort order of `columns:` and `tests:` array items. Can likely use PyYAML for this pretty easily.
- [ ] Create a Python script to check the sort order of `docs.md`, `columns.md`, and `shared_columns.md`. 
- [x] Create a GitHub Actions workflow to run both scripts.
- [x] Alphabetize any incorrect array keys or markdown headings.
- [x] Update the dbt README with alphabetization rules.

PR #578, opened 2024-08-27
Add GitHub action to check dbt yaml key sort order

wagnerlmichael:
This PR takes care of  https://github.com/ccao-data/data-architecture/issues/209. We add one action which checks for each set of files in the issue checklist. I set up what I think to be a pretty readable output that tells the user what is failing and where when there is data that is out of order. 

[Here](https://github.com/ccao-data/data-architecture/actions/runs/10709049327/job/29692780359?pr=578) is an example. At the bottom we have each file with unsorted data and the number of times it occurs per list that needs to be sorted - `file.yaml (# lists needed to be corrected)`. Above that we see the full list that is incorrect and arrows pointing to out of order observations.
```
. . .
. . .
----------------------------------------
In file: dbt/models/sale/docs.md
Headings not sorted:
- # flag
- # foreclosure
- # parameter
---> # group_mean
- # metadata
- # mydec
- # vw_ias_salesval_upload
----------------------------------------
The following files have unsorted columns:
dbt/models/ccbor/schema.yml (1)
dbt/models/iasworld/schema/iasworld.dweldat.yml (1)

The following files have unsorted data tests:
dbt/models/iasworld/schema/iasworld.dweldat.yml (3)

The following markdown files have unsorted headings:
dbt/models/other/docs.md (1)
dbt/models/sale/docs.md (1)

The following columns.md files have unsorted '##' headings:
dbt/models/proximity/columns.md (1)

The following shared_columns.md files have unsorted headings:
dbt/models/shared_columns.md (1)
```

The is a huge diff here but that is mostly due to me sorting and re-configuring yaml anchors to get the sorting fixed. The key files are `dbt/scripts/check_sort_dbt_yaml_files.py` and `.github/workflows/check_sort_dbt_yaml_files.yaml`

`dbt/scripts/check_sort_dbt_yaml_files.py` has a bunch of functions that handle different files to sort, and they are all called in `check_all_files()`


dbt/scripts/check_sort_column_names.py
```diff
@@ -0,0 +1,120 @@
...
+    error_files = []
+    for root, _, files in os.walk(directory):
+        if "venv" in root:  # Skip virtual environment directories
+            continue
+        for file in files:
+            if file.endswith(".yaml") or file.endswith(".yml"):
+                file_path = os.path.join(root, file)
+                # Temporary to solve the models/model/schema.yaml problem, there are
+                # different partitions of columns under the columns: key
+                # Not sure if it is worth it to build logic to handle that
```
wagnerlmichael, line 85: The column list in this file is partitioned into different sets of column within the same `columns:` key. This makes the alphanumeric sorting quite complicate. For now my most simple solution is just to remove these files from consideration. I'm curious to hear thoughts on this.

dbt/scripts/check_sort_column_names.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def check_columns(data, file_path, unsorted_files_dict, parent_key=None):
+        if isinstance(data, dict):
+            for key, value in data.items():
+                if key == "columns":
+                    column_names = [
+                        col.get("name")
+                        for col in value
+                        if isinstance(col, dict) and "name" in col
+                    ]
```
wagnerlmichael, line 42: Look at the `name` key value under `column` to grab the comparble list info

dbt/scripts/check_sort_data_tests.py
```diff
@@ -0,0 +1,123 @@
...
+            for key, value in data.items():
+                if key == "data_tests" and isinstance(value, list):
+                    data_test_names = []
+                    for test in value:
+                        if isinstance(test, dict):
+                            for test_type, test_details in test.items():
+                                if (
+                                    isinstance(test_details, dict)
+                                    and "name" in test_details
+                                ):
```
wagnerlmichael, line 46: Grab `name` two levels under data_tests to be able to sort:

```yaml
data_tests:
  - not_accepted_values:
      name: default_vw_pin_exempt_class_no_hyphens
```

dbt/scripts/check_sort_column_names.py
```diff
@@ -0,0 +1,120 @@
...
+        for file in files:
+            if file.endswith(".yaml") or file.endswith(".yml"):
+                file_path = os.path.join(root, file)
+                # Temporary to solve the models/model/schema.yaml problem, there are
+                # different partitions of columns under the columns: key
+                # Not sure if it is worth it to build logic to handle that
+                if file_path == os.path.join(
+                    "dbt", "models", "model", "schema.yml"
+                ):
+                    continue
```
jeancochrane, line 89: **[Suggestion, non-blocking]** Maybe we could print something about skipping any files that we configure the script to skip?

dbt/scripts/check_sort_column_names.py
```diff
@@ -0,0 +1,120 @@
...
+    if error_files:
+        print("\nThe following files could not be processed due to errors:")
+        for file in error_files:
+            print(file)
+
+    if unsorted_files_dict or error_files:
+        exit(1)  # Exit with a status code of 1 to indicate failure
+    else:
+        print("All files have sorted columns and no errors.")
+        exit(0)  # Exit with a status code of 0 to indicate success
```
jeancochrane, line 120: **[Suggestion, non-blocking]** Unfortunately, [Python system exit commands can have complicated side effects](https://stackoverflow.com/a/19747562)! A more typical pattern is to rely on exceptions for similar behavior:

```suggestion
    if unsorted_files_dict or error_files:
        raise ValueError(
            "Column name sort order check ran into failures, see logs above"
        )
    print("All files have sorted columns and no errors.")
```

jeancochrane (commented):
This is a great start! Thanks for doing the hard work to fix all of the existing problems. A few thoughts below on some improvements; so far the only blocking feedback for me is a few column entries that seem to have gotten lost during reordering.

I think your instinct is correct: We should eliminate the shared code between the two checks by consolidating them into one. This will make the `check_yaml_file` logic more complex, but it'll reduce the duplication in the code that supports that function, and I think it'll make things more maintainable going forward. So far I've only taken a quick glance at the code in both files, but request another review from me once you've refactored things to consolidate them, and I'll give the combined workflow a detailed pass!

dbt/scripts/check_sort_dbt_yaml_files.py
```diff
@@ -0,0 +1,465 @@
...
+    """
+    try:
+        with open(file_path, "r") as file:
+            first_line = file.readline().strip()
+            if first_line == "# disable-check-sort-order":
+                return {}, []
+            data = yaml.safe_load(file)
+    except yaml.YAMLError as error:
+        return {}, [f"Error processing {file_path}: {error}"]
+
```
wagnerlmichael, line 68: Add top level comment which ignores file

dbt/scripts/check_sort_dbt_yaml_files.py
```diff
@@ -0,0 +1,437 @@
...
+        unsorted_data_tests_files,
+        error_files,
+        unsorted_md_files,
+        unsorted_columns_md_files,
+        unsorted_shared_columns_md_files,
+    )
+
+
+if __name__ == "__main__":
+    directory = "dbt/"
```
jeancochrane, line 384: **[Suggestion, non-blocking]** Per my suggestion above, I think we should call this script from the context of the `dbt` subdirectory, which would allow us to remove this hardcoded directory and just check all files in the current working directory.

jeancochrane (commented):
Looking great! A couple more blocking changes, and a few small suggestions. Overall I think this is getting close, though.

The Python script seems like it has a bunch of opportunities for factoring out shared code; in particular, the YAML checks strike me as very similar, and the markdown checks share a lot of code, too. I think it's not worth spending a lot of time on right now, but it might be a useful exercise to take a look at the functions and think about how we could consolidate them in the future.

dbt/scripts/check_sort_dbt_yaml_files.py
```diff
@@ -0,0 +1,437 @@
...
+        file_path (str): The path to the YAML file to check.
+
+    Returns:
+        tuple: A dictionary with unsorted files and a list of errors encountered.
+    """
+    try:
+        with open(file_path, "r") as file:
+            data = yaml.safe_load(file)
+    except yaml.YAMLError as error:
+        return [error], [file_path]
```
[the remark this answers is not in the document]
wagnerlmichael: Added comments explaining this

dbt/models/proximity/columns.md
```diff
@@ -34,7 +34,7 @@ Estimated DNL for a PIN, assuming a baseline DNL of 50 ("quiet suburban") and
 adding predicted noise from O'Hare and Midway airports to that baseline.
 
 For more on DNL, see
-https://www.faa.gov/regulations_policies/policy_guidance/noise/basics
+<https://www.faa.gov/regulations_policies/policy_guidance/noise/basics>
```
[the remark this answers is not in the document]
wagnerlmichael: Interestingly, when I remove the them and save the file they are automatically added back. Perhaps is a weird vscode config I have?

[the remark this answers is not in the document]
jeancochrane: Huh, yeah, maybe! Either way, I don't mind leaving it like this.

wagnerlmichael:
I believe I added all of the comments. This PR is starting to get a bit long so let me know if I missed anything. I started on a python script refactor but realized it would take me a bit of time and testing so I didn't go through with it

dbt/scripts/check_sort_dbt_yaml_files.py
```diff
@@ -61,6 +61,11 @@ def check_columns(file_path):
     try:
         with open(file_path, "r") as file:
             first_line = file.readline().strip()
+            # This disregards files within the check_columns() function
+            # that have this tag at the top. This is limited to yaml
+            # files with column sorting, if we want to expand we will
+            # have to generalize it or manually add a check to the
+            # other functions
```
jeancochrane, line 68: **[Nitpick, blocking]** Hmm, I don't think this solves the `file.seek(0)` issue I was thinking about in https://github.com/ccao-data/data-architecture/pull/578/files/8fffe4512174dfb754ef13d9e01d3f8acedc69b7..e7c6802ce0c85725645320a644be6be41a44f8c0#r1746019955. Maybe we can discuss on Monday?

dbt/models/location/schema.yml
```diff
@@ -241,12 +241,7 @@ models:
 
     data_tests:
       # Unique by 10-digit PIN and year
-      - unique_combination_of_columns:
-          name: location_vw_pin10_location_unique_by_keys
-          combination_of_columns:
-            - pin10
-            - year
-      # GEOIDs are the correct length
+            # GEOIDs are the correct length
```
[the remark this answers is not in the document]
jeancochrane: @wagnerlmichael Seeems like this stray indent is still here, right?

jeancochrane (approved):
I found one small piece of feedback that I think got missed on the last pass, but otherwise this is good to go!

PR #579, opened 2024-09-04
Add framework for instant email notifications of failing tests

jeancochrane:
This PR adds support for a new `meta.notify` attribute to our dbt `schema.yml` syntax that can control automated email notifications for failing tests. It introduces a new test `iasworld_pardat_spot_is_null` that uses this system, along with a seed representing existing test failures; when combined, these two features allow us to send realtime notifications when a new value starts failing the test, without spamming us with notifications for known rows that are already failing.

I'll forward you a sample email so that you can see what the notification looks like. [Here](https://github.com/ccao-data/data-architecture/actions/runs/10727647674/job/29750441705) are the logs for the workflow that generated that notification.

Here's an overview for how this system works:

#### Sending a notification for a failing test

1. Get a dbt variable representing an SNS topic for the group of team members we want to email when this test fails
    * If no variable/topic combo exists yet for this group, a few config steps are required:
        1. Create the SNS topic
        2. Define a new empty dbt variable in `dbt_project.yml` that will store the ARN for the topic
        3. Set the ARN to a repo secret so that it is hidden but we can access it from GitHub workflows
        4. Update the `scripts/run_iasworld_data_tests.py` call in the `test-dbt-models` workflow to set the dbt variable you created in step 2 to the secret you created in step 3
    * If a variable/topic already exists for this group, no changes are necessary; just grab its dbt variable name
2. Update the test definition to add a `meta.notify` attribute whose value is the dbt variable name from step 1
3. When the `test-dbt-models` workflow runs, it will send a notification for any failures to the subscribers of the SNS topic from step 1

#### Restricting failure notifications to remove known failures

1. Add a seed representing the rows that we already know are failing for a given test
2. Adjust the interface for the generic test to add support for an `anti_join` argument that can join to another model and only return failures that are _not_ already represented in that model
3. Update the test to supply the `anti_join` argument added in step 2 and point it to the seed added in step 1
4. When the test runs, it will only fail for rows that do not have a representation in the seed from step 1

Note that we have sketched out a simplified design for this system in https://github.com/ccao-data/data-architecture/issues/595, but we're planning to merge this version anyway so that we can get started configuring tests for notifications as soon as possible.

### DIFF
--- a/dbt/README.md
+++ b/dbt/README.md
@@ -697,7 +697,7 @@ a major event in the Valuations calendar like the close of a township.
 
 The [`export_models` script](./scripts/export_models.py) is the foundation for
 our QC reports. The script expects certain Python requirements, which can be installed
-by running `pip install scripts/requirements.export_models.txt` in a virtual
+by running `pip install -r scripts/requirements.export_models.txt` in a virtual
 environment.
 
 The script exposes a few options that help to export the right data:


rross0:
This is really cool stuff!

.github/actions/publish_sns_topic/action.yaml
```diff
@@ -15,14 +15,9 @@ inputs:
...
     - name: Publish SNS topic
-      run: >
-        aws sns publish \
-          --topic-arn "${{ inputs.sns_topic_arn }}" \
-          --subject "${{ inputs.subject }}" \
-          --message "${{ inputs.body }}"
+      run: ./${{ github.action_path }}/publish.sh \
+          "${{ inputs.sns_topic_arn }}" \
+          "${{ inputs.subject }}" \
+          "${{ inputs.body }}"
```
jeancochrane, line 22: We want to reuse this SNS publishing logic when tests fail, but we can't dynamically spawn one or more actions from the context of a workflow (as far as I know), so I factored the logic out into a shared script instead.

.github/workflows/test_dbt_models.yaml
```diff
@@ -31,14 +31,16 @@ jobs:
         run: |
           python3 scripts/run_iasworld_data_tests.py \
             --target "$TARGET" \
-            --output-dir ./qc_test_results/
+            --output-dir ./qc_test_results/ \
+            --vars "{\"data_test_iasworld_commercial_sns_topic\": \"$COMMERCIAL_SNS_TOPIC\"}"
```
jeancochrane, line 35: One downside of this approach (setting topic ARNs as vars and overriding them with secrets during workflow execution) is that we'll have to update this bloated JSON dict every time we add a new topic. I don't think it's bad enough to outweigh the benefits of this approach, but it's a downside for sure.

.github/workflows/test_dbt_models.yaml
```diff
@@ -73,3 +75,18 @@ jobs:
...
+
+      - name: Send data test failure notification
+        run: |
+          failure_notifications_file="$PROJECT_DIR/qc_test_results/failure_notifications.json"
+          if [ -f "$failure_notifications_file" ]; then
+            echo "Sending failure notifications"
+            bash <( \
+              jq -r \
+              '.[] | "./.github/actions/publish_sns_topic/publish.sh \(.topic_arn|@sh) \(.subject|@sh) \(.body|@sh)"' \
+              "$failure_notifications_file" \
```
jeancochrane, line 87: This unholy jq command was the simplest way I could think of parsing failures by topic and queuing up each one to send, but I'm open to ideas that will be more understandable to laypeople!

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -145,11 +151,32 @@ def from_dict(cls, result_dict: typing.Dict) -> "TestResult":
...
         return f"TestResult({self.to_dict()!r})"
 
+    @property
+    def details(self) -> str:
+        """Returns a string that can be used to report on the status of
+        this test in detail. Strings are formatted like so:
+
+            - {table_name}: {description} ({status})
+                * {fail1_key1}: {fail1_value1}, {fail1_key2}: {fail1_value2}
+                * {fail2_key1}: {fail2_value1}, {fail2_key2}: {fail2_value2}
```
jeancochrane, line 167: The formatting here is not the most beautiful, but we're limited by the fact that [SNS only supports plaintext](https://stackoverflow.com/a/32243158). We could think about switching to SES if producing a beautiful email is a requirement, though.

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -904,25 +982,12 @@ def create(
...
-        if not var_year_start or not var_year_end:
-            with open("dbt_project.yml") as project_fobj:
-                project = yaml.safe_load(project_fobj)
-            var_year_start = (
-                var_year_start
-                or project["vars"]["data_test_iasworld_year_start"]
-            )
-            var_year_end = (
-                var_year_end or project["vars"]["data_test_iasworld_year_end"]
-            )
```
jeancochrane, line 925: I factored this logic out into a dedicated function so that we can reuse it for extracting values for SNS topic variables.

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -1228,6 +1325,19 @@ def get_test_categories(
...
         column_name = get_column_name_from_node(node)
         test_description = meta.get("description")
+        notify = meta.get("notify")
+        if notify:
+            # We can't use variables directly in dbt meta configs, because the
+            # `var()` function doesn't get called during compilation there.
+            # Instead, interpret the `notify` attribute as a string representing
+            # the name of the variable storing the SNS topic that we need
+            # to publish to
+            notify_var = get_var_value_from_run_results(notify, run_results)
```
jeancochrane, line 1335: This limitation is a huge bummer. Parsing the variables by name (rather than relying on `dbt compile` to build them into the DAG) felt like the best workaround to me, but I'm open to other ideas.

dbt/tests/generic/test_is_null.sql
```diff
@@ -1,13 +1,23 @@
 -- Confirm that a column is null. Essentially, the opposite of the `not_null`
 -- test.
-{% test is_null(model, column_name, additional_select_columns=[]) %}
+{% test is_null(model, column_name, additional_select_columns=[], anti_join={}) %}
```
jeancochrane, line 3: I wish there were a way to support anti-joins without having to update each individual generic to support it, but unfortunately I think this is the best approach. If we decide to roll it out more widely, we could think about taking a similar approach to the one we took with `additional_select_columns`, namely factoring out the shared logic for the param into a dedicated macro.

dbt/tests/generic/test_is_null.sql
```diff
@@ -1,13 +1,23 @@
...
-    where {{ column_name }} is not null
+    from
+        (
+            select model.*
+            from (select * from {{ model }}) as model
+            {% if anti_join -%}
+                left join
+                    {{ anti_join.table_name }} as anti_join
+                    on model.{{ anti_join.join_on }} = anti_join.{{ anti_join.join_on }}
+                where anti_join.{{ anti_join.join_on }} is null
```
jeancochrane, line 18: We could imagine extending `join_on` to support arrays of values too, or dicts if we want to join on two columns that don't have the same name. For now, I'm keeping it as simple as possible, and I figure we can extend that in the future if we need to.

jeancochrane:
Requesting @dfsnow for a preliminary look at the design before I polish things up and add docs.

.github/workflows/test_dbt_models.yaml
dfsnow: This is definitely a little bit cursed. What about doing it in Python? Also, let's add some conditionals on this (like those on L68) so that we don't accidentally spam folks if this workflow is triggered a bunch.

dbt/scripts/run_iasworld_data_tests.py
dfsnow: Let's start simple. IMO the formatting is readable enough.

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -869,6 +909,44 @@ def main() -> None:
...
+        body = "The following iasWorld data tests are failing:"
+        for test_result in test_results:
+            body += f"\n\n{test_result.details}"
+        failure_notifications.append(
+            {
+                "topic_arn": topic_arn,
+                "subject": "iasWorld data tests failed",
+                "body": body,
+            }
+        )
```
dfsnow, line 938: Just so I understand correctly: test failures are grouped by _topic_, such that different failing tests would end up in the same single email ala:

```
           - {table_name}: {description} ({status})
                * {fail1_key1}: {fail1_value1}, {fail1_key2}: {fail1_value2}
           - {table_name2}: {description} ({status})
                * {fail1_key1}: {fail1_value1}, {fail1_key2}: {fail1_value2}
                * {fail2_key1}: {fail2_value1}, {fail2_key2}: {fail2_value2}
```

Is that right?

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -869,6 +909,44 @@ def main() -> None:
...
+    # and `body`
+    failure_notifications: typing.List[typing.Dict] = []
+    for topic_arn, test_results in failures_by_topic.items():
+        body = "The following iasWorld data tests are failing:"
+        for test_result in test_results:
+            body += f"\n\n{test_result.details}"
+        failure_notifications.append(
+            {
+                "topic_arn": topic_arn,
+                "subject": "iasWorld data tests failed",
```
dfsnow, line 935: **suggestion (non-blocking):** Let's maybe add a count to the subject line i.e. `4 iasWorld data tests failed`.

.github/workflows/test_dbt_models.yaml
dfsnow: **suggestion (non-blocking):** Why not just make a jq call or similar that maps the dbt vars to their expected env var equivalent, such that if we add a new dbt var it will automatically search for the corresponding env var? E.g.

- Add `data_test_iasworld_commercial_sns_topic` searches for `$COMMERCIAL_SNS_TOPIC`
- Add `data_test_iasworld_asmt_sns_topic` searches for `$ASMT_SNS_TOPIC`

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -869,6 +909,44 @@ def main() -> None:
...
+        for test_result in test_category.test_results:
+            if test_result.status == Status.FAIL and test_result.notify:
+                if test_result.notify not in failures_by_topic:
+                    failures_by_topic[test_result.notify] = []
+                failures_by_topic[test_result.notify].append(test_result)
+
+    # Generate the notification body and send it to each SNS topic. Start
+    # by parsing out a message body and subject for each group of failures
+    # by topic into a list of objects with the keys `topic_arn`, `subject`,
+    # and `body`
```
dfsnow, line 926: **suggestion (non-blocking):** Let's add a call to action to the body as well. Something like "These were autogenerated by the Data Team. You will stop receiving these emails once the failures below are fixed" and then also our contact info at the bottom.

dfsnow (commented):
This is a great start! The variable passing stuff seems like a necessary evil, but the seed + anti-join stuff strikes me as a bit clunky. Let's brainstorm to see if we can't simplify things and then meet later this week. In the meantime, see my other comments for more minor actions.

dbt/scripts/run_iasworld_data_tests.py
jeancochrane: Yup, that's right! To make extra sure we're on the same page, "topic" here specifically means SNS topic, i.e. tests that are tagged with the same SNS topic representing a group of recipients will be merged into one email that goes out to those recipients, regardless of the category of the tests.

.github/workflows/test_dbt_models.yaml
jeancochrane: Hmm, this is an interesting idea, but I'm not totally convinced it's worth the complexity just yet. Let's plan to discuss in-person tomorrow.

.github/workflows/test_dbt_models.yaml
jeancochrane: Great ideas on both fronts! I gated this step behind a conditional in e5a615e24a91e436fb06014e4a822031353c703a and factored the esoteric jq call into a Python script in 8fbac8997901e26967054274df16adc9103704e0. Note that I decided to implement the gate via an input variable, so we'll have to update the Spark job in order to enable it. I think this is a feature rather than a bug, however, since it means we can merge this change without enabling it immediately.

dbt/scripts/run_iasworld_data_tests.py
jeancochrane: I like that, done in c5096a8aeefab0bb059b8275aaf625a302fb69d0. I added our names instead of contact info, to avoid committing our email addresses to the source code; let me know if you don't mind the risk of spam and I can add email addresses in.

dbt/scripts/run_iasworld_data_tests.py
jeancochrane: Good idea, done in c5096a8aeefab0bb059b8275aaf625a302fb69d0.

.github/workflows/test_dbt_models.yaml
jeancochrane: Based on our in-person discussion, I'm going to table this until we need to add more topics.

Issue #581, opened 2024-09-05
Ruff isn't configured to sort Python imports


jeancochrane:
I noticed while reviewing https://github.com/ccao-data/data-architecture/pull/578 that ruff doesn't seem to be sorting Python imports. It turns out that this feature is disabled by default, and [per maintainer instructions](https://github.com/astral-sh/ruff/issues/8926#issuecomment-1834048218), we need to enable it.

PR #582, opened 2024-09-05
Configure ruff to sort Python imports

jeancochrane:
This PR updates our ruff config to enable sorting Python imports, which is disabled by default. We also update the linter pre-commit step to use the `--fix` flag, which enables autofixing for linter errors ([docs](https://docs.astral.sh/ruff/linter/#fixes)).

Note that I cancelled the `build-and-test-dbt` workflow, since this PR changes the `ratio_stats` script and I didn't want to bother rerunning it. Merging this PR should also rerun `ratio_stats` in prod; since we have high-availability enabled, this shouldn't cause a problem, but I'll plan to merge at a slow time and babysit the deployment just in case.

Closes https://github.com/ccao-data/data-architecture/issues/581.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,5 +22,10 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.5.5
     hooks:
+      # Python linter. Ruff recommends running this before the formatter to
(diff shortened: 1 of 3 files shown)


dfsnow (approved):
Nice QoL fix!

PR #588, opened 2024-09-09
Integrate ias and mydec sales into vw_pin_sale_combined

wagnerlmichael:
## Motivation of this PR

During 2024, we have seen a delay in iasworld sales upload, leaving us with little to no sales to work with when we do EDA, modeling, sales val, etc for current year sales. We started mock up a way to bring in mydec sales that aren't in iasworld, which led to this PR. This should allow us to use this sale data has a very close proxy for iasworld sales data during periods in which we are missing iasworld sales.

## Details of change

This PR brings in mydec sales into `default.vw_pin_sale` only if the `doc_no` does not already exist in iasworld. We coalesce columns (the main sale data columns: doc no, date, etc.) with priority given to iasworld, and add a new column called `source` which lets us know which table the sale was pulled from; iasworld or mydec. 

If a mydec `doc_no` does not exist in iasworld, but does exist in mydec, we will bring that sale into this view. If the `doc_no` exists in both tables we will preference the iasworld sale.

We (discussed with @jeancochrane ) also [changed the error threshold](https://github.com/ccao-data/data-architecture/pull/588/commits/fe8443163b3a96f3c134b42e6ac2557ff4c6f0dc) for the `default_vw_pin_sale_reasonable_number_of_sales_per_year` test since we are letting in a bunch of sales that were previously no longer in iasworld. The two reasons for this increase in sales are that there is a small but significant amount of mydec sales that get brought into this view historically, and we let in multi-sales from mydec which we were not doing before. It pushes the test failures from 3997 to over 7000. 

We will need to make a large amount of PRs in other repos to accommodate this. Most likely a bunch of `where source ='iasworld'` filters.

View at `FROM z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default.vw_pin_sale`

### Total counts

```sql
SELECT source, count(*)
FROM "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
GROUP BY source;

#	source	_col1
1	mydec	318236
2	iasworld	2496144

. . . . . . . . . . . . 

select count(*) from default.vw_pin_sale

#	_col0
1	2496144


```

### Filter deed type
```sql
-- deed_type_filter
SELECT count(*)
FROM "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
where
source = 'iasworld'
and sale_filter_deed_type


#	_col0
1	116925

. . . . . . . . . . . . 

select count(*) from default.vw_pin_sale
where sale_filter_deed_type

#	_col0
1	116925
```

### Filter same sale within 365 days

There is an important point to highlight here.

`sale_filter_same_sale_within_365` in the new view differs from `sale_filter_same_sale_within_365` in the old view. This is because we allow mydec sales that aren't in the ias sales data to become a part of the computation of `sale_filter_same_sale_within_365`.  

What we have done is added a `sale_filter_same_iasworld_sale_within_365` column to the new view which is an exact replica of how it looks in the current `default.vw_pin_sale`. This was done in order to maintain parity with the old view when we invoke `source = 'iasworld`. This way if we decide to keep using exclusively iasworld sales in modeling, we can invoke the filter free of mydec sale influence.

```sql
select count(*), sale_filter_same_sale_within_365
from "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
where source = 'iasworld'
group by sale_filter_same_sale_within_365

#	_col0	sale_filter_same_sale_within_365
1	2464608	false
2	31536	true

. . . . . . . . . . . . 


select count(*), sale_filter_same_iasworld_sale_within_365
from "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
where source = 'iasworld'
group by sale_filter_same_iasworld_sale_within_365

#	_col0	sale_filter_same_iasworld_sale_within_365
1	2483476	false
2	12668	true

. . . . . . . . . . . . 


select count(*), sale_filter_same_sale_within_365
from "default"."vw_pin_sale"
group by sale_filter_same_sale_within_365


#	_col0	sale_filter_same_sale_within_365
1	2483476	false
2	12668	true

```

### `is_mydec_date` check

```sql

select count(*), is_mydec_date
from "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
where source = 'iasworld'
group by is_mydec_date

#	_col0	is_mydec_date
1	913241	true
2	1582903	false


. . . . . . . . . . . .

select count(*), is_mydec_date
from "default"."vw_pin_sale"
group by is_mydec_date

#	_col0	is_mydec_date
1	913241	true
2	1582903	false

```

## Comparisons of old/new iasworld sales

This compares the old and new views when the new view has `source = 'iasworld'` invoked. The new column explains if the data is from the old or the new view. Everything seems to match except `sale_filter_same_sale_within_365`, which is explained above.

|skim_type|skim_variable                                                     |new  |numeric.mean|n_missing|logical.count            |
|---------|------------------------------------------------------------------|-----|------------|---------|-------------------------|
|POSIXct  |sale_date                                                         |FALSE|          NA|      0  |NA                       |
|POSIXct  |sale_date                                                         | TRUE|          NA|      0  |NA                       |
|character|pin                                                               |FALSE|          NA|      0  |NA                       |
|character|pin                                                               | TRUE|          NA|      0  |NA                       |
|character|year                                                              |FALSE|          NA|      0  |NA                       |
|character|year                                                              | TRUE|          NA|      0  |NA                       |
|character|township_code                                                     |FALSE|          NA|      0  |NA                       |
|character|township_code                                                     | TRUE|          NA|      0  |NA                       |
|character|nbhd                                                              |FALSE|          NA|      0  |NA                       |
|character|nbhd                                                              | TRUE|          NA|      0  |NA                       |
|character|class                                                             |FALSE|          NA|      0  |NA                       |
|character|class                                                             | TRUE|          NA|      0  |NA                       |
|character|sale_price                                                        |FALSE|          NA|      0  |NA                       |
|character|sale_price                                                        | TRUE|          NA|      0  |NA                       |
|character|doc_no                                                            |FALSE|          NA|      0  |NA                       |
|character|doc_no                                                            | TRUE|          NA|      0  |NA                       |
|character|deed_type                                                         |FALSE|          NA|     56  |NA                       |
|character|deed_type                                                         | TRUE|          NA|     56  |NA                       |
|character|seller_name                                                       |FALSE|          NA| 156454  |NA                       |
|character|seller_name                                                       | TRUE|          NA| 156454  |NA                       |
|character|buyer_name                                                        |FALSE|          NA| 145404  |NA                       |
|character|buyer_name                                                        | TRUE|          NA| 145404  |NA                       |
|character|sale_type                                                         |FALSE|          NA| 321992  |NA                       |
|character|sale_type                                                         | TRUE|          NA| 321992  |NA                       |
|character|mydec_deed_type                                                   |FALSE|          NA|1773326  |NA                       |
|character|mydec_deed_type                                                   | TRUE|          NA|1773326  |NA                       |
|character|sv_outlier_reason1                                                |FALSE|          NA|2450223  |NA                       |
|character|sv_outlier_reason1                                                | TRUE|          NA|2450223  |NA                       |
|character|sv_outlier_reason2                                                |FALSE|          NA|2456873  |NA                       |
|character|sv_outlier_reason2                                                | TRUE|          NA|2456873  |NA                       |
|character|sv_outlier_reason3                                                |FALSE|          NA|2496144  |NA                       |
|character|sv_outlier_reason3                                                | TRUE|          NA|2496144  |NA                       |
|character|sv_run_id                                                         |FALSE|          NA|1830272  |NA                       |
|character|sv_run_id                                                         | TRUE|          NA|1830272  |NA                       |
|character|sv_version                                                        |FALSE|          NA|1830272  |NA                       |
|character|sv_version                                                        | TRUE|          NA|1830272  |NA                       |
|character|source                                                            |FALSE|          NA|2496144  |NA                       |
|character|source                                                            | TRUE|          NA|      0  |NA                       |
|logical  |is_mydec_date                                                     |FALSE|          NA|      0  |FAL: 1582903, TRU: 913241|
|logical  |is_mydec_date                                                     | TRUE|          NA|      0  |FAL: 1582903, TRU: 913241|
|logical  |is_multisale                                                      |FALSE|          NA|      0  |FAL: 1976266, TRU: 519878|
|logical  |is_multisale                                                      | TRUE|          NA|      0  |FAL: 1976266, TRU: 519878|
|logical  |sale_filter_same_sale_within_365                                  |FALSE|          NA|      0  |FAL: 2483476, TRU: 12668 |
|logical  |sale_filter_same_sale_within_365                                  | TRUE|          NA|      0  |FAL: 2464488, TRU: 31656 |
|logical  |sale_filter_less_than_10k                                         |FALSE|          NA|      0  |FAL: 2282851, TRU: 213293|
|logical  |sale_filter_less_than_10k                                         | TRUE|          NA|      0  |FAL: 2282851, TRU: 213293|
|logical  |sale_filter_deed_type                                             |FALSE|          NA|      0  |FAL: 2379219, TRU: 116925|
|logical  |sale_filter_deed_type                                             | TRUE|          NA|      0  |FAL: 2379219, TRU: 116925|
|logical  |sale_filter_is_outlier                                            |FALSE|          NA|      0  |FAL: 2450223, TRU: 45921 |
|logical  |sale_filter_is_outlier                                            | TRUE|          NA|      0  |FAL: 2450223, TRU: 45921 |
|logical  |sale_filter_ptax_flag                                             |FALSE|          NA|1650150  |FAL: 776894, TRU: 69100  |
|logical  |sale_filter_ptax_flag                                             | TRUE|          NA|1650150  |FAL: 776894, TRU: 69100  |
|logical  |mydec_property_advertised                                         |FALSE|          NA|1650150  |TRU: 632934, FAL: 213060 |
|logical  |mydec_property_advertised                                         | TRUE|          NA|1650150  |TRU: 632934, FAL: 213060 |
|logical  |mydec_is_installment_contract_fulfilled                           |FALSE|          NA|1650150  |FAL: 843944, TRU: 2050   |
|logical  |mydec_is_installment_contract_fulfilled                           | TRUE|          NA|1650150  |FAL: 843944, TRU: 2050   |
|logical  |mydec_is_sale_between_related_individuals_or_corporate_affiliates |FALSE|          NA|1650150  |FAL: 841748, TRU: 4246   |
|logical  |mydec_is_sale_between_related_individuals_or_corporate_affiliates | TRUE|          NA|1650150  |FAL: 841748, TRU: 4246   |
|logical  |mydec_is_transfer_of_less_than_100_percent_interest               |FALSE|          NA|1650150  |FAL: 845432, TRU: 562    |
|logical  |mydec_is_transfer_of_less_than_100_percent_interest               | TRUE|          NA|1650150  |FAL: 845432, TRU: 562    |
|logical  |mydec_is_court_ordered_sale                                       |FALSE|          NA|1650150  |FAL: 842113, TRU: 3881   |
|logical  |mydec_is_court_ordered_sale                                       | TRUE|          NA|1650150  |FAL: 842113, TRU: 3881   |
|logical  |mydec_is_sale_in_lieu_of_foreclosure                              |FALSE|          NA|1650150  |FAL: 845475, TRU: 519    |
|logical  |mydec_is_sale_in_lieu_of_foreclosure                              | TRUE|          NA|1650150  |FAL: 845475, TRU: 519    |
|logical  |mydec_is_condemnation                                             |FALSE|          NA|1650150  |FAL: 845939, TRU: 55     |
|logical  |mydec_is_condemnation                                             | TRUE|          NA|1650150  |FAL: 845939, TRU: 55     |
|logical  |mydec_is_short_sale                                               |FALSE|          NA|1650150  |FAL: 838921, TRU: 7073   |
|logical  |mydec_is_short_sale                                               | TRUE|          NA|1650150  |FAL: 838921, TRU: 7073   |
|logical  |mydec_is_bank_reo_real_estate_owned                               |FALSE|          NA|1650150  |FAL: 801823, TRU: 44171  |
|logical  |mydec_is_bank_reo_real_estate_owned                               | TRUE|          NA|1650150  |FAL: 801823, TRU: 44171  |
|logical  |mydec_is_auction_sale                                             |FALSE|          NA|1650150  |FAL: 841391, TRU: 4603   |
|logical  |mydec_is_auction_sale                                             | TRUE|          NA|1650150  |FAL: 841391, TRU: 4603   |
|logical  |mydec_is_seller_buyer_a_relocation_company                        |FALSE|          NA|1650150  |FAL: 842855, TRU: 3139   |
|logical  |mydec_is_seller_buyer_a_relocation_company                        | TRUE|          NA|1650150  |FAL: 842855, TRU: 3139   |
|logical  |mydec_is_seller_buyer_a_financial_institution_or_government_agency|FALSE|          NA|1650150  |FAL: 829159, TRU: 16835  |
|logical  |mydec_is_seller_buyer_a_financial_institution_or_government_agency| TRUE|          NA|1650150  |FAL: 829159, TRU: 16835  |
|logical  |mydec_is_buyer_a_real_estate_investment_trust                     |FALSE|          NA|1650150  |FAL: 845798, TRU: 196    |
|logical  |mydec_is_buyer_a_real_estate_investment_trust                     | TRUE|          NA|1650150  |FAL: 845798, TRU: 196    |
|logical  |mydec_is_buyer_a_pension_fund                                     |FALSE|          NA|1650150  |FAL: 845981, TRU: 13     |
|logical  |mydec_is_buyer_a_pension_fund                                     | TRUE|          NA|1650150  |FAL: 845981, TRU: 13     |
|logical  |mydec_is_buyer_an_adjacent_property_owner                         |FALSE|          NA|1650150  |FAL: 845302, TRU: 692    |
|logical  |mydec_is_buyer_an_adjacent_property_owner                         | TRUE|          NA|1650150  |FAL: 845302, TRU: 692    |
|logical  |mydec_is_buyer_exercising_an_option_to_purchase                   |FALSE|          NA|1650150  |FAL: 841495, TRU: 4499   |
|logical  |mydec_is_buyer_exercising_an_option_to_purchase                   | TRUE|          NA|1650150  |FAL: 841495, TRU: 4499   |
|logical  |mydec_is_simultaneous_trade_of_property                           |FALSE|          NA|1650150  |FAL: 845973, TRU: 21     |
|logical  |mydec_is_simultaneous_trade_of_property                           | TRUE|          NA|1650150  |FAL: 845973, TRU: 21     |
|logical  |mydec_is_sale_leaseback                                           |FALSE|          NA|1650150  |FAL: 844329, TRU: 1665   |
|logical  |mydec_is_sale_leaseback                                           | TRUE|          NA|1650150  |FAL: 844329, TRU: 1665   |
|logical  |mydec_is_homestead_exemption                                      |FALSE|          NA|1650150  |FAL: 762824, TRU: 83170  |
|logical  |mydec_is_homestead_exemption                                      | TRUE|          NA|1650150  |FAL: 762824, TRU: 83170  |
|logical  |sv_is_outlier                                                     |FALSE|          NA|1830272  |FAL: 619951, TRU: 45921  |
|logical  |sv_is_outlier                                                     | TRUE|          NA|1830272  |FAL: 619951, TRU: 45921  |
|logical  |sv_is_ptax_outlier                                                |FALSE|          NA|1830272  |FAL: 648200, TRU: 17672  |
|logical  |sv_is_ptax_outlier                                                | TRUE|          NA|1830272  |FAL: 648200, TRU: 17672  |
|logical  |sv_is_heuristic_outlier                                           |FALSE|          NA|1830272  |FAL: 619951, TRU: 45921  |
|logical  |sv_is_heuristic_outlier                                           | TRUE|          NA|1830272  |FAL: 619951, TRU: 45921  |
|logical  |sale_filter_same_iasworld_sale_within_365                         |FALSE|          NA|2496144  |:                        |
|logical  |sale_filter_same_iasworld_sale_within_365                         | TRUE|          NA|      0  |FAL: 2483476, TRU: 12668 |
|numeric  |sale_key                                                          |FALSE|8.569003e+07|      0  |NA                       |
|numeric  |sale_key                                                          | TRUE|8.569003e+07|      0  |NA                       |
|numeric  |num_parcels_sale                                                  |FALSE|2.986650e+00|      0  |NA                       |
|numeric  |num_parcels_sale                                                  | TRUE|2.986650e+00|      0  |NA                       |
|numeric  |mydec_homestead_exemption_general_alternative                     |FALSE|2.017056e+02|1650150  |NA                       |
|numeric  |mydec_homestead_exemption_general_alternative                     | TRUE|2.017056e+02|1650150  |NA                       |
|numeric  |mydec_homestead_exemption_senior_citizens                         |FALSE|3.525776e+01|1650150  |NA                       |
|numeric  |mydec_homestead_exemption_senior_citizens                         | TRUE|3.525776e+01|1650150  |NA                       |
|numeric  |mydec_homestead_exemption_senior_citizens_assessment_freeze       |FALSE|3.535516e+01|1650150  |NA                       |
|numeric  |mydec_homestead_exemption_senior_citizens_assessment_freeze       | TRUE|3.535516e+01|1650150  |NA                       |

## Glimpse at additional mydec sales entering view

This is the profile of the new mydec sales entering the view.

|skim_type|skim_variable                                                     |numeric.mean|n_missing|logical.count           |
|---------|------------------------------------------------------------------|------------|---------|------------------------|
|POSIXct  |sale_date                                                         |       NA   |     0   |NA                      |
|character|pin                                                               |       NA   |     0   |NA                      |
|character|year                                                              |       NA   |     0   |NA                      |
|character|township_code                                                     |       NA   |     1   |NA                      |
|character|nbhd                                                              |       NA   |     1   |NA                      |
|character|class                                                             |       NA   |     0   |NA                      |
|character|sale_price                                                        |       NA   |   833   |NA                      |
|character|doc_no                                                            |       NA   |     0   |NA                      |
|character|deed_type                                                         |       NA   |179151   |NA                      |
|character|seller_name                                                       |       NA   |   702   |NA                      |
|character|buyer_name                                                        |       NA   |   673   |NA                      |
|character|sale_type                                                         |       NA   |318236   |NA                      |
|character|mydec_deed_type                                                   |       NA   |179151   |NA                      |
|character|sv_outlier_reason1                                                |       NA   |318182   |NA                      |
|character|sv_outlier_reason2                                                |       NA   |318197   |NA                      |
|character|sv_outlier_reason3                                                |       NA   |318236   |NA                      |
|character|sv_run_id                                                         |       NA   |317545   |NA                      |
|character|sv_version                                                        |       NA   |317545   |NA                      |
|character|source                                                            |       NA   |     0   |NA                      |
|logical  |is_mydec_date                                                     |       NA   |     0   |TRU: 318236             |
|logical  |is_multisale                                                      |       NA   |     0   |FAL: 197097, TRU: 121139|
|logical  |sale_filter_same_sale_within_365                                  |       NA   |     0   |FAL: 238565, TRU: 79671 |
|logical  |sale_filter_same_iasworld_sale_within_365                         |       NA   |     0   |FAL: 318236             |
|logical  |sale_filter_less_than_10k                                         |       NA   |   833   |TRU: 177302, FAL: 140101|
|logical  |sale_filter_deed_type                                             |       NA   |     0   |TRU: 179151, FAL: 139085|
|logical  |sale_filter_is_outlier                                            |       NA   |     0   |FAL: 318182, TRU: 54    |
|logical  |sale_filter_ptax_flag                                             |       NA   |     0   |FAL: 308441, TRU: 9795  |
|logical  |mydec_property_advertised                                         |       NA   |     0   |FAL: 200407, TRU: 117829|
|logical  |mydec_is_installment_contract_fulfilled                           |       NA   |     0   |FAL: 317829, TRU: 407   |
|logical  |mydec_is_sale_between_related_individuals_or_corporate_affiliates |       NA   |     0   |FAL: 317333, TRU: 903   |
|logical  |mydec_is_transfer_of_less_than_100_percent_interest               |       NA   |     0   |FAL: 317883, TRU: 353   |
|logical  |mydec_is_court_ordered_sale                                       |       NA   |     0   |FAL: 317645, TRU: 591   |
|logical  |mydec_is_sale_in_lieu_of_foreclosure                              |       NA   |     0   |FAL: 318147, TRU: 89    |
|logical  |mydec_is_condemnation                                             |       NA   |     0   |FAL: 318212, TRU: 24    |
|logical  |mydec_is_short_sale                                               |       NA   |     0   |FAL: 317509, TRU: 727   |
|logical  |mydec_is_bank_reo_real_estate_owned                               |       NA   |     0   |FAL: 313124, TRU: 5112  |
|logical  |mydec_is_auction_sale                                             |       NA   |     0   |FAL: 317640, TRU: 596   |
|logical  |mydec_is_seller_buyer_a_relocation_company                        |       NA   |     0   |FAL: 317388, TRU: 848   |
|logical  |mydec_is_seller_buyer_a_financial_institution_or_government_agency|       NA   |     0   |FAL: 315309, TRU: 2927  |
|logical  |mydec_is_buyer_a_real_estate_investment_trust                     |       NA   |     0   |FAL: 318153, TRU: 83    |
|logical  |mydec_is_buyer_a_pension_fund                                     |       NA   |     0   |FAL: 318233, TRU: 3     |
|logical  |mydec_is_buyer_an_adjacent_property_owner                         |       NA   |     0   |FAL: 317979, TRU: 257   |
|logical  |mydec_is_buyer_exercising_an_option_to_purchase                   |       NA   |     0   |FAL: 317503, TRU: 733   |
|logical  |mydec_is_simultaneous_trade_of_property                           |       NA   |     0   |FAL: 318223, TRU: 13    |
|logical  |mydec_is_sale_leaseback                                           |       NA   |     0   |FAL: 317914, TRU: 322   |
|logical  |mydec_is_homestead_exemption                                      |       NA   |     0   |FAL: 305101, TRU: 13135 |
|logical  |sv_is_outlier                                                     |       NA   |317545   |FAL: 637, TRU: 54       |
|logical  |sv_is_ptax_outlier                                                |       NA   |317545   |FAL: 691                |
|logical  |sv_is_heuristic_outlier                                           |       NA   |317545   |FAL: 637, TRU: 54       |
|logical  |new                                                               |       NA   |     0   |TRU: 318236             |
|numeric  |sale_key                                                          |      NaN   |318236   |NA                      |
|numeric  |num_parcels_sale                                                  | 1.615188   |     0   |NA                      |
|numeric  |mydec_homestead_exemption_general_alternative                     |95.173511   |     0   |NA                      |
|numeric  |mydec_homestead_exemption_senior_citizens                         |19.101832   |     0   |NA                      |
|numeric  |mydec_homestead_exemption_senior_citizens_assessment_freeze       |19.151515   |     0   |NA                      |


dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -156,16 +154,20 @@ mydec_sales AS (
...
+            CAST(line_11_full_consideration AS BIGINT) AS sale_price,
+            line_2_total_parcels AS num_parcels_sale,
+            FALSE AS is_multisale,
             COALESCE(line_7_property_advertised = 1, FALSE)
                 AS mydec_property_advertised,
             COALESCE(line_10a = 1, FALSE)
                 AS mydec_is_installment_contract_fulfilled,
             COALESCE(line_10b = 1, FALSE)
-                AS mydec_is_sale_between_related_individuals_or_corporate_affiliates, -- noqa
+                AS mydec_is_sale_between_related_individuals_or_corporate_affiliates, --noqa
```
wagnerlmichael, line 170: Adding mydec sources of columns like `year`, `sale_price`, etc

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +244,166 @@ sales_val AS (
...
+        COALESCE(uq_sales.sale_date, md_sales.sale_date) AS sale_date,
+        COALESCE(uq_sales.sale_price, md_sales.sale_price) AS sale_price,
+        uq_sales.sale_key,
+        COALESCE(uq_sales.doc_no, md_sales.doc_no) AS doc_no,
+        COALESCE(uq_sales.deed_type, md_sales.mydec_deed_type) AS deed_type,
+        COALESCE(uq_sales.seller_name, md_sales.seller_name) AS seller_name,
+        COALESCE(uq_sales.is_multisale, md_sales.is_multisale) AS is_multisale,
+        COALESCE(uq_sales.num_parcels_sale, md_sales.num_parcels_sale) --noqa
+            AS num_parcels_sale,
+        COALESCE(uq_sales.buyer_name, md_sales.buyer_name) AS buyer_name,
```
wagnerlmichael, line 269: Coalescing ias/mydec data

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +244,166 @@ sales_val AS (
...
+        COALESCE(uq_sales.sale_type, NULL) AS sale_type,
```
wagnerlmichael, line 270: Any idea if there is a mydec equivalent for this? I looked but it wasn't obvious to me

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +244,166 @@ sales_val AS (
...
+),
+
+combined_sales AS (
+    SELECT
+        COALESCE(uq_sales.pin, md_sales.pin) AS pin,
+        COALESCE(uq_sales.year, md_sales.year) AS year,
+        COALESCE(uq_sales.township_code, tc.township_code) AS township_code,
+        COALESCE(uq_sales.nbhd, tc.nbhd) AS nbhd,
+        COALESCE(uq_sales.class, tc.class) AS class,
+        COALESCE(uq_sales.sale_date, md_sales.sale_date) AS sale_date,
```
wrridgeway, line 260: I think here we're going to want to favor `md_sales.date` before 2021, and `uq_sales.sale_date` during and after 2021.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +244,166 @@ sales_val AS (
...
-        FALSE
+    combined_sales.pin,
+    combined_sales.year,
+    combined_sales.township_code,
+    combined_sales.nbhd,
+    combined_sales.class,
+    combined_sales.sale_date,
+    (
+        combined_sales.source = 'mydec'
+        OR YEAR(combined_sales.sale_date) >= 2021
```
wrridgeway, line 359: This logic might need to be a bit more complicated. We're favoring mydec for dates a bit differently than for other columns. Before 2021 we always favor mydec when it's available, and for 2021 and on we favor iasworld if it's available.

wrridgeway (changes requested):
Looking great so far.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +244,176 @@ sales_val AS (
...
+            END,
+            md_sales.sale_date,
+            uq_sales.sale_date
+        ) AS sale_date,
+        CASE
+            WHEN md_sales.sale_date < '2021-01-01' THEN TRUE
+            WHEN uq_sales.sale_date >= '2021-01-01' THEN FALSE
+            WHEN md_sales.sale_date IS NOT NULL THEN TRUE
+            WHEN uq_sales.sale_date IS NOT NULL THEN FALSE
+        END AS is_mydec_sale,
```
wagnerlmichael, line 273: Is the `is_mydec_date` logic sufficient here? Or should we be doing another branch of conditional logic for both
- no ias sale date after 2021
- no mydec sale date before 2020

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +244,176 @@ sales_val AS (
...
+        COALESCE(uq_sales.nbhd, tc.nbhd) AS nbhd,
+        COALESCE(uq_sales.class, tc.class) AS class,
+        COALESCE(
+            CASE
+                WHEN md_sales.sale_date < '2021-01-01' THEN md_sales.sale_date
+                WHEN uq_sales.sale_date >= '2021-01-01' THEN uq_sales.sale_date
+            END,
+            md_sales.sale_date,
+            uq_sales.sale_date
+        ) AS sale_date,
```
wrridgeway, line 267: `CASE
WHEN uq_sales.year < '2021' THEN COALESCE(md_sales.sale_date, uq_sales.sale_date)
ELSE COALESCE(uq_sales.sale_date, md_sales.sale_date)
END AS sale_date`

This is how I would do sale date... does that make sense? I'd avoid comparing the sales dates themselves since they might be different and lead to weird behavior

dbt/models/default/default.vw_pin_sale.sql
wrridgeway: `CASE
    WHEN (uq_sales.year < '2021' OR uq_sales.sale_date IS NULL) AND md_sales.sale_date IS NOT NULL THEN TRUE
    WHEN (uq_sales.year >= '2021' OR md_sales.sale_date IS NULL) AND uq_sales.sale_date IS NOT NULL THEN FALSE
    END AS is_mydec_date`

If your logic aligns with this, and assuming I'm thinking correctly, it should be good.

dbt/models/default/default.vw_pin_sale.sql
wrridgeway: We're also gonna have to make sure we use the same logic for `sale_filter_same_sale_within_365`

dbt/models/default/default.vw_pin_sale.sql
wagnerlmichael: Yours is much more readable, I'll add it

dbt/models/default/default.vw_pin_sale.sql
wagnerlmichael: Can you think of a simpler solution than this?

```sql
CASE
    -- If there is a previous sale date within the same partition, we will perform a day difference calculation
    WHEN LAG(
        CASE
            WHEN uq_sales.year < '2021' THEN COALESCE(md_sales.sale_date, uq_sales.sale_date)
            ELSE COALESCE(uq_sales.sale_date, md_sales.sale_date)
        END
    ) OVER (
        -- Define the partition or grouping for the LAG function.
        -- Partition by pin and sale price
        PARTITION BY
            COALESCE(uq_sales.pin, md_sales.pin),
            COALESCE(uq_sales.sale_price, md_sales.sale_price)
        -- Order the sales by the sale date (using the same conditional logic as above) in ascending order
        ORDER BY
            CASE
                WHEN uq_sales.year < '2021' THEN COALESCE(md_sales.sale_date, uq_sales.sale_date)
                ELSE COALESCE(uq_sales.sale_date, md_sales.sale_date)
            END ASC
    ) IS NOT NULL
    -- If there is a previous sale, calculate difference in days between the current sale date and the previous sale date.
    THEN
        -- Use EXTRACT to compute the day difference between the current sale date and the previous sale date.
        EXTRACT(
            DAY FROM 
                -- The current sale date
                CASE
                    WHEN uq_sales.year < '2021' THEN COALESCE(md_sales.sale_date, uq_sales.sale_date)
                    ELSE COALESCE(uq_sales.sale_date, md_sales.sale_date)
                END
            -- Subtract the previous sale date, which is found using the LAG function.
            - LAG(
                CASE
                    WHEN uq_sales.year < '2021' THEN COALESCE(md_sales.sale_date, uq_sales.sale_date)
                    ELSE COALESCE(uq_sales.sale_date, md_sales.sale_date)
                END
            ) OVER (
                PARTITION BY
                    COALESCE(uq_sales.pin, md_sales.pin),
                    COALESCE(uq_sales.sale_price, md_sales.sale_price)
                ORDER BY
                    CASE
                        WHEN uq_sales.year < '2021' THEN COALESCE(md_sales.sale_date, uq_sales.sale_date)
                        ELSE COALESCE(uq_sales.sale_date, md_sales.sale_date)
                    END ASC
            )
        ) <= 365
    -- If there is no previous sale in the same partition (first sale or no prior sale at the same price for that pin), return FALSE.
    ELSE FALSE
END AS sale_filter_same_sale_within_365
```

wrridgeway: You could construct it outside of the `combined_sales` CTE, in the main `SELECT` statement, just using `combined_sales.sale_date` etc.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +239,179 @@ sales_val AS (
...
+                    'day',
+                    LAG(cte_s.sale_date_coalesced) OVER (
+                        PARTITION BY --noqa
+                            cte_s.pin_coalesced, cte_s.sale_price_coalesced
+                        ORDER BY cte_s.sale_date_coalesced ASC
+                    ),
+                    cte_s.sale_date_coalesced
+                ) <= 365
+            ELSE FALSE
+        END AS sale_filter_same_sale_within_365,
```
wagnerlmichael, line 344: I think this is a big improvement. Let me know what you think @wrridgeway

wrridgeway: Agreed. Let's do some testing - let's make sure that if we limit our source to iasWorld the new view looks like the old view, and that invoking filters still returns the same numbers of rows.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -217,16 +207,15 @@ mydec_sales AS (
             ) > 0 AS sale_filter_ptax_flag,
             COUNT() OVER (
                 PARTITION BY line_1_primary_pin, line_4_instrument_date
-            ) AS num_single_day_sales,
-            year_of_sale
```
wrridgeway, line 221: Rather than removing this, I'd use it as your source of mydec sale year. It's constructed using the instrument date during ingest. No reason to introduce additional calculations.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -217,16 +207,15 @@ mydec_sales AS (
...
         FROM {{ source('sale', 'mydec') }}
-        WHERE line_2_total_parcels = 1 -- Remove multisales
+        WHERE line_2_total_parcels = 1
     )
     /* Some sales in mydec have multiple rows for one pin on a given sale date.
     Sometimes they have different dates than iasworld prior to 2021 and when
     joined back onto unique_sales will create duplicates by pin/sale date. */
     WHERE num_single_day_sales = 1
-        OR (YEAR(mydec_date) > 2020)
+        OR (YEAR(sale_date) > 2020)
```
wrridgeway, line 218: We should probably just use the year column we create in the inner query, huh

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +238,192 @@ sales_val AS (
...
     FROM
-        {{ source('sale', 'flag') }}
-            AS sf
+        {{ source('sale', 'flag') }} AS sf
     INNER JOIN max_version_flag AS mv
         ON sf.meta_sale_document_num = mv.meta_sale_document_num
         AND sf.version = mv.max_version
+),
+
+-- Introducing cte_sales to precompute the coalesced values
```
wrridgeway, line 247:
```suggestion
-- CTE to coalesce iasworld and mydec values prior to constructing filters that depend on coalesced fields
```

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +238,192 @@ sales_val AS (
...
+cte_sales AS (
+    SELECT
+        COALESCE(uq_sales.pin, md_sales.pin) AS pin_coalesced,
+        CASE
+            WHEN md_sales.sale_date IS NOT NULL
```
wrridgeway, line 252: Let's add a comment as to why year (and date for that matter) is handled differently than the other coalesces.

wrridgeway (changes requested):

wrridgeway:
Just gonna nibble around the edges on this for the first bite. I'll look at the data and inspect the filters once this round of changes is made.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -19,8 +18,6 @@ WITH town_class AS (
         AND par.deactivat IS NULL
 ),
 
--- "nopar" isn't entirely accurate for sales associated with only one parcel,
--- so we create our own counter
```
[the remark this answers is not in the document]
wagnerlmichael: Ah thanks, this one slipped through

dbt/models/default/default.vw_pin_sale.sql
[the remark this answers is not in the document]
wagnerlmichael: Yep, my fault there. Let me know if I missed anything

wagnerlmichael:
> Just gonna nibble around the edges on this for the first bite. I'll look at the data and inspect the filters once this round of changes is made.

Impressively mouthy sentence

dbt/models/default/schema/default.vw_pin_sale.yml
wrridgeway, line 1: Could we add a note to the `Nuance` section that lays out how sales can come from iasWorld OR MyDec, but iasWorld is preferred?

dbt/models/default/default.vw_pin_sale.sql
wrridgeway: Let's take care of this before we merge.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -218,15 +219,15 @@ mydec_sales AS (
             COUNT() OVER (
                 PARTITION BY line_1_primary_pin, line_4_instrument_date
             ) AS num_single_day_sales,
-            year_of_sale
+            year_of_sale AS year
         FROM {{ source('sale', 'mydec') }}
-        WHERE line_2_total_parcels = 1 -- Remove multisales
```
wrridgeway, line 223: I think this line is going to prevent you from having any sales in your mydec sample that are multisales. I don't think we want to exclude multisales unless we invoke the `is_multisale` filter.

wrridgeway (changes requested):

dbt/models/default/schema/default.vw_pin_sale.yml
```diff
@@ -76,7 +78,7 @@ models:
             - year
           allowed_duplicates: 2
           config:
-            error_if: ">4032"
+            error_if: ">5032"
```
wrridgeway, line 81: @jeancochrane @wagnerlmichael This test compiles as:

```
select
    pin, year,
    count(*) as num_duplicates
from "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
group by pin, year
having count(*) > 2 + 1
```

Are we sure we only want to look for pin/year combos that show up _more than three times a year_? Wouldn't it make more sense to look for combos that show up more than twice?

jeancochrane: I think this is a question we should answer empirically: What's the maximum number of sales per PIN per year that appear in our data and are legitimate? And do we have enough confidence in our ability to determine legitimacy such that we can answer the question in the first place?

dbt/models/default/default.vw_pin_sale.sql
wagnerlmichael: Ah yep good catch

dbt/models/default/schema/default.vw_pin_sale.yml
wrridgeway: Yeah, I suppose if we're not testing a filter here then it's really a matter of expectations and experience rather than some logic that can be divined. I'm fine with it as, perhaps it's worth an issue of its own for investigation.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -218,15 +219,14 @@ mydec_sales AS (
...
     )
     /* Some sales in mydec have multiple rows for one pin on a given sale date.
     Sometimes they have different dates than iasworld prior to 2021 and when
     joined back onto unique_sales will create duplicates by pin/sale date. */
     WHERE num_single_day_sales = 1
-        OR (YEAR(mydec_date) > 2020)
+        OR YEAR(DATE_PARSE(year, '%Y')) > 2020
```
wrridgeway, line 229:
```suggestion
        OR year > 2020
```

Will this cause an issue? Might have to wrap 202o in single quotes.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +249,204 @@ sales_val AS (
...
+combined_sales AS (
+    SELECT
+        COALESCE(uq_sales.pin, md_sales.pin) AS pin_coalesced,
+        -- For many of the fields we used  simple coalesce statement,
+        -- but some data is a bit more complicated. Prior to 2021,
+        -- mydec sales and iasworld sales used different sale dates.
+        -- We preference the mydec sale as they are believed to be more
+        -- accurate. As of 2021, iasworld utilizes mydec sales, which means
+        -- we can prioritize iasworld data instead of mydec data.
+        CASE
```
wrridgeway, line 269: I think `year_coalesced` and `sale_date_coalesced` should follow the exact same logic - let's make sure `year_coalesced` respects the 2021 cut-off

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +249,204 @@ sales_val AS (
...
+            WHEN
+                uq_sales.year < '2021'
+                THEN COALESCE(md_sales.sale_date, uq_sales.sale_date)
+            ELSE COALESCE(uq_sales.sale_date, md_sales.sale_date)
+        END AS sale_date_coalesced,
+        CASE
+            -- If uq_sales.doc_no is not NULL, apply the COALESCE logic
+            WHEN uq_sales.doc_no IS NOT NULL
+                THEN
+                COALESCE(COALESCE(
```
wrridgeway, line 292: Do we need a double `COALESCE` here?

wrridgeway (changes requested):
Just a few tiny suggestions, we're really close.

dbt/models/default/default.vw_pin_sale.sql
wagnerlmichael: Nope

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +249,198 @@ sales_val AS (
...
+        -- If uq_sales.doc_no is not NULL, apply the COALESCE logic
+            WHEN uq_sales.doc_no IS NOT NULL
+                THEN COALESCE(
+                    md_sales.sale_date IS NOT NULL
+                    OR YEAR(uq_sales.sale_date) >= 2021,
+                    FALSE
+                )
+            ELSE
+                TRUE
+        END AS is_mydec_date,
```
wrridgeway, line 294: I'm not sure this logic is correct. In the past, `is_mydec_date` indicated at least 1 of 2 possible things:
- there was mydec data available for this sale, so we preferred the mydec date
- it was 2021 or later, when iasworld starts using mydec dates

if i grab the new view only for source = 'iasworld' and tally `is_mydec_date`, it currently doesn't match the old view.

I think you should only need the inner `coalesce` function and not the `case when`?

wrridgeway (changes requested):
This seems just about done.

wagnerlmichael:
I wanted to take one final look at this portion of the data checks in the body of the PR.

> Filter same sale within 365 days
This figure is different because after a talk with Billy it was decided that we should be
using doc_no's from mydec in the calculation of this filter even for ias doc_no observations.

Currently when invoking the `source = iasworld` filter when calling `vw_pin_sale` will return a replica of what the current production `vw_pin_sale` looks like, except for one column difference. The `sale_filter_same_sale_within_365` column will look different. This is because we are using mydec sales that aren't also in iasworld to calculate this metric. 

My concern is that (I think) we had decided at some point that we are using iasworld sales without any mydec sales to be consistent with a data source. Additionally, all invocations of `sale_filter_same_sale_within_365` will now return a slightly different number of observations. It will be slightly more since we have more sales to look back on when calculating the filter.

dbt/models/default/default.vw_pin_sale.sql
wagnerlmichael: Fixed this. I had accidentally removed a multi-parcel filter a few commits back

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -221,12 +221,13 @@ mydec_sales AS (
             ) AS num_single_day_sales,
             year_of_sale AS year
         FROM {{ source('sale', 'mydec') }}
+        WHERE line_2_total_parcels = 1
```
wrridgeway, line 224: Unfortunately, while this solves your filter issue, I believe it also drops all multisales from the mydec sample.

wrridgeway (changes requested):
@dfsnow  you're gonna want to weigh in on the `sale_filter_same_sale_within_365` question. My personal opinion is that it should include mydec sales and that we should have a new filter named `sale_filter_same_iasworld_sale_within_365` that ignores mydec sales and allows us to replicate the old view.

dbt/models/default/default.vw_pin_sale.sql
wrridgeway: In the previous view we weren't allow mydec sales into the view. We were only using them to improve dates for iasworld sales. Now that they're in the view, we need to treat them the same as iasworld sales.

wagnerlmichael: Right. It seems like in the `default.vw_pin_sale` we aren't using mydec sales to improve dates for iasworld if the iasworld sales are multi-sale - due to the `WHERE line_2_total_parcels = 1` in `default.vw_pin_sale`.

> Unfortunately, while this solves your filter issue, I believe it also drops all multisales from the mydec sample, which is something that should only happen when we set is_multisale = FALSE

If we remove the parcel filter and let in multi-sales from mydec, we will be allowing mydec date to override iasworld date for multi-sales, but this will make the following queries inconsistent, since we were not doing it for `default.vw_pin_sale`:

```sql
select count(*), is_mydec_date
from "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
where source = 'iasworld'
group by is_mydec_date

select count(*), is_mydec_date
from "default"."vw_pin_sale"
group by is_mydec_date
```

wrridgeway: So you're right about how the view has worked in the past, but this is ultimately still something we're going to have to address. We didn't let mydec multisales in because we were only joining mydec sales to iasworld sales and that would have been a one to many join on sale doc. But now what mydec sales are in the view in their own right, we'll need to figure out a way to allow in those multisales, without joining them to iasworld sales.

I think you can probably swing this with a couple of extra conditionals in `FULL OUTER JOIN mydec_sales AS md_sales ON uq_sales.doc_no = md_sales.doc_no` to make sure multisale mydec sales don't join onto iasworld sales.

wagnerlmichael: Awesome, I'll implement this and create a second 365 filter, hopefully that way when Dan looks a this we are good to go

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -43,9 +43,22 @@ unique_sales AS (
...
+        -- Compute sale_filter_same_iasworld_sale_within_365 using the same logic --noqa
+        COALESCE(
+            DATE_DIFF(
+                'day',
+                same_price_earlier_date,
+                sale_date
+            ) <= 365,
+            FALSE
+        ) AS sale_filter_same_iasworld_sale_within_365 --noqa
     FROM (
```
wagnerlmichael, line 62: Made an OG iasworld filter that matches the current view:

```sql 
SELECT 
    COUNT(*) AS count, 
    sale_filter_same_iasworld_sale_within_365
FROM 
    "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
WHERE 
    source = 'iasworld'
GROUP BY 
    sale_filter_same_iasworld_sale_within_365;

-  -  -
#	count	sale_filter_same_iasworld_sale_within_365
1	2483405	false
2	12668	true

```

```sql 
SELECT 
    COUNT(*) AS count, 
    sale_filter_same_sale_within_365
FROM 
    "default"."vw_pin_sale"
GROUP BY 
    sale_filter_same_sale_within_365;

-  -  -
#	count	sale_filter_same_sale_within_365
1	2483405	false
2	12668	true

```

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -249,93 +259,198 @@ sales_val AS (
...
+    FROM unique_sales AS uq_sales
+    -- This logic brings in mydec sales that aren't in iasworld.
+    -- If a doc_no exists in iasworld and mydec, we prioritize iasworld,
+    -- if it only exists in mydec, we will grab the doc_no from mydec. The
+    -- 'source' column lets us know which table the doc_no came from and allows
+    -- us to filter for only iasworld sales or for mydec sales that aren't in
+    -- iasworld already.
+    FULL OUTER JOIN mydec_sales AS md_sales
+        ON uq_sales.doc_no = md_sales.doc_no
+        AND md_sales.is_multisale = FALSE
```
wagnerlmichael, line 354: > So you're right about how the view has worked in the past, but this is ultimately still something we're going to have to address. We didn't let mydec multisales in because we were only joining mydec sales to iasworld sales and that would have potentially been a one to many join on sale doc in some cases. But now that mydec sales are in the view in their own right, we'll need to figure out a way to allow in those multisales, without joining them to iasworld sales. I think you can probably swing this with a couple of extra conditionals in FULL OUTER JOIN mydec_sales AS md_sales ON uq_sales.doc_no = md_sales.doc_no to make sure multisale mydec sales don't join onto iasworld sales but can exist if they don't have a matching iasworld sale.

Does this fix our problem of not letting multi-sale mydec date data join and override the iasworld multi-sale date? I think it at least mostly does:

```sql 
select count(*), is_mydec_date
from "z_ci_583_create_a_sales_view_that_combines_iasworld_mydec_and_ccrd_sales_default"."vw_pin_sale"
where source = 'iasworld'
group by is_mydec_date

- - -
#	_col0	is_mydec_date
1	1582983	false
2	913090	true
```
```sql 
select count(*), is_mydec_date
from "default"."vw_pin_sale"
group by is_mydec_date

- - -

#	_col0	is_mydec_date
1	1582903	false
2	913170	true
```

wrridgeway: i think you're gonna have to be a little bit more precise with the join. it'll have to be something like
```
FULL OUTER JOIN mydec_sales AS md_sales
        ON (uq_sales.doc_no = md_sales.doc_no AND md_sales.is_multisale = FALSE)
        OR uq_sales.doc_no != md_sales.doc_no
```

wrridgeway: I feel like this should do the trick, but i'm seeing a lot of small disagreements across columns in skim, now

wrridgeway: I think 
```
WHERE num_single_day_sales = 1
        OR year > '2020'
```
is causing the discrepancy. doc number 1607434035 seems to be an example of a mydec sale that is being excluded when it shouldn't be.

wagnerlmichael: Aside from these, which other columns were you looking at?

```R
skim <- bind_rows(old, new_filtered) %>%
  select(
    new,
    year,
    sale_filter_same_sale_within_365,
    sale_filter_less_than_10k,
    sale_filter_deed_type,
    sale_filter_is_outlier
  ) %>%
  group_by(new) %>%
  skim()
```

wrridgeway: Stuff like `seller_name`, `mydec_deed_type`. They're pretty small differences.

wrridgeway: 1607434035 presents a very awkward problem.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -216,17 +226,20 @@ mydec_sales AS (
...
-                PARTITION BY line_1_primary_pin, line_4_instrument_date
+                PARTITION BY
+                    line_1_primary_pin,
+                    line_4_instrument_date,
+                    line_2_total_parcels
             ) AS num_single_day_sales,
-            year_of_sale
+            year_of_sale AS year
         FROM {{ source('sale', 'mydec') }}
-        WHERE line_2_total_parcels = 1 -- Remove multisales
```
wrridgeway, line 223: Okay, I did a lot of digging today, and this is the source of the current discrepancies between new data where `source = 'iasworld'` and the old data (letting in mydec multisales). it affects how `num_single_day_sales` is built. I *believe* i've fixed this by adding an additional partition to the construction of `num_single_day_sales` and then only excluding pins with more than one sale per day if it isn't a multisale.

wrridgeway (approved):
This is great work @wagnerlmichael .  @jeancochrane or @dfsnow I worked with @wagnerlmichael on this and I think it's good to go, but I'm not sure what our plans are.  We could deploy it as is and we can still replicate the old table by using the new `source` column, but it will also start allowing mydec sales into anything that is downstream of `default.vw_pin_sale`. Might be worth sitting on it and hoping iasworld sales get updated in the meantime.

jeancochrane:
@wrridgeway @wagnerlmichael I like the idea of bringing this view into our lake so that we can decouple ourselves from the iasWorld sales ingest cycle going forward, but I also worry about making such a big change to an important view without testing out its downstream implications first. What if we move this logic into a new view named something like `default.vw_pin_sale_combined` and then rewrite `default.vw_pin_sale` so that it continues to select only the sales from `default.vw_pin_sale_combined` where `sale_type = 'iasworld'`? That would let us start opting into the combined set of sales for downstream consumers (e.g. modeling) without having to update _every_ consumer of the view to handle the new schema. Whatever we decide to do going forward, agreed that @dfsnow should be on board too.

wrridgeway:
I think renaming it for now is a good way to get it into the code base without making a commitment to it. Would make swapping it out for `vw_pin_sale` on the fly very easy.

wagnerlmichael:
I like Jean's idea as well.

ccao-jardine:
Sounds solid to me. Good thinking, @jeancochrane.

dbt/models/default/schema/default.vw_pin_sale_combined.yml
```diff
@@ -0,0 +1,102 @@
...
+      - name: sale_date
+        description: '{{ doc("shared_column_sale_date") }}'
+      - name: sale_filter_deed_type
+        description: Remove quit claim, executor, beneficiary and missing deed types
+      - name: sale_filter_is_outlier
+        description: Same as `sv_is_outlier`, but `NULL` values are replaced with `FALSE`
+      - name: sale_filter_less_than_10k
+        description: Indicator for whether sale is less than $10K FMW
+      - name: sale_filter_same_sale_within_365
+        description: Remove sale with the _same_ value (for the same PIN) within 365 days
```
wagnerlmichael, line 35: I think you left out `sale_filter_same_iasworld_sale_within_365`. I would also mention each column in the description of each as they kind of exist in relation to each other.

`sale_filter_same_sale_within_365` uses sales from both wheras the other is only ias

wagnerlmichael:
@jeancochrane if you could do a quick once over of this commit [dab3906](https://github.com/ccao-data/data-architecture/pull/588/commits/dab3906ad99f430ef63fd1244008067af26efee4) to ensure I properly implemented the movement of the test from pin_sale to qc. I saw you did this with `default.vw_pin_sale` [here](https://github.com/ccao-data/data-architecture/pull/638/files#r1833498531).

PR #589, opened 2024-09-09
Add scripts and schemas for iasWorld test and legacy files

dfsnow:
This PR adds:

- Scripts to parse and upload legacy mainframe fixed-width files to Athena
- dbt schema entries to expose the resulting legacy file tables in our dbt docs
- dbt schema entries for tables from the iasWorld test environment

### DIFF
--- a/dbt/models/iasworld_test/schema.yml
+++ b/dbt/models/iasworld_test/schema.yml
@@ -1,0 +1,122 @@
+sources:
+  - name: iasworld_test
+    loaded_at_field: date_parse(loaded_at, '%Y-%m-%d %H:%i:%S.%f')
+    tags:
+      - load_auto
+      - type_res
+
+    tables:
+      - name: addn
+        description: |
+          Test environment version of the production
+          [`addn` table](#!/source/source.ccao_data_athena.iasworld.addn)
+
+      - name: addrindx
+        description: |
+          Test environment version of the production
+          [`addrindx` table](#!/source/source.ccao_data_athena.iasworld.addrindx)
+
+      - name: aprval
+        description: |
+          Test environment version of the production
+          [`aprval` table](#!/source/source.ccao_data_athena.iasworld.aprval)
+
+      - name: asmt_hist
+        description: |
+          Test environment version of the production
+          [`asmt_hist` table](#!/source/source.ccao_data_athena.iasworld.asmt_hist)
+
+      - name: cname
+        description: |
+          Test environment version of the production
+          [`cname` table](#!/source/source.ccao_data_athena.iasworld.cname)
+
+      - name: comnt
+        description: |
+          Test environment version of the production
+          [`comnt` table](#!/source/source.ccao_data_athena.iasworld.comnt)
+
+      - name: cvleg
+        description: |
+          Test environment version of the production
+          [`cvleg` table](#!/source/source.ccao_data_athena.iasworld.cvleg)
+
+      - name: cvown
+        description: |
+          Test environment version of the production
+          [`cvown` table](#!/source/source.ccao_data_athena.iasworld.cvown)
+
+      - name: cvtran
+        description: |
+          Test environment version of the production
+          [`cvtran` table](#!/source/source.ccao_data_athena.iasworld.cvtran)
+
+      - name: dweldat
+        description: |
+          Test environment version of the production
+          [`dweldat` table](#!/source/source.ccao_data_athena.iasworld.dweldat)
+
+      - name: enter
+        description: |
+          Test environment version of the production
+          [`enter` table](#!/source/source.ccao_data_athena.iasworld.enter)
+
+      - name: exadmn
+        description: |
+          Test environment version of the production
+          [`exadmn` table](#!/source/source.ccao_data_athena.iasworld.exadmn)
+
+      - name: exapp
+        description: |
+          Test environment version of the production
+          [`exapp` table](#!/source/source.ccao_data_athena.iasworld.exapp)
+
+      - name: exdet
+        description: |
+          Test environment version of the production
+          [`exdet` table](#!/source/source.ccao_data_athena.iasworld.exdet)
+
+      - name: htdates
+        description: |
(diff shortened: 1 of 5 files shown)


dbt/models/iasworld_test/schema.yml
```diff
@@ -0,0 +1,122 @@
...
+    loaded_at_field: date_parse(loaded_at, '%Y-%m-%d %H:%i:%S.%f')
+    tags:
+      - load_auto
+      - type_res
+
+    tables:
+      - name: addn
+        description: |
+          Test environment version of the production
+          [`addn` table](#!/source/source.ccao_data_athena.iasworld.addn)
```
dfsnow, line 12: Hacky internal linking as suggested [by the docs](https://docs.getdbt.com/reference/resource-properties/description#link-to-another-model-in-a-description).

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
dfsnow, line 1: This whole file is just a big bunch of boilerplate to manually parse pipe-delimited (usually) fixed-width text files.

dbt/models/iasworld_test/schema.yml
```diff
@@ -0,0 +1,122 @@
+sources:
```
dfsnow, line 1: We don't have access to the full set of tables that we do from the prod environment. I also don't know to what extend the test table schemas differ from prod. A brief look shows that they mostly seem identical, but I know they sometimes add new columns in test first.

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
```diff
@@ -0,0 +1,264 @@
...
+        c(year, tax_year),
+        \(x) ifelse(substr(x, 1, 1) == "9", paste0("19", x), paste0("20", x))
+      ),
+      across(starts_with("applicant_"), \(x) str_trim(str_squish(x))),
+      applicant_zip = str_sub(applicant_zip, 1, 5),
+      birth_date = lubridate::mdy(birth_date),
+      first_app_received_date = lubridate::ymd(first_app_received_date),
+      base_value_year_class = str_sub(base_value_year_class, 4, 6),
+      qualified_date = lubridate::ymd(qualified_date),
+      source_file = {{ file }}
```
jeancochrane, line 95: **[Question, non-blocking]** R newbie question: Based on the [dplyr docs](https://dplyr.tidyverse.org/articles/programming.html#indirection), I think I understand that we're embracing the `file` symbol here because we want to use an env-variable value to reference a data-variable. Is that right? If so, where is `file` defined in the environment?

jeancochrane (approved):
Nice work here! A couple nitpicks and questions below, but nothing serious.

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
dfsnow: Whoops, I'm dumb and didn't change `file` to `f` after I changed the function arg name. Fixed in 9f6f41b.

The double curly braces actually aren't necessary in this case since we want the `f` arg to evaluate to a string (instead of a [quosure](https://rlang.r-lib.org/reference/topic-quosure.html#foreign-and-local-expressions)), but it also doesn't hurt. One benefit of the double-curly braces here is that we could have a function argument with the same name as an existing column but still use it in assignment. For example:

```r
df <- tibble(a = 1, b = 2)

test <- function(data, a) {
  data %>%
    mutate(a = a)
}
test(df, a = 3)
```

Would return a tibble where `a = 1`, while the below would return a tibble where `a = 3`:

```r
df <- tibble(a = 1, b = 2)

test <- function(data, a) {
  data %>%
    mutate(a = {{ a }})
}
test(df, a = 3)
```

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
```diff
@@ -0,0 +1,264 @@
...
+      pct_of_shares = col_integer(),
+      sf_percent = col_integer()
+    )
+  ) %>%
+    mutate(
+      across(
+        c(year, tax_year),
+        \(x) ifelse(substr(x, 1, 1) == "9", paste0("19", x), paste0("20", x))
+      ),
+      across(starts_with("applicant_"), \(x) str_trim(str_squish(x))),
```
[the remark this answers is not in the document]
dfsnow: We don't, just bein defensive.

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
```diff
@@ -0,0 +1,264 @@
...
+output_bucket <- file.path(AWS_S3_WAREHOUSE_BUCKET, "ccao", "legacy")
+
+
+##### CC_DLI_SENFRR #####
+files_cc_dli_senfrr <- aws.s3::get_bucket_df(
+  bucket = AWS_S3_RAW_BUCKET,
+  prefix = "ccao/legacy/CC_DLI_SENFRR",
+  max = Inf
+) %>%
+  filter(Size > 0)
```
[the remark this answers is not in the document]
dfsnow: The actual folders are returned as 0b objects e.g. `BUCKET/ccao/legacy/`. This is just a hacky way to drop them.

dbt/seeds/ccao/docs.md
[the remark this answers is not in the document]
dfsnow: Whoops! Good catch. Fixed in ed2fe98.

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
jeancochrane: Very clear explanation, thanks!

PR #586, opened 2024-09-06
Add script to export town close QC reports and print command for refreshing tables

jeancochrane:
This PR adds a new script `dbt/scripts/export_qc_town_close_reports.py` with two primary functions:

1. Exports the QC town close reports
2. When the `--print-table-refresh-command` argument is provided, prints a command that can be run using the [`service-spark-iasworld`](https://github.com/ccao-data/service-spark-iasworld/) project to refresh data for iasWorld tables that the report depends on, and quits instead of exporting reports

Closes https://github.com/ccao-data/data-architecture/issues/576.

PR #590, opened 2024-09-09
Factor shared dbt script logic out into `utils` module

jeancochrane:
This PR builds on https://github.com/ccao-data/data-architecture/pull/586, implementing [one specific piece of feedback](https://github.com/ccao-data/data-architecture/pull/586#discussion_r1750638576):

> Why not move some of the export functionality into a `utils/` dir/module. Then `dbt/scripts/export_models.py` can be a simple wrapper for the `utils/` code, and both `export_models.py` and `export_qc_town_close_reports.py` can share the utils code.

Accordingly, this PR creates a `dbt/scripts/utils/` module that stores some shared logic for our dbt scripts.


dbt/scripts/utils/export.py
```diff
@@ -0,0 +1,230 @@
...
+            "dev"
+        * select (list[str]): One or more dbt --select statements to
+            use for filtering models
+        * selector (str): A selector name to use for filtering
+            models, as defined in selectors.yml. One of `select` or `selector`
+            must be set, but they can't both be set
+        * rebuild (bool): Rebuild models before exporting, defaults to False
+        * where (str): Optional SQL expression representing a WHERE clause to
+            filter models
+    """
```
jeancochrane, line 43: This is the only part of the code that changed during the move -- I added a docstring since the code is no longer colocated with the argparse arguments that were previously providing documentation.

dbt/scripts/export_models.py
```diff
@@ -54,6 +42,12 @@ def parse_args():
         # get preserved
         formatter_class=argparse.RawTextHelpFormatter,
     )
+    parser.add_argument(
+        *constants.TARGET_ARGUMENT_ARGS, **constants.TARGET_ARGUMENT_KWARGS
+    )
+    parser.add_argument(
+        *constants.REBUILD_ARGUMENT_ARGS, **constants.REBUILD_ARGUMENT_KWARGS
+    )
```
jeancochrane, line 50: Although this wasn't in the scope of [the comment that inspired this PR](https://github.com/ccao-data/data-architecture/pull/586#discussion_r1750638576), I took the opportunity to factor out shared argparse argument definitions into the new `utils` module as well.

dfsnow (approved):
Nice refactor! This seems like a much cleaner way to share the export code.

PR #592, opened 2024-09-10
Disable natural language check in super-linter workflow

jeancochrane:
The `lint` workflow runs [super-linter](https://github.com/super-linter/super-linter), which enables the [`terminology` rule](https://github.com/sapegin/textlint-rule-terminology) for the [textlint linter](https://textlint.github.io/). Per [the super-linter 
 textlint template](https://github.com/super-linter/super-linter/blob/058b9239d47af5565e7c80f55ec097259ad00673/TEMPLATES/.textlintrc#L1-L8), this is actually the _only_ rule that super-linter enables for textlint. Super-linter calls this the "natural language" check. 

I find the `terminology` rule unhelpful, in that it mostly just fails CI for my PRs due to small nitpicks like the use of "website" instead of "site" or "repo" instead of "repository". See [this failed workflow run](https://github.com/ccao-data/data-architecture/actions/runs/10795697411/job/29942969002#step:4:745) for an example. If we were publishing docs that needed to adhere to a strict styleguide, I might find this helpful, but instead I think it just slows down our velocity without clarifying our writing very much. As such, this PR disables textlint, and therefore the `terminology` rule, in our super-linter config.

See [the `terminology` rule code](https://github.com/sapegin/textlint-rule-terminology/blob/master/terms.jsonc) for the list of replacements that this PR will disable. There doesn't seem to be a page for these replacements in [the docs](https://github.com/sapegin/textlint-rule-terminology), so reading the code is indeed the recommended way of browsing them.

See [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/10796716422/job/29946341776?pr=592#step:4:233) for evidence of the `Natural Language` check not running, and the super-linter workflow passing in spite of the presence of `website` and `repo` in [the README](https://github.com/ccao-data/data-architecture/pull/592/commits/ef28368d8a90b549b08d7da57670b1c1f4a32f24).

My opinion about the helpfulness of this change is weakly held, so I'm open to pushback if either of you feel strongly that we should keep the natural language check!

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -39,8 +39,5 @@ jobs:
           VALIDATE_PYTHON_ISORT: false
           VALIDATE_PYTHON_PYLINT: false
           VALIDATE_PYTHON_RUFF: false
-          # The natural language check slows down our velocity and is rarely
-          # helpful
-          VALIDATE_NATURAL_LANGUAGE: false
           LINTER_RULES_PATH: /
           YAML_CONFIG_FILE: .yamllint.yml
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -39,5 +39,8 @@ jobs:
           VALIDATE_PYTHON_ISORT: false
           VALIDATE_PYTHON_PYLINT: false
           VALIDATE_PYTHON_RUFF: false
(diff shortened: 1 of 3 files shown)


dfsnow (approved):
I'm fine with this and find myself annoyed with the same spelling nitpicks. Let's nix it.

PR #594, opened 2024-09-11
Fix socrata upload actionworkflow

wrridgeway:
This github workflow is currently failing, this PR is an attempt to fix it.

# BEFORE MERGING
- Remove limits from sql queries
- Change parcel universe asset id to actual open data portal asset id

### DIFF
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -32,7 +32,6 @@ on:
       by_township:
         type: boolean
         description: Chunk Socrata upload by township
-        default: 'false'
         required: false
 
 jobs:
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -36,7 +36,7 @@ on:
 
 jobs:
   upload-open-data-assets:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     permissions:
       # contents:read and id-token:write permissions are needed to interact
       # with GitHub's OIDC Token endpoint so that we can authenticate with AWS
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -36,7 +36,7 @@ on:
 
 jobs:
   upload-open-data-assets:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     permissions:
       # contents:read and id-token:write permissions are needed to interact
       # with GitHub's OIDC Token endpoint so that we can authenticate with AWS
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -42,6 +42,7 @@ jobs:
       # contents:read and id-token:write permissions are needed to interact
       # with GitHub's OIDC Token endpoint so that we can authenticate with AWS
       contents: read
+      id-token: write
 
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


wrridgeway:
I'm marking this as ready for review since the workflow and script are now working as anticipated, but I have some questions about it that need to be addressed before it's merged in.

- I had @dfsnow add a `AWS_REGION` variable to the repo - was this superfluous? Is there already a `AWS_DEFAULT_REGION` variable?
- I have our s3 staging directory hard coded because I don't know the variable name for it.
- Do I have superfluous workflow steps?

.github/workflows/socrata_upload.yaml
```diff
@@ -32,27 +32,31 @@ on:
       by_township:
         type: boolean
         description: Chunk Socrata upload by township
-        default: false
         required: false
 
 jobs:
   upload-open-data-assets:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
jeancochrane, line 39: **[Question, non-blocking]** What's the story here?

.github/workflows/socrata_upload.yaml
```diff
@@ -62,13 +66,13 @@ jobs:
         env:
           AWS_IAM_ROLE_TO_ASSUME_ARN: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
           AWS_ACCOUNT_ID: ${{ secrets.AWS_ACCOUNT_ID }}
-          AWS_ATHENA_S3_STAGING_DIR: ${{ secrets.AWS_ATHENA_S3_STAGING_DIR }}
-          AWS_REGION: ${{ secrets.AWS_REGION }}
+          AWS_ATHENA_S3_STAGING_DIR: ${{ vars.AWS_ATHENA_S3_STAGING_DIR }}
+          AWS_REGION: ${{ vars.AWS_REGION }}
```
jeancochrane, line 70: **[Question, non-blocking]** Elsewhere we just hardcode this to `"us-east-1"`, any reason not to do that here as well?

https://github.com/ccao-data/data-architecture/blob/ac90562f411757e53b74349f2eace76621f81f7d/dbt/scripts/run_iasworld_data_tests.py#L1206-L1210

socrata/socrata_upload.py
```diff
@@ -12,7 +12,7 @@
 
 # Connect to Athena
 cursor = connect(
-    s3_staging_dir=str(os.getenv("AWS_ATHENA_S3_STAGING_DIR")) + "/",
+    s3_staging_dir="s3://ccao-athena-results-us-east-1/",
```
jeancochrane, line 15: **[Question, non-blocking]** If we're not going to read `AWS_ATHENA_S3_STAGING_DIR` from the environment, do we need to pass it in as an env var in the workflow definition above?

jeancochrane (approved):
A couple questions below, but no serious concerns from my end!

.github/workflows/socrata_upload.yaml
wrridgeway: I was struggling with the github action telling me it couldn't find most versions of python. I came across some advice suggesting that the version of ubuntu was the culprit and it worked. I'll try again with latest and see if this solution is still needed.

socrata/socrata_upload.py
wrridgeway: I was getting an error that the staging directory wasn't a proper S3 location. My only thought was that perhaps I'm not passing the script an environmental var that actually exists?

.github/workflows/socrata_upload.yaml
wrridgeway: Okay, seems to no longer be an issue.

socrata/socrata_upload.py
wrridgeway: @jeancochrane This still causes an error for some reason... to you know if the github var actually exists?

PR #601, opened 2024-09-17
Remove +1 increment on unique combination of columns test

wagnerlmichael:


jeancochrane:
@wagnerlmichael If I recall correctly, I believe we realized this was a misunderstanding, so should we go ahead and close this?

PR #603, opened 2024-09-17
Modify dbt test workflow and script to allow control over test selection and artifact output

jeancochrane:
## Background

This PR adjusts a few behaviors of the `test-dbt-models` workflow and its underlying script `run_iasworld_data_tests` that I've come to find frustrating in the process of developing data tests:

* The script and the workflow always run all data tests, even when I only want to run a subset of them
* The workflow doesn't implement [deferral](https://docs.getdbt.com/reference/node-selection/defer), meaning it can't run on a branch unless you manually create all of the missing nodes
* The script and the workflow always parse run metadata to produce output artifacts, even when I don't need them; as a result, script/workflow runs can take ~20 minutes longer than they otherwise would
* The workflow always uploads test result artifacts to S3, even when the output is not significant (that is, basically any run except for the sqoop-bot's runs); as a result, we use S3 storage space that we don't need

## Changes

We implement the following changes to improve the developer experience of using the script and the workflow:

* Allow the caller to select which tests will run
    * Adds `--select` and `--selector` arguments to the `run_iasworld_data_tests.py` script that override the default `--selector select_data_test_iasworld` behavior
    * Adds a `select` input variable to the `test-dbt-models` workflow that passes through to `run_iasworld_data_tests.py`
* Support deferral so that the workflow can run on branches with no extra intervention
    * Adds `--defer` and `--state` arguments to the `run_iasworld_data_tests.py` script that pass through to the underlying `dbt test` call
    * Updates the `test-dbt-models` workflow to pull state from the remote cache and pass it through via `--defer` and `--state` on cache hit (similar to how the `build-and-run-dbt` workflow works)
* Allow the caller to limit artifact creation and upload
    * Adds a `--skip-artifacts`/`--no-skip-artifacts` boolean option to `run_iasworld_data_tests.py` with the following behavior:
        * `--skip-artifacts`: Exits once tests have completed
        * `--no-skip-artifacts` (default): Parses test metadata and produces output artifacts (workbook and Parquet files) 
    * Adds an `upload_test_results` input variable to the `test-dbt-models` workflow that controls whether `--skip-artifacts` or `--no-skip-artifacts` gets passed to the underlying script

## Testing

See these three workflow runs for examples of each set of behaviors:

* `select="--select iasworld_htagnt_unique_by_agent"` and `upload_test_results=false`: [One test runs, no artifacts generated or uploaded](https://github.com/ccao-data/data-architecture/actions/runs/10948222694/job/30398766311#step:5:43)
* `select="--selector data_test_iasworld"` and `upload_test_results=false`: [All tests run, no artifacts generated or uploaded](https://github.com/ccao-data/data-architecture/actions/runs/10948635383/job/30400096194#step:6:54)
* `select="--selector data_test_iasworld"` and `upload_test_results=true`: [All tests run, artifacts generated and uploaded](https://github.com/ccao-data/data-architecture/actions/runs/10949039658/job/30401403752#step:6:54)


.github/workflows/test_dbt_models.yaml
```diff
@@ -3,7 +3,19 @@ name: test-dbt-models
...
 # by the Data Department's sqoop data extraction process upon completion. See
 # https://github.com/ccao-data/service-sqoop-iasworld
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      select:
+        description: dbt select statement for tests to run
+        required: false
+        default: --selector select_data_test_iasworld
```
jeancochrane, line 12: It's annoying that this design requires us to include the option name, but I think it's simpler to merge the `--select` and `--selector` at this level of abstraction rather than requiring the workflow caller to handle two separate variables. I'm open to other design suggestions if you have them, though!

.github/workflows/test_dbt_models.yaml
```diff
@@ -3,7 +3,19 @@ name: test-dbt-models
...
+        type: string
+      upload_test_results:
+        description: Upload test results to S3
+        required: false
+        default: true
```
jeancochrane, line 17: Setting this to `true` by default means that we don't have to update `service-spark-iasworld` to handle this change.

.github/workflows/test_dbt_models.yaml
```diff
@@ -22,25 +34,49 @@ jobs:
         with:
           role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
 
+      - name: Restore dbt state cache
+        id: cache
+        uses: ./.github/actions/restore_dbt_cache
+        with:
+          path: ${{ env.PROJECT_DIR }}/${{ env.STATE_DIR }}
+          key: ${{ env.CACHE_KEY }}
```
jeancochrane, line 42: I adapted this logic from the `build-and-test-dbt` workflow:

https://github.com/ccao-data/data-architecture/blob/ac90562f411757e53b74349f2eace76621f81f7d/.github/workflows/build_and_test_dbt.yaml#L35-L40

.github/workflows/test_dbt_models.yaml
```diff
@@ -22,25 +34,49 @@ jobs:
...
+              $SKIP_ARTIFACTS_OPTION \
+              --defer --state "$STATE_DIR"
+          else
+            # shellcheck disable=SC2086
+            python3 scripts/run_iasworld_data_tests.py \
+              --target "$TARGET" \
+              --output-dir ./qc_test_results/ \
+              $SELECT_OPTION \
+              $SKIP_ARTIFACTS_OPTION
+          fi
```
jeancochrane, line 66: It's annoying that these commands are identical, except that we add the `--defer` and `--state` flags in the case when `CACHE_HIT` is `true`. Is there a cleaner way to do this in bash that's still readable?

dbt/scripts/export_models.py
```diff
@@ -49,19 +49,10 @@ def parse_args():
...
+        *constants.SELECT_ARGUMENT_ARGS, **constants.SELECT_ARGUMENT_KWARGS
     )
     parser.add_argument(
-        "--selector",
-        required=False,
-        help=(
-            "A selector name to use for filtering models, as defined in "
-            "selectors.yml. One of --select or --selector must be set, "
-            "but they can't both be set"
-        ),
```
jeancochrane, line 64: Since I'm reusing these arguments in another script, I factor them out into the `constants` module.

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -855,6 +905,10 @@ def main() -> None:
             ["run_year"],
         ),
     ]:
+        if not metadata_list:
+            print(f"{tablename} is empty, skipping metadata output")
+            continue
```
jeancochrane, line 910: This fixes a bug that I hadn't noticed until implementing the `--select` and `--selector` options: If a test run passes, pyarrow will fail to create a Parquet file from `test_run_failing_row`, since there's no `run_year` to partition on.

dbt/scripts/utils/export.py
```diff
@@ -33,21 +33,21 @@ def export_models(
...
     if not select and not selector:
         raise ValueError("One of --select or --selector is required")
 
-    if select and selector:
-        raise ValueError("--select and --selector cannot both be set")
-
-    select_args = ["--select", *select] if select else ["--selector", selector]  # type: ignore
+    select_args = (
+        ["--selector", selector] if selector else ["--select", *select]  # type: ignore
+    )
```
jeancochrane, line 50: The dbt CLI seems to give `selector` precedence over `select`, so I've flipped our logic here to do the same.

.github/workflows/test_dbt_models.yaml
```diff
@@ -22,31 +34,58 @@ jobs:
...
           s3_prefix="s3://ccao-data-warehouse-us-east-1/qc"
           local_prefix="qc_test_results/metadata"
           for dir in "test_run" "test_run_result" "test_run_failing_row"; do
-            echo "Copying ${dir} metadata to S3"
-            aws s3 sync "${local_prefix}/${dir}" "${s3_prefix}/${dir}"
+            dirpath="${local_prefix}/${dir}"
+            if [ -e "$dirpath" ]; then
+              echo "Copying ${dirpath} metadata to S3"
+              aws s3 sync "$dirpath" "${s3_prefix}/${dir}"
+            fi
```
jeancochrane, line 88: See below for a description of a bug I fixed by sometimes not outputting `test_run_failing_row`.

dbt/scripts/transform_dbt_test_results.py
jeancochrane, line 1: This file has been deprecated for a while and replaced by `run_iasworld_data_tests.py`, but it looks like I neglected to delete it.

dfsnow (approved):

.github/workflows/test_dbt_models.yaml
jeancochrane: Good point! I realized that the `--selector` argument is unnecessary anyway, since we only have one selector (`select_data_test_iasworld`) and we can just default to it if the `select` argument is empty. If we ever add more selectors it should be easy to add support for a `--selector` argument here, but in the meantime, one input variable that overrides the default selector with a custom `--select` statement feels a lot clearer to me.

.github/workflows/test_dbt_models.yaml
jeancochrane: > Otherwise, won't we constantly be uploading new results every time we trigger this workflow?

Not necessarily, unless workflow callers are leaving it checked. But I hear your point that it's a better nudge to leave it unchecked, and it's an easy change to update `service-spark-iasworld` to provide the argument anyway. I made that change and will put up a corresponding `service-spark-iasworld` PR.

.github/workflows/test_dbt_models.yaml
```diff
@@ -22,31 +34,58 @@ jobs:
...
         run: |
-          python3 scripts/run_iasworld_data_tests.py \
-            --target "$TARGET" \
-            --output-dir ./qc_test_results/
+          if [[ $CACHE_HIT == 'true' ]]; then
+            # shellcheck disable=SC2086
+            python3 scripts/run_iasworld_data_tests.py \
+              --target "$TARGET" \
+              --output-dir ./qc_test_results/ \
+              $SELECT_OPTION \
```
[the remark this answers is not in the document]
jeancochrane: Hmm, that's a great question! It certainly opens us up to workflow callers providing arbitrary options, although I don't actually think callers could short circuit the command itself and run an entirely different one. I made a few different attempts to hijack the command, which appeared to confirm that GitHub escapes everything correctly as a string:

* `iasworld_pardat_class_equals_luc; echo oops; exit 1; echo` ([logs](https://github.com/ccao-data/data-architecture/actions/runs/11020740704/job/30606210015#step:7:41))
* `"iasworld_pardat_class_equals_luc"; echo "oops"; exit 1;` ([logs](https://github.com/ccao-data/data-architecture/actions/runs/11020782703/job/30606352546#step:7:41))
* `"iasworld_pardat_class_equals_luc" && echo "oops" && exit 1 &&` ([logs](https://github.com/ccao-data/data-architecture/actions/runs/11020801880/job/30606420046#step:7:41))

In each case, you can see that the workflow correctly interpreted each space-separated token as an argument to `--select` and not as its equivalent bash command or keyword. Can you think of another approach that might work for a bad actor who gained access to our workflows?

Either way, I refactored the input variable so that we always add a `--select` prefix, which should help prevent a caller from passing another option that isn't supported.

.github/workflows/test_dbt_models.yaml
jeancochrane: Great idea, done!

jeancochrane:
This should be ready for one more look once you're back next week @dfsnow!

.github/workflows/test_dbt_models.yaml
```diff
@@ -29,6 +30,20 @@ jobs:
...
+      - name: Validate and parse input variables
+        id: parse-inputs
+        run: |
+          # Default to running all tests
+          SELECT_OPTION="--selector select_data_test_iasworld"
+          if [[ -n "$SELECT" ]]; then
+            SELECT_OPTION="--select $SELECT"
+          fi
+          echo "Setting select option to '$SELECT_OPTION'"
+          echo "select-option=$SELECT_OPTION" >> "$GITHUB_OUTPUT"
```
dfsnow, line 42: **question (non-blocking):** Is GitHub okay with hyphenated env vars/outputs? This definitely wouldn't work in bash.

dfsnow (approved):
The update looks fine to me @jeancochrane!

PR #604, opened 2024-09-18
Update mode calculation for `res_report_summary` view

dfsnow:
Our `build-daily-dbt-models` workflow [started failing this morning](https://github.com/ccao-data/data-architecture/actions/runs/10921987102/job/30325348700) due to an obscure error resulting from the `reporting.res_report_summary` view:

> Query exhausted resources at this scale factor.

I traced the error back to the calculation of the modal class code within the view. The mode calculation seems pretty inefficient since it has to ORDER BY many small groups multiple times. Swapping the count -> sort method for the one from [this SO post](https://stackoverflow.com/a/66954462) fixes the error and has the same query time/data amount.


dbt/macros/assessment_progress_pin_count.sql
```diff
@@ -1,4 +1,4 @@
--- Macros that counts the number of the number of PINs with values and the total
+-- Macro that counts the number of the number of PINs with values and the total
```
dfsnow, line 1: This was just a grammar error I found while poking around the macros. Not related to this PR.

dbt/models/reporting/reporting.res_report_summary.sql
```diff
@@ -133,59 +133,6 @@ sales AS (
...
-            PARTITION BY assessment_stage, townnbhd, year, property_group
-            ORDER BY group_townnbhd_count DESC, class DESC
-        ) AS group_townnbhd_mode,
-        FIRST_VALUE(class) OVER (
-            PARTITION BY assessment_stage, township_code, year
-            ORDER BY group_townnbhd_count DESC, class DESC
-        ) AS town_mode,
-        FIRST_VALUE(class) OVER (
-            PARTITION BY assessment_stage, townnbhd, year
-            ORDER BY group_townnbhd_count DESC, class DESC
```
dfsnow, line 184: The modal class values were actually wrong before, since we weren't ordering by the `_count` column associated with each group.

jeancochrane (approved):
This is some wizard stuff. Nice work! I have less context on what this table is for, so I'll let @wrridgeway decide if the transformation itself seems equivalent.

wrridgeway:
Truly, I had no idea `histogram` was even a presto function. Neat stuff.

wrridgeway (approved):

PR #606, opened 2024-09-18
Tweak Res Edit town close QC report to store percentage columns as numbers

jeancochrane:
I got a call from Valuations this morning to report that column sorting for the percentage columns in the `Res Edit` town close report wasn't behaving as they expected. It turns out that we're casting those values to strings, which causes Excel to sort them as strings instead of numbers.

This PR adds a new `number_format` option to the `export_format` config blocks that we use to format columns in the QC reports. It uses `number_format` to format the `Res Edit` percent columns as percentages, which causes sort behavior to work as expected.

### DIFF
--- a/dbt/scripts/utils/export.py
+++ b/dbt/scripts/utils/export.py
@@ -177,54 +177,73 @@ def export_models(
                 )
                 sheet.add_table(table)
 
+                # Parse column format settings by col index. Since column
+                # formatting needs to be applied at the cell level, we'll
+                # first parse all format settings for each column, and then
+                # we'll iterate every cell once to apply all formatting at the
+                # same time
+                column_format_by_index = {}
+
                 # If a parid column exists, format it explicitly as a
                 # 14-digit number to avoid Excel converting it to scientific
                 # notation or stripping out leading zeros
                 if "parid" in model_df or "pin" in model_df:
                     parid_field = "parid" if "parid" in model_df else "pin"
                     parid_index = model_df.columns.get_loc(parid_field)
-                    # Skip header row when applying formatting. We need to
-                    # catch the special case where there is only one row, or
-                    # else we will iterate the _cells_ in that row instead of
-                    # the row when slicing it from 2 : max_row
-                    non_header_rows = (
-                        [sheet[2]]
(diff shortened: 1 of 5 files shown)


dbt/scripts/utils/export.py
```diff
@@ -177,54 +177,73 @@ def export_models(
                 )
                 sheet.add_table(table)
 
+                # Parse column format settings by col index. Since column
+                # formatting needs to be applied at the cell level, we'll
+                # first parse all format settings for each column, and then
+                # we'll iterate every cell once to apply all formatting at the
+                # same time
+                column_format_by_index = {}
```
jeancochrane, line 185: The refactor in this file is more substantial than I was expecting, and we could probably have gotten away with a smaller change, but my motivation was to reduce the number of times we iterate the rows of the spreadsheet to apply formatting. By parsing the column formats ahead of time, we can apply them all in one pass, in contrast to the previous approach that required a separate pass for each format config.

dbt/scripts/utils/export.py
```diff
@@ -177,54 +177,73 @@ def export_models(
...
-                                raise ValueError(
-                                    "'index' attribute is required when "
-                                    "'horizontal_align' is set on "
-                                    "export_format.columns config for "
-                                    f"model {model_name}"
-                                )
-                            idx = column_index_from_string(col_letter) - 1
-                            # Skip header row
-                            for row in sheet[2 : sheet.max_row]:
-                                row[idx].alignment = horizontal_alignment
```
jeancochrane, line 228: In contrast to the `parid` formatting above, the previous version of this loop wouldn't have handled single-row spreadsheets properly (i.e. empty sheets with only a header row). We hadn't yet run into a case where this loop needed to format an empty sheet, so it hadn't caused problems yet, but the refactor will handle this case much better if it ever comes up.

dfsnow (changes requested):
@jeancochrane The code here looks good, however, the column formatting for `Res edit.xlsx` still seems wrong. It looks like the number formatting _is_ applied (per the Number dropdown in the Home menu), but the percentage number is still stored as text.

dfsnow (approved):

PR #609, opened 2024-09-27
Add missing tax columns to socrata column transformations

wrridgeway:
These columns were accidentally left out in the past from being transformed on the open data portal


jeancochrane (approved):
Nice catch!

PR #611, opened 2024-10-03
Add homeowner exemption legacy extract to data lake

jeancochrane:
This PR follows up on Dan's work in https://github.com/ccao-data/data-architecture/pull/589 to add legacy data extracts from MV to the data lake in order to enable our work to compare homeowner exemptions between data sources.

### DIFF
--- a/etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
@@ -287,12 +287,12 @@ cc_pifdb_piexemptre_ownr <- map_dfr(files_cc_pifdb_piexemptre_ownr$Key, \(f) {
       segment_code = c(24, 24),
       printed_indicator = c(26, 26),
       response = c(28, 28),
-      year_applied = c(30, 33),  # The header file thinks this is 2 chars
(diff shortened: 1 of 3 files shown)


etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-legacy.R
```diff
@@ -262,3 +262,88 @@ cc_pifdb_piexemptre_dise %>%
...
   )
+
+
+##### CC_PIFDB_PIEXEMPTRE_OWNR #####
+files_cc_pifdb_piexemptre_ownr <- aws.s3::get_bucket_df(
+  bucket = AWS_S3_RAW_BUCKET,
+  prefix = "ccao/legacy/CC_PIFDB_PIEXEMPTRE_OWNR",
+  max = Inf
+) %>%
+  filter(Size > 0)
```
jeancochrane, line 273: All of the code in this file is very similar to the logic to extract the other exemptions that Dan established in https://github.com/ccao-data/data-architecture/pull/589. The only bit that is specific to the homeowner data are the column positions and types that we define below, along with the transformations we apply to those columns.

wrridgeway (approved):
Looks good!

PR #617, opened 2024-10-07
Add IDOT data to data-lake

Damonamajor:
This uploads the IDOT data from the following [location ](https://apps1.dot.illinois.gov/gist2/)to our data lake in order to process it into future characteristics.

- The transformation first unions roads with identical characteristics into the same line string.
- It then joins roads with the same name and missing features by averaging the features of intersecting roads with the same name.

PR #620, opened 2024-10-09
Create IDOT features

Damonamajor:
This PR takes the output of [617](https://github.com/ccao-data/data-architecture/pull/617) and transforms the data into proximity features.
Basic Steps:
- Identifies the proximity characteristics for each road typology. For example, nearest_highway_lanes would be the number of lanes for the nearest highway. It also identifies the distance of the closest highway with the standard dist_ft suffix.
- Using the distance feature, it uses a [macro ]([dbt/macros/nearest_feature_aggregation.sql](https://github.com/ccao-data/data-architecture/pull/620/files#diff-d955c80b10f691c0d9044b23cb57226d54233ac36f2883744809881f9159f183)) to identify the smallest distance value. With that distance value, it creates features for nearest_road_lanes, nearest_road_dist_ft, etc..

Ongoing questions:
- The local roads still times out when run for all characteristics. Is this something we want to investigate. Currently we only identify the local_road features where there is a valid average daily traffic.
--  Local roads could use a structure like the one present in `dist_pin_to_pin`. But, the macro as currently structured doesn't seem to support it. Is this a priority to fix?
-- The method could subset the data to speed up the process would be to include only roads and parcels in the same township. Is that a structure that we would want?
- Building on this, would we want to separate out major and secondary roads as we did before? Something like highways / freeways / collectors/ arterial / local makes sense in my head. At minimum, it is probably necessary to remove local roads from this query. If a secondary road and a local road are both abutting a street, we would probably want to make sure the secondary road is accounted for even if it is 5 ft farther away. 
- Do we want to keep the nearest feature as a macro since it could be used again (with a little tuning)?
-- I also did some playing with it, and it seemed the easiest way to do it without a macro would be doing a case when for each new variable. So in this sense, the macro does make sense.


dbt/models/proximity/proximity.dist_pin_to_traffic_master.sql
```diff
@@ -0,0 +1,340 @@
...
+    FROM {{ source('spatial', 'traffic') }}
+    WHERE road_type = 'Local Road or Street'
+        AND daily_traffic IS NOT NULL
+),
+
+traffic_collector AS (  -- noqa: ST03
+    SELECT *
+    FROM {{ source('spatial', 'traffic') }}
+    WHERE road_type = 'Major Collector'
+        OR road_type = 'Minor Collector'
```
Damonamajor, line 39: Minor collectors aren't present in all years and create error messages on joins.

dbt/models/proximity/proximity.dist_pin_to_traffic_master.sql
```diff
@@ -0,0 +1,340 @@
...
+    nearest_other_road_surface_type,
+    nearest_other_road_lanes,
+    {{ nearest_feature_aggregation(
+        [
+            "nearest_minor_road_dist_ft", 
+            "nearest_interstate_road_dist_ft", 
+            "nearest_other_road_dist_ft",
+            "nearest_freeway_road_dist_ft",
+            "nearest_local_road_dist_ft",
+            "nearest_collector_road_dist_ft"
```
Damonamajor, line 327: I ran some tests and this should work if we don't want to include the macro.
```
case
        when nearest_minor_road_dist_ft = least(
            nearest_minor_road_dist_ft, 
            nearest_interstate_road_dist_ft, 
            nearest_other_road_dist_ft,
            nearest_freeway_road_dist_ft,
            nearest_local_road_dist_ft,
            nearest_collector_road_dist_ft
        ) then nearest_minor_road_name
        when nearest_interstate_road_dist_ft = least(
            nearest_minor_road_dist_ft, 
            nearest_interstate_road_dist_ft, 
            nearest_other_road_dist_ft,
            nearest_freeway_road_dist_ft,
            nearest_local_road_dist_ft,
            nearest_collector_road_dist_ft
        ) then nearest_interstate_road_name
        when nearest_other_road_dist_ft = least(
            nearest_minor_road_dist_ft, 
            nearest_interstate_road_dist_ft, 
            nearest_other_road_dist_ft,
            nearest_freeway_road_dist_ft,
            nearest_local_road_dist_ft,
            nearest_collector_road_dist_ft
        ) then nearest_other_road_name
        when nearest_freeway_road_dist_ft = least(
            nearest_minor_road_dist_ft, 
            nearest_interstate_road_dist_ft, 
            nearest_other_road_dist_ft,
            nearest_freeway_road_dist_ft,
            nearest_local_road_dist_ft,
            nearest_collector_road_dist_ft
        ) then nearest_freeway_road_name
        when nearest_local_road_dist_ft = least(
            nearest_minor_road_dist_ft, 
            nearest_interstate_road_dist_ft, 
            nearest_other_road_dist_ft,
            nearest_freeway_road_dist_ft,
            nearest_local_road_dist_ft,
            nearest_collector_road_dist_ft
        ) then nearest_local_road_name
        when nearest_collector_road_dist_ft = least(
            nearest_minor_road_dist_ft, 
            nearest_interstate_road_dist_ft, 
            nearest_other_road_dist_ft,
            nearest_freeway_road_dist_ft,
            nearest_local_road_dist_ft,
            nearest_collector_road_dist_ft
        ) then nearest_collector_road_name
        else null
    end as closest_road_name,
    year
FROM final_aggregation
```

dbt/models/proximity/proximity.dist_pin_to_traffic_master.sql
```diff
@@ -0,0 +1,340 @@
+{{
+    config(
+        materialized='table',
+        partitioned_by=['year'],
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH traffic_minor AS (  -- noqa: ST03
```
wrridgeway, line 10: do we need all these noqa's still?

dbt/models/proximity/proximity.dist_pin_to_traffic_master.sql
```diff
@@ -0,0 +1,340 @@
...
+
+distinct_pins AS (
+    SELECT DISTINCT
+        x_3435,
+        y_3435,
+        pin10
+    FROM {{ source('spatial', 'parcel') }}
+),
+
+distinct_years AS (
```
wrridgeway, line 56: This CTE seems to only get used once on line 769, where we grab the min year from it. We can probably just do that without the CTE.

dbt/models/proximity/proximity.dist_pin_to_traffic_master.sql
```diff
@@ -0,0 +1,340 @@
...
+            local_road.year,
+            collector.year
+        )
+        = other.year
+    WHERE COALESCE(
+            minor.year, interstate.year, freeway.year,
+            local_road.year, collector.year,
+            other.year
+        )
+        >= (SELECT MIN(year) FROM distinct_years)
```
wrridgeway, line 272:
```suggestion
        >= (SELECT MIN(year) FROM "awsdatacatalog"."spatial"."traffic"
    WHERE road_type IS NOT NULL)
```

dbt/models/proximity/proximity.dist_pin_to_traffic_master.sql
```diff
@@ -0,0 +1,340 @@
...
+        other.nearest_other_road_data_year,
+        other.nearest_other_road_daily_traffic,
+        other.nearest_other_road_speed_limit,
+        other.nearest_other_road_surface_type,
+        other.nearest_other_road_lanes,
+        COALESCE(
+            minor.year, interstate.year, freeway.year,
+            local_road.year, collector.year, other.year)
+            AS year
+    FROM nearest_minor AS minor
```
wrridgeway, line 230: I think rather than doing all of this coalescing for `pin10` you could just start with `spatial.parcel` as your base for pins and years and left join everything on to that. My guess is that this will be less resource intensive.

wrridgeway (changes requested):
This is looking really good. I've got a few requests, but I don't think there's a ton to do.

Damonamajor:
@wrridgeway

Questions:

- At the moment highways are coded as highway_roads. This doesn't make sense linguistically, but does make sense when keeping all the data in the same structure. Thoughts?
- The features in the added model are coded as nearest_collector_road_lanes. Do we want to keep them in proximity, since it is proximity to nearest road or have them in another chunk (maybe environment or it's on chunk)?
- The initial dataset was named as traffic. It makes sense to rename this to roads now, but the workflow is wonky with aws uploads? Do we want to rename this in a separate PR?
- A lot of highways have 1 lane. Do we want to just filter to highways with more than 1 lane since these are going to be the service lanes on the side of the road?

wrridgeway:
> At the moment highways are coded as highway_roads. This doesn't make sense linguistically, but does make sense when keeping columns in the same structure / if someone wants to find road columns. For example prox_nearest_highway_road_dist_ft. Thoughts?

I'd actually prefer a road _prefix_ rather than suffix for all the roads, highway included.

> The features in the added model in model.vw_pin_shared_input.sql are coded as nearest_collector_road_lanes. Do we want to keep them in proximity, since it is proximity to nearest road or have them in another chunk (maybe environment or it's own chunk)?

I'm fine with it as is.

> The initial dataset was named as traffic. It makes sense to rename this to roads now, but the workflow is wonky with aws uploads? Do we want to rename this in a separate PR?

I realize it's a pain in the ass, but getting the names right before we merge things into master is preferable. I'm happy to run anything if you need me to.

> A lot of highways have 1 lane. Do we want to just filter to highways with more than 1 lane since these are mostly service lanes on the side of the road?

Let's leave it for now and then we can open a new PR after we're done here investigating this kind of additional processing.

dbt/models/proximity/proximity.dist_pin_to_road_arterial.sql
```diff
@@ -0,0 +1,32 @@
...
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH arterial AS (  -- noqa: ST03
+    SELECT *
+    FROM {{ source('spatial', 'road') }}
+    WHERE road_type = 'Minor Arterial'
+        OR road_type = 'Other Principal Arterial'
```
wrridgeway, line 14:
```suggestion
    WHERE road_type IN ('Minor Arterial', 'Other Principal Arterial')
```

dbt/models/proximity/proximity.dist_pin_to_road_collector.sql
```diff
@@ -0,0 +1,32 @@
...
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH collector AS (  -- noqa: ST03
+    SELECT *
+    FROM {{ source('spatial', 'road') }}
+    WHERE road_type = 'Major Collector'
+        OR road_type = 'Minor Collector'
```
wrridgeway, line 14:
```suggestion
    WHERE road_type IN ('Major Collector', 'Minor Collector')
```

dbt/models/proximity/proximity.dist_pin_to_road_highway.sql
```diff
@@ -0,0 +1,32 @@
...
+        bucketed_by=['pin10'],
+        bucket_count=1
+    )
+}}
+
+WITH highway AS (  -- noqa: ST03
+    SELECT *
+    FROM {{ source('spatial', 'road') }}
+    WHERE road_type = 'Interstate'
+        OR road_type = 'Freeway and Expressway'
```
wrridgeway, line 14:
```suggestion
    WHERE road_type IN ('Interstate', 'Freeway and Expressway')
```

wrridgeway (approved):
Wildly nit-picky last few changes. Merge them in, make sure it builds, then feel free to merge in your work. Awesome job on this.

Issue #621, opened 2024-10-16
Correct `CDU` column for `iasworld.dweldat` in `default.vw_pin_status`


wrridgeway:
The dweldat column feeding `dweldat_cdu` in `default.vw_pin_status` should be `user16` rather than `cdu`.

wrridgeway:
See https://github.com/ccao-data/data-architecture/pull/624

PR #622, opened 2024-10-16
Correct `CDU` column for `iasworld.dweldat` in `default.vw_pin_status`

wrridgeway:
See https://github.com/ccao-data/data-architecture/issues/621

### DIFF
--- a/dbt/macros/aggregate_cdu.sql
+++ b/dbt/macros/aggregate_cdu.sql
@@ -3,7 +3,9 @@
     select
         parid,
         taxyr,
-        array_join(array_sort(array_distinct(array_agg({{ cdu_column }}))), ', ') as cdu
+        array_join(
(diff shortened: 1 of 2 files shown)


dbt/macros/aggregate_cdu.sql
```diff
@@ -3,7 +3,9 @@
     select
         parid,
         taxyr,
-        array_join(array_sort(array_distinct(array_agg({{ cdu_column }}))), ', ') as cdu
+        array_join(
+            array_sort(array_distinct(array_agg(upper({{ cdu_column }})))), ', '
```
wrridgeway, line 7: Apparently there are some lowercase CDUs, as well. Might as well fix that while we're in here. 🤦

wrridgeway:
See https://github.com/ccao-data/data-architecture/pull/624

PR #626, opened 2024-10-23
Update QC town close export script to prep for automation

jeancochrane:
This PR makes a few quality-of-life improvements to the QC town close export script to enable automation via any distribution mechanism (VM, OneDrive, or S3). The changes include:

* Define a new `Township` class that owns a few different configurations for each township:
    * Township code and human-readable name
    * Tri code and human-readable name
    * Start/end dates for "active" periods, indicating periods when towns are eligible for QC
* Support for a config file at `scripts/utils/town_active_schedule.csv` that can optionally configure the list of available `Township`s
* Change the `--township` option so that it accepts one or more township codes
    * When the caller provides multiple codes, the script will output one workbook for each report and workbook readers can filter the `township_code` column for the township they are interested in
    * When the caller provides one code, the script will prepend the human-readable township name to the filename for each output workbook (e.g. "Res edit.xlsx" will become "Jefferson Res edit.xlsx"). This feature automates the renaming operation that we perform when exporting these reports manually
* Change the `--township` option so that it is optional, and when omitted, falls back to all towns that are active per the `town_active_schedule`
* Add  an `--output-dir` option that controls the directory where the script will save reports

With these changes, we can follow these steps to define an automated process to export QC reports on a regular schedule:

* Create a `scripts/utils/town_active_schedule.csv` config file on the machine that will run the process (or download it from S3 in an ephemeral environment like a GitHub workflow) with a defined activity schedule
* Schedule a process to run the `export_qc_town_close.py` script at regular intervals
    * Omit the `--township` argument so that the script will export all active towns
    * Optionally set the `--output-dir` option on environments like the VM that need to write to specific locations but cannot chain an `mv` call to the scheduled `export_qc_town_close.py` call


dbt/scripts/utils/town_active_schedule.csv.example
```diff
@@ -0,0 +1 @@
+township_code,township_name,tri_code,tri_name,active_start_date,active_end_date
```
jeancochrane, line 1: The fact that we'd have to keep this schedule up-to-date based on Valuations' QC requirements is the biggest downside of this approach, I think. Still, it strikes me as better than producing reports for every township every day, since the query load would actually be quite extensive if we wanted to do that.

One alternative might be to refactor the way that the `export_qc_town_close_reports.py` script works so that it queries data for all towns in one Athena query and then splits up the data for export by township in Python. This might actually be a more sustainable solution from an operational perspective, but since it's a bigger refactor, I'm curious to hear what you think.

dfsnow: Having looked through this a little bit more, I actually like this first approach more than the second one. It's nice to be able to run a single town-level report ad-hoc, which the second option wouldn't really support.

dfsnow: **suggestion (non-blocking):** May want to add an example line in here, just to demo the format of dates and such.

dbt/scripts/utils/townships.py
```diff
@@ -0,0 +1,383 @@
...
+    # so that the script still works without access to the town_active_schedule
+    TOWNSHIPS = [
+        Township(
+            township_code="10",
+            township_name="Barrington",
+            tri_code="2",
+            tri_name="North",
+            active_start_date=None,
+            active_end_date=None,
+        ),
```
dfsnow, line 81: **suggestion (non-blocking):** Could you DRY this up a bit by just looping through a list of tuples with the positional values for each township? i.e.

```python
TOWNSHIPS_DATA = [
    ("10", "Barrington", "2", "North"),
    ("11", "Berwyn", "3", "South"),
    # blah blah blah
]

TOWNSHIPS = [
    Township(
        township_code=code,
        township_name=name,
        tri_code=tri_code,
        tri_name=tri_name,
        active_start_date=None,
        active_end_date=None,
    )
    for code, name, tri_code, tri_name in TOWNSHIPS_DATA
]
```

dfsnow: Better yet, why not just have a default `town_active_schedule.csv` file with all this info but the start and end date blank, then just instantiate the townships based on what's available in that sheet.

dfsnow (approved):
I tested everything and it all works great. Think we could simplify the township instantiation a little bit but nothing major/blocking.

dbt/scripts/export_qc_town_close_reports.py
```diff
@@ -203,13 +191,63 @@ def main():
...
+                # Filter the query results for only this town, but only if the
+                # query results are not empty, since otherwise they will have
+                # no columns
+                town_df = (
+                    model.df
+                    if model.df.empty
+                    else model.df[
+                        model.df["township_code"] == town.township_code
+                    ]
+                )
```
jeancochrane, line 228: There were two ways I thought about filtering results by township:

1. As part of the Athena query that pulls results, as the previous iteration of this PR did
2. _After_ pulling data from Athena, as this code block does

Option 2 allows us to more easily export a model for each town, since filtering in Pandas allows us to avoid incurring the overhead of a bunch of extra Athena queries.

dbt/scripts/utils/export.py
```diff
@@ -40,6 +48,69 @@ def export_models(
...
+        select=select,
+        selector=selector,
+        rebuild=rebuild,
+        where=where,
+    )
+    output_paths: list[pathlib.Path] = []
+    for model in models_for_export:
+        output_path = save_model_to_workbook(model, output_dir)
+        output_paths.append(output_path)
+    return output_paths
```
jeancochrane, line 67: I don't feel 100% great about the way that I've cut up this function to expose internals that `export_qc_town_close_reports` can use to modify the query data. In particular, I think it's kind of cumbersome to deal with the `ModelForExport` abstraction that acts as the interface between the `query` and `save` functions. I'm curious if you have thoughts on a better way to separate the concerns.

dbt/scripts/utils/export.py
```diff
@@ -101,149 +173,190 @@ def export_models(
...
-            shutil.copyfile(template_path, output_path)
-        else:
-            print("No template file exists; creating a workbook from scratch")
+    return models_for_export
+
+
+def save_model_to_workbook(
+    model: ModelForExport,
+    output_dir: str | None = None,
+) -> pathlib.Path:
```
jeancochrane, line 225: This interface in particular feels a bit brittle to me, since in order to add or change a parameter you have to edit the `ModelForExport` interface as well. A caller also has to understand `ModelForExport` and construct if not calling `query_models_for_export` first or if modifying query results between the `query` and the `save` state. Not perfect, but it was the best incremental design improvement I could think of.

dbt/scripts/utils/export.py
```diff
@@ -101,149 +173,190 @@ def export_models(
...
+    Returns a pathlib.Path object representing the path to the file that the
+    function created.
+    """
+    # Compute the full filepath for the output
+    output_dirpath = (
+        pathlib.Path(output_dir)
+        if output_dir
+        else pathlib.Path("export/output")
+    )
+    output_path = output_dirpath / model.export_filename
```
jeancochrane, line 247: Beyond this step, very little of the logic of this function has changed, it's just present in the diff because we've extracted it out from the old `export_models()` implementation. I'll leave a comment on anything major that needs attention.

dbt/scripts/utils/townships.py
```diff
@@ -0,0 +1,78 @@
...
+    township_name: str
+    tri_code: str
+    tri_name: str
+
+    def is_reassessed_during(self, year: int) -> bool:
+        """Helper function to determine if a town in a given year is undergoing
+        triennial reassessment"""
+        # 2024 is the City reassessment year (tri code 1), so
+        # ((2024 - 2024) % 3) + 1 == 1, and so on for the other two tris
+        return str(((year - 2024) % 3) + 1) == self.tri_code
```
jeancochrane, line 19: This abstraction really seems like it should live in a `ccao` package, but absent that package, I think it's useful for the purposes of the `export` code.

dbt/scripts/utils/townships.py
```diff
@@ -0,0 +1,78 @@
...
+    ("39", "Worth", "3", "South"),
+    ("70", "Hyde Park", "1", "City"),
+    ("71", "Jefferson", "1", "City"),
+    ("72", "Lake", "1", "City"),
+    ("73", "Lake View", "1", "City"),
+    ("74", "North Chicago", "1", "City"),
+    ("75", "Rogers Park", "1", "City"),
+    ("76", "South Chicago", "1", "City"),
+    ("77", "West Chicago", "1", "City"),
+]
```
jeancochrane, line 63: I wonder if this data would make sense as a dbt seed? Not something that I think we need to do now, but something I'm thinking about for the future.

dbt/scripts/utils/townships.py
jeancochrane: I've refactored the design to move away from the activity schedule, so this point is moot now.

dbt/scripts/utils/town_active_schedule.csv.example
jeancochrane: I've refactored the design to move away from the activity schedule approach, so this point is moot now.

jeancochrane:
@dfsnow I finished a pretty major refactor of the PR, so I think it's worth taking a fresh look at everything. Instead of the no-filter default being a set of active towns determined by a schedule, the script will now default to exporting reports for _all_ towns, using pandas to speed up the township filtering operation. Testing this locally, it takes about 30 minutes to run an export for all towns, and we scan the same amount of data in Athena since our tables are not partitioned by township anyway.

dbt/scripts/utils/townships.py
dfsnow: We should make it a seed _and_ have it as a built-in dataset in the `ccao` package once that's setup.

dbt/scripts/utils/townships.py
dfsnow: Yep! We can move it there once that package is finished.

dfsnow (approved):
You were right, this setup is definitely a bit simpler, even if it requires splitting things into query and save components. I think this is fine for now, it's a good incremental improvement. If we need to refactor it later it should be straightforward.

dbt/scripts/utils/export.py
dfsnow: Kind of a bizarre idea, but maybe something like:

- Have a `query` for each model
- Have a `filter` for each model

Create a unique set of all the `query` values across models, run them, then populate the models using the results (and apply the filter to each one).

Might be more complicated? Maybe save it for the next incremental improvement.

jeancochrane: I like this idea! I'll open up an issue witih this design as an iterative improvement, and then pick it up once we've validated that the ability to run reports for all towns is an important feature for the project.

jeancochrane: Issue open here: https://github.com/ccao-data/data-architecture/issues/637

PR #633, opened 2024-11-05
Switch dependency management to pyproject.toml

dfsnow:
This PR moves us from using `requirements.txt` files for Python dependency management to using `pyproject.toml` files. \*insert Drake meme\*

This change allows us to more easily define optional dependencies of our various data architecture projects. For example, instead of having three `requirements.txt` files for three dbt testing workflows, we can instead have a single `pyproject.toml` file with `optional-dependencies` keys containing the dependencies for each of the three workflows.

This change also: lets us define a minimum required Python version, cleans up some Actions scripts, and consolidates some overlapping requirements.


.github/actions/install_dbt_requirements/action.yaml
```diff
@@ -5,10 +5,6 @@ inputs:
     description: Path to the directory containing the dbt project.
     required: false
     default: ./dbt
-  requirements_file_path:
-    description: Path to Python requirements file.
-    required: false
-    default: ./dbt/requirements.txt
```
dfsnow, line 11: Is there a context in which this is ever called? I couldn't find one so I removed it in favor of just setting a working directory.

.github/workflows/check_sort_dbt_yaml_files.yaml
```diff
@@ -10,15 +10,22 @@ jobs:
...
+        python-version: '3.11'
+        cache: pip
+
+    - name: Install Python dependencies
+      working-directory: dbt
+      shell: bash
       run: |
-        python -m pip install --upgrade pip
-        pip install pyyaml
+        # shellcheck disable=SC2102
```
dfsnow, line 26: Shellcheck gets mad about the `[]` around the optional deps, so need to disable this rule.

.github/workflows/check_sort_dbt_yaml_files.yaml
dfsnow, line 1: I did a little bit of Actions cleanup while I was in a lot of these scripts. Just things like giving steps consistent names, enabled caching where available, etc.

dbt/pyproject.toml
dfsnow, line 1: This guy defines the dependencies related to our dbt projects and tests. I consolidated the three different requirements files into a single `dbt_tests` optional dependency since they had a lot of overlap.

Note that `pyproject` files are hierarchical, so the tool settings and such are still inherited from the root `pyproject.toml`.

dbt/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[project]
+name = "dbt"
+version = "0.0.1"
```
dfsnow, line 3: You unfortunately _have_ to set a version key, even if it's not really applicable to your project. I just set all of ours to `0.0.1`.

dbt/pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+readme = "README.md"
+requires-python = ">=3.11"
```
dfsnow, line 5: I matched the required/used Python version across all our CI jobs and `pyproject` files to 3.11. Everyone _should_ be up to that as the default now on the server. If not, I can bump the global version.

dbt/pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+  "pyarrow==15.0.2",
+  "pyathena~=3.8.3",
+  "simplejson~=3.19.2",
+]
+ci_checks = [
+  "pyyaml==6.0.2",
+]
+
+[tool.setuptools]
+packages = []
```
dfsnow, line 27: This little addendum is required or pip will try to install all your subdirectories as modules/packages.

dbt/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[project]
+name = "dbt"
+version = "0.0.1"
+readme = "README.md"
+requires-python = ">=3.11"
+dependencies = [
+  "dbt-athena-community==1.8.2",
+  "dbt-core==1.8.3",
```
dfsnow, line 8: Note that I basically copied the versions for packages verbatim from the requirements.txt.

pyproject.toml
```diff
@@ -1,3 +1,12 @@
+[project]
+name = "data-architecture"
+version = "0.0.1"
+readme = "README.md"
+requires-python = ">=3.11"
+
+[tool.pyright]
+reportUnusedImport = false
```
dfsnow, line 8: Adding this to the global tool config since `ruff` already handles checking for unused imports.

.github/actions/install_dbt_requirements/action.yaml
jeancochrane: **[Praise]** I think it's fine to consolidate this with the working directory. I was just being a bit overly defensive following [GitHub's advice to define all filepaths as environment variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables).

.github/workflows/check_sort_dbt_yaml_files.yaml
jeancochrane: **[Suggestion, non-blocking]** Could we avoid disabling the rule if we quoted the `.[ci_checks]` argument, per [the shellcheck docs](https://www.shellcheck.net/wiki/SC2102#exceptions)?

jeancochrane (approved):
Nice work! This is indeed way simpler than the prior solution using multiple `requirements.txt` files. Looking forward to seeing the follow-up uv PR.

.github/workflows/check_sort_dbt_yaml_files.yaml
dfsnow: Done in 40245bf!

PR #635, opened 2024-11-06
Switch to uv for Python management

dfsnow:
This PR switches to using [`uv`](https://docs.astral.sh/uv/) for environment and dependency management in our data architecture repo and Actions.

Locally, this (optionally) allows users to:
- Use `uv` for virtual env management (via `uv venv && source .venv/bin/activate`) with a per-subdirectory virtualenv (i.e. dbt has its own `dbt/.venv` directory)
- Install the pinned Python version (3.11) into the virtualenv using `uv python install`
- Install packages into that virtualenv using `uv pip install .`

For Actions:
- `uv` now handles caching and installing pip dependencies
- The standard, GitHub-provided setup Action uses the pinned Python version in each project
- We ignore virtual environments, instead using `uv` to install directly to the system library (just as we did before with plain `pip`)

Why make this switch?
- `uv` is _significantly_ faster when installing Python dependencies in Actions (10-100x faster)
- It versions and isolates the dependencies used by each subproject (`dbt/`, `socrata/`, etc.) via `uv.lock` files
- It pins the Python version used by each subproject and uses the same version within Actions via `.python-version` files

The downsides are:
- It's something new to learn for users. However, the commands are simpler than standard `venv` and users can still use plain `pip` and `venv` if they want to
- It's still slightly hacky to install into the system environment within Actions using `uv`


.github/actions/install_dbt_requirements/action.yaml
```diff
@@ -8,16 +8,25 @@ inputs:
...
-        uses: actions/setup-python@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
         with:
-          python-version: '3.11'
-          cache: pip
+          enable-cache: true
+          cache-dependency-glob: |
+            ${{ inputs.dbt_project_dir }}/pyproject.toml
+            ${{ inputs.dbt_project_dir }}/uv.lock
```
dfsnow, line 17: Unlike before, we have to explicitly specify the cache keys for pip dependencies. This is actually probably a good thing insofar as it lets us isolate each of the subproject dependencies in their own cache.

.github/actions/install_dbt_requirements/action.yaml
```diff
@@ -8,16 +8,25 @@ inputs:
...
+        with:
+          python-version-file: ${{ inputs.dbt_project_dir }}/.python-version
+
+      - name: Install Python dependencies
+        working-directory: ${{ inputs.dbt_project_dir }}
         shell: bash
-        run: python -m pip install .
+        run: uv pip install .
+        env:
+          UV_SYSTEM_PYTHON: 1
```
dfsnow, line 29: This env var [is required](https://docs.astral.sh/uv/guides/integration/github/#using-uv-pip) if we want to install the pip dependencies into the system library rather than a `uv` virtual env. We _could_ use the `uv` virtual env but we'd need to switch all the run commands from `python $SCRIPT` to `uv run $SCRIPT` and I thought this was easier.

.github/actions/install_dbt_requirements/action.yaml
```diff
@@ -8,16 +8,25 @@ inputs:
...
+          cache-dependency-glob: |
+            ${{ inputs.dbt_project_dir }}/pyproject.toml
+            ${{ inputs.dbt_project_dir }}/uv.lock
 
-      - name: Install python requirements
-        working-directory: dbt
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: ${{ inputs.dbt_project_dir }}/.python-version
```
dfsnow, line 22: Having this file lets us keep our Python version synchronized with `uv`. If you want to use the Python version associated with the project locally, you just `uv python install` with the virtualenv active.

jeancochrane (approved):
Really nice! I tested this out locally and it works great. No concerns from my end.

PR #639, opened 2024-11-07
Exclude new uv venv path from pre-commit

jeancochrane:
This PR updates our pre-commit config to exclude the [new virtualenv path that we recommend folks use with uv](https://github.com/ccao-data/data-architecture/tree/master/dbt#install-dependencies) from linting. Without this change, sqlfmt will format files in the dbt virtualenv that uv creates for anyone who is using our new virtualenv system.


dfsnow (approved):
Whoops, nice catch. Knew I'd miss something or other.

PR #644, opened 2024-11-18
Move `check_sort_dbt_yaml_files` linter to pre-commit

jeancochrane:
The dbt YAML sort order linter we introduced in https://github.com/ccao-data/data-architecture/pull/578 is awesome, but I find it annoying that I only get notified of failures after I've already pushed a commit. This PR updates our config to move the linter into pre-commit so that we can be notified of failures more proactively.

In order to speed up the pre-commit hook when installed locally, this PR also adjusts the interface for the `check_sort_dbt_yaml_files.py` script so that it can accept a list of filepaths from pre-commit to check for errors. When testing locally on one YAML file changed to have an incorrect sort order, the check fails nearly instantly.


.pre-commit-config.yaml
```diff
@@ -30,3 +30,12 @@ repos:
...
       - id: ruff-format
+  - repo: local
+    hooks:
+      - id: check-sort-dbt-yaml-files
+        name: check-sort-dbt-yaml-files
+        entry: python3 dbt/scripts/check_sort_dbt_yaml_files.py
+        language: system
+        types_or: [yaml, markdown]
+        files: ^dbt/
+        exclude: venv
```
jeancochrane, line 41: I double-checked the filenames generated by this config when `--all-files` is present by comparing it to the output of this `find` command and confirming they match:

```
find dbt/ \( -path "dbt/.venv" -o -path "dbt/node_modules" \) -prune -o -type f \( -name "*.yaml" -o -name "*.md" -o -name "*.yml" \)
```

I don't really understand how `node_modules` gets excluded from pre-commit, but it does, and the file paths match 🤷🏻‍♀️

dbt/scripts/check_sort_dbt_yaml_files.py
```diff
@@ -340,42 +341,68 @@ def check_all_files(directory):
...
+            unsorted_columns_md = check_columns_md_file(file_path)
+            if unsorted_columns_md:
+                unsorted_columns_md_files.append(unsorted_columns_md)
+        elif os.path.basename(file_path) == "shared_columns.md":
+            unsorted_shared_columns_md = check_shared_columns_md_file(
+                file_path
+            )
+            if unsorted_shared_columns_md:
+                unsorted_shared_columns_md_files.append(
+                    unsorted_shared_columns_md
```
jeancochrane, line 404: No logic here has changed, I've just changed indentation and some variable references to match the new function structure and arguments.

dfsnow (approved):
Looks good to me @jeancochrane. Nice QoL fix.

PR #647, opened 2024-11-18
Further improvements to dbt sort order linter

jeancochrane:
https://github.com/ccao-data/data-architecture/pull/644 got merged in before we could implement two pieces of feedback:

* Adjusting the `exclude` attribute of the pre-commit hook to be more precise and only exclude files in a directory matching the string `venv/` rather than any file containing the substring `venv`, which could exist without being in a venv dir ([comment](https://github.com/ccao-data/data-architecture/pull/644#discussion_r1847173185))
* Adding a docstring to explain the different cases in which the linter runs, and how it expects arguments ([comments](https://github.com/ccao-data/data-architecture/pull/644#discussion_r1847175457))

In the course of implementing these two pieces of feedback, I also noticed that the logic to check for properly sorted tests is only looking at data tests and so is ignoring the unit tests we added in https://github.com/ccao-data/data-architecture/pull/638. This PR also tweaks that logic so that we are properly linting unit tests.

Sample output including both types of unsorted tests:

```
$ python dbt/scripts/check_sort_dbt_yaml_files.py dbt/models/default/schema/default.vw_pin_universe.yml
In file: dbt/models/default/schema/default.vw_pin_universe.yml
Top level: models
data_tests in this group are not sorted:
---> default_vw_pin_universe_unique_by_14_digit_pin_and_year
---> default_vw_pin_universe_row_count
----------------------------------------
In file: dbt/models/default/schema/default.vw_pin_universe.yml
unit_tests in this group are not sorted:
---> default_vw_pin_universe_filters_invalid_pins
---> default_vw_pin_universe_class_strips_non_alphanumerics
----------------------------------------

The following files have unsorted tests:
dbt/models/default/schema/default.vw_pin_universe.yml (2)
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,4 +38,4 @@ repos:
         language: system
         types_or: [yaml, markdown]
         files: ^dbt/
-        exclude: venv
+        exclude: venv/
--- a/dbt/scripts/check_sort_dbt_yaml_files.py
+++ b/dbt/scripts/check_sort_dbt_yaml_files.py
@@ -147,9 +147,10 @@ def check_columns(file_path):
     return unsorted_files_dict, []
 
 
-def check_data_tests(file_path):
+def check_tests(file_path):
     """
-    Check if the 'data_tests' sections in a YAML file are sorted.
+    Check if the 'data_tests' and 'unit_tests' sections in a YAML file are
+    sorted.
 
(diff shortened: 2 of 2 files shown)


dfsnow (approved):
Thanks for the cleanup @jeancochrane 🙏 Sorry for the early merge!

PR #651, opened 2024-11-19
Update `build-and-test-dbt` workflow to rebuild children of modified assets in main

jeancochrane:
This PR adjusts the `Build models` stage in the `build-and-test-dbt` workflow to rebuild all downstream models in the DAG whenever a model or seed gets modified on the main branch. This is intended to help solve edge cases like the one revealed in [this workflow](https://github.com/ccao-data/data-architecture/actions/runs/11919381878/job/33218963393#step:7:2289) where a modified asset undergoes a change to a column's data type, causing its downstream models to become stale if dbt does not also rebuild them.

Per the comment in the code, we only do this on the main branch since otherwise we could end up rebuilding compute-intensive models on every single commit to a PR if that PR modifies a commonly-referenced ancestor. We may want to re-evaluate this decision in the future if child staleness becomes a problem for PR developers, but in the meantime it seems like a reasonable way to start.

I tested this change by swapping `prod` for `ci` in the `$STATE` variable check, then making a trivial modification to `model.vw_pin_shared_input` to make sure its two downstream models `model.vw_pin_condo_input` and `model.vw_card_res_input` got rebuilt too. See the [workflow logs](https://github.com/ccao-data/data-architecture/actions/runs/11923197596/job/33231056840#step:8:57) for evidence. Note that I canceled the job before it completed since I only wanted to see whether the state selection worked, I don't have any reason to believe this change would affect the model builds themselves.

### DIFF
--- a/.github/workflows/build_and_test_dbt.yaml
+++ b/.github/workflows/build_and_test_dbt.yaml
@@ -68,7 +68,7 @@ jobs:
             if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
               echo "Running build on manually selected resources"
               dbt build -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR" --resource-types model seed
-            elif [[ $TARGET == 'ci' ]]; then
+            elif [[ $TARGET == 'prod' ]]; then
               # Rebuild children of modified resources in prod, since schema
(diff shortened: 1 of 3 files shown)


dfsnow (approved):
All set @jeancochrane. Let's see how this goes.

Issue #653, opened 2024-11-22
Refactor `res_report_summary` to fix failing daily dbt model builds


jeancochrane:
Our `build-daily-dbt-models` workflow is [failing every day](https://github.com/ccao-data/data-architecture/actions/runs/11972489764/job/33379507755) because `reporting.res_report_summary` keeps running out of resources:

```
Query exhausted resources at this scale factor. You may need to manually clean the data at location 's3://ccao-athena-results-us-east-1/tables/ec8046e9-8e2c-4849-81b8-f81a22bc5f6c' before retrying. Athena will not delete data in your account.
```

Let's refactor the view to reduce the resources that this job uses.

PR #641, opened 2024-11-13
Add provisional assessment stages to PIN value views

jeancochrane:
## Background

This PR updates the `default.vw_pin_value` and `reporting.vw_pin_value_long` views using our new knowledge of [null procnames](https://github.com/ccao-data/data-architecture/issues/640) to add two new assessment stages:

* `PRE-MAILED`: Provisional values that will be mailed once first-pass desk review completes
* `ASSESSOR PRE-CERTIFIED`: Provisional values that will be set for second-pass once appeals complete

Closes https://github.com/ccao-data/data-architecture/issues/640.

## Open questions

* It's important that these values not be published in any public-facing assets, since they are not final and so are subject to change. **Are there other models that we need to change to make sure we don't accidentally publish these values?** For example, `reporting.vw_assessment_roll_muni` filters by `stage_name` such that these new stages will not be included, but `reporting.vw_assessment_roll` does not. Do we need to update the latter to filter out provisional stages?
* As far as I know, we can only determine provisional values for open towns in the current assessment year; this is because our queries for provisional values rely on the `cur = 'Y'` condition, which automatically becomes false once the record gets stamped with a procname. (This doesn't stop us from determining the values for past assessment stages, however, because we can use the `valclass IS NULL` condition in those cases to find the final stamped record for each stage.) **Is it acceptable that provisional values are only available for currently open towns**?

## Testing

I added a few unit tests in this PR to cover obvious problems with the new assessment stages, but I also ran some one-off QC queries to make sure that this PR doesn't accidentally change any non-provisional values that are already in prod. Expand each of the sections below to see the checks I ran in detail.

<details>
<summary>Make sure <b>total counts of records</b> with non-provisional assessment stages match between prod and dev</summary>

Check `default.vw_pin_value`:

```sql
with dev_count as (
    select 'dev' as source, count(*) as count
    from default.vw_pin_value
    where stage_name in ('MAILED', 'ASSESSOR CERTIFIED', 'BOARD CERTIFIED')
),

prod_count as (
    select 'prod' as source, count(*) as count
    from default.vw_pin_value
)

select * from dev_count
union
select * from prod_count
```

Check `reporting.vw_pin_value_long`:

```sql
with dev_count as (
    select 'dev' as source, count(*) as count
    from z_dev_jecochr_reporting.vw_pin_value_long
    where stage_name in ('MAILED', 'ASSESSOR CERTIFIED', 'BOARD CERTIFIED')
),

prod_count as (
    select 'prod' as source, count(*) as count
[...]

PR #654, opened 2024-11-22
Fix `reporting.res_report_summary` builds by turning input CTEs to tables

jeancochrane:
## Background

This PR fixes our broken `build-daily-dbt-models` workflow by improving build performance for the `reporting.res_report_summary` table, which has been exhausting Athena query resources ever since we merged https://github.com/ccao-data/data-architecture/pull/641.

The approach in this PR is to take the CTEs that the query reuses for aggregation and split them out into input tables that we materialize on the same schedule as `reporting.res_report_summary`. Tables improve performance over CTEs in this case because [Athena treats CTEs the same as subqueries and views](https://docs.aws.amazon.com/athena/latest/ug/select.html#select-parameters), that is to say, it will rerun the underlying query code every time the CTE is referenced.

Closes https://github.com/ccao-data/data-architecture/issues/653.

## Tradeoffs

While materializing the input data to the model speeds it up and reduces its memory usage, it also requires that we remember to refresh the inputs any time we want to refresh the model. This should not be a problem for the daily build task, since we can just tag the inputs as `daily` and dbt will understand the correct order to run the models, but it might be a gotcha if someone is ever asked to refresh this table on command.

One idea to mitigate this downside might be to add a when-loaded column to `res_report_summary` and its inputs, then add a test to the model confirming that the when-loaded column for its inputs are within some small bound (say, 15 minutes). This could provide a layer of protection against the models getting out of sync since the `build-and-test-dbt` workflow that we use to rebuild production models also runs tests for any models it builds. If you think this is a good idea, I'll go ahead and set it up before merging.

## Alternatives I considered

Other alternatives I considered was to improve the memory use of the query itself without factoring out any extra tables. The most memory-intensive part of the query is the [repeated unions that aggregate from the source data at varying levels of granularity](https://github.com/ccao-data/data-architecture/blob/1ae727ebdf592e8887fd1e8509650c64067ecb47/dbt/models/reporting/reporting.res_report_summary.sql#L136-L180), and I spent a while brainstorming ways of doing this without having to aggregate different groups and union them, but I couldn't think of anything better.

## Testing

See [this workflow run](https://github.com/ccao-data/data-architecture/actions/runs/11979289781/job/33401150667) for evidence that `build-daily-dbt-models` now runs successfully.

I built the version of the table and its ancestors that existed prior to https://github.com/ccao-data/data-architecture/pull/641 in the `z_dev_jecochr_reporting` database to test that the contents of the view don't change following this refactor.

I checked to make sure row counts match in the two assets:

```sql
select
    assessment_stage,
    prod_count,
    dev_count
from (
    select
        assessment_stage,
        count(*) as prod_count
    from z_dev_jecochr_reporting.res_report_summary
    group by assessment_stage
) prod
left join (
    select
        assessment_stage,
        count(*) as dev_count
    from z_ci_jeancochrane_653_refactor_res_report_summary_to_fix_failing_daily_dbt_model_builds_reporting.res_report_summary
    where assessment_stage in ('ASSESSOR CERTIFIED', 'MAILED', 'BOR CERTIFIED', 'model')
    group by assessment_stage
) dev
using (assessment_stage)
```
| assessment_stage | prod_count | dev_count |
| ---------------- | ---------- | --------- |
| MAILED | 12306 | 12306 |
| BOR CERTIFIED | 9329 | 9329 |
| model | 8920 | 8920 |
| ASSESSOR CERTIFIED | 11702 | 11702 |
| ASSESSOR PRE-CERTIFIED | 0 | 260 |

Confirming that PIN and sale counts are the same across the two versions of the table:

```sql
select
    triad,
    geography_type,
    geography_id,
    property_group,
    assessment_stage,
    year,
    dev.pin_n as dev_pin_n,
    prod.pin_n as prod_pin_n,
    abs(coalesce(dev.pin_n, 0) - coalesce(prod.pin_n, 0)) as pin_n_mismatch,
    dev.sale_n as dev_sale_n,
    prod.sale_n as prod_sale_n,
    abs(coalesce(dev.sale_n, 0) - coalesce(prod.sale_n, 0)) as sale_n_mismatch
from z_ci_jeancochrane_653_refactor_res_report_summary_to_fix_failing_daily_dbt_model_builds_reporting.res_report_summary dev
full outer join z_dev_jecochr_reporting.res_report_summary prod
using ( 
    triad,
    geography_type,
    geography_id,
    property_group,
    assessment_stage,
    year
)
where assessment_stage not in ('ASSESSOR PRE-CERTIFIED')
and (
  coalesce(dev.pin_n, 0) != coalesce(prod.pin_n, 0) or
  coalesce(dev.sale_n, 0) != coalesce(prod.sale_n, 0)
)
```

| triad | geography_type | geography_id | property_group | assessment_stage | year | dev_pin_n | prod_pin_n | pin_n_mismatch | dev_sale_n | prod_sale_n | sale_n_mismatch |
| ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- | ----- |

### DIFF
--- a/dbt/models/reporting/docs.md
+++ b/dbt/models/reporting/docs.md
@@ -32,12 +32,52 @@ Feeds public reporting assets.
 # res_report_summary
 
 {% docs table_res_report_summary %}
-Materialized version of `reporting.vw_res_report_summary`.
+Aggregates statistics on characteristics, classes, AVs, and sales
+by assessment stage, property groups, year, and various geographies.
+Feeds public reporting assets.
 
-Materialized to speed up queries for Tableau.
+Materialized once per day to speed up queries for Tableau.
+
+## Nuance
+
+- Model and assessment values are gathered independently and
+  aggregated via a union rather than a join, so it's important to keep in mind
+  that years for model and assessment stages do NOT need to match, i.e. we can
+  have 2023 model values in the table before there are any 2023 assessment
+  values to report on. Sales are added via a lagged join, so `sale_year` should
+  always be `year - 1`. It is also worth nothing that "model year" is
+  incremented by 1 solely for the sake of reporting in this table, meaning that
+  models with a `meta_year` value of 2022 in `model.assessment_pin` will
+  populate the table with a value of 2023 for `year`.
 
 **Primary Key**: `year`, `geography_type`, `geography_id`, `assessment_stage`,
 `property_group`
+{% enddocs %}
+
+# res_report_summary_sales_input
+
+{% docs table_res_report_summary_sales_input %}
+Input table for `reporting.res_report_summary` that produces the raw
+sales data that `res_report_summary` aggregates.
+
+We split these input data out into a separate table to speed up query time for
+`res_report_summary`, since otherwise it needs to rerun the query logic
+for every possible geography and reporting group combination.
+
+**Primary Key**: `pin`, `doc_no`
+{% enddocs %}
+
+# res_report_summary_values_input
+
+{% docs table_res_report_summary_values_input %}
+Input table for `reporting.res_report_summary` that produces the raw
+characteristic and value data that `res_report_summary` aggregates.
+
+We split these input data out into a separate table to speed up query time for
+`res_report_summary`, since otherwise it needs to rerun the query logic
+for every possible geography and reporting group combination.
+
+**Primary Key**: `pin`, `year`
 {% enddocs %}
 
 # vw_assessment_roll
@@ -118,17 +158,6 @@ The assessment stages are:
   levels of assessment, building splits, etc.
 
 **Primary Key**: `year`, `pin`, `stage_name`
-{% enddocs %}
-
-# vw_res_report_summary
-
-{% docs view_vw_res_report_summary %}
-Aggregates statistics on characteristics, classes, AVs, and sales
-by assessment stage, property groups, year, and various geographies.
-Feeds public reporting assets.
-
-**Primary Key**: `year`, `geography_type`, `geography_id`, `assessment_stage`,
-`property_group`
 {% enddocs %}
 
 # vw_top_5
--- a/dbt/models/reporting/docs.md
+++ b/dbt/models/reporting/docs.md
@@ -60,7 +60,7 @@ Materialized once per day to speed up queries for Tableau.
 Input table for `reporting.res_report_summary` that produces the raw
 sales data that `res_report_summary` aggregates.
 
-We split these input data out into a separate table to speed up query time for
+We split these input data out into a separate table to reduce resource use in
 `res_report_summary`, since otherwise it needs to rerun the query logic
 for every possible geography and reporting group combination.
 
@@ -73,7 +73,7 @@ for every possible geography and reporting group combination.
 Input table for `reporting.res_report_summary` that produces the raw
 characteristic and value data that `res_report_summary` aggregates.
 
-We split these input data out into a separate table to speed up query time for
+We split these input data out into a separate table to reduce resource use in
 `res_report_summary`, since otherwise it needs to rerun the query logic
 for every possible geography and reporting group combination.
 
--- a/dbt/models/reporting/reporting.res_report_summary.sql
+++ b/dbt/models/reporting/reporting.res_report_summary.sql
@@ -27,111 +27,14 @@ Intended to be materialized daily through a GitHub action.
     )
 }}
 
--- AVs and model values
-WITH all_fmvs AS (
-    SELECT
-        ap.meta_pin AS pin,
-        ap.year,
-        'model' AS assessment_stage,
-        ap.pred_pin_final_fmv_round AS total
-    FROM {{ source('model', 'assessment_pin') }} AS ap
-    INNER JOIN {{ ref('model.final_model') }} AS fm
-        ON ap.run_id = fm.run_id
-        AND ap.year = fm.year
-        AND (
-            -- If reassessment year, use different models for different towns
-            (
-                CONTAINS(fm.township_code_coverage, ap.township_code)
-                AND ap.meta_triad_code = fm.triad_code
-            )
-            -- Otherwise, just use whichever model is "final"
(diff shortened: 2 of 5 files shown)


dbt/models/reporting/docs.md
```diff
@@ -32,14 +32,54 @@ Feeds public reporting assets.
 # res_report_summary
 
 {% docs table_res_report_summary %}
-Materialized version of `reporting.vw_res_report_summary`.
+Aggregates statistics on characteristics, classes, AVs, and sales
+by assessment stage, property groups, year, and various geographies.
+Feeds public reporting assets.
```
jeancochrane, line 37: Previously, this text was included in an unused `view_res_report_summary` docs block that I'm guessing we forgot to remove when we switched to materializing the table. I removed that docs block and consolidated its text over here.

dbt/models/reporting/docs.md
```diff
@@ -32,14 +32,54 @@ Feeds public reporting assets.
...
+
+- Model and assessment values are gathered independently and
+  aggregated via a union rather than a join, so it's important to keep in mind
+  that years for model and assessment stages do NOT need to match, i.e. we can
+  have 2023 model values in the table before there are any 2023 assessment
+  values to report on. Sales are added via a lagged join, so `sale_year` should
+  always be `year - 1`. It is also worth nothing that "model year" is
+  incremented by 1 solely for the sake of reporting in this table, meaning that
+  models with a `meta_year` value of 2022 in `model.assessment_pin` will
+  populate the table with a value of 2023 for `year`.
```
jeancochrane, line 51: This nuance comes from the model docstring. I think it's useful, so I copied it into the dbt docs as well.

dbt/models/reporting/docs.md
```diff
@@ -32,14 +32,54 @@ Feeds public reporting assets.
...
+
+{% docs table_res_report_summary_values_input %}
+Input table for `reporting.res_report_summary` that produces the raw
+characteristic and value data that `res_report_summary` aggregates.
+
+We split these input data out into a separate table to speed up query time for
+`res_report_summary`, since otherwise it needs to rerun the query logic
+for every possible geography and reporting group combination.
+
+**Primary Key**: `pin`, `year`
```
jeancochrane, line 80: Note that, due to the problem I shared in our chat, this view isn't actually _truly_ unique by `pin` and `year` since there's one PIN in the 2023 condo model that has two rows in `model.assessment_pin`. I think that's a minor detail and I doubt anyone will need to rely on the uniqueness of this table, however, so I left that context out of the docs -- happy to put it back in if you think it's appropriate.

dbt/models/reporting/reporting.res_report_summary.sql
```diff
@@ -157,25 +60,25 @@ aggregated_values AS (
 ),
 
 -- Aggregate and stack stats on sales for each reporting group
-all_sales AS (
+aggregated_sales AS (
```
jeancochrane, line 63: No logic has changed in this CTE; I've just renamed it to more closely match the naming pattern used by `aggregated_values` and its ancestors.

dbt/models/reporting/reporting.res_report_summary_sales_input.sql
```diff
@@ -0,0 +1,39 @@
...
+LEFT JOIN {{ ref('reporting.vw_pin_township_class') }} AS tc
+    ON vwps.pin = tc.pin
+    AND vwps.year = tc.year
+WHERE NOT vwps.is_multisale
+    AND NOT vwps.sale_filter_is_outlier
+    AND NOT vwps.sale_filter_deed_type
+    AND NOT vwps.sale_filter_less_than_10k
+    AND NOT vwps.sale_filter_same_sale_within_365
+    AND tc.property_group IS NOT NULL
+    AND tc.triad_name IS NOT NULL
```
jeancochrane, line 39: The logic in this model should be identical to this block from the old `res_report_summary` query, with the only difference being that we add `pin` and `doc_no` to the list of selected columns so that the table has a distinct primary key:

https://github.com/ccao-data/data-architecture/blob/1ae727ebdf592e8887fd1e8509650c64067ecb47/dbt/models/reporting/reporting.res_report_summary.sql#L113-L132

dbt/models/reporting/reporting.res_report_summary_values_input.sql
```diff
@@ -0,0 +1,106 @@
...
+    ON fmvs.pin = vptc.pin
+    AND fmvs.year = vptc.year
+INNER JOIN chars
+    ON fmvs.pin = chars.pin
+    AND fmvs.year = chars.year
+LEFT JOIN {{ ref('default.vw_pin_land') }} AS vpl
+    ON fmvs.pin = vpl.pin
+    AND fmvs.year = vpl.year
+WHERE vptc.property_group IS NOT NULL
+    AND vptc.triad_name IS NOT NULL
```
jeancochrane, line 106: The logic in this model should be identical to this block from the old `res_report_summary`, with the only change being that we move `year` to the bottom of the list of selected columns in order to enable its use as a partition key:

https://github.com/ccao-data/data-architecture/blob/1ae727ebdf592e8887fd1e8509650c64067ecb47/dbt/models/reporting/reporting.res_report_summary.sql#L30-L111

dbt/models/reporting/docs.md
dfsnow: That's fine. I'll debug the dupe PIN. I'm going to add a follow-up issue for data sets on the `model` DB assets.

dfsnow (approved):
I don't love our recent pattern of separating things into `_input` tables to get around Athena resource constraints. It's just more stuff to maintain, more opportunities for things to get out-of-sync, etc. I'm positive that with some clever querying we could probably make the original view work.

That said, I think this is a perfectly fine stop-gap until we can either: 

- Find an optimized SQL query that works
- Refactor this into a Python model, which is what it probably deserves due to its complexity

@jeancochrane Let's merge this ASAP to unblock our daily builds.

PR #656, opened 2024-11-26
Fix socrata api github action primary key handling

wrridgeway:
Well, this PR seems to address the issues that were blocking deployment of out Socrata SODA API github action from working on `default.vw_pin_sale` and its Open Data asset "Parcel Sales", but the outcome doesn't bode well for deploying the API script as an alternative for the socrata agent running on the server.

I can run the github action now [without any errors](https://github.com/ccao-data/data-architecture/actions/runs/12127852104/job/33813006378) _on small samples of data_. Anything large, we get our [connection reset](https://github.com/ccao-data/data-architecture/actions/runs/12127715772/job/33812565042) by Socrata's server. Perhaps I'm misunderstanding what's happening in those logs, but if I'm not, I'll have to reach out to them and see if there's any solution for this issue. In the meantime, these changes should get merged into master.

One more important thing to note: `default.vw_pin_sale` is no longer unique by pin and document number. We need to fix this.

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -84,12 +84,12 @@ def build_query(
     columns = cursor.execute("show columns from " + athena_asset).as_pandas()
 
     # Limit pull to columns present in open data asset
-    new_url = (
+    asset_url = (
         "https://datacatalog.cookcountyil.gov/resource/"
         + asset_id
         + ".json?$limit=1"
     )
-    asset_columns = requests.get(new_url).json()[0].keys()
+    asset_columns = requests.get(asset_url).json()[0].keys()
 
     columns = columns[columns["column"].isin(asset_columns)]
 
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -89,6 +89,7 @@ def build_query(
         + asset_id
         + ".json?$limit=1"
     )
+    print(requests.get(asset_url).json())
     asset_columns = requests.get(asset_url).json()[0].keys()
     columns = columns[columns["column"].isin(asset_columns)]
 
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -114,13 +114,10 @@ def build_query(
         query = query
 
     elif years is not None and not township:
-        query += " WHERE year = %(year)s LIMIT 10000"
+        query += " WHERE year = %(year)s"
 
     elif years is not None and township is not None:
-        query += (
-            " WHERE year = %(year)s"
-            + " AND township_code = %(township)s LIMIT 10000"
-        )
+        query += " WHERE year = %(year)s" + " AND township_code = %(township)s"
 
     return query
 
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -89,8 +89,14 @@ def build_query(
         + asset_id
         + ".json?$limit=1"
     )
-    print(requests.get(asset_url).json())
-    asset_columns = requests.get(asset_url).json()[0].keys()
+
+    asset_columns = (
+        requests.get(
+            asset_url, headers={"X-App-Token": os.getenv("SOCRATA_APP_TOKEN")}
+        )
+        .json()[0]
+        .keys()
+    )
     columns = columns[columns["column"].isin(asset_columns)]
 
     # Array type columns are not compatible with the json format needed for
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -63,7 +63,9 @@ def get_asset_info(socrata_asset):
     return athena_asset, asset_id, row_identifier
 
 
-def build_query(athena_asset, row_identifier, years=None, township=None):
+def build_query(
+    athena_asset, asset_id, row_identifier, years=None, township=None
+):
     """
     Build an Athena compatible SQL query. Function will append a year
     conditional if `years` is non-empty. Many of the CCAO's open data assets are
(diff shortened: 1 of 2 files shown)


socrata/socrata_upload.py
```diff
@@ -72,11 +75,30 @@ def build_query(athena_asset, row_identifier, years=None, township=None):
     passed to `row_identifiers`.
     """
 
-    row_identifier = f"CONCAT({', '.join(row_identifier)}) AS row_id"
+    if len(row_identifier) > 1:
```
wrridgeway, line 78: Needed to address primary keys that are singular

socrata/socrata_upload.py
```diff
@@ -72,11 +75,30 @@ def build_query(athena_asset, row_identifier, years=None, township=None):
...
+        row_identifier = f"CONCAT(CAST({' AS varchar), CAST('.join(row_identifier)} AS varchar)) AS row_id"
+    else:
+        row_identifier = f"CAST({row_identifier[0]} AS varchar) AS row_id"
 
     # Retrieve column names and types from Athena
     columns = cursor.execute("show columns from " + athena_asset).as_pandas()
 
+    # Limit pull to columns present in open data asset
```
wrridgeway, line 86: Can't pass the API columns it isn't looking for.

socrata/socrata_upload.py
```diff
@@ -72,11 +75,30 @@ def build_query(athena_asset, row_identifier, years=None, township=None):
...
+
+    asset_columns = (
+        requests.get(
+            asset_url, headers={"X-App-Token": os.getenv("SOCRATA_APP_TOKEN")}
+        )
+        .json()[0]
+        .keys()
+    )
+    columns = columns[columns["column"].isin(asset_columns)]
+
```
wrridgeway, line 101: Can't pass the API columns it isn't looking for.

socrata/socrata_upload.py
```diff
@@ -139,31 +161,34 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
...
 
+    # We grab the data before uploading it so we can make sure timestamps are
+    # properly formatted
+    input_data = cursor.execute(sql_query, query_conditionals).as_pandas()
+    date_columns = input_data.select_dtypes(include="datetime").columns
+    input_data[date_columns] = input_data[date_columns].map(
+        lambda x: x.strftime("%Y-%m-%dT%X")
+    )
+    input_data = input_data.to_json(orient="records")
+
```
wrridgeway, line 172: Socrata expects are very particularly formatted timestamp.

socrata/socrata_upload.py
jeancochrane: **[Thought, non-blocking]** This works I suppose, but it feels brittle that we need to send a web request to the asset to figure out what columns we want to push to it. It seems to me that our code should be the authority about what goes to the asset rather than the other way around. Plus, if this request fails, the whole script will fail; we could handle this by adding retry logic, but then we're potentially building in a bunch of complexity just so we can get information we should already know about the asset.

I wonder if there's a deeper, structural fix we could implement for restricting the columns that we send to open data? Some ideas:

* Add an optional list of columns to include under a key like `meta.columns` in the exposure configuration
* Create new views under a dedicated schema like `open_data` that build on the `default` views but restrict to only the data that we want to upload to open data

I'm happy to merge this in as-is to get things working, especially since it's not totally clear to us yet whether this is going to work at all, but in the long term I bet there's a more robust foundation we can build on.

socrata/socrata_upload.py
jeancochrane: **[Question, non-blocking]** How does the Socrata agent handle this problem? Does it perform this conversion for us?

Continuing my thought above, this is another kind of thing that might be best solved by having dedicated views for our open data assets that exist on top of `default` views.

socrata/socrata_upload.py
```diff
@@ -304,6 +332,7 @@ def socrata_upload(
     print(f"Total upload in {toc - tic:0.4f} seconds")
 
 
+# %%
```
jeancochrane, line 335: **[Nitpick, non-blocking]** Stray comment here?

socrata/socrata_upload.py
```diff
@@ -72,11 +75,30 @@ def build_query(athena_asset, row_identifier, years=None, township=None):
     passed to `row_identifiers`.
     """
 
-    row_identifier = f"CONCAT({', '.join(row_identifier)}) AS row_id"
+    if len(row_identifier) > 1:
+        row_identifier = f"CONCAT(CAST({' AS varchar), CAST('.join(row_identifier)} AS varchar)) AS row_id"
+    else:
+        row_identifier = f"CAST({row_identifier[0]} AS varchar) AS row_id"
```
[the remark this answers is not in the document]
jeancochrane: **[Suggestion, non-blocking]** It took me a couple scans to parse this logic, here's something I think might be slightly cleaner (weak preference):

```python
row_identifier_sql_parts = [f"CAST({col} AS varchar)" for col in row_identifier]
row_identifier_sql_joined = (
    row_identifier_sql_parts[0]
    if len(row_identifier_sql_parts) == 1
    else f"CONCAT({', '.join(row_identifier_sql_parts)})"
)
row_identifier = f"{row_identifier_sql_joined} AS row_id"
```

jeancochrane (approved):
I'm a bit worried about the long-term maintainability of some of the code here, but I also acknowledge that right now we're just trying to debug things to make sure we can use this approach at all. Happy to punt my bigger-picture worries till after we've got things up and running and confirmed we want to use the API in the first place.

You mention in the PR body that we're restricted in the number of rows we can upload at one time, but it doesn't seem like this PR implements any logic around that. How small do the batches have to be? And does it seem like we'll be able to adjust this script to automatically perform batching?

jeancochrane:
One more thought -- my understanding is that this comment is out of date, and `default.vw_pin_sale` should be the same as it's always been while `default.vw_pin_sale_combined` should have both mydec and ias sales, correct?

> One more important thing to note: `default.vw_pin_sale` is no longer unique by pin and document number. We need to fix this.

socrata/socrata_upload.py
wrridgeway: I've been thinking about this as well, and there's going to be another problem that pops up very similar to the one you're outlining - we have changed API field names for some assets, and the API is going to crap out because of that. Either we need to make sure every api field name is that same as the column name in the athena view that feeds the asset, or we need to create a crosswalk table similar to what you're outlining above.

My current thought it this - creating these athena tables is probably the more robust approach, and I've considered it, but I'm also deeply unsure that this API solution is actually going to end up being the solution we were hoping it might be given the issues I'm having with being kicked out by socrata's server due to time or memory. If I can resolve those issues and get to the point where I'm able to update an entire year of just the sales asset, it'll be worth sketching out a better solution to the column and api field name issue.

wrridgeway:
Regarding the non-uniqueness of `default.vw_pin_sale`, that doesn't have anything to do with the new combined view, it's a problem that seems to have arisen at some point in the recent past within the sales view.

PR #657, opened 2024-11-27
Update docs to reflect new card code 5C

jeancochrane:
This PR updates our docs for the `chgrsn`/`mktrsn` fields to reflect the new 5C value, which indicates a one-year Board reduction that is _not_ implemented via occupancy.

Note that I also double-checked our tests and QC reports to confirm that nothing else needs to change based on this new code.

### DIFF
--- a/dbt/models/iasworld/columns.md
+++ b/dbt/models/iasworld/columns.md
@@ -228,8 +228,15 @@ Reason code for change of value.
 Possible values for this variable are:
 
 - `4` = Regular reduction (lasts until next triennial reassessment)
-- `5` = One year only market value relief (occupancy/vacancy reduction)
-- `5B` = One year only market value relief, issued by the Board of Review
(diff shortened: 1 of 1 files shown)


dbt/models/iasworld/columns.md
```diff
@@ -228,8 +228,15 @@ Reason code for change of value.
 Possible values for this variable are:
 
 - `4` = Regular reduction (lasts until next triennial reassessment)
-- `5` = One year only market value relief (occupancy/vacancy reduction)
-- `5B` = One year only market value relief, issued by the Board of Review
+- `4B` = Regular reduction issued by the Board of Review (lasts until next
+  triennial reassessment)
```
jeancochrane, line 232: It seemed like we were previously missing this code, so I added it.

dfsnow (approved):
LGTM 👍

PR #660, opened 2024-12-03
Test urllib fix

wrridgeway:
This PR makes sure all column names are retrieved from an open data asset and that if any chunk sent to socrata is larger than 10,000 rows it gets sub-chunked into chunks with less than 10k rows.

It also uses a [session object from the requests library](https://requests.readthedocs.io/en/latest/user/advanced/) in order to pool connections. We were getting 500 errors in the past, possibly because too many requests were being sent, and switching to connection pooling resolved that.

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -187,41 +187,21 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
         for i in range(0, input_data.shape[0], 10000):
             print(print_message)
             print(f"Rows {i + 1}-{i + 10000}")
-            if method == "put":
-                response = session.put(
-                    url=url,
-                    data=input_data.iloc[i : i + 10000].to_json(
-                        orient="records"
-                    ),
-                    headers={"X-App-Token": app_token},
-                )
-
-            elif method == "post":
-                response = session.post(
-                    url=url,
-                    data=input_data.iloc[i : i + 10000].to_json(
-                        orient="records"
-                    ),
-                    headers={"X-App-Token": app_token},
-                )
+            response = getattr(session, method)(
+                url=url,
+                data=input_data.iloc[i : i + 10000].to_json(orient="records"),
+                headers={"X-App-Token": app_token},
+            )
 
             print(response.content)
 
     else:
         print(print_message)
-        if method == "put":
-            response = session.put(
-                url=url,
-                data=input_data.to_json(orient="records"),
-                headers={"X-App-Token": app_token},
-            )
-
-        elif method == "post":
-            response = session.post(
-                url=url,
-                data=input_data.to_json(orient="records"),
-                headers={"X-App-Token": app_token},
-            )
+        response = getattr(session, method)(
+            url=url,
+            data=input_data.to_json(orient="records"),
+            headers={"X-App-Token": app_token},
+        )
 
         print(response.content)
 
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -95,19 +95,16 @@ def build_query(
     # Retrieve column names and types from Athena
     columns = cursor.execute("show columns from " + athena_asset).as_pandas()
 
-    # Limit pull to columns present in open data asset
-    asset_url = (
-        "https://datacatalog.cookcountyil.gov/resource/"
-        + asset_id
-        + ".json?$limit=1"
-    )
-
+    # Limit pull to columns present in open data asset - shouldn't change anything, but prevents failure if columns have become misaligned.
     asset_columns = (
-        s.get(
-            asset_url, headers={"X-App-Token": os.getenv("SOCRATA_APP_TOKEN")}
+        requests.get(
+            f"https://datacatalog.cookcountyil.gov/resource/{asset_id}"
         )
-        .json()[0]
-        .keys()
+        .headers["X-SODA2-Fields"]
+        .replace('"', "")
+        .strip("[")
+        .strip("]")
+        .split(",")
     )
     columns = columns[columns["column"].isin(asset_columns)]
 
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -2,7 +2,6 @@ import contextlib
 import io
 import json
 import os
-import socket
 import time
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


socrata/socrata_upload.py
```diff
@@ -87,19 +95,16 @@ def build_query(
...
-            asset_url, headers={"X-App-Token": os.getenv("SOCRATA_APP_TOKEN")}
+            f"https://datacatalog.cookcountyil.gov/resource/{asset_id}"
         )
-        .json()[0]
-        .keys()
+        .headers["X-SODA2-Fields"]
+        .replace('"', "")
+        .strip("[")
+        .strip("]")
+        .split(",")
```
wrridgeway, line 107: I was very naively pulling asset column names before. If the first row of an asset had `NULL` values for a column, we wouldn't get back that column header. This ensures we get all column headers.

socrata/socrata_upload.py
```diff
@@ -10,6 +10,14 @@
 from pyathena import connect
 from pyathena.pandas.cursor import PandasCursor
 
+# Create a session object so HTTP requests can be pooled
+s = requests.Session()
+s.verify = True
```
jeancochrane, line 15: **[Nitpick, non-blocking]** I think this should be [`True` by default](https://docs.python-requests.org/en/latest/api/#requests.Session.verify), no?

socrata/socrata_upload.py
```diff
@@ -10,6 +10,14 @@
...
+s.auth = (
+    str(os.getenv("SOCRATA_USERNAME")),
+    str(os.getenv("SOCRATA_PASSWORD")),
+)
```
jeancochrane, line 19: **[Question, non-blocking]** I see that we were doing this before, so I get that this auth practice is already on the main branch, but now that my memory is refreshed it's making me wonder -- is there a way to auth to the API using only a token instead of also needing to set our account username/password? Without understanding the details, it strikes me as a super weird pattern (the token should be able to perform auth on its own), and I'm a bit nervous reading and sending username/password given that a leaked username/password could cause more damage than a leaked token.

socrata/socrata_upload.py
```diff
@@ -10,6 +10,14 @@
 from pyathena import connect
 from pyathena.pandas.cursor import PandasCursor
 
+# Create a session object so HTTP requests can be pooled
+s = requests.Session()
```
jeancochrane, line 14: **[Nitpick, non-blocking]** I dislike single-character variable names in Python because they often conflict with debugger commands (in this case [`s` for `s(tep)`](https://docs.python.org/3/library/pdb.html#pdbcommand-step)). Plus, a word is typically more descriptive than a single letter anyway:

```suggestion
session = requests.Session()
```

socrata/socrata_upload.py
```diff
@@ -172,31 +176,56 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
...
-        url=url,
-        data=input_data,
-        auth=auth,
+    s.get(
+        (
+            "https://datacatalog.cookcountyil.gov/resource/"
+            + asset_id
+            + ".json?$limit=1"
+        ),
+        headers={"X-App-Token": os.getenv("SOCRATA_APP_TOKEN")},
```
jeancochrane, line 187: **[Nitpick, non-blocking]** Didn't we already read this env var into a local variable on line 141?

```suggestion
        headers={"X-App-Token": app_token},
```

jeancochrane, line 187: **[Question, non-blocking]** Also, why do we use the `headers` kwarg here, while we template the token directly into the `url` string on line 147? I'm inclined to prefer `headers` in both cases but I don't totally understand the tradeoff.

socrata/socrata_upload.py
```diff
@@ -172,31 +176,56 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
...
+                    ),
+                )
+
+            elif method == "post":
+                response = s.post(
+                    url=url,
+                    data=input_data.iloc[i : i + 10000].to_json(
+                        orient="records"
+                    ),
+                )
```
jeancochrane, line 210: **[Nitpick, non-blocking]** Slightly simpler:

```suggestion
               response = getattr(s, method)(
                    url=url,
                    data=input_data.iloc[i : i + 10000].to_json(
                        orient="records"
                    ),
                )
```

socrata/socrata_upload.py
```diff
@@ -172,31 +176,56 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
...
 
-    elif method == "post":
-        response = requests.post(
-            url=url,
-            data=input_data,
-            auth=auth,
-        )
+            print(response.content)
```
jeancochrane, line 212: **[Question, non-blocking]** What sort of content do we expect to get back from the response? I just want to make absolutely sure that it can't contain anything sensitive.

socrata/socrata_upload.py
```diff
@@ -172,31 +176,56 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
...
+            response = s.put(
+                url=url,
+                data=input_data.to_json(orient="records"),
+            )
+
+        elif method == "post":
+            response = s.post(
+                url=url,
+                data=input_data.to_json(orient="records"),
+            )
```
jeancochrane, line 226: **[Nitpick, non-blocking]** Do we need this `if`/`else` conditional at all? Or could we just chunk all input using the `for i in range(0, input_data.shape[0], 10000)` loop, and when the input data is less than 10k rows, there will just be one iteration?

jeancochrane (approved):
Looking good! No huge concerns, but a bunch of small questions below.

socrata/socrata_upload.py
wrridgeway: It is. Removed.

socrata/socrata_upload.py
wrridgeway: Good call. I genuinely have no idea why I switched between methods, but i'll move everything to the kwargs method.

socrata/socrata_upload.py
wrridgeway: This may be up to something I don't understand about auth'ing for their API, but the token does not suffice on it's own. I will get an `authentication_required` error if I don't auth here.

socrata/socrata_upload.py
wrridgeway: This content is pretty harmless (I think):

`b'{\n  "Errors" : 0,\n  "Rows Created" : 0,\n  "Rows Updated" : 10000,\n  "Rows Deleted" : 0\n}\n'`

socrata/socrata_upload.py
wrridgeway: This is rad.

socrata/socrata_upload.py
wrridgeway: I think you're right we don't need it. Removed.

PR #397, opened 2024-04-18
Add `class_dict.csv` dbt seed

wrridgeway:
We've got some new classes this year that need to be incorporated into iasworld error reporting so as not to trigger false-positive bad-class detection.

PR #667, opened 2024-12-05
Pull class codes from Athena instead of `ccao` package in `land_nbhd_rate` ETL

jeancochrane:
We're removing the `class_dict` data variable from the `ccao` package in https://github.com/ccao-data/ccao/pull/34, so we have to change the reference to it in the ETL script for `land_nbhd_rate`. We switch the reference to point to the `ccao.class_dict` table that we added to our data lake in https://github.com/ccao-data/data-architecture/pull/397.

> [!WARNING]  
> For some reason the diff in this PR is reporting that the entire `land_nbhd_rate` script changed, even though I only adjusted one line. I can't figure out why this is the case (maybe some sort of change in file encoding) so I'm going to just comment on the lines that I changed.


dfsnow (commented):
This is undoubtedly a line-ending change. Can you use `unix2dos` or a similar tool to flip the line endings back so we get a true diff? Otherwise, this looks totally fine.

jeancochrane:
Good catch @dfsnow, looks like my editor converted the file from DOS to Unix file format automatically. 303be387d1d8359f5d21ed9360132069227f8943 fixes this issue and the diff now looks like what I expected before.

dfsnow (approved):
Sweet! Thanks for the line ending fix, looks good to me now.

PR #671, opened 2024-12-05
Split out `test-iasworld-data` workflow from `test-dbt-models` so that data integrity tests will fail properly

jeancochrane:
There is currently a bug in our weekly scheduled data integrity tests: Failing tests do not cause the workflow to fail, so we never get notified of them. ([See here for an example workflow run that should have failed](https://github.com/ccao-data/data-architecture/actions/runs/12118092672/job/33781887419#step:7:265).) This makes sense in that the `test-dbt-models` workflow was originally intended to be used exclusively to run iasWorld tests, so it doesn't fail the workflow if tests fail, because we expect some iasWorld tests to fail. This doesn't work for data integrity tests, however, since in the case the workflow _should_ fail if any test fails.

I thought about modifying the `test-dbt-models` workflow and its underlying `run_iasworld_data_tests.py` script to add a new flag to exit with a nonzero status if a test fails, but I realized that this would bloat `run_iasworld_data_tests.py` with a feature that it doesn't need. What's more, it's unreasonable to use the `run_iasworld_data_tests.py` script to run data integrity tests in the first place, since iasWorld tests and data integrity tests have been categorically different since https://github.com/ccao-data/data-architecture/pull/638. In my opinion, this is the root cause of the bug, and will continue causing problems until we disentangle the two testing workflows.

In order to fix the problem at the root, this PR splits out `test-dbt-models` into two workflows with different uses:

* `test-iasworld-data`, which runs iasWorld tests via workflow dispatch using the `run_iasworld_data_tests.py` script
* `test-dbt-models`, which runs any dbt tests via workflow dispatch or runs data integrity tests on a schedule using direct `dbt test` calls

As a result, we can configure `test-dbt-models` to fail if any tests fail, while `test-iasworld-data` can maintain its current behavior by _not_ failing when tests fail.

Examples of the two workflows running as expected:

* [`test-iasworld-data`](https://github.com/ccao-data/data-architecture/actions/runs/12189460666/job/34004643684)
* [`test-dbt-models`](https://github.com/ccao-data/data-architecture/actions/runs/12189451367/job/34004613229) (Note that this workflow failed, but that's what we expect)

> [!NOTE]
> Since this PR changes the purpose of the `test-dbt-models` workflow, we need to change the `service-spark-iasworld` to dispatch the new `test-iasworld-data` workflow instead. See this companion PR: https://github.com/ccao-data/service-spark-iasworld/pull/25

### DIFF
--- a/.github/workflows/test_dbt_models.yaml
+++ b/.github/workflows/test_dbt_models.yaml
@@ -104,8 +104,8 @@ jobs:
         run: echo "TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%S")" >> "$GITHUB_ENV"
         shell: bash
 
-      # Only triggered when run on a schedule. Otherwise, notifications go to
-      # whoever dispatched the workflow
+      # Only triggered when run on a schedule. Otherwise, whoever dispatched the
+      # workflow is notified via GitHub (instead of SNS)
       - name: Send failure notification
         if: github.event_name == 'schedule' && failure()
         uses: ./.github/actions/publish_sns_topic
--- a/.github/workflows/test_iasworld_data.yaml
+++ b/.github/workflows/test_iasworld_data.yaml
@@ -26,8 +26,6 @@ on:
         required: false
         default: false
         type: boolean
-  pull_request:
-    branches: [main, master]
 
 env:
   PYTHONUNBUFFERED: "1"
--- a/.github/workflows/test_iasworld_data.yaml
+++ b/.github/workflows/test_iasworld_data.yaml
@@ -1,26 +1,18 @@
-name: test-dbt-models
+name: test-iasworld-data
 
 # This workflow is not configured to run on PRs, because PRs have their own CI
-# process defined by the `build-and-test-dbt` workflow. It runs under two
-# circumstances:
+# process defined by the `build-and-test-dbt` workflow. Instead, the Data
+# Department's Spark data extraction process dispatches it via API once per
+# day after finishing the daily ingest of iasWorld data. For more detail, see:
 #
-#   1. When manually triggered by a Data Department team member, in order to
-#      run specific dbt tests using the `select` or `selector` input (runs all
-#      data integriy tests by default, if both `select` and `selector`
-#      are blank)
-#
-#   2. On the cron schedule set below, which runs data integrity tests
-#      weekly in order to proactively alert the team of any problems. Since
-#      the cron schedule cannot pass input variables to the workflow, we
-#      configure the input variables to fall back to defaults that support
-#      the scheduled job, namely the data integrity tests
+# https://github.com/ccao-data/service-spark-iasworld
 on:
   workflow_dispatch:
     inputs:
       select:
         description: >
           Optional space-separated list of tests to run (defaults to all
-          non-iasWorld data tests)
+          iasWorld data tests)
         required: false
         type: string
       selector:
@@ -29,16 +21,18 @@ on:
           over the above list of tests if both are present)
         required: false
         type: string
-  schedule:
-    # Every Monday at 11am UTC (6am UTC-5)
-    - cron: '0 11 * * 1'
+      upload_test_results:
+        description: Upload test results to S3
+        required: false
+        default: false
+        type: boolean
 
 env:
   PYTHONUNBUFFERED: "1"
   UV_SYSTEM_PYTHON: 1
 
 jobs:
-  test-dbt-models:
+  test-iasworld-data:
     runs-on: ubuntu-latest
     # These permissions are needed to interact with GitHub's OIDC Token endpoint
     # so that we can authenticate with AWS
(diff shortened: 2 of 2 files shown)


.github/workflows/test_iasworld_data.yaml
```diff
@@ -0,0 +1,141 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Validate and parse input variables
+        id: parse-inputs
+        run: |
+          # Default to no select option, which will fall back to the default
+          # behavior of the underlying Python script
+          SELECT_OPTION=""
```
jeancochrane, line 51: This block is different from the old implementation of `test-dbt-models`: Now that this workflow is dedicated to running iasWorld tests, we can fall back to the default behavior of the underlying `run_iasworld_data_tests.py` script, which means we can avoid passing in the `--select` or `--selector` options unless the caller specifies one of them.

.github/workflows/test_iasworld_data.yaml
```diff
@@ -0,0 +1,141 @@
+name: test-iasworld-data
+
+# This workflow is not configured to run on PRs, because PRs have their own CI
+# process defined by the `build-and-test-dbt` workflow. Instead, the Data
+# Department's Spark data extraction process dispatches it via API once per
+# day after finishing the daily ingest of iasWorld data. For more detail, see:
+#
+# https://github.com/ccao-data/service-spark-iasworld
```
jeancochrane, line 8: This comment is slightly different, to reflect the new conditions in which this workflow runs compared to the old `test-dbt-models` workflow.

.github/workflows/test_iasworld_data.yaml
```diff
@@ -0,0 +1,141 @@
...
+      # Only triggered when called by a bot. Otherwise, notifications
+      # go to whoever called the workflow
+      - name: Send failure notification
+        if: github.event_name == 'workflow_dispatch' && github.triggering_actor == 'sqoop-bot[bot]' && failure()
+        uses: ./.github/actions/publish_sns_topic
+        with:
+          sns_topic_arn: ${{ secrets.AWS_SNS_NOTIFICATION_TOPIC_ARN }}
+          subject: "iasWorld tests errored for workflow run: ${{ github.run_id }}"
+          body: |
+            iasWorld tests raised an error for workflow ${{ github.run_id }}, run on ${{ env.TIMESTAMP }} UTC
```
jeancochrane, line 138: This subject and body are now slightly different from the old implementation ("iasWorld tests" instead of "dbt tests") to make it clear that the message specifically relates to iasWorld tests.

.github/workflows/test_iasworld_data.yaml
```diff
@@ -0,0 +1,141 @@
+name: test-iasworld-data
```
jeancochrane, line 1: This workflow should be nearly identical to [the old implementation of `test-dbt-models`](https://github.com/ccao-data/data-architecture/blob/38b066ed2b4e1c03f21bd6a8ef9d249197e65518/.github/workflows/test_dbt_models.yaml#L1-L146), but GitHub can't pick up on the diff because we're keeping `test-dbt-models` and adapting it for a slightly more specific purpose. I'll call out every change I made in comments below.

.github/workflows/test_iasworld_data.yaml
```diff
@@ -0,0 +1,141 @@
...
+          # shellcheck disable=SC2086
+          python scripts/run_iasworld_data_tests.py \
+            --target "$TARGET" \
+            --output-dir ./qc_test_results/ \
+            $SELECT_OPTION \
+            $SKIP_ARTIFACTS_OPTION \
+            $DEFER_OPTION
+        working-directory: ${{ env.PROJECT_DIR }}
+        shell: bash
+        env:
```
dfsnow, line 96: **question (non-blocking):** Why pass these in as env vars rather than just interpolating them directly into the bash?

dfsnow (approved):
Alright @jeancochrane, this looks good to me. I think the separation here makes sense. Let's get this and the Spark PR merged today.

One tiny blocking issue: we need to update the README to reflect the new workflows and when they're run.

.github/workflows/test_iasworld_data.yaml
jeancochrane: Good question! There's no functional reason, I just felt like it was clearer in this particular case. For `USER`, `GIT_SHA`, `GIT_REF`, and `GIT_AUTHOR`, the script expects these environment variables to be set, so we need to convert the GitHub variables to env vars anyway. For `SELECT_OPTION`, `CACHE_HIT`, and `SKIP_ARTIFACTS_OPTION`, I felt like the GitHub variable references were a bit long/esoteric, and it feels easier to me to read them as env vars. Definitely a weak preference on my part, I'm open to changing any of these if you think a the GitHub variables would be clearer.

.github/workflows/test_dbt_models.yaml
```diff
@@ -90,51 +90,24 @@ jobs:
...
 
       - name: Get current time
         if: failure()
         run: echo "TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%S")" >> "$GITHUB_ENV"
         shell: bash
 
-      # Only triggered when called by a bot. Otherwise, notifications
-      # go to whoever called the workflow
+      # Only triggered when run on a schedule. Otherwise, notifications go to
+      # whoever dispatched the workflow
```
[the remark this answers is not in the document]
jeancochrane: Much clearer, thanks! I updated the other file to match in https://github.com/ccao-data/data-architecture/pull/671/commits/6cb2d36c9a9e09602d2f4b9d4b2cd944bf4c9f08.

Issue #661, opened 2024-12-03
Add QC views for daily commercial queries


jeancochrane:
IC runs some daily queries to use as reference files for their own QC reporting. Let's create some views and define a scheduled job to publish them every day so that IC doesn't have to do it manually anymore.

The steps here:

- [ ] Create the views
- [ ] Update https://github.com/ccao-data/service-spark-iasworld to fix data types for `iasworld.land` columns that are being truncated (like `orate`)
- [ ] Schedule a daily process

PR #678, opened 2024-12-10
Add models for exporting daily Commercial QC reference files

jeancochrane:
This PR completes the first task in https://github.com/ccao-data/data-architecture/issues/661 by creating dbt models corresponding to the reference files that the Commercial team exports daily during desk review. 

A subsequent PR will fix a data type bug in https://github.com/ccao-data/service-spark-iasworld that is causing slight discrepancies in values in four columns in the land details sheet. Once both PRs have landed, we will be ready to schedule a daily process on the VM.


dbt/macros/.sqlfluff
```diff
@@ -1,7 +1,7 @@
 # These rules are required to support sqlfmt. See:
 # https://docs.sqlfmt.com/integrations/sqlfluff
 [sqlfluff]
-exclude_rules = layout.indent, layout.cte_bracket, layout.select_targets, ambiguous.column_count, structure.column_order, RF04, ST05
+exclude_rules = layout.indent, layout.cte_bracket, layout.select_targets, layout.end_of_file, ambiguous.column_count, structure.column_order, RF04, ST05
```
jeancochrane, line 4: I'm not sure why this started failing now, but it seems like sqlfluff and sqlfmt disagree on whether there should be a trailing newline at the end of files. Even more mysteriously, sqlfluff also only seems upset about `dbt/macros/insert_hyphens.sql`, even though no other SQL files end with a trailing newline. I think the root fix here is probably to upgrade both of these dependencies, because I suspect this is a bug in one of the packages, but upgrading causes a huge diff so we should do it in a follow-up PR.

dbt/macros/insert_hyphens.sql
```diff
@@ -0,0 +1,86 @@
...
+-- Call the macro like so:
+--
+--     insert_hyphens("'123456798'", 2, 4, 7)
+--
+-- Remember to wrap any identifiers in strings before passing them to the
+-- macro, since otherwise the jinja parser won't be able to resolve the
+-- reference. For example, if you wanted to transform a column named
+-- `database.column`, you would call the macro like so:
+--
+--     insert_hyphens("database.column", 2, 4, 7)
```
jeancochrane, line 23: Do the motivation and interface for this macro make sense to you?

dbt/models/qc/qc.vw_ic_reference_all_non_res_land_details.sql
```diff
@@ -0,0 +1,55 @@
...
+    land.influ,
+    land.code,
+    land.note1,
+    land.allocpct,
+    land.influ2,
+    CONCAT(
+        COALESCE(land.infl1, ''),
+        CASE WHEN land.infl1 IS NOT NULL THEN ':' ELSE '' END,
+        COALESCE(infl1_reascd.description, '')
+    ) AS infl1,
```
jeancochrane, line 27: Might be useful to make a macro out of this logic at some point, but I skipped it for now since we use this logic much less frequently than the `insert_hyphens` logic.

dbt/models/qc/schema.yml
```diff
@@ -221,6 +221,244 @@ models:
...
+    meta:
+      export_name: AllNonRes_LandDetails
+      export_template: &qc_report_ic_reference_export_template
+        start_row: 3
+        add_table: false
+      export_format:
+        columns:
+          - name: ACRES
+            index: J
+            data_type: float
```
jeancochrane, line 238: I'm a little confused why basically all of our fields get written to Excel as strings with `df.to_excel()`, given that all of these fields have correct types in iasWorld and we're using `unload=True` when querying them. Specifying `data_type` fixes this issue, but do you have insight into why it might be happening in the first place?

dbt/models/qc/schema.yml
```diff
@@ -823,7 +1061,8 @@ models:
         - qc_report_town_close
     meta:
       export_name: 0 land value
-      export_template: assessed_and_market_values
+      export_template:
+        name: assessed_and_market_values
```
jeancochrane, line 1065: This interface has changed in the export script, so we have to migrate existing instances.

dbt/seeds/ccao/ccao.commercial_subclass_1.csv
```diff
@@ -0,0 +1,7 @@
+code,description
```
jeancochrane, line 1: Does it make sense to you to add these descriptions as seeds, similar to `reascd`? If so, what do you think about the name of the seeds?

dbt/seeds/ccao/docs.md
```diff
@@ -35,14 +35,28 @@ descriptions. Also describes which classes are included in residential
 regressions and reporting classes.
 
 Derived from the 2023
-[PDF](https://prodassets.cookcountyassessor.com/s3fs-public/form_documents/Definitions%20for%20Classifications_2023.pdf)
+[PDF](https://prodassets.cookcountyassessor.com/s3fs-public/form_documents/Definitions%20for%20Classifications_2023.pdf).
+
+To find the level of assessment (LoA) for each class, see the `ccao.loa` table.
```
jeancochrane, line 40: Fixing a couple pre-existing typos I noticed in this docs file.

pyproject.toml
```diff
@@ -73,3 +73,4 @@ load_macros_from_path = "dbt/macros"
 # that reference those variables if they are used in code that sqlfluff
 # lints
 get_s3_dependency_dir = "{% macro get_s3_dependency_dir() %}s3://bucket{% endmacro %}"
+insert_hyphens = "{% macro insert_hyphens(str) %}{% set _ = varargs %}'{{ str }}'{% endmacro %}"
```
jeancochrane, line 76: `{% set _ = varargs %}` is kind of goofy here, but we need to reference `varargs` or else dbt's jinja templater will complain about `insert_hyphens` being called with arbitrary positional arguments.

dbt/models/qc/schema.yml
```diff
@@ -221,6 +221,244 @@ models:
...
+            index: T
+            data_type: int
+
+  - name: qc.vw_ic_reference_all_towns_com_bldg_detail
+    description: '{{ doc("view_vw_ic_reference_all_towns_com_bldg_detail") }}'
+    config: *qc_report_ic_reference_tags
+    meta:
+      export_name: AllTowns_ComBldgDetail
+      export_template: *qc_report_ic_reference_export_template
+      export_format:
```
dfsnow, line 357: **issue (blocking):** I'm getting a `ValueError` when running the following:

```bash
uv run python scripts/export_models.py --select tag:qc_report_ic_reference --where "taxyr = '2024'"
```

```
ValueError: invalid literal for int() with base 10: '477.1'
```

I'm guessing it has to do with the changed data types for land? Either way, may be a non-issue once the data type issue (above) is sorted.

dfsnow (commented):
@jeancochrane Data type issues aside this is looking great! Nice work. Looking forward to getting these out.

dbt/models/qc/schema.yml
jeancochrane: PyAthena doesn't seem to work the way I would expect based on the docs. It seems like `unload` should perform this type conversion for us, without the need for a custom converter:

> If the unload option is enabled, the Parquet file itself has a schema, so the conversion is done to the dtypes according to that schema, and the mappings and types settings of the Converter class are not used.

However, in practice, `PandasCursor` with `unload=True` seems to always return `Object` type. However, `ArrowCursor` performs the conversion I would expect (e.g. `comdat.ovrrcnld` gets converted to `pyarrow.decimal128(10, 0)`). I suspect this may be exposing a bug in the PyAthena implementation, but I don't want to spend too much time isolating it right now.

Most likely due to the automatic conversion that the docs describe, the custom converter also doesn't seem to do anything when set on a cursor with `unload=True`. We could switch to a different cursor type, but I do think the pandas cursor with `unload=True` is the easiest one to work with in this case. I  also don't think a custom converter would substantially reduce our code even if we could get it working, because our Athena types are often ambiguous (e.g. `DECIMAL` sometimes indicates a float and sometimes indicates an integer,  and `VARCHAR` sometimes indicates a decimal stored as a string). We might save a little bit of space on the schema specification, but I don't think it's worth it for this particular application. Definitely open to other opinions though!

dfsnow: Alright, let's just go with the existing setup then, as it works and is perfectly clear.

dbt/macros/tests/test_insert_hyphens.sql
```diff
@@ -0,0 +1,35 @@
...
+    {% do test_insert_hyphens_one_position() %}
+    {% do test_insert_hyphens_multiple_positions() %}
+    {% do test_insert_hyphens_raises_when_missing_arguments() %}
+{% endmacro %}
+
+{% macro test_insert_hyphens_one_position() %}
+    {{
+        assert_equals(
+            "test_insert_hyphens_one_position",
+            insert_hyphens("class", 1),
```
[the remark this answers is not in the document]
dfsnow: I'm a dummy and made the comment before reading your comments/docs about quotes. Disregard.

jeancochrane:
@dfsnow All your edits are in, let me know if this is ready to go!

dbt/scripts/utils/export.py
```diff
@@ -365,6 +366,8 @@ def save_model_to_workbook(
                             type_func = int
                         elif data_type == "float":
                             type_func = float
+                        elif data_type == "decimal":
+                            type_func = decimal.Decimal
```
dfsnow, line 370: **question (non-blocking):** Just out of curiousity, what's the deal with the decimal type here? Does it maintain the fixed precision of the Parquet/spark types?

dfsnow (approved):
@jeancochrane Sorry for the long delay! This looks great to me. All set.

dbt/scripts/utils/export.py
jeancochrane: Pretty much! We have to process a couple of fields that are stored as strings in iasWorld, but actually represent arbitrary numerics. The Python `decimal.Decimal` type should let us handle those columns gracefully, without the pitfalls of both `int` (which will truncate any decimals) and `float` (which will add decimals to ints). Annoying that we have to do this, but at least it works.

jeancochrane:
I snuck in one quick change before merging: https://github.com/ccao-data/data-architecture/pull/678/commits/37b49ade0f58b838e0152c1035d20c069ced4e47 passes the `target` argument through from `export_models` to the helper function `query_models_for_export`. This is how `export_models` was always supposed to work, to the extent that the `query_models_for_export` function expected a `target` kwarg even before this change; I think I just forgot to pass the arg through in https://github.com/ccao-data/data-architecture/pull/626.

PR #680, opened 2024-12-11
Strip non-numerics from PINs in `default.vw_pin_appeal`

jeancochrane:
_TK_


jeancochrane:
Closing because this is not actually correct, we do want to remove the appeals with non-numeric PINs.

PR #686, opened 2024-12-20
Add dbt tests to model databases

Damonamajor:
This PR provides DBT tests which fail at a set number of errors. There are three tests which currently fail which should be able to be fixed by deleting the duplicated data.

- model_timing: relaxed-Tristan and stupefied-maya
- model_parameter_final: clever-kyra 
- model_feature_importance: sad-tristan

The other tests are set to warnings with a set value. If we upload duplicated data based on the keys, warnings will transition to error messages, and we will know to further debug. These were set a few days ago, and didn't error during the final commit, so it doesn't seem to be a recurring issue.

Keys are updated with two new variables which should cause the data to be unique.


dbt/models/model/schema.yml
```diff
@@ -4,61 +4,192 @@ sources:
...
+              meta:
+                description: shap should be unique by year, run_id, meta_pin, and meta_card_num
         tags:
           - load_auto
 
       - name: test_card
         description: '{{ doc("table_test_card") }}'
+        data_tests:
+          - unique_combination_of_columns:
+              name: model_test_card_unique
```
Damonamajor, line 169: This one has a super high error level. Is there a key that is missing?

dbt/models/model/docs.md
```diff
@@ -34,7 +34,7 @@ Overall feature importance by model run (`run_id`).
 Includes metrics such as gain, cover, and frequency. This is the output
 of the built-in LightGBM/XGBoost feature importance methods.
 
-**Primary Key**: `year`, `run_id`, `model_predictor_name_all`
+**Primary Key**: `year`, `run_id`, `model_predictor_all_name`
```
dfsnow, line 37: Ah, thank you for updating these docs.

dbt/models/model/docs.md
```diff
@@ -188,4 +188,4 @@ View to compile PIN-level model inputs shared between the residential
 (`model.vw_card_res_input`) and condo (`model.vw_pin_condo_input`) model views.
 
 **Primary Key**: `year`, `run_id`, `meta_pin`
```
dfsnow, line 190: **suggestion:** None of the `model.vw_` views actually have `run_id` as a component of their primary key. Would be nice to remove it for those views.

dbt/models/model/schema.yml
```diff
@@ -4,61 +4,192 @@ sources:
...
+        data_tests:
+          - unique_combination_of_columns:
+              name: model_assessment_card_unique_by_pin_card_and_year
+              combination_of_columns:
+                - meta_pin
+                - meta_card_num
+                - meta_year
+                - run_id
+              config:
+                  error_if: ">5748"
```
dfsnow, line 16: **suggestion:** It would be handy to have a comment here about why we're setting this threshold i.e. their are past errors that we don't intend to fix but we want to prevent future dupes.

dbt/models/model/schema.yml
```diff
@@ -4,61 +4,192 @@ sources:
...
         description: '{{ doc("table_parameter_range") }}'
+        data_tests:
+          - unique_combination_of_columns:
+              name: model_parameter_range_unique_by_year_run_id_and_parameter_name
+              combination_of_columns:
+                - year
+                - run_id
+                - parameter_name
+              meta:
+                description: parameter range should be unique by year run_id and parameter_name
```
dfsnow, line 88:
```suggestion
                description: parameter range should be unique by year, run_id, and parameter_name
```

dbt/models/model/schema.yml
```diff
@@ -4,61 +4,192 @@ sources:
...
         tags:
           - load_auto
 
       - name: parameter_search
         description: '{{ doc("table_parameter_search") }}'
+        data_tests:
+          - unique_combination_of_columns:
+              name: model_parameter_search_unique_by_year_run_id_and_iteration
+              combination_of_columns:
```
dfsnow, line 97: **issue:** I believe the full primary key here is `year`, `run_id`, `iteration`, `configuration`, `fold_id`. Let's update this test and the docs to reflect that.

dbt/models/model/schema.yml
```diff
@@ -4,61 +4,192 @@ sources:
...
+              name: model_assessment_card_unique_by_pin_card_and_year
+              combination_of_columns:
+                - meta_pin
+                - meta_card_num
+                - meta_year
+                - run_id
+              config:
+                  error_if: ">5748"
+              meta:
+                description: assessment card should be unique by pin, card, year, and run_id
```
dfsnow, line 18: **suggestion:** I don't think we actually need a description for these tests, since that's mainly used as an input for constructing Excel QC workbooks. Since these tests are exclusive to GitHub Actions (i.e. are only data tests), the description isn't needed.

dbt/models/model/schema.yml
dfsnow: That's pretty bizarre. I don't _think_ you're missing any key columns here. It's possible that in the past we had separate test set data frames for the linear vs lgbm model results, but I don't see any column that would separate them. IMO as long as more recent model runs aren't adding to the error count it's fine.

dfsnow (commented):
@Damonamajor This looks great. Some super minor changes to make and then this is good to merge.

dbt/models/model/docs.md
```diff
@@ -138,7 +138,7 @@ The test set is the out-of-sample data used to evaluate model performance.
 Predictions in this table are trained using only data _not in this set
 of sales_.
 
-**Primary Key**: `year`, `run_id`, `meta_pin`, `meta_card_num`
+**Primary Key**: `year`, `run_id`, `meta_pin`, `meta_card_num`, `document_number`
```
dfsnow, line 141: **nitpick (non-blocking):** There's not (yet) a doc number column in this dataset.

dfsnow (approved):
@Damonamajor Let's add failure thresholds to the last remaining failing tests, then have @jeancochrane take a look at this. After that it should be set to go.

Damonamajor:
> @Damonamajor Let's add failure thresholds to the last remaining failing tests, then have @jeancochrane take a look at this. After that it should be set to go.
@dfsnow 
The remaining failing tests should be easily fixed (with the notes in the original comment). Do we want to fix them or add the thresholds?

dbt/models/model/docs.md
Damonamajor: This is a variable do we want it?
meta_sale_document_num

PR #687, opened 2024-12-23
Create baby version of `default.vw_pin_universe`

wrridgeway:
This is a stopgap measure to make updating the open data 'Parcel Universe' manageable. Presently it is so large that doing things like adding columns leads to the asset getting orphaned while attempting to load the data into memory, which is a prerequisite for publishing updates to the data. It might not ultimately work, but it's worth a try. Not a longterm solution as we plan to move to a one-year only version of Parcel Universe.

Shouldn't affect what the public sees on the open data portal.

### DIFF
--- a/dbt/models/default/default.vw_pin_universe_temp.sql
+++ b/dbt/models/default/default.vw_pin_universe_temp.sql
@@ -1,3 +1,3 @@
 -- Small, managable version of pin universe for the purporse of updating the
 -- Parcel Universe asset on open data.
-SELECT * FROM {{ ref('location.vw_pin10_location') }} LIMIT 100000
+SELECT * FROM {{ ref('default.vw_pin_universe') }} LIMIT 100000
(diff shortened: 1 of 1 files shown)


wrridgeway:
I can add more documentation if we'd like, but I plan on deleting this view by the end of the week if things go according to plan.

dbt/models/default/default.vw_pin_universe_temp.sql
```diff
@@ -0,0 +1,3 @@
+-- Small, managable version of pin universe for the purporse of updating the
+-- Parcel Universe asset on open data.
+SELECT * FROM {{ ref('location.vw_pin10_location') }} LIMIT 100000
```
jeancochrane, line 3: **[Question, non-blocking]** Does it matter that this view won't have the same schema as `default.vw_pin_universe`?

jeancochrane (approved):
I don't totally get the constraints we're under here, but I'm willing to give this approach a try!

dbt/models/default/default.vw_pin_universe_temp.sql
wrridgeway: Yes, thank you for catching this Jean. I completely got this wrong.

PR #689, opened 2024-12-24
Remove `default.vw_pin_universe_temp`

wrridgeway:
Unfortunately, using this small version of the pin universe view didn't help update its associated open portal asset in the way I'd hoped.


jeancochrane (approved):
Bummer that it didn't work, but it was worth a shot. For posterity, this PR is reverting https://github.com/ccao-data/data-architecture/pull/687.

PR #691, opened 2024-12-26
Create `open_data` athena db

wrridgeway:
⚠️ URLs pointing to the old parcel universe asset in exposures, the wiki, and the open data portal stories need to be updated before this PR is merged. ⚠️ 

This PR creates a new athen db, `open_data`. This db will feed all of our open data assets that use the API for updating - this only excludes `ccao.commercial_data`. The main goal here is to make sure all of the columns available through the API for a given open data asset match the columns in the athena asset that feeds it. `row_id` columns are now constructed in the view that feeds an open data asset rather than during uploads through the API.

There are a lot of discrepancies between our current athena views and what is making it onto the open data portal. Some of those are by design, but clearly some are a product of the athena views having grown while the open data sets didn't keep up. I think it's worth opening another PR after this in order to try and bring the open data assets into parity with the views that (ultimately) feed them.


socrata/socrata_upload.py
```diff
@@ -62,7 +62,7 @@ def get_asset_info(socrata_asset):
 
     model = pd.json_normalize(model)
     model = model[model["label"] == socrata_asset]
-    athena_asset = model.iloc[0]["depends_on.nodes"][0].split(".")[-1]
+    athena_asset = model.iloc[0]["depends_on.nodes"][0].split(".", 2)[-1]
```
wrridgeway, line 65: This makes sure the athena db is included in all queries.

socrata/socrata_upload.py
```diff
@@ -96,7 +96,7 @@ def build_query(
 
     # Limit pull to columns present in open data asset - shouldn't change anything, but prevents failure if columns have become misaligned.
     asset_columns = (
-        requests.get(
+        session.get(
```
wrridgeway, line 99: Accidentally omitted switch to `session`

.github/workflows/socrata_upload.yaml
```diff
@@ -7,15 +7,16 @@ on:
...
-        - Residential Condominium Unit Characteristics
-        - Parcel Sales
-        - Assessed Values
         - Appeals
+        - Assessed Values
         - Parcel Addresses
         - Parcel Proximity
+        - Parcel Sales
         - Property Tax-Exempt Parcels
+        - Parcel Universe
```
wrridgeway, line 16: Depending on how we merge all these PRs in, we may want this changed to "Parcel Universe Historic" prior to merging.

socrata/socrata_upload.py
wrridgeway, line 1: Ripped row id construction out of this script since row id will already be in every open data view

dbt/models/open_data/docs.md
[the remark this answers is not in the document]
jeancochrane: I don't think we need a ton of detail in these descriptions, since the views are all pretty simple wrappers around single assets. I'd be fine with something like:

```markdown
{% docs view_vw_appeal %}
Thin wrapper around `default.vw_pin_appeal` that powers the `Appeals` open data asset
{% enddocs %}

File: .github/workflows/socrata_upload.yaml
Comment (wrridgeway) [inline .github/workflows/socrata_upload.yaml:1]: Adding both new assets and renaming the parcel universe assets here.

File: dbt/models/default/docs.md
Comment (wrridgeway) [inline dbt/models/default/docs.md:1]: This combo uniquely identifies rows and feels more intuitive to me.

File: dbt/models/default/docs.md [hunk 4]
```diff
@@ -128,7 +128,7 @@ View containing aggregate land square footage for all PINs.
 View containing building permits organized by PIN, with extra metadata
 recorded by CCAO permit specialists during the permit processing workflow.
 
-**Primary Key**: `pin`, `date_issued`
+**Primary Key**: `pin`, `permit_number`
```
wrridgeway, line 131: This combo uniquely identifies rows and feels more intuitive to me:

`select count(*) from default.vw_pin_permit group by pin, permit_number having count(*) > 1`

Yields zero rows. I also needed to make sure that none of the columns I used for PKs had any NULL values, and this combo suits that condition.

dbt/models/open_data/exposures.yml
wrridgeway, line 1: Mostly just a lift and shift from the `default` folder, but also added the two new assets and changed refs to point to the `open_data` db.

socrata/socrata_upload.py
```diff
@@ -113,6 +97,8 @@ def build_query(
         .strip("]")
         .split(",")
     )
+    # row id won't show up here since it's hidden on the open data portal assets
+    asset_columns += ["row_id"]
```
wrridgeway, line 101: `row_id` is hidden by the API, but still necessary to include.

dbt/models/default/docs.md
jeancochrane: **[Praise]** Yup, this new version is correct. Thanks!

dbt/models/open_data/open_data.vw_assessed_value.sql
```diff
@@ -0,0 +1,49 @@
...
+    twoyr_pri_mailed_land
+    twoyr_pri_mailed_tot
+    twoyr_pri_certified_bldg
+    twoyr_pri_certified_land
+    twoyr_pri_certified_tot
+    twoyr_pri_board_bldg
+    twoyr_pri_board_land
+    twoyr_pri_board_tot
+    twoyr_pri_change_reason
+*/
```
jeancochrane, line 30: **[Thought, non-blocking]** I'm a little bit unsure how useful it is to document these missing fields, mostly because they're at high risk of becoming out of date if we change the underlying views and neglect to update these lists. Is there a big value-add to duplicating these fields in a docstring vs. requiring interested code authors to navigate to the underlying view to see which fields are missing? I'm genuinely not sure and don't have strong feelings about it, but I'm curious what you think!

jeancochrane (approved):
Thanks for this!

dbt/models/open_data/open_data.vw_assessed_value.sql
wrridgeway: I'm very inclined to agree with you, I just figured others would disagree with that take. I'm more than happy to rip these commented columns out.

PR #692, opened 2024-12-27
Fix `minor_subclass` join in `qc.vw_ic_reference_all_towns_com_bldg_detail`

jeancochrane:
IC identified a small bug in one of the reference queries we started exporting for them in https://github.com/ccao-data/data-architecture/pull/678: We're joining the `ccao.commercial_minor_subclass` table to the _major_ subclass code rather than the minor code, which is producing the wrong human-readable descriptions for minor subclasses. This PR is a one-line change that fixes that bug.

I tested this change by rerunning the table and confirming the output for the `minor_subclass` column displays the correct human-readable description.


dfsnow (approved):
Whoops! I should've caught this in review. Good on the IC folks.

PR #696, opened 2025-01-02
Allow non-default db queries in socrata API script

wrridgeway:
Previously the script could query the default db, but no other DBs. This PR fixes that.

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -62,7 +62,7 @@ def get_asset_info(socrata_asset):
 
     model = pd.json_normalize(model)
     model = model[model["label"] == socrata_asset]
-    athena_asset = model.iloc[0]["depends_on.nodes"][0].split(".")[-1]
+    athena_asset = model.iloc[0]["depends_on.nodes"][0].split(".", 2)[-1]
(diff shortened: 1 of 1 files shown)


socrata/socrata_upload.py
```diff
@@ -62,7 +62,7 @@ def get_asset_info(socrata_asset):
 
     model = pd.json_normalize(model)
     model = model[model["label"] == socrata_asset]
-    athena_asset = model.iloc[0]["depends_on.nodes"][0].split(".")[-1]
+    athena_asset = model.iloc[0]["depends_on.nodes"][0].split(".", 2)[-1]
```
wrridgeway, line 65: This makes sure the athena db is included in all queries.

socrata/socrata_upload.py
```diff
@@ -96,7 +96,7 @@ def build_query(
 
     # Limit pull to columns present in open data asset - shouldn't change anything, but prevents failure if columns have become misaligned.
     asset_columns = (
-        requests.get(
+        session.get(
```
wrridgeway, line 99: Accidentally omitted switch to session

jeancochrane (approved):
Good catch!

PR #699, opened 2025-01-07
Ensure socrata workflow inputs are handled correctly

wrridgeway:
The default workflow value for years, 'all', was not being handled correctly since it becomes a list during ingest. Additionally, spaces in the years input could lead to year values not being handled correctly.

Since the columns that get updated currently depend on how the response from the API and the column names of the athena view that feeds the asset match up, it'd be nice to be able to confirm that all expected columns are in fact going to be updated.

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -337,6 +337,6 @@ def socrata_upload(
 socrata_upload(
     socrata_asset=os.getenv("SOCRATA_ASSET"),
     overwrite=os.getenv("OVERWRITE"),
-    years=str(os.getenv("YEARS")).split(","),
+    years=str(os.getenv("YEARS").replace(" ", "")).split(","),
     by_township=os.getenv("BY_TOWNSHIP"),
 )
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -337,6 +337,6 @@ def socrata_upload(
 socrata_upload(
     socrata_asset=os.getenv("SOCRATA_ASSET"),
     overwrite=os.getenv("OVERWRITE"),
-    years=str(os.getenv("YEARS").replace(" ", "")).split(","),
+    years=str(os.getenv("YEARS")).replace(" ", "").split(","),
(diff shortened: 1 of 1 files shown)


socrata/socrata_upload.py
```diff
@@ -333,6 +337,6 @@ def socrata_upload(
 socrata_upload(
     socrata_asset=os.getenv("SOCRATA_ASSET"),
     overwrite=os.getenv("OVERWRITE"),
-    years=str(os.getenv("YEARS")).split(","),
+    years=str(os.getenv("YEARS")).replace(" ", "").split(","),
```
wrridgeway, line 340: Without this entering "2023, 2024" in the github action will include the space with 2024 and not actually pull that year.

socrata/socrata_upload.py
```diff
@@ -213,6 +215,8 @@ def generate_groups(athena_asset, years=None, by_township=False):
             .to_list()
         )
 
+    # Ensure township codes aren't available if they shouldn't be
+    township_codes = None
```
wrridgeway, line 219: Shouldn't be an issue running the script as a job, but if you're ever working interactively township codes can be present in your environment when you need it not to be.

socrata/socrata_upload.py
```diff
@@ -204,7 +206,7 @@ def generate_groups(athena_asset, years=None, by_township=False):
     if not years and by_township:
         raise ValueError("Cannot set 'by_township' when 'years' is None")
 
-    if years == "all":
+    if years == ["all"]:
```
wrridgeway, line 209: the default `'all'` is turned into `['all']` when inherited through the environment and this logic wasn't properly reacting to it.

socrata/socrata_upload.py
jeancochrane: **[Nitpick, non-blocking]** In the case when `by_township` is truthy, we define `township_codes` as a list on line 221, right? Might be more intuitive if we match that data type here:

```suggestion
    township_codes = []
```

jeancochrane (approved):
Looks good, thanks!

PR #702, opened 2025-01-09
Ensure API upload script throws error/warning if Athena and Socrata columns mismatch

wrridgeway:
This PR makes sure we don't accidentally update _part_ of an open data asset.  If there are any columns in an open data asset that wouldn't get updated by the script, it will throw an error.

For now, it will only log a warning if there are columns in the Athena view feeding an asset that are not also in the asset. This is done to give us flexibility for now since not all columns in a view are intended to end up on the open data portal. This can be adjusted in the future if we shift the views feeding all assets to the `open_data` db since we'll want those views and assets to mirror one another 1:1.

Here are examples of [expected failure](https://github.com/ccao-data/data-architecture/actions/runs/12695581200/job/35387775210) and [expected success with a warning](https://github.com/ccao-data/data-architecture/actions/runs/12697420561/job/35393657163).

### DIFF
--- a/socrata/pyproject.toml
+++ b/socrata/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
   "PyAthena~=3.9.0",
-  "ndg-httpsclient>=0.5.1",
+  "ndg-httpsclient~=0.5.1",
   "pandas~=2.2.2",
   "pyarrow~=17.0.0",
   "requests~=2.32.3",
--- a/socrata/pyproject.toml
+++ b/socrata/pyproject.toml
@@ -4,6 +4,7 @@ version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
   "PyAthena~=3.9.0",
+  "ndg-httpsclient>=0.5.1",
   "pandas~=2.2.2",
   "pyarrow~=17.0.0",
   "requests~=2.32.3",
--- a/socrata/pyproject.toml
+++ b/socrata/pyproject.toml
@@ -4,7 +4,6 @@ version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
   "PyAthena~=3.9.0",
-  "ndg-httpsclient~=0.5.1",
   "pandas~=2.2.2",
   "pyarrow~=17.0.0",
   "requests~=2.32.3",
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -110,7 +110,13 @@ def build_query(
     )
     asset_columns.sort()
 
-    # If Athena and Socrata columns don't match, abort upload and inform user of discrepancies.
+    # If there are columns on Socrata that are not in Athena, abort upload and
+    # inform user of discrepancies. The script should not run at all in this
+    # circumstance since it will update some but not all columns in the open
+    # data asset.
+    # If there are columns in Athena but not on Socrata, it may be the case that
+    # they should be added, but there are also cases when not all columns for an
+    # Athena view that feeds an open data asset need to be part of that asset.
     if athena_columns != asset_columns:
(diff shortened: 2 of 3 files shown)


socrata/pyproject.toml
```diff
@@ -4,6 +4,7 @@ version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
   "PyAthena~=3.9.0",
+  "ndg-httpsclient>=0.5.1",
```
wrridgeway, line 7: This is an attempt to address an issue with an error I was getting well into uploading large datasets, [`requests.exceptions.SSLError: EOF occurred in violation of protocol`](https://stackoverflow.com/questions/33410577/python-requests-exceptions-sslerror-eof-occurred-in-violation-of-protocol/39758581#39758581)

socrata/pyproject.toml
```diff
@@ -4,6 +4,7 @@ version = "0.0.1"
 requires-python = ">=3.11"
 dependencies = [
   "PyAthena~=3.9.0",
+  "ndg-httpsclient~=0.5.1",
```
wrridgeway, line 7: This is an attempt to address an issue with an error [I was getting](https://github.com/ccao-data/data-architecture/actions/runs/12695906661/job/35388829830) well into uploading large datasets, [`requests.exceptions.SSLError: EOF occurred in violation of protocol`](https://stackoverflow.com/questions/33410577/python-requests-exceptions-sslerror-eof-occurred-in-violation-of-protocol/39758581#39758581)

[It seems to have worked.](https://github.com/ccao-data/data-architecture/actions/runs/12697744731/job/35394712217)

socrata/socrata_upload.py
```diff
@@ -105,6 +111,30 @@ def build_query(
...
+    if athena_columns != asset_columns:
+        columns_not_on_socrata = set(athena_columns) - set(asset_columns)
+        columns_not_in_athena = set(asset_columns) - set(athena_columns)
+        exception_message = (
+            f"Columns on Socrata and in Athena do not match for {athena_asset}"
+        )
+
+        if len(columns_not_on_socrata) > 0:
+            exception_message += f"\nColumns in Athena but not on Socrata: {columns_not_on_socrata}"
+            logging.warning(exception_message)
```
jeancochrane, line 132: **[Thought, non-blocking]** Not necessary to change, but I thought you might be interested in some Python code style: It's actually somewhat uncommon to call methods on the `logging` module directly like this. Per [the docs](https://docs.python.org/3/library/logging.html), typical usage is to instantiate a module-level logger object that we then use to call logging methods:

```python
# __name__ is a reserved keyword storing the name of the current module
logger = logging.getLogger(__name__)
logger.warning("Your warning message here")
```

When we call a method like `warning()` directly on the `logging` object that we import into the module, we're calling it on the "root" logger, which is the default logging object that the `logging` module exports.

There are a few ways that a module-level logger object offers more flexibility than the root logger, but the most important one is that the log output contains a reference to the module that emitted the log:

```python
>>> import logging
>>> logging.warning("foo")

WARNING:root:foo

>>> logger = logging.getLogger(__name__)
>>> logger.warning("foo")

# __main__ is a special module name referring to the top-level code environment, see:
# https://docs.python.org/3/library/__main__.html
WARNING:__main__:foo
```

Since all of the code for `socrata/socrata_upload.py` lives in one module, this feature doesn't matter much, since we can be confident that any logs emitted by the root logger are coming from this script. If we ever decide to refactor the code to pull out certain pieces into other modules, however, module-level loggers would help us know which part of the code is emitting the logs that we see.

jeancochrane (approved):
Thanks for separating this from the SSL error business that I was unsure about!

socrata/socrata_upload.py
wrridgeway: Thank you @jeancochrane, fixed!

PR #706, opened 2025-01-14
Ensure sales view row id is string

wrridgeway:
Quick fix to https://github.com/ccao-data/data-architecture/pull/691 that makes sure the only row id that isn't a product of a `concat` operation is a string.

image


jeancochrane (approved):
Oops, thanks!

PR #707, opened 2025-01-15
Improve Socrata API upload log formatting

wrridgeway:
Tiny formatting tweak to make output a little more readable.

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -134,7 +134,7 @@ def build_query(athena_asset, asset_id, years=None, township=None):
         + columns[columns["type"] == "array(varchar)"]["column"]
     )
 
-    print(f"The following columns will be updated: {columns}")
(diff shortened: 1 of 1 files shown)


socrata/socrata_upload.py
```diff
@@ -134,7 +134,7 @@ def build_query(athena_asset, asset_id, years=None, township=None):
         + columns[columns["type"] == "array(varchar)"]["column"]
     )
 
-    print(f"The following columns will be updated: {columns}")
+    print(f"The following columns will be updated:\n{columns}")
```
jeancochrane, line 137: **[Suggestion, non-blocking]** We can take advantage of the fact that `print()` adds a newline at the end to simplify this a little bit:

```suggestion
    print("The following columns will be updated:")
    print(columns)
```

jeancochrane (approved):
Makes sense!

PR #709, opened 2025-01-15
Update IC reference files to match format for missing data and sheet names with Inquire

jeancochrane:
The IC team noticed two small but consequential differences between the output of an Inquire query and the output of the `export_models` script that we use to power our daily IC reference file exports (https://github.com/ccao-data/data-architecture/pull/678):

1. Inquire stores missing data as empty strings, whereas our exports store missing data as nulls
2. Inquire stores data in a sheet named `Query Results`, whereas our exports use the default sheet name `Sheet1`

The Excel workbook interface doesn't distinguish between nulls and empty strings when displaying missing values, but the Power Query editor _does_ distinguish between them, and the discrepancy meant that IC staff needed to define two separate processing steps in their power queries to account for the fact that our data sources had slightly different formats for missing values compared to Inquire. Similarly, IC needed to define two different steps to account for the different sheet names containing data.

This PR adds two new config options to the `export_models` script to resolve these differences:

* `export_format.format_blanks_as_empty_strings`: When `True`, outputs empty strings for missing values instead of nulls (defaults to `False`)
* `export_template.sheet_name`: The name of the sheet in the template to use to store data (defaults to `Sheet1`, the Excel default)
 
Using these new config attributes, we update the model configs for our daily IC reference files to store blanks empty strings and use the sheet name `Query Results` so that our workbooks exactly match the corresponding Inquire output.


dbt/export/templates/qc.vw_ic_reference_all_non_res_land_details.xlsx
jeancochrane, line 1: I updated the sheet names from `Sheet1` to `Query Results` in these template files. Otherwise, they haven't changed.

dbt/models/qc/schema.yml
```diff
@@ -231,7 +231,9 @@ models:
       export_template: &qc_report_ic_reference_export_template
         start_row: 3
         add_table: false
-      export_format:
+        sheet_name: Query Results
+      export_format: &qc_report_ic_reference_export_format
+        format_blanks_as_empty_string: true
```
jeancochrane, line 236: The sheet name feels to me like a template configuration option, while the missingness value feels like a format configuration option. The boundary here is somewhat arbitrary, however, so I'm open to different approaches.

dbt/scripts/utils/export.py
```diff
@@ -389,17 +404,26 @@ def save_model_to_workbook(
...
-                                if row[idx].value != ""
-                                and row[idx].value is not None
-                                else None
-                            )
-                        else:
-                            setattr(row[idx], attr, val)
-
-        print(f"Exported model {model.name} to {output_path}")
-        return output_path
+                for cell in row:
```
jeancochrane, line 407: Iterating every cell in the workbook is very slightly slower than our previous approach, but not enough to make a substantial difference since most of our models are much longer than they are wide.

dbt/scripts/utils/export.py
```diff
@@ -379,6 +382,18 @@ def save_model_to_workbook(
...
+            # default, openpyxl will output empty cells as nulls, but we allow
+            # users to override this behavior to instead output an empty string
+            # using the `format_blanks_as_empty_string` config
+            format_blanks_as_empty_string = format_config.get(
+                "format_blanks_as_empty_string"
+            )
+            # openpyxl does not support writing empty strings, so we write a
+            # formula that evaluates to an empty string as a workaround. See:
+            # https://foss.heptapod.net/openpyxl/openpyxl/-/issues/2174
+            blank_value = '=""' if format_blanks_as_empty_string else None
```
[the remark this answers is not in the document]
dfsnow: Woof yeah this is really bizarre. Nice workaround though.

dfsnow (approved):
Looks good @jeancochrane!

PR #710, opened 2025-01-21
Move socrata column transformations to `open_data` db

wrridgeway:
This PR moves column transformations that used to be applied on the open data portal into the `open_data` Athena db. This lets us better version control these transformations, apply more intricate transformations, and avoid the multiple nested transformations that could happen from altering transformation on the open data portal without replacing the entire dataset. Additionally, some transformation on the open data portal randomly seemed to stop being applied recently and we couldn't get a straight answer as to why.

The PR also fixes from functionality that was broken in the API upload script.


socrata/socrata_upload.py
```diff
@@ -30,6 +30,20 @@
     cursor_class=PandasCursor,
 ).cursor(unload=True)
 
+# Set global counter value
+count = 0
```
wrridgeway, line 34: Multiple functions need to update this counter, so it's easiest to make it global.

socrata/socrata_upload.py
```diff
@@ -30,6 +30,20 @@
...
+
+
+def parse_years(years):
```
wrridgeway, line 37: `None` was not getting handled correctly before, it was getting turning into `['None']`

socrata/socrata_upload.py
```diff
@@ -125,15 +139,6 @@ def build_query(athena_asset, asset_id, years=None, township=None):
     # Limit pull to columns present in open data asset
     columns = columns[columns["column"].isin(asset_columns)]
 
-    # Array type columns are not compatible with the json format needed for
-    # Socrata uploads. Automatically convert any array type columns to string.
-    columns.loc[columns["type"] == "array(varchar)", "column"] = (
```
wrridgeway, line 130: This is now handled in athena.

socrata/socrata_upload.py
```diff
@@ -188,9 +195,8 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
     # properly formatted
     input_data = cursor.execute(sql_query, query_conditionals).as_pandas()
     date_columns = input_data.select_dtypes(include="datetime").columns
-    input_data[date_columns] = input_data[date_columns].map(
-        lambda x: x.strftime("%Y-%m-%dT%X")
-    )
+    for i in date_columns:
```
wrridgeway, line 198: This does the same thing as the old map function did, but that function was giving me difficulties setting `NaT` values. Prior to testing the permit view, I hadn't had to deal with missing timestamp values.

socrata/socrata_upload.py
```diff
@@ -188,9 +195,8 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
...
+        input_data[i] = input_data[i].fillna("").dt.strftime("%Y-%m-%dT%X")
```
wrridgeway, line 199: This does the same thing as the old map function did, but that function was giving me difficulties setting `NaT` values. Prior to testing the permit view, I hadn't had to deal with missing timestamp values.

socrata/socrata_upload.py
```diff
@@ -207,12 +213,14 @@ def upload(method, asset_id, sql_query, overwrite, year=None, township=None):
     for i in range(0, input_data.shape[0], 10000):
         print(print_message)
         print(f"Rows {i + 1}-{i + 10000}")
+        if count > 0:
+            method = "post"
```
wrridgeway, line 217: Making sure mandatory chunking based on row size doesn't just keep overwriting.

socrata/socrata_upload.py
```diff
@@ -30,6 +30,20 @@
...
+
+def parse_years(years):
+    """
+    Make sure the years environmental variable is formatted correctly.
+    """
+
+    if years is not None:
+        years = str(years).replace(" ", "").split(",")
+
+    return years
```
wrridgeway, line 45: `None` was not getting handled correctly before, it was getting turning into `['None']`

socrata/socrata_upload.py
```diff
@@ -125,15 +139,6 @@ def build_query(athena_asset, asset_id, years=None, township=None):
...
 
-    # Array type columns are not compatible with the json format needed for
-    # Socrata uploads. Automatically convert any array type columns to string.
-    columns.loc[columns["type"] == "array(varchar)", "column"] = (
-        "ARRAY_JOIN("
-        + columns[columns["type"] == "array(varchar)"]["column"]
-        + ", ', ') AS "
-        + columns[columns["type"] == "array(varchar)"]["column"]
-    )
-
```
wrridgeway, line 136: This is now handled in athena.

dbt/models/open_data/open_data.vw_permit.sql
```diff
@@ -2,17 +2,29 @@
 -- Some columns from the feeder view may not be present in this view.
 
 SELECT
-    CONCAT(pin, permit_number) AS row_id,
+    CONCAT(
+        pin, COALESCE(permit_number, '')
+    ) AS row_id,
```
wrridgeway, line 7: I really thought I tested for this before, but since we've got some `NULL` permit numbers and `CONCAT` returns `NULL` in those cases, we need to coalesce permit number and an empty string to make sure we don't have any `NULL` row id's.

socrata/socrata_upload.py
jeancochrane: **[Nitpick, blocking]** Let's chat about this IRL before we merge, global mutable variables are [generally considered an antipattern](https://softwareengineering.stackexchange.com/questions/148108/why-is-global-state-so-evil) so we should see if we have any other options before moving forward with this design.

dbt/models/open_data/open_data.vw_parcel_universe_current.sql
```diff
@@ -95,28 +95,42 @@ SELECT
...
-    tax_library_district_name,
-    tax_park_district_num,
-    tax_park_district_name,
-    tax_sanitation_district_num,
-    tax_sanitation_district_name,
-    tax_special_service_area_num,
-    tax_special_service_area_name,
-    tax_tif_district_num,
-    tax_tif_district_name,
+    ARRAY_JOIN(tax_municipality_num, ', ') AS tax_municipality_num,
```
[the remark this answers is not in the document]
wrridgeway: A great idea.

dbt/models/open_data/open_data.vw_permit.sql
```diff
@@ -21,11 +33,704 @@ SELECT
...
+        WHEN job_code_secondary = '566' THEN '2002 HISTORICAL LANDMARK'
+        WHEN job_code_secondary = '911' THEN 'OCCUPANCY CODE WILL NOT BE USED'
+    END AS job_code_secondary,
     work_description,
-    improvement_code_1,
-    improvement_code_2,
-    improvement_code_3,
-    improvement_code_4
+    CASE
+        WHEN improvement_code_1 = '110' THEN 'NEW CONSTRUCTION - MAJOR IMPRVT'
```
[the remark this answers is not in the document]
wrridgeway: > Never attribute to malice that which is adequately explained by stupidity.

dbt/models/open_data/open_data.vw_permit.sql
```diff
@@ -21,11 +33,704 @@ SELECT
...
+        WHEN improvement_code_4 = '392' THEN 'MINOR IMPROVEMENT - BURNOUT'
+        WHEN improvement_code_4 = '400' THEN 'EXEMPT PROPERTIES'
+        WHEN improvement_code_4 = '411' THEN 'LEASEHOLD'
+        WHEN improvement_code_4 = '510' THEN 'RECHECKS'
+        WHEN improvement_code_4 = '511' THEN 'OCCUPANCY'
+        WHEN improvement_code_4 = '512' THEN 'PRIOR YEAR PARTIAL ASSESSMENT'
+        WHEN improvement_code_4 = '515' THEN 'MANUALLY ISSUED PERMIT'
+        WHEN improvement_code_4 = '516' THEN 'PRIOR YEAR RECHECK'
+        WHEN improvement_code_4 = '900' THEN 'OTHER'
+    END AS improvement_code_4
```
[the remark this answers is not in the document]
wrridgeway: I never think to use jinja. I need to internalize that it's available.

socrata/socrata_upload.py
wrridgeway: I just went ahead and switched over to passing count as an argument to the `upload` function so that it has access to it.

wrridgeway:
Seems like there's a lot of room for adding seeds here. I can create a new issue for that once this is wrapped.

dbt/models/open_data/open_data.vw_permit.sql
```diff
@@ -2,17 +2,29 @@
 -- Some columns from the feeder view may not be present in this view.
 
 SELECT
-    CONCAT(pin, permit_number) AS row_id,
+    CONCAT(
+        pin, COALESCE(permit_number, ''), COALESCE(date_issued, '')
```
wrridgeway, line 6: As you said, there are currently three null date_issued values right now, so unfortunately we'll need to coalesce.

jeancochrane (approved):
Awesome work, thanks!

PR #712, opened 2025-01-22
Add uniqueness test for `default.vw_pin_permit` view

jeancochrane:
I noticed while reviewing https://github.com/ccao-data/data-architecture/pull/710 that our new open data asset relies on a definition of uniqueness for the `default.vw_pin_permit` view that is not guaranteed to hold. This PR adds a data integrity test for that assumption so that we will get alerted if any new data come in that violate it.

Note that we may want to add `date_issued` to this uniqueness definition as well (discussion here: https://github.com/ccao-data/data-architecture/pull/710#discussion_r1925696992) but I'm holding off on doing that until we resolve our discussion, since we'll also want to change the docs and the open data model definition if we decide to do that.


wrridgeway (approved):
Thank you for catching this!!

PR #715, opened 2025-01-27
Fixup tests and views for the 2025 rollover

jeancochrane:
Now that we have 2025 data in Athena, [some of our data integrity tests are failing](https://github.com/ccao-data/data-architecture/actions/runs/12987644431/job/36216966648) due to views that aren't properly equipped to handle data following a rollover. This PR updates those views and bumps the `data_test_iasworld_year_start` variable to 2025 so that all of our tests are looking at the latest year of data.

Data integrity test output: https://github.com/ccao-data/data-architecture/actions/runs/13000380643/job/36257605843

iasWorld data test output: https://github.com/ccao-data/data-architecture/actions/runs/13000382805/job/36257611283


dbt/macros/tests/test_all.sql
```diff
@@ -5,6 +5,5 @@
     {% do test_generate_schema_name() %}
     {% do test_generate_alias_name() %}
     {% do test_format_additional_select_columns() %}
-    {% do test_pre_stage_filters() %}
```
jeancochrane, line 8: These tests weren't really useful in the first place, and are even less useful now that our two pre-stage macros don't even share common logic. I got rid of the tests to reduce the future maintenance burden for this view.

dbt/models/default/default.vw_pin_history.sql
```diff
@@ -6,6 +6,12 @@ SELECT
     leg.user1 AS township_code,
     town.township_name,
     REGEXP_REPLACE(par.nbhd, '([^0-9])', '') AS nbhd,
+    -- Add pre-mailed values to cover us for the period after the rollover but
+    -- before the year's assessment cycle starts
+    vwpv.pre_mailed_class,
+    vwpv.pre_mailed_bldg,
+    vwpv.pre_mailed_land,
+    vwpv.pre_mailed_tot,
```
jeancochrane, line 14: The lack of these values was causing all 2025 records in the `reporting.vw_top_5_muni` view to have null values for stage and `ccao_av`. We have to add them into this source view so that we can properly pull them into `vw_top_5_muni`.

dbt/models/default/default.vw_pin_value.sql
```diff
@@ -16,10 +16,25 @@ WITH stages AS (
...
+                    -- always cast the array to null
+                    ELSE ''
+                END
+            ),
+            ''
+        ) AS procnames
     FROM {{ source('iasworld', 'asmt_all') }}
-    WHERE procname IN ('CCAOVALUE', 'CCAOFINAL', 'BORVALUE')
-        AND rolltype != 'RR'
+    WHERE rolltype != 'RR'
```
jeancochrane, line 37: Another subtle bug that I missed due to the lack of existing pre-mailed values when I merged this code: In the period before we stamp a PIN procname for mail, it has no not-null procnames in `asmt`, so the result set for this CTE does not include a value for the PIN. As a consequence, when we right join this CTE to the main query in order to run `CONTAINS()` checks on the procnames and look for provisional values, the `procnames` field will be a null instead of an empty array and `CONTAINS()` will always return null. The diff here ensures that we will always have a record for every PIN and that `procnames` will be an empty list for PINs that have not yet mailed.

dbt/models/default/schema/default.vw_pin_value.yml
```diff
@@ -118,28 +118,31 @@ models:
           - `3` = Board certified
       - name: year
         description: '{{ doc("shared_column_year") }}'
+        data_tests:
+          - not_null:
+              name: default_vw_pin_value_stage_num_not_null
```
jeancochrane, line 123: Added this test in while I was here, since it was quick and easy and the equivalent `stage_name` test ended up being helpful in catching some problems.

dbt/models/default/schema/default.vw_pin_value.yml
```diff
@@ -118,28 +118,31 @@ models:
...
 
     data_tests:
     - not_null:
         name: default_vw_pin_value_board_class_not_null
         column_name: board_class
         config:
           where: CAST(year AS int) < {{ var('data_test_iasworld_year_start') }} - 1
-          error_if: ">1260"
+          error_if: ">1261"
```
jeancochrane, line 131: Annoyingly, bumping `data_test_iasworld_year_start` revealed one PIN in 2023 stuck without a board value. I'm not totally sure how much time to spend on bugging folks to fix this, since we already have 1,260 other PINs in other years with seemingly-impossible stages.

dbt/models/reporting/reporting.vw_top_5_muni.sql
```diff
@@ -63,8 +63,11 @@ most_recent_values AS (
         pin,
         year,
         oneyr_pri_board_tot AS prior_bor_av,
-        COALESCE(certified_tot, mailed_tot) AS ccao_av,
+        COALESCE(certified_tot, mailed_tot, pre_mailed_tot) AS ccao_av,
         CASE
+            WHEN
+                (certified_tot IS NULL AND mailed_tot IS NULL)
+                THEN 'pre-mailed'
```
jeancochrane, line 70: Only adding the pre-mailed provisional stage here to avoid null AVs for towns that haven't mailed yet, per the test that was failing. It wouldn't be too hard to add the pre-certified provisional stage too, but I'm avoiding it until we decide it's necessary.

dbt/models/reporting/schema.yml
```diff
@@ -344,6 +344,16 @@ models:
...
+    columns:
+      - name: ccao_stage_used
+        data_tests:
+          - accepted_values:
+              name: reporting_vw_top_5_muni_ccao_stage_used_certified_in_prior_years
+              values:
+                - certified
+              config:
+                where: >
+                  CAST(year AS int) < {{ var('data_test_iasworld_year_start') }}
```
jeancochrane, line 356: Clearer regression test for the problem we saw with null AVs in `reporting.vw_top_5_muni`.

dbt/models/ccao/ccao.vw_time_util.sql
```diff
@@ -0,0 +1,4 @@
+-- View that supplies dynamic datetime values in a way that we can mock for
+-- unit tests. See:
+-- https://discourse.getdbt.com/t/dynamic-dates-in-unit-tests/16883/2
+SELECT CURRENT_DATE AS date_today
```
jeancochrane, line 4: Kind of a bummer that dbt unit tests don't offer a simpler way of mocking dynamic values, but at least it works.

dbt/macros/pre_stage_filters.sql
```diff
@@ -1,21 +1,18 @@
...
+-- that query already filters rows where `cur = 'Y'`
+{% macro pre_mailed_filters(tablename) %}
     {{ tablename }}.procname is null
     and {{ tablename }}.cur = 'Y'
-    and not contains(stages.procnames, '{{ stage_name }}')
-{% endmacro %}
-
-{% macro pre_mailed_filters(tablename) %}
-    {{- pre_stage_filters(tablename, "CCAOVALUE") -}}
+    and cardinality(stages.procnames) = 0
```
dfsnow, line 10: **suggestion:** I would maybe add a brief comment here explaining what `cardinality()` is doing. I had to look it up and then take a moment to grok how it relates to `procnames`. Copying and pasting your PR comment below about the logic is fine IMO.

dbt/models/default/default.vw_pin_value.sql
dfsnow: Woof, tricky tricky. Nice catch.

dbt/models/default/schema/default.vw_pin_value.yml
dfsnow: Don't worry about it IMO.

dbt/models/default/schema/default.vw_pin_value.yml
```diff
@@ -226,12 +229,58 @@ unit_tests:
...
       - input: ref("ccao.aprval_reascd")
         rows:
           - {reascd: "28"}
+      - input: ref("ccao.vw_time_util")
+        rows:
+          - {date_today: "2024-01-01"}
     expect:
       rows:
         - {mailed_class: "211A"}
+  - name: default_vw_pin_value_stage_name_matches_procname
```
dfsnow, line 246: **praise:** This is such a good idea/use of the unit tests. Awesome work here.

dfsnow (approved):
Amazing work here @jeancochrane. These are definitely some tricky issues. Glad we'll have tests to catch them going forward.

dbt/macros/pre_stage_filters.sql
jeancochrane: Good call, done in 28b78f8048172ceca1626013d295f1a6cd7deb80.

PR #718, opened 2025-02-03
Fix error in Assessed Values open data asset test

wrridgeway:


wrridgeway:
@jeancochrane I'm not sure we can strip out the `year_field` logic quite yet - it was never defined for the permit asset, but I think it should have be there as `assessment_year`. We could change this if we'd like, but `tax_year` was an old nomenclature we moved away from. I'm not sure if this is the same case?

jeancochrane (approved):
Fine by me, I think the permits asset is an indication that this config interface is still useful!

PR #719, opened 2025-02-03
Fix missing years of filled location data

jeancochrane:
During model QC, we realized that a number of location features have unusual patterns of missingness over time (https://github.com/ccao-data/model-res-avm/issues/336).

I tested this PR with two new unit tests, one for `location.vw_pin10_location_fill` and one for `location.crosswalk_year_fill`, along with thorough QC of the two tables. In particular, I checked two things:

* The `year`, `{feature_name}_fill_year`, and `{feature_name}_data_year` columns in `location.crosswalk_year_fill` make sense for each feature
* The counts of missing values for each feature in `location.vw_pin10_location_fill` look reasonable when grouped by year
    * By "reasonable", I mean to say that the number of missing records corresponds to the number of parcels that do not have a shape for the fill year

Note that this PR only fixes the problem for `location` features; `proximity` features are filled in a similar way, but in a distinct set of queries. I'm planning to put up a fast-follow to this PR that makes the same change for the `proximity` queries.

### DIFF
--- a/dbt/models/location/location.crosswalk_year_fill.sql
+++ b/dbt/models/location/location.crosswalk_year_fill.sql
@@ -50,7 +50,7 @@ Take this hypothetical row as an example:
 | ---- | ---------------- | ---------------- |
 | 2019 | 2015             | 2017             |
 
-This row means that for a PIN in year 2010, a data consumer can get a filled
+This row means that for a PIN in year 2019, a data consumer can get a filled
 value for a `location.census` feature by joining to the representation of that
 PIN in the `location.census` table with year 2015, and that this particular row
 of data in `location.census` uses a value that was recorded in the raw data in
--- a/dbt/models/location/location.crosswalk_year_fill.sql
+++ b/dbt/models/location/location.crosswalk_year_fill.sql
@@ -107,15 +107,33 @@ WITH unfilled AS (
             {%- set outer_loop_last = loop.last %}
             {% for fieldname in fieldnames %}
                 {%- set inner_loop_last = loop.last %}
-                    {#-
-                        Use MAX to remove nulls in cases of years where some PINs
-                        have no match in the external data, since the DISTINCT
-                        subqueries in the list of joins below will still return
-                        those nulls
-                    -#}
-                MAX({{ tablename }}.year) AS {{ fieldname }}_fill_year,
+                {#
+                    Use MAX to remove nulls in cases of years where some PINs
+                    have no match in the external data, since the DISTINCT
+                    subqueries in the list of joins below will still return
+                    those nulls
+                #}
                 MAX({{ tablename }}.{{ fieldname }}_data_year)
-                    AS {{ fieldname }}_data_year
+                    AS {{ fieldname }}_data_year,
+                {#
+                    Only pull the year as the fill year for years where we also
+                    have a data year, since otherwise we risk pulling fill
+                    years for years that don't actually have data. This is
+                    a particular risk for tables that contain multiple
+                    data fields (i.e. `location.political`) because it's
+                    possible for one field to have data for a year while
+                    another field does not, in which case a naive `MAX(year)`
+                    call would pull a fill year for the first field even though
+                    data only exists for that year in the second field
+                #}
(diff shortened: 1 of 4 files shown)


dbt/models/location/location.vw_pin10_location_fill.sql
```diff
@@ -1,17 +1,5 @@
...
--- values for env_airport_noise_data_year won't match values for year
-WITH cyf_airport_noise AS (
-    SELECT
-        year,
-        CASE
-            WHEN env_airport_noise_data_year = 'omp' THEN year ELSE
-                env_airport_noise_data_year
-        END AS env_airport_noise_data_year
-    FROM {{ ref('location.crosswalk_year_fill') }}
-)
```
jeancochrane, line 13: We're not joining on `data_year` anymore, so this transformation only affects output columns and not join columns. As such, I moved it to a `CASE WHEN` statement in the `SELECT` clause of the query, which I think is simpler to read.

dbt/models/location/location.crosswalk_year_fill.sql
```diff
@@ -6,271 +6,234 @@ equivalent location data are filled thus:
...
+        "env_flood_fema",
+        "env_flood_fs",
+        "env_ohare_noise_contour",
+        "env_airport_noise"
+    ],
+    "school": ["school"],
+    "tax": ["tax"],
+    "access": ["access_cmap_walk"],
+    "other": ["misc_subdivision"]
+} -%}
```
jeancochrane, line 109: I noticed while editing this model that it contains a ton of boilerplate code, so I factored out all of the duplicated code using Jinja loops and this central data structure. Hopefully I haven't traded off clarity for brevity; if you think I have, I'm happy to revert back to the duplicative version of the code.

dbt/models/location/location.tax.sql
```diff
@@ -230,3 +230,4 @@ LEFT JOIN wide
             ELSE pcl.year
         END = wide.year
     )
+WHERE pcl.year >= (SELECT MIN(year) FROM wide)
```
jeancochrane, line 233: Previously we had been missing this filter, which meant that `location.tax` had a different universe of years than the rest of the inputs to `location.vw_pin10_location_fill` (specifically, it contained all parcel years, even for years before the first year of tax data). I don't think this should affect other queries that consume this particular table, since prior to this change, everything before the minimum year was totally empty anyway.

dbt/models/location/schema.yml
```diff
@@ -264,3 +264,162 @@ models:
   - name: location.vw_pin10_location_fill
     description: '{{ doc("view_vw_pin10_location_fill") }}'
     columns: *vw_location_columns
+
+unit_tests:
```
jeancochrane, line 268: I bet there are more tests that we could add here to be more defensive going forward, but I'm keeping them pretty lean for now in the interest of time.

dfsnow: This is a _perfect_ use-case for unit tests though, I gotta say. Very glad you added these.

dfsnow (approved):
@jeancochrane This is awesome, nice work. I built all the downstream views/tables in my dev env and alls looks good. My concerns over the ACS joins were wrong, I forgot we also do forward-filling in the ACS5 view.

The unit tests here definitely give good peace of mind, but IMO it's probably worth refactoring some of this complex filling logic to Python models at some point in the medium-term future.

dbt/models/location/location.crosswalk_year_fill.sql
```diff
@@ -6,271 +6,234 @@ equivalent location data are filled thus:
...
+3. `{feature_name}_data_year` is the original year in the raw data that
+   produced the value that exists in the fill year
+
+Take this hypothetical row as an example:
+
+| year | census_fill_year | census_data_year |
+| ---- | ---------------- | ---------------- |
+| 2019 | 2015             | 2017             |
+
+This row means that for a PIN in year 2010, a data consumer can get a filled
```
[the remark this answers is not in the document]
jeancochrane: Yup, good catch!

PR #726, opened 2025-02-04
Fix foreclosure script to work with values of 0

Damonamajor:
This fixes an issue discovered in [336](https://github.com/ccao-data/model-res-avm/issues/336), where PINs with 0 foreclosures within 1/2 a mile returned values of NA. This coalesces the result to 0 if there are 0 pins in that radius.

### DIFF
--- a/dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
+++ b/dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
@@ -67,11 +67,16 @@ SELECT
         0
     ) AS num_foreclosure_in_half_mile_past_5_years,
     COALESCE(pc.num_pin_in_half_mile, 1) AS num_pin_in_half_mile,
-    ROUND(
-        CAST(pib.num_foreclosure_in_half_mile_past_5_years AS DOUBLE) / (
-            CAST(pc.num_pin_in_half_mile AS DOUBLE) / 1000
-        ), 2
-    ) AS num_foreclosure_per_1000_pin_past_5_years,
+    CASE
+        WHEN
+            COALESCE(pib.num_foreclosure_in_half_mile_past_5_years, 0) = 0
+            THEN 0
+        ELSE ROUND(
+                CAST(pib.num_foreclosure_in_half_mile_past_5_years AS DOUBLE)
+                / (CAST(COALESCE(pc.num_pin_in_half_mile, 1) AS DOUBLE) / 1000),
+                2
+            )
+    END AS num_foreclosure_per_1000_pin_past_5_years,
     CONCAT(
         CAST(CAST(pl.year AS INT) - 5 AS VARCHAR),
         ' - ',
--- a/dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
+++ b/dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
@@ -67,16 +67,11 @@ SELECT
         0
     ) AS num_foreclosure_in_half_mile_past_5_years,
     COALESCE(pc.num_pin_in_half_mile, 1) AS num_pin_in_half_mile,
-    CASE
-        WHEN
-            COALESCE(pib.num_foreclosure_in_half_mile_past_5_years, 0) = 0
-            THEN 0
-        ELSE ROUND(
-                CAST(pib.num_foreclosure_in_half_mile_past_5_years AS DOUBLE)
-                / (CAST(COALESCE(pc.num_pin_in_half_mile, 1) AS DOUBLE) / 1000),
-                2
-            )
-    END AS num_foreclosure_per_1000_pin_past_5_years,
+    ROUND(
+        CAST(COALESCE(pib.num_foreclosure_in_half_mile_past_5_years, 0) AS DOUBLE)
+        / (CAST(pc.num_pin_in_half_mile AS DOUBLE) / 1000),
+        2
+    ) AS num_foreclosure_per_1000_pin_past_5_years,
     CONCAT(
         CAST(CAST(pl.year AS INT) - 5 AS VARCHAR),
         ' - ',
--- a/dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
+++ b/dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
@@ -68,7 +68,9 @@ SELECT
     ) AS num_foreclosure_in_half_mile_past_5_years,
     COALESCE(pc.num_pin_in_half_mile, 1) AS num_pin_in_half_mile,
     ROUND(
-        CAST(COALESCE(pib.num_foreclosure_in_half_mile_past_5_years, 0) AS DOUBLE)
+        CAST(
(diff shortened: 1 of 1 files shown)


Damonamajor:
```
query <- glue_sql(
  "SELECT * 
  from z_ci_724_modify_foreclosure_script_for_values_of_0_proximity.cnt_pin_num_foreclosure"
)

data_1 <- dbGetQuery(AWS_ATHENA_CONN_NOCTUA, query)

query <- glue_sql(
  "SELECT * 
  from proximity.cnt_pin_num_foreclosure"
)

data_2 <- dbGetQuery(AWS_ATHENA_CONN_NOCTUA, query)

data_3 <- data_1 %>%
  inner_join(data_2, by = c("pin10", "year")) %>%
  arrange(desc(year))
             
diff <- data_3 %>%
  filter(num_foreclosure_per_1000_pin_past_5_years.x != num_foreclosure_per_1000_pin_past_5_years.y |
         num_foreclosure_in_half_mile_past_5_years.x != num_foreclosure_in_half_mile_past_5_years.y)
```

Returns a value of 0, which excludes NA's by default.

dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
```diff
@@ -67,11 +67,16 @@ SELECT
...
+    CASE
+        WHEN
+            COALESCE(pib.num_foreclosure_in_half_mile_past_5_years, 0) = 0
+            THEN 0
+        ELSE ROUND(
+                CAST(pib.num_foreclosure_in_half_mile_past_5_years AS DOUBLE)
+                / (CAST(COALESCE(pc.num_pin_in_half_mile, 1) AS DOUBLE) / 1000),
+                2
+            )
+    END AS num_foreclosure_per_1000_pin_past_5_years,
```
jeancochrane, line 79: **[Suggestion, non-blocking]** Two thoughts about this, neither of them particularly important but I'm curious if you agree and think it's worth pushing another commit to address:

1. Do we need the `CASE WHEN` statement? If we have 0 in the numerator thanks to the new coalesce statement, I figure the division will always result in 0 whether we use `CASE WHEN` or not.
2. I wonder if it's better _not_ to coalesce `num_pin_in_half_mile` to 1 and instead add a data test to ensure that this field is never null? It seems better to me to fail loudly in the presence of a null than produce an incorrect value that we miss due to coalescing to 1. We could also add a data test to make sure that `num_foreclosure_in_half_mile_past_5_years` never ends up null again, too.

Here's the suggested diff for this block:

```suggestion
    ROUND(
        CAST(COALESCE(pib.num_foreclosure_in_half_mile_past_5_years, 0) AS DOUBLE)
        / (CAST(pc.num_pin_in_half_mile AS DOUBLE) / 1000),
        2
    ) AS num_foreclosure_per_1000_pin_past_5_years,
```

And here's what two data tests might look like in [`proximity/schema.yml`](https://github.com/ccao-data/data-architecture/blob/master/dbt/models/proximity/schema.yml):

```diff
--- a/dbt/models/proximity/schema.yml
+++ b/dbt/models/proximity/schema.yml
@@ -4,6 +4,15 @@ models:

   - name: proximity.cnt_pin_num_foreclosure
     description: '{{ doc("table_cnt_pin_num_foreclosure") }}'
+    columns:
+      - name: num_pin_in_half_mile
+        data_tests:
+          - not_null:
+              name: proximity_cnt_pin_num_foreclosure_num_pin_in_half_mile_not_null
+      - name: num_foreclosure_per_1000_pin_past_5_years
+        data_tests:
+          - not_null:
+              name: proximity_cnt_pin_num_foreclosure_num_foreclosure_per_1000_pin_past_5_years_not_null

   - name: proximity.cnt_pin_num_school
     description: '{{ doc("table_cnt_pin_num_school") }}'
```

jeancochrane (approved):
Awesome work! One optional change, but I don't feel strongly about it.

Query to prove that the new table does indeed replace nulls with 0s:

```sql
with dev as (
    select
        year,
        sum(
            case
                when num_foreclosure_per_1000_pin_past_5_years = 0 then 1 else 0
            end
        ) as null_count,
        sum(
            case
                when num_foreclosure_per_1000_pin_past_5_years != 0 then 1 else 0
            end
        ) as not_null_count
    from z_ci_724_modify_foreclosure_script_for_values_of_0_proximity.cnt_pin_num_foreclosure
    group by year
),

prod as (
    select
        year,
        sum(
            case
                when num_foreclosure_per_1000_pin_past_5_years is null then 1 else 0
            end
        ) as null_count,
        sum(
            case
                when num_foreclosure_per_1000_pin_past_5_years is not null then 1 else 0
            end
        ) as not_null_count
    from proximity.cnt_pin_num_foreclosure
    group by year
)

select
    dev.year,
    dev.null_count as dev_null_count,
    prod.null_count as prod_null_count,
    dev.not_null_count as dev_not_null_count,
    prod.not_null_count as prod_not_null_count
from dev
inner join prod
    on dev.year = prod.year
where dev.null_count != prod.null_count
    or dev.not_null_count != prod.not_null_count
order by year
```

This returns no results, which means we're good!

dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
Damonamajor: > Do we need the CASE WHEN statement? If we have 0 in the numerator thanks to the new coalesce statement, I figure the division will always result in 0 whether we use CASE WHEN or not.

Yup that makes sense, I just pushed it.

> I wonder if it's better not to coalesce num_pin_in_half_mile to 1 and instead add a data test to ensure that this field is never null? It seems better to me to fail loudly in the presence of a null than produce an incorrect value that we miss due to coalescing to 1. We could also add a data test to make sure that num_foreclosure_in_half_mile_past_5_years never ends up null again, too.

I thought about this for a second. I don't know if there are any pins which don't have any pins within a half mile (Ohare?). Would we want that value to return 0 or NA?

jeancochrane: > I had thought about this. I don't know if there are any pins which don't have any pins within a half mile (Ohare?). Would we want that value to return 0 or NA?

Hmm, good question. I see now that you were probably just following the pattern set by the line that defines the output value for `num_pin_in_half_mile`:

https://github.com/ccao-data/data-architecture/blob/fe8c98b9ee14014ed3483c1ea31bea0d38978741/dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql#L69

I also see we have 76 PINs in all years with a value of 1 for `num_pin_in_half_mile`, which are likely actually NA values that are getting coalesced to 1:

```sql
select count(*)
from proximity.cnt_pin_num_foreclosure
where num_pin_in_half_mile = 1
```
```
> 76
```

I guess by definition none of these PINs could possibly have a `num_foreclosure_in_half_mile_past_5_years` greater than 0, since they don't have any PINs at all in the nearest half mile. (A quick query confirms this is currently the case in practice.) So `COALESCE(num_pin_in_half_mile, 1)` makes sense in the denominator insofar as it avoids division-by-zero errors and will never return anything other than 0.

All of that to say, I think I'm on board with your original change to the denominator and I don't think we need to tweak it per my original comment!

PR #728, opened 2025-02-05
Fix missing years of filled proximity data

jeancochrane:
Companion PR to https://github.com/ccao-data/data-architecture/pull/719 that implements the same change to `proximity` models to fix missing years of data.

In order to reduce duplication, we also factor out the logic that #719 had defined in `location.crosswalk_year_fill` into a new `crosswalk_year_fill` macro that we use as a basis for both the `location` crosswalk and the `proximity` crosswalk.

### DIFF
--- a/dbt/macros/crosswalk_year_fill.sql
+++ b/dbt/macros/crosswalk_year_fill.sql
@@ -1,0 +1,202 @@
+/*
+Create a crosswalk for a set of features in a certain schema that a data
+consumer can use in order to back- and forward-fill values for missing years.
+
+We make two assumptions about input tables to this query:
+
+1. The input data should already be forward-filled; that is to say, there
+   should be a representation for every `year` starting from the earliest year
+   of available data (i.e. the `data_year`) and going up through the current
+   year
+    * For example, if the first `data_year` is 2017, we expect there to be
+      records for all years from 2017 to present
+2. The input data should _not_ be backfilled; that is to say, the earliest
+   `year` of data in the input table should correspond to the earliest
+   `data_year`
+    * For example, if the first `data_year` is 2017, we expect there to be
+      no records with `year` values prior to 2017
+    * We make an exemption for input tables that contain multiple variables
+      (like `location.political`), in which case we expect the earliest `year`
+      to correspond to the earliest `data_year` among _all_ of the variables.
+      For example, if the earliest `chicago_ward_data_year` is 2017, but there
+      exists a row with an `evanston_ward_data_year` of 2016, then we assume
+      that the earliest `year` in the view will be 2016, and any rows with this
+      `year` will have a null value for `chicago_ward_data_year`
+
+Taken together, these two assumptions mean that this query _only_ performs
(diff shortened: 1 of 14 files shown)


dbt/macros/crosswalk_year_fill.sql
```diff
@@ -0,0 +1,202 @@
+/*
```
jeancochrane, line 1: The comment and code in this macro is identical to the previous `location.crosswalk_year_fill` definition in [#719](https://github.com/ccao-data/data-architecture/pull/719/files#diff-4c39447a5466b05068a38c9b0cbda6eb4e973ea42af22ddc19ea9b389ed329a1), just with the addition of a macro interface that requires the caller to specify the `schema` string and the `fields` data object.

dbt/models/location/schema.yml
```diff
@@ -271,7 +271,7 @@ unit_tests:
     model: location.crosswalk_year_fill
     given:
       - input: source("spatial", "parcel")
-        rows: &crosswalk-location-fill-unit-test-pin-and-year
+        rows: &location-crosswalk-fill-year-unit-test-pin-and-year
```
jeancochrane, line 274: Slightly clearer anchor naming than what I implemented in https://github.com/ccao-data/data-architecture/pull/719.

dbt/macros/crosswalk_year_fill.sql
```diff
@@ -0,0 +1,202 @@
...
+Create a crosswalk for a set of features in a certain schema that a data
+consumer can use in order to back- and forward-fill values for missing years.
+
+We make two assumptions about input tables to this query:
+
+1. The input data should already be forward-filled; that is to say, there
```
Damonamajor, line 7: Would a check for this / increasing number of NULL values to be implemented somewhere?

dbt/models/proximity/proximity.cnt_pin_num_foreclosure.sql
```diff
@@ -72,11 +72,7 @@ SELECT
             CAST(pc.num_pin_in_half_mile AS DOUBLE) / 1000
         ), 2
     ) AS num_foreclosure_per_1000_pin_past_5_years,
-    CONCAT(
-        CAST(CAST(pl.year AS INT) - 5 AS VARCHAR),
-        ' - ',
-        CAST(pl.year AS VARCHAR)
-    ) AS num_foreclosure_data_year,
+    pl.year AS num_foreclosure_data_year,
```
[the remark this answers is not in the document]
dfsnow: No I agree, the range is implied by the column name and this is a good fix.

dfsnow (approved):
Excellent work here @jeancochrane. Love the macro abstraction and unit tests. Much cleaner and clearer!

dbt/macros/crosswalk_year_fill.sql
jeancochrane: That's a good idea, I'll put up a fast follow PR with data tests for the input tables.

PR #669, opened 2024-12-05
Materialize `model.vw_pin_shared_input` view to speed up daily dbt builds

jeancochrane:
This PR implements the easiest incremental step for for resolving our [sporadically failing daily dbt model builds](https://github.com/ccao-data/data-architecture/actions/runs/12068826927/job/33654774064): Materializing the `model.vw_pin_shared_input` view. This view is queried by two of the most expensive model queries in the daily build (`model.vw_pin_condo_input` and `model.vw_card_res_input`), one of which is the culprit for sporadic failures. By materializing the output of the view, we expect to reduce overall resource usage in the daily build.

Evidence that these changes work: https://github.com/ccao-data/data-architecture/actions/runs/12188851207/job/34002787866#step:8:56

PR #730, opened 2025-02-07
Tag `model.vw_pin_shared_input` for daily materialization

jeancochrane:
In https://github.com/ccao-data/data-architecture/pull/669 we changed `model.vw_pin_shared_input` to materialize as a table, but we neglected to tag it with the `daily` tag so that it would get rebuilt every day by the [`build-daily-dbt-models` workflow](https://github.com/ccao-data/data-architecture/actions/workflows/build_daily_dbt_models.yaml). This PR adds that missing tag.

Proof that the tag works:

```
(dbt) ~/data-architecture/dbt$ dbt list --select tag:daily
22:43:32  Running with dbt=1.9.1
22:43:33  Registered adapter: athena=1.8.4
ccao_data_athena.model.model.vw_card_res_input
ccao_data_athena.model.model.vw_pin_condo_input
ccao_data_athena.model.model.vw_pin_shared_input
...
```

### DIFF
--- a/dbt/models/model/schema.yml
+++ b/dbt/models/model/schema.yml
@@ -217,6 +217,9 @@ models:
 
   - name: model.vw_pin_shared_input
     description: '{{ doc("view_vw_pin_shared_input") }}'
+    config:
+      tags:
+        - daily
     columns:
       - &acs5_count_sex_total
         name: acs5_count_sex_total


dfsnow (approved):
Thank you much 👍

PR #731, opened 2025-02-10
Loosen PRB test threshold for `pre-mailed` assessment stage

wrridgeway:
Quick PR to suppress PRB test failures in `reporting.ratio_stats` for the pre-mailed stage. We've been getting a test failure for Cicero this year and after inspection the sales data looks fine - there's nothing to be corrected. Pre-mailed is a bit of an odd stage so for now we'll loosen the PRB test threshold for that stage to avoid constant failure alerts while maintaining the stricter alerts for other stages.

[Here's our quick and dirty sales investigation.](https://github.com/user-attachments/files/18734929/prb.zip)

### DIFF
--- a/dbt/models/reporting/schema.yml
+++ b/dbt/models/reporting/schema.yml
@@ -37,7 +37,10 @@ models:
           expression: |
             cod >= 0
             AND prd >= 0
-            AND prb BETWEEN -1 AND 1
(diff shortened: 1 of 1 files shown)


jeancochrane (approved):
Thanks for looking into this!

PR #733, opened 2025-02-11
Add assessable permits flag to `vw_pin_status` view

jeancochrane:
This PR adds a new column `has_recent_assessable_permit` to the `default.vw_pin_status` view. This column looks for assessable permits issued between January 1st of the year two years prior to the parcel year and December 31st of the parcel year, returning `TRUE` if any such permits exist and `FALSE` otherwise.

This PR doesn't yet add the new column to the `open_data.vw_parcel_status` view. Do we want to do that now, or wait till later @wrridgeway?

Connects https://github.com/ccao-data/model-res-avm/issues/340.

### DIFF
--- a/dbt/models/default/default.vw_pin_status.sql
+++ b/dbt/models/default/default.vw_pin_status.sql
@@ -142,7 +142,7 @@ SELECT
     pdat.note2 AS pardat_note,
     pdat.class = '999' AS is_filler_class,
     pdat.parid LIKE '%999%' AS is_filler_pin,
-    permit.num_assessable_permits > 0 AS has_assessable_permit_past_3_years
+    permit.num_assessable_permits > 0 AS has_recent_assessable_permit
 FROM {{ source('iasworld', 'pardat') }} AS pdat
 LEFT JOIN {{ source('spatial', 'corner') }} AS colo
     ON SUBSTR(pdat.parid, 1, 10) = colo.pin10
--- a/dbt/models/default/default.vw_pin_status.sql
+++ b/dbt/models/default/default.vw_pin_status.sql
@@ -89,12 +89,15 @@ WITH ahsap AS (
     GROUP BY par.parid, par.taxyr
 ),
 
+-- Get a table with the number of assessable permits in the past 3 years for
+-- all PINs in all years, which we use to power a "recent assessable permit"
+-- flag
 assessable_permits_past_3_years AS (
     SELECT
         pardat.parid AS pin,
         pardat.taxyr AS year,
         SUM(
-            CASE WHEN permit.year IS NULL THEN 0 ELSE 1 END
+            CASE WHEN assessable_permit.year IS NULL THEN 0 ELSE 1 END
         ) AS num_assessable_permits
     FROM {{ source('iasworld', 'pardat') }} AS pardat
     LEFT JOIN (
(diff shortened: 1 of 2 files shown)


dbt/models/default/default.vw_pin_status.sql
```diff
@@ -87,6 +87,29 @@ WITH ahsap AS (
...
+            SUBSTR(date_issued, 1, 4) AS year
+        FROM {{ ref('default.vw_pin_permit') }}
+        WHERE assessable = 'A'
+            AND date_issued IS NOT NULL
+        GROUP BY pin, SUBSTR(date_issued, 1, 4)
+    ) AS permit
+        ON pardat.parid = permit.pin
+        AND CAST(permit.year AS INT)
+        BETWEEN CAST(pardat.taxyr AS INT) - 2 AND CAST(pardat.taxyr AS INT)
+    GROUP BY pardat.parid, pardat.taxyr
```
jeancochrane, line 112: This query looks a bit gnarly, but I tested it against the current prod view and confirmed that it doesn't change the runtime or amount of queried data by very much.

dbt/models/default/default.vw_pin_status.sql
```diff
@@ -118,7 +141,8 @@ SELECT
     ddat.cdu_description AS dweldat_cdu_description,
     pdat.note2 AS pardat_note,
     pdat.class = '999' AS is_filler_class,
-    pdat.parid LIKE '%999%' AS is_filler_pin
+    pdat.parid LIKE '%999%' AS is_filler_pin,
+    permit.num_assessable_permits > 0 AS has_recent_assessable_permit
```
jeancochrane, line 145: I wonder if we want to preserver the number of assessable permits? I decided to lean in the direction of YAGNI. I'm already not totally convinced that this PIN status will be super useful outside the modeling code, but it will be easy to strip out in the future if that ends up being true.

dbt/models/default/schema/default.vw_pin_status.yml
```diff
@@ -13,6 +13,18 @@ models:
...
         description: CDU code description for `dweldat_cdu_description`
+      - name: has_recent_assessable_permit
+        description: |
+          Parcel has at least one assessable permit in the prior 3 years.
+
+          The date range that defines "the prior 3 years" starts on
+          January 1st of the year 2 years before the parcel year, and ends
+          on December 31st of the parcel year. For example, for a parcel
+          with year 2024, this column will count all assessable permits
+          issued between January 1st, 2022, and December 31st, 2024.
```
jeancochrane, line 24: Does this date range logic makes sense? It's perhaps not the most intuitive logic, since a data consumer might expect "prior 3 years" to mean exactly three years before a specific date, but it makes the most sense in the context of desk review.

dbt/models/default/default.vw_pin_status.sql
```diff
@@ -87,6 +87,29 @@ WITH ahsap AS (
...
+        ) AS num_assessable_permits
+    FROM {{ source('iasworld', 'pardat') }} AS pardat
+    LEFT JOIN (
+        SELECT
+            pin,
+            SUBSTR(date_issued, 1, 4) AS year
+        FROM {{ ref('default.vw_pin_permit') }}
+        WHERE assessable = 'A'
+            AND date_issued IS NOT NULL
+        GROUP BY pin, SUBSTR(date_issued, 1, 4)
```
dfsnow, line 107: **question (blocking):** What is the `GROUP BY` doing here? Wouldn't you want there to be one row per permit, such that the `SUM()` in the outer query gets the count of joined rows?

dbt/models/default/schema/default.vw_pin_status.yml
dfsnow: Yes! Seems reasonable to me.

dbt/models/default/default.vw_pin_status.sql
jeancochrane: Oops, good catch! Fixed in b39fcf066744264ef2dff55a7c36914380347249.

jeancochrane:
Down for one more look @dfsnow?

dfsnow (approved):
Looks great now!

PR #734, opened 2025-02-18
Add 2025 final models to `model.final_model` table

jeancochrane:
This PR updates the seed that powers the `model.final_model` table to include metadata for our 2025 final models.

Query my dev view to inspect the new rows in a friendlier table format:

```sql
select *
from z_dev_jecochr_model.final_model
where year = '2025'
```

Some important assumptions I made that are worth double-checking:

* "Best performance" is a fine enough `reason_chosen` for `reason_chosen`, given that we didn't have more than one candidate per assessment group this year
* No `reference_run_id` is necessary for either model
* The `note` column should correspond to the `run_note` for the final model

Let me know if any of these assumptions are incorrect and I can fix them!


dfsnow (approved):
Perfect!

PR #735, opened 2025-02-19
Remove township as a choice for socrata chunking

wrridgeway:
This PR removes chunking by township as an option for socrata uploads. Now that the script automatically chunks any group larger than 10,000 rows being able to dictate which years we'd like to upload is more than enough user choice.

I'm going to have to add functionality for handling multiple assets in one go in order to automate socrata uploads in a future PR, and simplifying this script as much as possible in preparation for that will save some time and energy.

I tested all the possible configurations of overwriting and year values, and all of the jobs ran as expected:

image
https://github.com/ccao-data/data-architecture/actions/runs/13421072113/job/37493591749

image
https://github.com/ccao-data/data-architecture/actions/runs/13421200605/job/37494008961

image
https://github.com/ccao-data/data-architecture/actions/runs/13421301622/job/37494345438

image
https://github.com/ccao-data/data-architecture/actions/runs/13422411118/job/37497958718

image
https://github.com/ccao-data/data-architecture/actions/runs/13422509341/job/37498265818

image
https://github.com/ccao-data/data-architecture/actions/runs/13423091271/job/37500179006

### DIFF
--- a/dbt/models/open_data/exposures.yml
+++ b/dbt/models/open_data/exposures.yml
@@ -199,7 +199,7 @@ exposures:
       name: Data Department
     meta:
       test_row_count: true
-      asset_id: vgzx-68gb
+      asset_id: a8dq-ppc7
       primary_key:
         - pin
         - year
--- a/dbt/models/open_data/exposures.yml
+++ b/dbt/models/open_data/exposures.yml
@@ -199,7 +199,7 @@ exposures:
       name: Data Department
     meta:
       test_row_count: true
-      asset_id: a8dq-ppc7
+      asset_id: vgzx-68gb
       primary_key:
         - pin
         - year
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -85,7 +85,7 @@ def get_asset_info(socrata_asset):
     return athena_asset, asset_id
 
 
-def build_query(athena_asset, asset_id, years=None, township=None):
+def build_query(athena_asset, asset_id, years=None):
     """
     Build an Athena compatible SQL query. Function will append a year
     conditional if `years` is non-empty. Many of the CCAO's open data assets are
@@ -144,18 +144,13 @@ def build_query(athena_asset, asset_id, years=None, township=None):
     if not years:
         query = query
 
-    elif years is not None and not township:
+    elif years is not None:
         query += " WHERE year = %(year)s"
-
-    elif years is not None and township is not None:
-        query += " WHERE year = %(year)s" + " AND township_code = %(township)s"
 
     return query
 
 
-def upload(
-    method, asset_id, sql_query, overwrite, count, year=None, township=None
-):
+def upload(method, asset_id, sql_query, overwrite, count, year=None):
     """
     Function to perform the upload to Socrata. `puts` or `posts` depending on
     user's choice to overwrite existing data.
@@ -171,21 +166,10 @@ def upload(
     if not year:
         query_conditionals = {}
         print_message = print_message + " all years for asset " + asset_id
-    if year is not None and not township:
+    if year is not None:
         query_conditionals = {"year": year}
         print_message = (
             print_message + " year: " + year + " for asset " + asset_id
-        )
-    if year is not None and township is not None:
-        query_conditionals = {"year": year, "township": township}
-        print_message = (
-            print_message
-            + " township: "
-            + township
-            + ", year: "
-            + year
-            + " for asset "
-            + asset_id
         )
 
     # We grab the data before uploading it so we can make sure timestamps are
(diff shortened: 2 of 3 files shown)


socrata/socrata_upload.py
```diff
@@ -314,52 +281,33 @@ def socrata_upload(
...
+                "sql_query": sql_query,
+                "overwrite": overwrite,
+                "count": count,
+                "year": item,
+            }
             if count == 0 and overwrite:
                 upload("put", **upload_args)
             else:
                 upload("post", **upload_args)
+            count += 1
```
wrridgeway, line 303: I don't think overwriting + multiple groups was working properly before. This should make sure we don't overwrite the group that was just written when we set `overwrite = true`

dbt/models/open_data/exposures.yml
```diff
@@ -199,7 +199,7 @@ exposures:
       name: Data Department
     meta:
       test_row_count: true
-      asset_id: vgzx-68gb
+      asset_id: a8dq-ppc7
```
wrridgeway, line 202: This needs to be reverted before merging. But I'm leaving it now in case whoever reviews this would like to test the workflow. Just make sure you select `Property Tax-Exempt Parcels`!

socrata/socrata_upload.py
```diff
@@ -221,56 +220,33 @@ def upload(
...
-            flag = None
         else:
             groups = years
-            flag = "years"
 
-    return flag, groups
+    else:
+        groups = None
+
+    return groups
```
wrridgeway, line 246: `flag` is no longer needed since we only have one variable that can define groups.

socrata/socrata_upload.py
```diff
@@ -221,56 +220,33 @@ def upload(
...
-
-    else:
-        if not years:
-            groups = None
-            flag = None
         else:
             groups = years
-            flag = "years"
 
-    return flag, groups
```
wrridgeway, line 268: `flag` is no longer needed since we only have one variable that can define groups.

socrata/socrata_upload.py
```diff
@@ -314,52 +281,33 @@ def socrata_upload(
...
-            )
-
-        if flag == "both":
-            sql_query = build_query(
-                athena_asset=athena_asset,
-                asset_id=asset_id,
-                years=years,
-                township=by_township,
-            )
+    elif groups is not None:
```
jeancochrane, line 284: **[Suggestion, non-blocking]** It makes me a little nervous that there's no `else` block here, since at some point in the future we might accidentally introduce a code path by which `years` could be not-null but `groups` could be null, in which case neither of the upload conditional blocks would trigger and the workflow wouldn't do any uploads. If we decide to switch from `groups` to something like `years_list` using a renamed `parse_years_list()` function, it might make sense to have the two conditional branch checks be `if not years_list` and `else`. Does that make sense?

socrata/socrata_upload.py
```diff
@@ -221,56 +220,33 @@ def upload(
         print(response.content)
 
 
-def generate_groups(athena_asset, years=None, by_township=False):
+def generate_groups(athena_asset, years=None):
```
jeancochrane, line 223: **[Question, non-blocking]** It seems like this function is now just generating the set of years to iterate. Maybe we rename it something like `parse_years_list()` instead, for clarity? I think that might also help for readability in the `socrata_upload()` function.

socrata/socrata_upload.py
jeancochrane: **[Question, non-blocking]** Doesn't the `upload` function handle the `count` increment? Or is the problem that the `upload()` function won't reliably reach this line in all cases?

https://github.com/ccao-data/data-architecture/blob/cfb4d880734a9428d5f653d63e09100f2ef59f05/socrata/socrata_upload.py#L219

jeancochrane (approved):
Much simpler! I have a few questions and thoughts below, but nothing serious.

socrata/socrata_upload.py
wrridgeway: It handles it, but only inside that function. I don't think count ever increased outside that function, so every time `upload` got done chunking the group it was passed, it would get a new `count` value of 0 again.

socrata/socrata_upload.py
wrridgeway: Completely agree.

socrata/socrata_upload.py
wrridgeway: Alright, I went with a solution where the updated count it passed back in order to ensure it persists. That way we always know exactly how and where that variable is being changed.

jeancochrane (approved):
Nice refactor, thanks!

PR #738, opened 2025-02-20
Align open data db transformations with socrata

wrridgeway:
While investigating every open data asset for parity with the open data db, I came upon two columns that need to be but are not currently being transformed. These were the only two I could find.

```
select char_attic_fnsh, char_gar1_area, count(*) as count
from z_ci_align_open_data_db_columns_with_socrata_open_data.vw_sf_mf_improvement_char
group by char_attic_fnsh, char_gar1_area
```

char_attic_fnsh | char_gar1_area | count
-- | -- | --
Living Area | No | 3134847
  | No | 3472
Living Area | Yes | 179158
Partial | Yes | 1387
Partial | No | 181195
Partial |   | 99
  | Yes | 1609
Living Area | | 954
None |   | 4910
None | Yes | 2048661
  |   | 64803
None | No | 23931299

### DIFF
--- a/dbt/models/open_data/open_data.vw_sf_mf_improvement_char.sql
+++ b/dbt/models/open_data/open_data.vw_sf_mf_improvement_char.sql
@@ -50,17 +50,17 @@ SELECT
         WHEN char_apts = '6' THEN 'None'
     END AS char_apts,
     CASE
-        WHEN char_attic_fnsh = "1" THEN "Living Area"
-        WHEN char_attic_fnsh = "2" THEN "Partial"
-        WHEN char_attic_fnsh = "3" THEN "None"
+        WHEN char_attic_fnsh = '1' THEN 'Living Area'
+        WHEN char_attic_fnsh = '2' THEN 'Partial'
+        WHEN char_attic_fnsh = '3' THEN 'None'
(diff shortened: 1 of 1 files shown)


dbt/models/open_data/open_data.vw_sf_mf_improvement_char.sql
```diff
@@ -122,11 +129,6 @@ SELECT
         WHEN char_site = '3' THEN 'Detracts From Value'
     END AS char_site,
     char_ncu,
-    CASE
-        WHEN char_renovation = '1' THEN 'Yes'
-        WHEN char_renovation = '2' THEN 'No'
-    END AS char_renovation,
-    char_recent_renovation AS recent_renovation,
```
jeancochrane, line 129: **[Question, non-blocking]** Is this removal intentional?

jeancochrane (approved):
Nice catch!

dbt/models/open_data/open_data.vw_sf_mf_improvement_char.sql
wrridgeway: It is, but good eye! We have never published this column since it's built rather than taken from iasworld.

PR #739, opened 2025-02-20
Revert char_renovation change

wrridgeway:
@jeancochrane you were so close to saving me from myself. I did mean to get rid of `char_recent_renovation` in the previous pull request, but not `char_renovation`.


jeancochrane (approved):
Glad we caught it fast!

PR #740, opened 2025-02-21
Add new features to ratio stats

Damonamajor:
Adds med_ratio_met column referencing the new function in assesspy
Adds a  "Vertical Equity met" column; if either PRB or PRD is met, then true, else false.
Adds a column of sales_excluded_n

### DIFF
--- a/dbt/models/reporting/reporting.ratio_stats.py
+++ b/dbt/models/reporting/reporting.ratio_stats.py
@@ -18,8 +18,8 @@ CCAO_MIN_SAMPLE_SIZE = 20.0
 # to determine the final column order
 SPARK_SCHEMA = (
     "year string, triad string, geography_type string, property_group string, "
-    "assessment_stage string, geography_id string, sale_year string, sale_n bigint, "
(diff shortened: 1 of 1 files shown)


Damonamajor:
Some tests:

select distinct(prb_met, prd_met, vertical_equity_met)
from z_dev_damajor_reporting.ratio_stats

image


SELECT *
FROM z_dev_damajor_reporting.ratio_stats
WHERE (sale_n - prd_n) <> sales_removed
   OR (sale_n - prb_n) <> sales_removed
   OR (sale_n - mki_n) <> sales_removed
   OR (sale_n - cod_n) <> sales_removed

returns 0 results

jeancochrane (approved):
I love how simple this was! Once we've confirmed that the new output hasn't changed substantially from the old output for PRB/PRD, we're good to merge from my perspective. I'll let @wrridgeway weigh in first as well.

wrridgeway (approved):
Awesome work. If you want to investigate floating point oddness in another PR that's fine with me.

PR #744, opened 2025-02-26
Pull prior year AVs to match prior year MVs in prior year val IC QC view

jeancochrane:
This PR fixes a small bug that Valuations noticed in the QC view that IC uses for comparing prior year values. The view should return prior year AVs and MVs for all parcels in the current year, but it's returning current year AVs instead. The source of the bug is that the query currently joins `pardat` to `aprval` on `aprval.taxyr + 1 = pardat.taxyr` in order to get prior year MVs for parcels in the current year, but then it joins `asmt_all` to `pardat` using `taxyr`, which returns current year AVs. By joining `asmt_all` to `aprval` using `taxyr` instead, we return prior year AVs to match the prior year MVs.

Valuations has confirmed that this fix works on their end.


dfsnow (approved):
Tricky tricky. Nice catch.

PR #745, opened 2025-02-26
Simplify Socrata API script and allow multiple assets to be updated

wrridgeway:
This started as trying to allow this script to accept multiple assets, but working with it was a chore and it was clear it could be simplified and improved.

This PR really just shifts a lot of the looping and year handling to the query building and upload functions and removes a lot of `overwrite` parameter handling involving a `count` variable by only allowing `overwrite` to be true the first time it's handled.

### DIFF
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -6,19 +6,9 @@ on:
       socrata_asset:
         type: string
         description: >
-          Comma-separated Socrata assets to update. Options are:
-          Appeals |
-          Assessed Values |
-          Parcel Addresses |
-          Parcel Proximity |
-          Parcel Sales |
-          Parcel Status |
-          Permits |
-          Property Tax-Exempt Parcels |
-          Parcel Universe (Historic) |
-          Parcel Universe (Current Year) |
-          Residential Condominium Unit Characteristics |
-          Single and Multi-Family Improvement Characteristics
+          Comma-separated Socrata assets to update. Run
+          `dbt list --resource-types exposure --output json --output-keys label`
+          to see options.
         required: true
       years:
         # Comma separated list of years
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -25,8 +25,8 @@ on:
         type: string
         description: >
           Years to update or overwrite (comma-separated). Leaving this field
-          blank will upload an entire asset in one piece. Enter 'all' to upload
-          all years individually.
+          blank will upload an entire asset in one piece. Enter 'all' to chunk
+          the upload by year for all years.
         required: false
       overwrite:
         # True for overwrite, False for update
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -6,19 +6,19 @@ on:
       socrata_asset:
         type: string
         description: |
-          Target Socrata asset. Options are:
-          - Appeals
-          - Assessed Values
-          - Parcel Addresses
-          - Parcel Proximity
-          - Parcel Sales
-          - Parcel Status
-          - Permits
-          - Property Tax-Exempt Parcels
-          - Parcel Universe (Historic)
-          - Parcel Universe (Current Year)
-          - Residential Condominium Unit Characteristics
-          - Single and Multi-Family Improvement Characteristics
+          Target Socrata asset. Options are:\n
+          * Appeals\n
+          * Assessed Values\n
+          * Parcel Addresses\n
+          * Parcel Proximity\n
+          * Parcel Sales\n
+          * Parcel Status\n
+          * Permits\n
+          * Property Tax-Exempt Parcels\n
+          * Parcel Universe (Historic)\n
+          * Parcel Universe (Current Year)\n
+          * Residential Condominium Unit Characteristics\n
+          * Single and Multi-Family Improvement Characteristics
         required: true
       overwrite:
         # True for overwrite, False for update
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -5,19 +5,19 @@ on:
     inputs:
       socrata_asset:
         type: string
-        description: |
-          Target Socrata asset. Options are:
-          Appeals
-          Assessed Values
-          Parcel Addresses
-          Parcel Proximity
-          Parcel Sales
-          Parcel Status
-          Permits
-          Property Tax-Exempt Parcels
-          Parcel Universe (Historic)
-          Parcel Universe (Current Year)
-          Residential Condominium Unit Characteristics
+        description: >
+          Comma-separated Socrata assets to update. Options are:
+          Appeals |
+          Assessed Values |
+          Parcel Addresses |
+          Parcel Proximity |
+          Parcel Sales |
+          Parcel Status |
+          Permits |
+          Property Tax-Exempt Parcels |
+          Parcel Universe (Historic) |
+          Parcel Universe (Current Year) |
+          Residential Condominium Unit Characteristics |
           Single and Multi-Family Improvement Characteristics
         required: true
       overwrite:
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -6,9 +6,9 @@ on:
       socrata_asset:
         type: string
         description: >
-          Comma-separated Socrata assets to update. Run
-          `dbt list --resource-types exposure --output json --output-keys label`
-          to see options.
+          Comma-separated list of names of Socrata assets to update.
+          Asset names are the same as the asset title in Socrata (or dbt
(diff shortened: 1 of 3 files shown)


socrata/socrata_upload.py
```diff
@@ -88,6 +90,9 @@ def get_asset_info(socrata_asset):
     Simple helper function to retrieve asset-specific information from dbt.
     """
 
+    if not os.path.isdir("./dbt"):
```
wrridgeway, line 93: I can remove this, but it makes working on it interactively easier.

socrata/socrata_upload.py
```diff
@@ -202,52 +208,39 @@ def upload(method, asset_id, sql_query, overwrite, count, year=None):
...
-    session.get(
-        (
-            "https://datacatalog.cookcountyil.gov/resource/"
-            + asset_id
-            + ".json?$limit=1"
-        ),
-        headers={"X-App-Token": app_token},
-    ).raise_for_status()
-
-    session.get(url=url, headers={"X-App-Token": app_token}).raise_for_status()
```
wrridgeway, line 233: We were checking for a bad URL twice.

socrata/socrata_upload.py
```diff
@@ -202,52 +208,39 @@ def upload(method, asset_id, sql_query, overwrite, count, year=None):
...
+                data=input_data.iloc[i : i + 10000].to_json(orient="records"),
+                headers={"X-App-Token": app_token},
+            )
+            overwrite = False
+            print(response.content)
 
-    # Return the updated count so that if this function is called in a loop
-    # the updated count persists.
-    return count
+        overwrite = False
```
wrridgeway, line 243: Ensuring overwrite can never be `true` more than once.

socrata/socrata_upload.py
```diff
@@ -128,8 +133,7 @@ def get_asset_info(socrata_asset):
 
 def build_query(athena_asset, asset_id, years=None):
     """
-    Build an Athena compatible SQL query. Function will append a year
-    conditional if `years` is non-empty. Many of the CCAO's open data assets are
+    Build a dictionary of Athena compatible SQL queries and their associated years. Many of the CCAO's open data assets are
```
wrridgeway, line 136: This is the meat of what's changing here.

.github/workflows/socrata_upload.yaml
```diff
@@ -4,22 +4,21 @@ on:
...
+          Parcel Addresses |
+          Parcel Proximity |
+          Parcel Sales |
+          Parcel Status |
+          Permits |
+          Property Tax-Exempt Parcels |
+          Parcel Universe (Historic) |
+          Parcel Universe (Current Year) |
+          Residential Condominium Unit Characteristics |
+          Single and Multi-Family Improvement Characteristics
```
wrridgeway, line 21: This is gross, but I can't think of a better way to inform users of their choices since multiple choice isn't an input type.

socrata/socrata_upload.py
```diff
@@ -88,6 +101,9 @@ def get_asset_info(socrata_asset):
     Simple helper function to retrieve asset-specific information from dbt.
     """
 
+    if not os.path.isdir("./dbt"):
```
wrridgeway, line 104: Only helpful locally, unecessary for github deployment.

.github/workflows/socrata_upload.yaml
```diff
@@ -4,34 +4,35 @@ on:
...
       overwrite:
         # True for overwrite, False for update
         type: boolean
         description: Overwrite socrata asset
         required: true
-      years:
-        # Comma separated list of years
-        type: string
-        description: Years to update or overwrite (comma-separated)
-        default: 'all'
```
wrridgeway, line 33: Removed the default and explained user's choices.

socrata/socrata_upload.py
```diff
@@ -31,11 +31,24 @@
...
-def parse_years(years):
+def parse_assets(assets):
+    """
+    Make sure the asset environmental variable is formatted correctly.
+    """
+
+    assets = str(assets).split(",")
+    assets = [i.strip() for i in assets]
+
+    return assets
```
jeancochrane, line 42: **[Nitpick, non-blocking]** We could do this in one line, might be more readable? Also, this is a personal preference thing, but when I'm looping through an iterator in Python I usually prefer to assign each object in the iterator to a variable name that's descriptive of what the object represents, e.g. `asset` in this case. `i` or `idx` makes sense when we're enumerating an object and looping through its indices (in an expression like `for i, asset in enumerate(assets)` or `for i in range(n)`), but since a `for` loop in Python returns objects and not indices by default, I find it more descriptive to hint at what those objects mean.

```suggestion
    return [asset.strip() for asset in str(assets).split(",")]
```

socrata/socrata_upload.py
```diff
@@ -36,6 +36,8 @@ def parse_years(years):
     Make sure the years environmental variable is formatted correctly.
     """
 
+    if years == "":
```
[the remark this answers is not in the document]
jeancochrane: **[Praise]** I like the defensiveness!

socrata/socrata_upload.py
jeancochrane: **[Suggestion, non-blocking]** I think this is knowledge that it would make sense to preserve for the future:

```suggestion
    # When running locally, we will probably be inside the socrata/ dir, so
    # switch back out to find the dbt/ dir
    if not os.path.isdir("./dbt"):
```

socrata/socrata_upload.py
```diff
@@ -128,9 +144,9 @@ def get_asset_info(socrata_asset):
 
 def build_query(athena_asset, asset_id, years=None):
```
jeancochrane, line 145: **[Suggestion, non-blocking]** Since this now returns a dict rather than a query, how about we rename it to better hint at its return type? Something like `build_query_dict`?

socrata/socrata_upload.py
```diff
@@ -177,21 +193,23 @@ def build_query(athena_asset, asset_id, years=None):
...
+    # Build a dictionary with queries for each year requested, or no years
     if not years:
-        query = query
+        query = {None: query}
 
-    elif years is not None:
-        query += " WHERE year = %(year)s"
+    else:
+        query = [query + " WHERE year = '" + year + "'" for year in years]
+        query = dict([(k, v) for k, v in zip(years, query)])
```
jeancochrane, line 207: **[Suggestion, non-blocking]** A couple things I find slightly confusing about this block:

1. We're mutating the `query` variable continuously, which makes it unclear what each transformation step does and could potentially make it confusing to step through the code using logging or a breakpoint
2. I admire the clever list comprehensions! But I think the combination of the first list comprehension and the second using `zip()` is a bit hard to get your head around quickly. I would prefer slightly more verbose code for the sake of readability

Here's how I would suggest we resolve both nits:

```python
if not years:
    query_dict = {None: query}
else:
    query_dict = {year: f"{query} WHERE year = '{year}'" for year in years}

return query_dict
```

Also, keep in mind that templating user input into SQL strings like this is a common cause of [SQL injection](https://en.wikipedia.org/wiki/SQL_injection) vulnerabilities. I think it's OK in this case, since we don't allow external code authors to run workflows, but I would still recommend we test an attack input for `year` like `2004'; select 1;` to make sure we're properly escaping user input in our queries. Let me know if you want help with this and we can pair on it together.

socrata/socrata_upload.py
```diff
@@ -264,54 +269,26 @@ def socrata_upload(socrata_asset, overwrite=False, years=None):
...
-    socrata_asset=os.getenv("SOCRATA_ASSET"),
-    overwrite=check_overwrite(os.getenv("OVERWRITE")),
-    years=parse_years(os.getenv("YEARS")),
-)
+for asset in all_assets:
+    socrata_upload(
+        socrata_asset=asset,
+        overwrite=check_overwrite(os.getenv("OVERWRITE")),
+        years=parse_years(os.getenv("YEARS")),
+    )
```
jeancochrane, line 294: **[Nitpick, non-blocking]** I probably should have noticed this before, but if we're planning to run this module as a Python script, let's add the Pythonic [`if __name__ == "__main__"` Python script entrypoint](https://stackoverflow.com/a/419185):

```python
if __name__ == "__main__":
    # Retrieve asset(s)
    all_assets = parse_assets(os.getenv("SOCRATA_ASSET"))

    ...
```

jeancochrane (approved):
Nice cleanup!

socrata/socrata_upload.py
wrridgeway: I didn't know I could build a dict this way...

wrridgeway: Do you think [this](https://github.com/ccao-data/data-architecture/pull/745/commits/5e9e2874eb949559c62852e9f6ebaaf5b7ad6ebe#) is enough to anticipate any bad behavior? Ignore the stray interactive cell.

jeancochrane: Yup, I think that's great, thanks!

PR #746, opened 2025-02-27
Add year column to permit views

wrridgeway:
This new column allows the the socrata api script to treat the permit view the same as every other view. There are unfortunately three null values in the `year` column, which we can usually assume is complete. This feels like something we should ask the permit team to address rather than scripting out a workaround.

### DIFF
--- a/dbt/models/default/default.vw_pin_permit.sql
+++ b/dbt/models/default/default.vw_pin_permit.sql
@@ -25,7 +25,7 @@ permit_addresses AS (
 
 SELECT
     permit.parid AS pin,
-    YEAR(CAST(permit.permdt AS TIMESTAMP)) AS year,
+    SUBSTR(permit.permdt, 1, 4) AS year,
     permit.num AS permit_number,
     permit.user28 AS local_permit_number,
     permit.permdt AS date_issued,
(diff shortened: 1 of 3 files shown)


dbt/models/default/default.vw_pin_permit.sql
```diff
@@ -25,6 +25,7 @@ permit_addresses AS (
 
 SELECT
     permit.parid AS pin,
+    SUBSTR(permit.permdt, 1, 4) AS year,
```
wrridgeway, line 28: Using `SUBSRT()` rather than converting to a timestamp and then using `YEAR()` because we want a varchar as the result.

jeancochrane (changes requested):
Do you mind documenting this field real quick?

wrridgeway:
🤦 Done.

dbt/models/default/default.vw_pin_permit.sql
jeancochrane: **[Nitpick, non-blocking]** Before we merge, let's double-check how this works for nulls. I tried `select substr(null, 1, 4)`, which throws an error, so I'm worried it may not work as we expect. I would expect `year` to be null in cases where `date_issued` is null, so if that's true, then we're good to go.

jeancochrane (approved):
Thanks!

wrridgeway:
```
select year, date_issued
from z_ci_standardize_permit_year_column_for_open_data_default.vw_pin_permit
where date_issued is null
```

is this what you had in mind?

image

jeancochrane:
Yup, that's it! My bad for forgetting I could just query the staging table myself 🤦🏻‍♀️

PR #749, opened 2025-02-28
Move permit `date_issued` not-null test to data integrity test

jeancochrane:
There was a small mistake in https://github.com/ccao-data/data-architecture/pull/747: By defining the test on `iasworld.permit`, it inherited the `data_test_iasworld` tag, and so got included in our [nightly iasworld tests](https://github.com/ccao-data/data-architecture/actions/runs/13582846933/job/37971819096#step:7:499) rather than our weekly data integrity tests. This PR moves it to `default.vw_pin_permit`, which will move it to our weekly data integrity tests and better reflect the purpose of the test.

Proof that this works:

```console
$ dbt list --selector select_data_test_non_iasworld | grep date_issued_not_null
ccao_data_athena.default.schema.default_vw_pin_permit_date_issued_not_null
$ dbt list --selector select_data_test_iasworld | grep date_issued_not_null
$
```

### DIFF
--- a/dbt/models/default/schema/default.vw_pin_permit.yml
+++ b/dbt/models/default/schema/default.vw_pin_permit.yml
@@ -25,6 +25,11 @@ models:
       - name: date_issued
         description: >
           Date that the municipality issued the permit to the applicant
+        data_tests:
+          - not_null:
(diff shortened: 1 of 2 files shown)


wrridgeway (approved):
Thanks for catching this

PR #755, opened 2025-03-05
Automate socata api updates

wrridgeway:
This PR adds a schedule event to trigger the `socrata_upload` workflow. It updates behavior for no inputs to accommodate this new event.


jeancochrane:
Ah right, I forgot about that limitation @wrridgeway 😅 What do you think about deploying with the test asset IDs + a schedule that will trigger immediately after merging, waiting for the schedule to trigger to test to make sure it works, and then putting up a fast-follow PR to revert the test asset IDs back to prod and set the schedule to its permanent cadence? Not the most elegant deployment plan, but as long as we're fine being unable to push to the prod assets for however long the testing takes, I don't think it'll be a big deal.

wrridgeway:
That sounds good to me. I'm not sure it'll go quite that smoothly but i'll have a fast-follow PR ready to merge in regardless.

socrata/socrata_upload.py
```diff
@@ -32,12 +32,54 @@
...
+            DBT.invoke(dbt_list_args)
+
+        model = [
+            json.loads(model_dict_str)
+            for model_dict_str in dbt_output.getvalue().split("\n")
+            # Filter out empty strings caused by trailing newlines
+            if model_dict_str
+        ]
+
+        os.chdir("..")
```
jeancochrane, line 75: **[Suggestion, non-blocking]** Seems like this code mostly shares the same logic as the similar `dbt list` operation in `get_asset_info()`. Since `dbt list` is a bit on the slower side (usually takes a couple seconds to return), we could speed things up slightly and reduce duplication by consolidating the two calls. I think it might make sense to tweak `parse_assets()` so that it returns a list of asset objects that each contains all the info we pull from `get_asset_info()`, in which case `parse_assets()` would be the only function responsible for retreiving asset metadata. Then, we could pass each asset object returned by `parse_assets()` through to `socrata_upload()` so that `socrata_upload()` doesn't need to call `get_asset_info()` again. This would also have the nice benefit of reducing runtime in cases where we are uploading many assets, since the current code will run a `dbt list` call for each asset even though we only really need one call to get the full metadata object.

My proposed changes represent a medium-sized refactor compared to the rest of this PR, though, so I'm comfortable merging it in with the duplication and moving towards consolidation later if we find the extra `dbt list` call to be especially slow or annoying.

.github/workflows/socrata_upload.yaml
```diff
@@ -83,4 +88,5 @@ jobs:
           SOCRATA_ASSET: ${{ inputs.socrata_asset }}
           OVERWRITE: ${{ inputs.overwrite }}
           YEARS: ${{ inputs.years }}
+          WORKFLOW: ${{ github.event_name }}
```
jeancochrane, line 91: **[Nitpick, non-blocking]** `WORKFLOW` feels a bit ambiguous to me as a variable name; if I didn't already know what was going on here, I might assume it referred to the name of the workflow rather than the name of the event that triggered the workflow. What do you think about the more verbose `WORKFLOW_EVENT_NAME`?

jeancochrane (changes requested):
Looking good! Only one blocking change (I think we may be referencing the `$OVERWRITE` variable where we instead want to reference the `$WORKFLOW` variable in `parse_years_list`). Once we resolve that issue, feel free to choose whether or not to act on my other changes before you request my review again.

socrata/socrata_upload.py
wrridgeway: I think I've implemented what you've described here.

PR #758, opened 2025-03-07
Initial push of geocoding test doc

Damonamajor:
TK

### DIFF
--- a/geocoding_test.qmd
+++ b/geocoding_test.qmd
@@ -1,0 +1,251 @@
+---
+title: "Test of geocoding"
+format: html
+output: html_document
+
+---
+
+# This is a test of two geocoding techniques. 
+
+## Problem Statement
+
+Between 2015 and our current datset there are 15676 missing geographies. We are attempting to geocode these geographies, using 
+
+
+```{r}
+library(DBI)
+library(dplyr)
+library(noctua)
+library(tidygeocoder)
+library(readr)
+library(sf)
+library(leaflet)
+library(purrr)
+
+
+AWS_S3_WAREHOUSE_BUCKET <- Sys.getenv("AWS_S3_WAREHOUSE_BUCKET")
+output_bucket <- file.path(AWS_S3_WAREHOUSE_BUCKET, "spatial", "parcel")
+
+
+AWS_ATHENA_CONN_NOCTUA <- dbConnect(
+  noctua::athena()
+)
+
+# Pull missing geographies and the information for the year prior and following.
+
+missing_geographies <- dbGetQuery(
+  AWS_ATHENA_CONN_NOCTUA,
+  "SELECT
+    m.pin,
+    m.year AS missing_year,
+    m.x_3435 AS missing_x,
+    m.y_3435 AS missing_y,
+    p.year AS previous_year,
+    p.x_3435 AS previous_x,
+    p.y_3435 AS previous_y,
+    n.year AS following_year,
+    n.x_3435 AS following_x,
+    n.y_3435 AS following_y,
+    -- Address fields for the missing year
+    addr_m.prop_address_full AS missing_prop_address_full,
+    addr_m.prop_address_city_name AS missing_prop_address_city_name,
+    addr_m.prop_address_state AS missing_prop_address_state,
+    addr_m.prop_address_zipcode_1 AS missing_prop_address_zipcode_1,
+    -- Address fields for the previous year
+    addr_p.prop_address_full AS previous_prop_address_full,
+    addr_p.prop_address_city_name AS previous_prop_address_city_name,
+    addr_p.prop_address_state AS previous_prop_address_state,
+    addr_p.prop_address_zipcode_1 AS previous_prop_address_zipcode_1,
+    -- Address fields for the following year
+    addr_n.prop_address_full AS following_prop_address_full,
+    addr_n.prop_address_city_name AS following_prop_address_city_name,
+    addr_n.prop_address_state AS following_prop_address_state,
+    addr_n.prop_address_zipcode_1 AS following_prop_address_zipcode_1
+  FROM default.vw_pin_universe m
+  -- Join for previous year record
+  LEFT JOIN default.vw_pin_universe p
+    ON m.pin = p.pin AND CAST(p.year AS integer) = CAST(m.year AS integer) - 1
+  -- Join for following year record
+  LEFT JOIN default.vw_pin_universe n
+    ON m.pin = n.pin AND CAST(n.year AS integer) = CAST(m.year AS integer) + 1
+  -- Join to get address for missing year
+  LEFT JOIN vw_pin_address addr_m
+    ON m.pin = addr_m.pin AND m.year = addr_m.year
+  -- Join to get address for previous year
+  LEFT JOIN vw_pin_address addr_p
+    ON p.pin = addr_p.pin AND p.year = addr_p.year
+  -- Join to get address for following year
+  LEFT JOIN vw_pin_address addr_n
+    ON n.pin = addr_n.pin AND n.year = addr_n.year
+  WHERE (m.x_3435 IS NULL OR m.y_3435 IS NULL)
+    AND CAST(m.year AS integer) >= 2015")
+
+
+# Concatenate addresses into a single value
+missing_geographies <- missing_geographies %>%
+  mutate(
+    missing_prop_address_full = paste(
+      missing_prop_address_full,
+      missing_prop_address_city_name,
+      missing_prop_address_state,
+      missing_prop_address_zipcode_1,
+      sep = " "
+    ),
+    previous_prop_address_full = paste(
+      previous_prop_address_full,
+      previous_prop_address_city_name,
+      previous_prop_address_state,
+      previous_prop_address_zipcode_1,
+      sep = " "
+    ),
+    following_prop_address_full = paste(
+      following_prop_address_full,
+      following_prop_address_city_name,
+      following_prop_address_state,
+      following_prop_address_zipcode_1,
+      sep = " "
+    )
+  )
+```
+
+
+```{r}
+# Identify missing geographies that can be imputed by averaging the previous and following years
+# We use a small window .0001 difference in both Latitude and longitude for this
+missing_geographies_imputed <- missing_geographies %>%
+  mutate(
+    both_missing = is.na(missing_x) & is.na(missing_y) &
+      !is.na(previous_x) & !is.na(previous_y) &
+      !is.na(following_x) & !is.na(following_y) &
+      abs(previous_x - following_x) <= 0.0001 &
+      abs(previous_y - following_y) <= 0.0001,
+    missing_x = if_else(both_missing, (previous_x + following_x) / 2, missing_x),
+    missing_y = if_else(both_missing, (previous_y + following_y) / 2, missing_y)
+  ) %>%
+  select(-both_missing)
+
+# Goecoding requires batches of 1000 to do, so we create a function
+batch_list <- split(missing_geographies_imputed,
+                    ceiling(seq_len(nrow(missing_geographies_imputed)) / 1000))
+
+geocode_batch <- function(batch) {
+  batch %>%
+    geocode(
+      street = missing_prop_address_full,
+      city   = missing_prop_address_city_name,
+      state  = missing_prop_address_state,
+      method = "census"
+    ) %>%
+    filter(!is.na(long) & !is.na(lat)) %>%
+    st_as_sf(coords = c("long", "lat")) %>%
+    st_set_crs(4326) %>%
+    st_transform(3435)
+}
+
+missing_geographies_geocoded <- map(batch_list, geocode_batch) %>%
+  do.call(rbind, .)
+
+missing_geographies_imputed_sf <- missing_geographies_imputed %>%
+  filter(!is.na(missing_x) & !is.na(missing_y)) %>%
+  st_as_sf(coords = c("missing_x", "missing_y"), crs = 3435) %>%
+  mutate(
+    technique = "imputed",
+    current_year_address = paste(missing_prop_address_full,
+                                 missing_prop_address_city_name,
+                                 missing_prop_address_state, sep = ", ")
+  )
+
+missing_geographies_geocoded <- missing_geographies_geocoded %>%
+  mutate(
+    technique = "geocoded",
+    current_year_address = paste(missing_prop_address_full,
+                                 missing_prop_address_city_name,
+                                 missing_prop_address_state, sep = ", ")
+  )
+
+# Bind them back to one dataset
+combined_geographies <- bind_rows(
+  missing_geographies_imputed_sf,
+  missing_geographies_geocoded
+)
+
+```
+
+Number of geographies filled by imputing is `r nrow(missing_geographies_imputed_sf)` and number of geographies filled by geocoding is `r nrow(missing_geographies_geocoded)`.
+
+To test how these work, we join them back together again and create a distance in ft column.
+```{r}
+combined_geographies <- combined_geographies %>%
+  group_by(pin) %>%
+  filter(n() == 2) %>%  # keep only pins that have both techniques
+  ungroup() %>%
+  st_transform(crs = 3435) %>%
+  group_by(pin) %>%
+  mutate(distance_ft = as.numeric(st_distance(first(geometry), last(geometry)))) %>%
+  ungroup() %>%
+  st_transform(crs = 4326)
+
+pal <- colorFactor(palette = "Set1", domain = combined_geographies$technique)
+```
+
+## Map of all pins with 2 techniques
+
+```{r}
+leaflet(data = combined_geographies) %>%
+  addTiles() %>%
+  addCircleMarkers(
+    color = ~pal(technique),
+    popup = ~paste(
+      "<strong>Technique:</strong>", technique, "<br>",
+      "<strong>Address:</strong>", current_year_address, "<br>",
+      ifelse(!is.na(distance_ft),
+             paste("<strong>Distance (ft):</strong>", round(distance_ft, 2)),
+             "")
+    )
+  )
+```
+## Pins with Long Distances (>1000 ft)
+
+```{r}
+# Look at long distance pins
+subset <- combined_geographies %>%
+  filter(distance_ft > 1000)
+
+
+leaflet(data = subset) %>%
+  addTiles() %>%
+  addCircleMarkers(
+    color = ~pal(technique),
+    popup = ~paste(
+      "<strong>Technique:</strong>", technique, "<br>",
+      "<strong>Address:</strong>", current_year_address, "<br>",
+      ifelse(!is.na(distance_ft),
+             paste("<strong>Distance (ft):</strong>", round(distance_ft, 2)),
+             "")
+    )
+  )
+```
+
+
+## Histogram of Distance (ft) Between Techniques
+The vast majority of pins are within 300 feet of eachother. 
+```{r}
+combined_geographies <- combined_geographies %>%
+  mutate(distance_ft = as.numeric(as.character(distance_ft))) %>%
+  filter(distance_ft < 1000)
+
+p <- ggplot(combined_geographies, aes(x = distance_ft)) +
+  geom_histogram(binwidth = 100, fill = "steelblue", color = "white") +
+  labs(
+    title = "Histogram of Distance (ft) Between Techniques",
+    x = "Distance (ft)",
+    y = "Frequency"
+  ) +
+  theme_minimal()
+
+print(p)
+
+```
+# Conclusion:
+In general, the two techniques are very similar. I'd have to go through them a bit more in depth, but at the moment, the geocoding looks a bit more accurate. This mostly comes from looking at the long distance pin map.
+


Damonamajor:
@jeancochrane @wrridgeway 
Draft is up

- In total 4327 observations can be encoded by using previous and following year x/y coordinates
- 12101 observations can be encoded using geolocation from OSM.
- X/Y coordinates are probably slightly more accurate since we have known issues with our address data.
- A comparison between the two (when there are observations for both imputed and geolocated) found that the vast majority of pins have negligible differences <300 ft. This is often the distance from the pin to the street. There are 5 out of a possible ~4,000 matched observations with distances > 1000 ft.

jeancochrane (commented):
Thanks for this! I agree with your general conclusion: Results seem pretty similar with both approaches, so let's prioritize the simpler one (filling using prior/future years when present) and fall back to geocoding in cases where that's not feasible.

I think we should keep this PR up to aid in discussion with @wrridgeway, but let's close it instead of merging it once we've decided on a path forward -- I don't think we need to persist this analysis indefinitely in this repo.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -394,6 +399,186 @@ process_parcel_file <- function(s3_bucket_uri,
...
+    # Keep pin since address will be coded to that level
+    distinct(pin, year, .keep_all = TRUE) %>%
+    # removes missing columns since geocoding duplicates lat
+    select(-c("x_3435", "y_3435", "lon", "lat"))
+
+  # Split the missing coordinates into batches of 1000 rows for tidygeocoder
+  batch_list <- split(missing_geographies,
+                      ceiling(seq_len(nrow(missing_geographies)) / 1000))
+
+  cook_county <- dbGetQuery(
```
Damonamajor, line 467: I put this test in because I know we geocode one property outside of Cook County. Do we want this for all parcels?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -394,6 +399,186 @@ process_parcel_file <- function(s3_bucket_uri,
...
+      y_3435_imputed = y_3435,
+      lon_imputed = lon,
+      lat_imputed = lat
+    ) %>%
+    group_by(prop_address_full, pin10) %>%
+    arrange(year) %>%
+    fill(x_3435_imputed, y_3435_imputed, lon_imputed,
+         lat_imputed, .direction = "updown") %>%
+    ungroup() %>%
+    # We want pins for addresses, but location data is on the pin10 level
```
Damonamajor, line 446: I did this test here. There are 3 PIN10s which have different x-y coordinates. I'll take a deeper look into them. We have 0 PIN10s which have NA and coordinates with coordinates with values, which is what I was assuming would happen.
```
missing_geographies_imputed <- missing_geographies_imputed %>%
+     filter(missing == "TRUE")
> test <- missing_geographies_imputed %>%
+     group_by(pin10, year)
> View(test)
> missing_geographies_imputed %>% 
+     group_by(pin10, year) %>%
+     summarise(
+         unique_x = n_distinct(x_3435_imputed),
+         unique_y = n_distinct(y_3435_imputed)
+     ) %>%
+     filter(unique_x > 1 | unique_y > 1)
`summarise()` has grouped output by 'pin10'. You can override using the `.groups` argument.
# A tibble: 3 × 4
# Groups:   pin10 [3]
  pin10      year  unique_x unique_y
  <chr>      <chr>    <int>    <int>
1 1710107018 2016         2        2
2 1710135038 2016         2        2
3 1715309032 2016         2        2
> mixed_na <- missing_geographies_imputed %>% 
+     group_by(pin10, year) %>%
+     summarise(
+         mixed_x = any(is.na(x_3435_imputed)) & any(!is.na(x_3435_imputed)),
+         mixed_y = any(is.na(y_3435_imputed)) & any(!is.na(y_3435_imputed))
+     ) %>%
+     filter(mixed_x | mixed_y)
`summarise()` has grouped output by 'pin10'. You can override using the `.groups` argument.
> 
> mixed_na
# A tibble: 0 × 4
# Groups:   pin10 [0]
# ℹ 4 variables: pin10 <chr>, year <chr>, mixed_x <lgl>, mixed_y <lgl>
```

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -394,6 +399,186 @@ process_parcel_file <- function(s3_bucket_uri,
...
+           x_3435_imputed, y_3435_imputed, lon_imputed, lat_imputed)
+
+  missing_geographies <- missing_geographies %>%
+    mutate(
+      # Calculate Euclidean distance using the projected coordinates
+      distance = sqrt((x_3435_imputed - avg_x_3435_geocoded)^2 +
+                        (y_3435_imputed - avg_y_3435_geocoded)^2),
+
+      # Replace x_3435 with the imputed value if both available
+      # and within 1000 feet, else choose available value or NA
```
Damonamajor, line 526: I bumped this up to 1000 ft for testing, mainly because there was a giant PIN10 which had a bunch of values 600 ft apart. For context a City of Chicago block is about 330 x 660. We should probably collectively decide what we want this to be.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -394,6 +399,186 @@ process_parcel_file <- function(s3_bucket_uri,
...
+    select(pin10, year, x_3435, y_3435, lon, lat, source) %>%
+    filter(!is.na(x_3435) & !is.na(y_3435) & !is.na(lon) & !is.na(lat))
+
+  spatial_df_final <- spatial_df_final %>%
+    mutate(source = "raw")
+
+  spatial_df_final <- bind_rows(spatial_df_final, missing_geographies)
+
+  # Test to make sure we don't duplicate any pins
+  duplicate_keys <- spatial_df_final %>%
```
Damonamajor, line 573: Test for duplicating keys because it made sense to rbind values back into the dataset.

wrridgeway:
I personally don't think we should be engineering solutions that depend on arbitrary thresholds and assumptions for what is essentially a vanishingly tiny portion of our dataset. IMO filling and geo-coding is all we should be doing here and if there are any parcels this doesn't address that is a data problem that needs to fixed, not something we should be engineering in our ETL. If most our imputation comes down to one or two PIN10w we should just reach out to GIS and let them know the data has an issue.

But if you two think this is essential for our data in some way I'll defer.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
jeancochrane: In that case, can we hardcode the correct coordinates for that particular address, and remove the county restriction?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
Damonamajor: They are big complexes, and it's a bit strange because the code doesn't seem to _lend_ itself to differences, but they are very small. 

test_1 <- test %>%
 distinct(lon, lat, x_3435, y_3435)

image

jeancochrane (commented):
Much simpler! I think there's still one spot where we could potentially simplify, but I'm not totally confident in it, so I leave it up to you and Billy. I'll let Billy be the final reviewer here.

Relatedly, is there a test dataset available that I can QC before we merge?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -394,6 +397,139 @@ process_parcel_file <- function(s3_bucket_uri,
...
+  # Grab all years of data for parcels which are missing in any year
+  spatial_subset <- spatial_df_final %>%
+    filter(pin10 %in% missing_geographies$pin10) %>%
+    # Add property address to the subset to make sure PINs are consistent.
+    # This will expand the dataset since prop_addresses are not unique
+    # by PIN10.
+    left_join(all_geographies %>% select(year, pin10, prop_address_full),
+              by = c("year", "pin10"))
+
+  imputed <- bind_rows(spatial_subset, missing_geographies)
```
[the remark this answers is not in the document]
Damonamajor: The issue is that a PIN10 will have multiple addresses. 
100 Clark St Unit A
100 Clark St Unit B
or 102 Clark Street
104 Clark Street

Since we are joining based on addresses and PIN10s need to match, we want to make sure that all addresses are kept associated with the PIN10s.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -394,6 +397,139 @@ process_parcel_file <- function(s3_bucket_uri,
...
+               vpa.prop_address_zipcode_1,
+               pd.class,
+               pd.parid
+            from default.vw_pin_address vpa
+            left join iasworld.pardat pd
+              on vpa.pin = pd.parid and vpa.year = pd.taxyr
+            where vpa.year >= '2000'
+              ")
+
+  # This will be larger than the parcel dataframe since it's on PIN level
```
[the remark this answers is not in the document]
Damonamajor: parcels is on the pin10 level, this query grabs things on the pin level so there are more observations.

Damonamajor:
> Much simpler! I think there's still one spot where we could potentially simplify, but I'm not totally confident in it, so I leave it up to you and Billy. I'll let Billy be the final reviewer here.
> 
> Relatedly, is there a test dataset available that I can QC before we merge?

I think your idea works, so I'll try to implement it.
I'll re-run the report I have when everything feels wrapped up to make sure it looks OK before merging in.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +415,170 @@ pwalk(parcel_files_df, function(...) {
...
+      object = ..1,
+      bucket = AWS_S3_WAREHOUSE_BUCKET
+    )
+    df$year <- ..2
+    df
+  })
+
+pre_geocoding_data <- pre_geocoding_data %>%
+  mutate(source = "raw",
+         uploaded_before_geocoding = TRUE)
```
Damonamajor, line 445: This was the language you recommended, but to me it makes a bit more sense to reverse it. Thoughts on PIN10_has_undergone_geocoding = FALSE and then TRUE when it is finished?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +436,169 @@ pwalk(parcel_files_df, function(...) {
...
+  dbGetQuery(
+    conn = con, "
+    SELECT DISTINCT
+        pd.parid,
+        vpa.year,
+        vpa.prop_address_full,
+        vpa.prop_address_city_name,
+        vpa.prop_address_state,
+        vpa.prop_address_zipcode_1,
+        pd.class
```
wrridgeway, line 464: Do we need class here? If we don't, we can kill *pardat* entirely from the pull, if we've made sure it's got the same row count as `vw_pin_address`. Which I would *hope* it does.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +436,169 @@ pwalk(parcel_files_df, function(...) {
...
   )
 })
+
+gc()
+
+# Ingest processed parcel files into one dataframe
+pre_geocoding_data <- open_dataset(
+  file.path(AWS_S3_WAREHOUSE_BUCKET, "spatial", "parcel_test")
+) %>%
+  collect() %>%
```
wrridgeway, line 446: I replaced the map function here with `open_dataset()` and `collect()`

wrridgeway (changes requested):
Some minor changes and then we'll do a more thorough review. Main thing I'm going to be inspecting here is that we aren't needlessly using `pin` if `pin10` suffices at points.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -73,336 +76,356 @@ process_parcel_file <- function(s3_bucket_uri,
...
-    spatial_df_raw <- st_read(local_spatial_file)
-    tictoc::toc()
+  file_year_processed <- aws.s3::get_bucket_df(
+    bucket = AWS_S3_WAREHOUSE_BUCKET,
+    prefix = file.path("spatial", "parcel_test")
+  ) %>%
+    filter(Size > 0, str_detect(Key, file_year), !str_detect(Key, "test2"))
+
+  # Skip processing if expected data already exists in warehouse
+  if (nrow(file_year_processed) != 39) {
```
wrridgeway, line 86: This will let us skip already processed years in the past while still loading them below.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +436,169 @@ pwalk(parcel_files_df, function(...) {
...
+  mutate(
+    lat = if_else(pin10 == "1819200021" &
+      year == "2000", as.numeric("-87.896805"), lat),
+    lon = if_else(pin10 == "1819200021" &
+      year == "2000", as.numeric("41.766003"), lon),
+    x_3435 = if_else(pin10 == "1819200021" &
+      year == "2000", as.numeric("1573797"), x_3435),
+    y_3435 = if_else(pin10 == "1819200021" &
+      year == "2000", as.numeric("-46628780"), y_3435)
+  ) %>%
```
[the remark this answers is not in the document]
Damonamajor: I'm pretty sure this was a Jean choice.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
Damonamajor: Yeah I don't think it's needed. I'll take it out.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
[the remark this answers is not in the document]
wrridgeway: Alright, @jeancochrane are you sure on this? I'll defer to you, but I'd worry this list will slowly grow over time.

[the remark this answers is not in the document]
Damonamajor: She said to switch it back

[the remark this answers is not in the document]
jeancochrane: Yeah, sorry for the mixup! I was hoping we could avoid having an external dependency, but I hear your argument that the hardcoding will be a maintenance burden.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +442,170 @@ pwalk(parcel_files_df, function(...) {
...
+  fill(x_3435, y_3435, lon, lat, .direction = "updown") %>%
+  ungroup() %>%
+  # Remove duplicate rows based on pin10 and year
+  distinct(pin10, year, .keep_all = TRUE) %>%
+  # Only keep rows where none of the four fields are missing
+  filter(!is.na(x_3435) & !is.na(y_3435) & !is.na(lon) & !is.na(lat)) %>%
+  # Keep observations which were originally missing
+  filter(missing) %>%
+  mutate(source = "imputed") %>%
+  select(year, pin10, x_3435, y_3435, lon, lat, source)
```
wrridgeway, line 534: Okay, I think I understand what's going on here, but before I keep going on this, is there any reason we can't just do something like:

```
imputed <- pre_geocoding_data %>%
  full_join(
    all_addresses %>%
      distinct(pin10, year, .keep_all = TRUE),
    by = c("pin10" ,"year")) %>%
  group_by(pin10, prop_address_full) %>%
  mutate(missing = is.na(x_3435) | is.na(y_3435)) %>%
  arrange(year) %>%
  fill(x_3435, y_3435, lon, lat, .direction = "updown") %>%
  ungroup()
```
and avoid all these joins and different dataframes and multiple steps? I would actually make `all_addresses` distinct by `pin10` immediately after pulling it, and not even bother pulling `pin`.

I understand we're arbitrarily choosing one pin14 addresses per pin10 per year, but I can't think of why that would be an issue off the top of my head.

wrridgeway (changes requested):
Got some questions we can work through tomorrow if you'd like.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +442,170 @@ pwalk(parcel_files_df, function(...) {
...
+      prop_address_city_name,
+      prop_address_state,
+      prop_address_zipcode_1
+    FROM default.vw_pin_address
+    WHERE year >= '2000'
+  "
+)
+
+all_addresses <- all_addresses %>%
+  mutate(year = as.character(year))
```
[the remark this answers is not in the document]
Damonamajor: It's only necessary if it doesn't run otherwise. I think the rbind wasn't working without it, but happy to take it out and test without.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
wrridgeway: I feel like we can entirely avoid doing anything at the pin14 level if choosing an arbitrary address per pin10 doesn't have any adverse effect on the *actual* geocoding/non-filling part of this process.

Damonamajor:
I'm going to run taking out the as.character for year and see if it still works., but that shouldn't be blocking

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
Damonamajor: The observation counts seem pretty logical without any PIN14 values

Damonamajor:
> I'm going to run taking out the as.character for year and see if it still works., but that shouldn't be blocking

```
> missing_geographies <- pre_geocoding_data %>%
+   full_join(
+     all_addresses %>%
+       distinct(pin10, year, .keep_all = TRUE),
+     by = c("pin10", "year")) %>%
+   group_by(pin10, prop_address_full) %>%
+   mutate(missing = is.na(x_3435) | is.na(y_3435)) %>%
+   arrange(year) %>%
+   fill(x_3435, y_3435, lon, lat, .direction = "updown") %>%
+   ungroup() %>%
+   # Keep only values which in our input had no x-y coordinates
+   filter(missing)
Error in `full_join()`:
! Can't join `x$year` with `y$year` due to incompatible types.
ℹ `x$year` is a <integer>.
ℹ `y$year` is a <character>.
```

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +442,146 @@ pwalk(parcel_files_df, function(...) {
...
+
+cook_county <- geoarrow::read_geoparquet_sf(
+  "s3://ccao-data-warehouse-us-east-1/spatial/ccao/county/2019.parquet"
+) %>%
+  st_transform(3435)
+
+missing_geographies <- pre_geocoding_data %>%
+  full_join(
+    all_addresses %>%
+      distinct(pin10, year, .keep_all = TRUE),
```
wrridgeway, line 501: Do we need this `distinct()`? We expect the window function in the sql query to handle this, yeah?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +442,146 @@ pwalk(parcel_files_df, function(...) {
...
+# Split the missing parcels into those which were encoded with inputed technique
+# and those which still need to be calculated with geocoding
+imputed <- missing_geographies %>%
+  filter(!is.na(x_3435) & !is.na(y_3435) & !is.na(lon) & !is.na(lat)) %>%
+  select(pin10, year, x_3435, y_3435, lon, lat) %>%
+  mutate(source = "imputed")
+
+missing_geographies <- missing_geographies %>%
+  filter(is.na(x_3435) & is.na(y_3435) & is.na(lon) & is.na(lat)) %>%
+  distinct(pin10, year, .keep_all = TRUE) %>%
```
wrridgeway, line 521: Ditto here on `distinct`. Let's only use it if we need it.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +442,146 @@ pwalk(parcel_files_df, function(...) {
...
+    st_transform(3435) %>%
+    mutate(
+      x_3435 = st_coordinates(.)[, 1],
+      y_3435 = st_coordinates(.)[, 2]
+    )
+}
+
+# Apply geocoding to each batch and combine results
+geocoded <- map(batch_list, geocode_batch) %>%
+  do.call(rbind, .)
```
wrridgeway, line 554:
```suggestion
  do.call(rbind, .)
  bind_rows()
```

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +442,146 @@ pwalk(parcel_files_df, function(...) {
...
+
+geocoded <- geocoded %>%
```
wrridgeway, line 556:
```suggestion
bind_rows() %>%
```

wrridgeway (approved):
This looks really good! I took a look at what's in athena at it seems like everything's making it in there as expected.

Take a look at my suggested changes, and let's see if we can get rid of unnecessary `distinct()` calls.

Lastly, let's revert all of the `parcel_test`/`parcel_test2` syntax to `parcel`.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +442,146 @@ pwalk(parcel_files_df, function(...) {
...
+  # We can have address data from current year, before parcel data
+  # is available.
+  filter(year <= max(pre_geocoding_data$year, na.rm = TRUE))
+
+cook_county <- geoarrow::read_geoparquet_sf(
+  "s3://ccao-data-warehouse-us-east-1/spatial/ccao/county/2019.parquet"
+) %>%
+  st_transform(3435)
+
+missing_geographies <- pre_geocoding_data %>%
```
wrridgeway, line 499: Let's get a one-liner explaining this chunk of code.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -1,22 +1,30 @@
 library(arrow)
 library(aws.s3)
+library(ccao)
```
wrridgeway, line 3: Do we still need this?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
Damonamajor: Yup, and spatial.parcel is distinct, which I assume there are tests for.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
Damonamajor: I don't think so, I think it was from the Cook County shapefile which we switched out.

Damonamajor:
I should have done everything except take out test. 
I also talked with Jean, and rather than using a Cook County Boundary, we match the first four digits of the Sidwell Grid, which should produce more consistent results.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -490,26 +490,33 @@ all_addresses <- dbGetQuery(
   # is available.
   filter(year <= max(pre_geocoding_data$year, na.rm = TRUE))
 
-# Use the county shapefile as a bounding box to ensure no parcels are geo-coded outside of where we'd expect them to be
-cook_county <- geoarrow::read_geoparquet_sf(
-  "s3://ccao-data-warehouse-us-east-1/spatial/ccao/county/2019.parquet"
+# We will use the Sidwell grid to compare the first 4 digits
+# of the geocoded PIN10s against their expected location.
```
wrridgeway, line 494: Can we mention why we do this rather than just a county bounding box for posterity?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
wrridgeway, line 3:
```suggestion

```

wrridgeway (approved):
Looks good to me. Thanks for all the hard work on this.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +441,155 @@ pwalk(parcel_files_df, function(...) {
...
+  group_by(pin10, year) %>%
+  filter(n() > 1)
+
+if (nrow(duplicate_keys) > 0) {
+  stop("Duplicate rows found pin10 and year combinations. Check rbind")
+}
+
+# Write final dataframe to dataset on S3, partitioned by town and year
+post_geocoding_data %>%
+  mutate(uploaded_before_geocoding = FALSE) %>%
```
jeancochrane, line 587: **[Question, non-blocking]** What do we imagine we'll do with this `uploaded_before_geocoding` flag, given that we're expecting to run imputation/geocoding during ingest for all years going forward? I'm wondering if this field is substantively different from `source = 'raw'` or if we can get the same info from that field?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +438,168 @@ pwalk(parcel_files_df, function(...) {
...
+  "
+)
+
+all_addresses <- all_addresses %>%
+  mutate(year = as.character(year))
+
+cook_county <- dbGetQuery(
+  conn = con,
+  "SELECT * FROM spatial.county"
+) %>%
```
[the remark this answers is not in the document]
jeancochrane: **[Question, non-blocking]** I'm guessing this question/response is outdated now that we're using the Sidwell grid instead of the County shape, right?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +441,155 @@ pwalk(parcel_files_df, function(...) {
...
+)
+
+# Function to geocode each batch
+geocode_batch <- function(batch) {
+  batch %>%
+    geocode(
+      street = prop_address_full,
+      city   = prop_address_city_name,
+      state  = prop_address_state,
+      method = "census"
```
jeancochrane, line 546: **[Question, non-blocking]** I don't have strong opinions, but I'm curious: Any reason we chose `census` here over the other geocoding providers?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -413,3 +441,155 @@ pwalk(parcel_files_df, function(...) {
...
+  stop("Duplicate rows found pin10 and year combinations. Check rbind")
+}
+
+# Write final dataframe to dataset on S3, partitioned by town and year
+post_geocoding_data %>%
+  mutate(uploaded_before_geocoding = FALSE) %>%
+  relocate(year, town_code, .after = last_col()) %>%
+  group_by(year, town_code) %>%
+  write_partitions_to_s3(
+    "s3://ccao-data-warehouse-us-east-1/spatial/parcel",
```
jeancochrane, line 591: **[Nitpick, optional]** Shouldn't we use the `output_bucket` variable here, for consistency with the rest of the script?

```suggestion
    output_bucket,
```

jeancochrane (approved):
Nice work here, you two!

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -73,336 +80,357 @@ process_parcel_file <- function(s3_bucket_uri,
...
-
-    # Read local geojson file
-    tictoc::tic(paste("Read file for:", file_year))
-    spatial_df_raw <- st_read(local_spatial_file)
-    tictoc::toc()
+  file_year_processed <- aws.s3::get_bucket_df(
+    bucket = AWS_S3_WAREHOUSE_BUCKET,
+    prefix = file.path("spatial", "parcel")
+  ) %>%
+    filter(Size > 0, str_detect(Key, file_year), !str_detect(Key, "test2"))
```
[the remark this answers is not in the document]
Damonamajor: No we want to drop this

Damonamajor:
> Actually, one final request before we merge: Can we add documentation for the `source` column (and any other new columns we decide to add) to the docs for this table?
> 
> https://github.com/ccao-data/data-architecture/blob/87e8009430799ffdb113feb66bc907ef1979f2d8/dbt/models/spatial/schema.yml#L134-L135
> 
> I'm also curious if you two have a deployment plan? I assume we'll delete `spatial.parcel` and move `spatial.parcel_test2` into its place, but correct me if I'm wrong.

I think we need to run the script again. I'm not sure spatial.parcel_test2 includes the sidwell grid check, although we can do it in that way.

wrridgeway:
Yes, deployment plan is to run the script again. I still have the huge local files from the first part saved.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
```diff
@@ -73,336 +80,357 @@ process_parcel_file <- function(s3_bucket_uri,
...
+
+  # Skip processing if expected data already exists in warehouse
+  if (nrow(file_year_processed) != 39) {
```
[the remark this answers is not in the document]
wrridgeway: Thank you for catching this.  The presence of data for 2000 in `spatial/parcel_test` messed up the logic for 2000. We should always have 38 towns + 1 HIVE_DEFAULT file. 2000 has 63, but everything else has 39. Running it again with `spatial/parcel_test` deleted yields the expected results, and we don't expect that to be an issue moving forward. We can also remove the test for 'test2' strings.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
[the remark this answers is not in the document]
wrridgeway: It was referencing the `parcel_test2` files in s3. We should remove this test.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-parcel.R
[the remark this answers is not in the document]
wrridgeway:
```
hug <- map(("2000":"2024"), \(file_year) {
  aws.s3::get_bucket_df(
    bucket = AWS_S3_WAREHOUSE_BUCKET,
    prefix = file.path("spatial", "parcel")
  ) %>%
    filter(Size > 0, str_detect(Key, as.character(file_year)))
})
```
image

PR #764, opened 2025-03-12
Add more printing to socrata API script

wrridgeway:
The job did [run as scheduled](https://github.com/ccao-data/data-architecture/actions/runs/13815182863/job/38646411048) without throwing an error, but didn't seem to do anything. I'm just trying to make it easier to figure out what went wrong here since I can't seem to replicate the issue locally.


jeancochrane (approved):
Makes sense!

PR #765, opened 2025-03-13
Remove unecessary aws credentials step and fix buffered workflow logs

wrridgeway:
This PR should address [two issues](https://github.com/ccao-data/data-architecture/actions/runs/13817905136) that surfaced during our first test of running the socrata api script on a scheduled workflow:

- After one hour the runner's AWS token expired and it could no longer query Athena. This was because I added an unnecessary `Configure AWS credentials` step to the workflow what reduced how long the aws token is valid from 4 to 1 hours.
- The workflow logs are not necessarily printed in order which makes debugging difficult
---
Putting this here since it's [poorly documented](https://github.com/laughingman7743/PyAthena/blob/2787190871d6be6f18f7ff1fa61c8e570cd2d7e4/pyathena/util.py#L41) and was hard to find despite us not needing it for this PR:

```
from pyathena.util import RetryConfig

# Connect to Athena
cursor = connect(
    s3_staging_dir="s3://ccao-athena-results-us-east-1/",
    region_name="us-east-1",
    cursor_class=PandasCursor,
    # Limit cursor retry attempts
    retry_config=RetryConfig(attempt=1),
).cursor(unload=True)
```

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -11,7 +11,6 @@ import requests
 from dbt.cli.main import dbtRunner
 from pyathena import connect
 from pyathena.pandas.cursor import PandasCursor
-from pyathena.util import RetryConfig
 
 logger = logging.getLogger(__name__)
 
@@ -30,7 +29,6 @@ cursor = connect(
     s3_staging_dir="s3://ccao-athena-results-us-east-1/",
     region_name="us-east-1",
     cursor_class=PandasCursor,
-    retry_config=RetryConfig(attempt=1),
 ).cursor(unload=True)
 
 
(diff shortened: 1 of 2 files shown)


socrata/socrata_upload.py
```diff
@@ -29,6 +30,7 @@
     s3_staging_dir="s3://ccao-athena-results-us-east-1/",
     region_name="us-east-1",
     cursor_class=PandasCursor,
+    retry_config=RetryConfig(attempt=1),
```
wrridgeway, line 33: I believe this should do what we want, but the function is [not very documented](https://github.com/laughingman7743/PyAthena/blob/2787190871d6be6f18f7ff1fa61c8e570cd2d7e4/pyathena/util.py#L41)...

jeancochrane (approved):
Let's try it!

PR #768, opened 2025-03-13
Revert temporary socrata upload changes to steady-state

wrridgeway:
This PR:
- resets our open asset IDs from the temporary test asset IDs
- sets our normal update schedule moving forward
- makes sure at least one year of data for every open data asset will be updated


dbt/models/open_data/exposures.yml
```diff
@@ -180,7 +180,7 @@ exposures:
       # Disabled until the asset is public
       test_row_count: false
```
jeancochrane, line 181: **[Suggestion, non-blocking]** Same here:

```suggestion
      test_row_count: true
```

dbt/models/open_data/exposures.yml
```diff
@@ -136,7 +136,7 @@ exposures:
     meta:
       # Disabled until the asset is public
       test_row_count: false
```
jeancochrane, line 138: **[Suggestion, non-blocking]** How about we go ahead and enable these tests, since this asset should be public by this time next month?

```suggestion
      test_row_count: true
```

jeancochrane (approved):
Awesome work! I double-checked the cron, all of the asset IDs, and the asset links. I'm excited that this is finally going live!

PR #769, opened 2025-03-14
Update town close year columns and fix filter behaviour

wagnerlmichael:


dbt/export/templates/qc.vw_report_town_close_289s.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** Seems like we didn't actually change anything in this template, is that right? If so, can we reset it so that the diff for this PR only shows files that changed?

dbt/export/templates/qc.vw_report_town_close_bldg_parcel_class_mismatch.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** This file seems like it also didn't change, is that right?

dbt/export/templates/qc.vw_report_town_close_class_does_not_equal_luc.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** Seems like the columns here all got resized to fit the width of the header text, which makes e.g. the `PARID` column narrow enough that it won't actually fit a full PIN once we add data to it. It's a pain, but do you mind comparing the column widths to the original template and tweaking everything so that it's at least as wide as it was in the original template? Totally fine for columns with new text to be wider if it's necessary to fit the new text, I just want to make sure we don't make columns narrower in a way that will make data appear truncated.

dbt/export/templates/qc.vw_report_town_close_neg_asmt_value.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** Seems like this file also did not change, except for removing the background color from the `PARID` column (A), which seems to be a mistake. Let's remove it from the diff.

dbt/export/templates/qc.vw_report_town_close_prior_year_card_code_5s_comdat.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** Looks like the `Curr. Year Total AV` column (AA) is slightly too narrow so the new header text is getting cut off.

dbt/export/templates/qc.vw_report_town_close_prior_year_card_code_5s_oby.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** Two small nits here:

1. Missing a space in `Curr. YearLMV` (W)
2. Column containing `Curr. Year Total AV` (AB) is too narrow and the text is getting cut off

dbt/export/templates/qc.vw_report_town_close_res_edit.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** Columns containing `Curr. Year Dwelling MV` (X) and `Curr. Year Dwelling Total` (Y) are too narrow and the text is getting cut off.

dbt/export/templates/qc.vw_report_town_close_res_parcels_not_set_to_cost_approach.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** This template also seems to not have changed, so let's remove it from the diff.

dbt/scripts/utils/export.py
```diff
@@ -213,7 +213,7 @@ def query_models_for_export(
         template_name = template_config.get("name") or model_name
         sheet_name = template_config.get("sheet_name") or "Sheet1"
         start_row = template_config.get("start_row") or 2
-        add_table = template_config.get("add_table") or False
+        add_table = template_config.get("add_table", True)
```
jeancochrane, line 216: **[Nitpick, blocking]** Nice work tracking this down! I forgot to update the issue, but I actually chatted with Valuations last week and confirmed that they _don't_ in fact want to use tables for these workbooks. Sorry for the wild goose chase! Next time I'll update the issue as soon as I get new information.

```suggestion
        add_table = template_config.get("add_table") or False
```

jeancochrane (changes requested):
Nice work here! I like to keep these templates as clean as possible so that they need no tweaking after export; in that spirit, a few nitpicks below that will add an extra layer of polish to these changes.

In that spirit, do you mind also double-checking the workbooks that we want to include in the diff to make sure the cursor is highlighting the top left cell (A1)?

dbt/scripts/utils/export.py
wagnerlmichael: Ah, got it. No worries!

dbt/export/templates/qc.vw_report_town_close_res_edit.xlsx
jeancochrane, line 1: **[Nitpick, blocking]** One tiny edit here, looks like we may have accidentally duplicated "Curr. Year" instead of "Prior Year" for a couple of dwelling values:

Screenshot 2025-03-18 141740

Compare this to the version of the template on the main branch:

Screenshot 2025-03-18 141752

jeancochrane (approved):
Nice work adjusting the column widths! Everything looks perfect to me, just one final typo. Re-request me once that's fixed and we'll get this merged in 👍🏻

PR #774, opened 2025-03-24
Add municipality to permit views

wrridgeway:
Quick PR to add a municipality column to our permits views for easy filtering/aggregation. No rush to review this, just figured we'd have it ready to go if we decide we want to move forward with adding the column to the open data asset.

```
select
  count(*) as count,
  'new' as new
from z_ci_add_municipalities_to_permits_view_default.vw_pin_permit
union all
select
  count(*) as count,
  'old' as new
from default.vw_pin_permit
```

count | new
-- | --
554250 | old
554250 | new

```
select
  year,
  sum(cast(municipality is null as int)) as null_muni,
  sum(cast(municipality is not null as int)) as non_null_muni
from z_ci_add_municipalities_to_permits_view_default.vw_pin_permit
where year between '1999' and '2025'
group by year
order by year desc
```

year | null_muni | non_null_muni
-- | -- | --
2025 | 297 | 3379
2024 | 5046 | 93977
2023 | 4624 | 99013
2022 | 3807 | 96151
2021 | 5016 | 112950
2020 | 4573 | 103904
2019 | 448 | 15166
2018 | 9 | 3000
2017 | 0 | 109
2016 | 0 | 4
2015 | 0 | 1
2014 | 0 | 3
2013 | 0 | 2
2012 | 1 | 6
2011 | 1 | 6
2010 | 0 | 7
2009 | 0 | 2
2008 | 0 | 2
2007 | 0 | 6
2006 | 1 | 7
2005 | 1 | 0
2004 | 1 | 0
2002 | 16 | 0
2001 | 7 | 0
2000 | 20 | 0
1999 | 1 | 0

I don't think there are any new tests that need to be added here? We could do something with `iasworld.permit.permdt` since it's part of the municipality join and is kind of dirty, but I'm not entirely sure what.

### DIFF
--- a/dbt/models/default/default.vw_pin_permit.sql
+++ b/dbt/models/default/default.vw_pin_permit.sql
@@ -47,12 +47,12 @@ SELECT
         address.address_street_name,
         address.address_suffix_1,
         address.address_suffix_2
-    ) AS location_address_full,
-    address.address_street_dir AS location_address_street_dir,
-    address.address_street_number AS location_address_street_number,
-    address.address_street_name AS location_address_street_name,
-    address.address_suffix_1 AS location_address_suffix_1,
-    address.address_suffix_2 AS location_address_suffix_2,
+    ) AS prop_address_full,
+    address.address_street_dir AS prop_address_street_dir,
+    address.address_street_number AS prop_address_street_number,
+    address.address_street_name AS prop_address_street_name,
+    address.address_suffix_1 AS prop_address_suffix_1,
+    address.address_suffix_2 AS prop_address_suffix_2,
     -- Replace double commas that are present in the note2 field. We need
     -- to get rid of two types of double commas: trailing double commas,
     -- which need to be removed, and double commas inside an address
--- a/dbt/models/default/default.vw_pin_permit.sql
+++ b/dbt/models/default/default.vw_pin_permit.sql
@@ -61,7 +61,7 @@ SELECT
         REGEXP_REPLACE(permit.note2, ',,$'),
         ',,',
         ', '
-    ) AS mailing_address,
+    ) AS mail_address,
     permit.user21 AS applicant_name,
     permit.why AS job_code_primary,
     permit.user42 AS job_code_secondary,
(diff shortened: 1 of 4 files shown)


jeancochrane (commented):
This seems like a good approach, but I don't think we should introduce it to either the `default` view or the open data asset until we've discussed with Data Integrity to make sure it's reasonable. I'm fine leaving this PR open until then!

wrridgeway:
This should align with the issue it's attached to now, and with our usual addresses prefixes used in `default`.

jeancochrane (commented):
Looking good! What do you think about my suggested column name changes? Happy to approve if you feel strongly about the existing names, but I want to call them out explicitly in case you haven't seen my comments.

wrridgeway:
Totally fine with your name suggestions.

dbt/models/default/default.vw_pin_permit.sql
```diff
@@ -38,33 +38,31 @@ SELECT
...
+    NULLIF(ARRAY_JOIN(vpu.tax_municipality_name, ', '), '')
+        AS tax_municipality_name,
+    CONCAT_WS(
+        ' ',
+        CAST(address.address_street_number AS VARCHAR),
+        address.address_street_dir,
+        address.address_street_name,
+        address.address_suffix_1,
+        address.address_suffix_2
+    ) AS prop_address_full,
```
jeancochrane, line 51: **[Suggestion, optional]** I like the choice to use `NULLIF()` when expanding out the `tax_municipality_name` array. What do you think about doing the same for `prop_address_full`? It's not a dealbreaker that the field is currently an empty string when no property address components are present, but I think a null might be more intuitive, particularly if that's how `tax_municipality_name` works.

```suggestion
    NULLIF(
        CONCAT_WS(
            ' ',
            CAST(address.address_street_number AS VARCHAR),
            address.address_street_dir,
            address.address_street_name,
            address.address_suffix_1,
            address.address_suffix_2
        ),
        ''
    ) AS prop_address_full,
```

jeancochrane (approved):
Excellent! One suggestion for how to represent empty values for `prop_address_full`, but otherwise this is good to go.

Issue #776, opened 2025-03-25
Open Data cleanup


wrridgeway:
Cook County GIS reports a few issues with Open Data. Let's resolve:

- [x] **1. Assessed Values.** This is higher priority. [CookViewer](https://maps.cookcountyil.gov/cookviewer/) links from a specific PIN to "Historical Assessed Values" no longer work, as of March 20, 2025. Here's an [example](https://datacatalog.cookcountyil.gov/Property-Taxation/Assessor-Assessed-Values/uzyt-m557/explore/query/SELECT%0A%20%20%60pin%60%2C%0A%20%20%60year%60%2C%0A%20%20%60class%60%2C%0A%20%20%60township_code%60%2C%0A%20%20%60township_name%60%2C%0A%20%20%60mailed_bldg%60%2C%0A%20%20%60mailed_land%60%2C%0A%20%20%60mailed_tot%60%2C%0A%20%20%60certified_bldg%60%2C%0A%20%20%60certified_land%60%2C%0A%20%20%60certified_tot%60%2C%0A%20%20%60board_bldg%60%2C%0A%20%20%60board_land%60%2C%0A%20%20%60board_tot%60%0AWHERE%20caseless_eq%28%60pin%60%2C%20%22%7BPIN14%7D%22%29%0AORDER%20BY%20%60year%60%20DESC%20NULL%20LAST/page/filter). Did something on our end change?
- [x] **2. Parcel Universe (Current Year Only) has inconsistent field structure compared to Parcel Universe.** There are differences in column names; year versus tax_year, for example. We should standardize. (I prefer "tax year" but am open to suggestions.) 
- [x] **3. Permits: "assessment year"**. It is unclear whether this is the same as tax year / year. If it's the same as tax year, let's make this consistent.

wrridgeway:
Removed assessment_year from the permits asset, brought current parcel universe asset column names in line with historical parcel universe column names, Josh sorted out the broken cook viewer link.

PR #777, opened 2025-03-25
Remove `assessment_year` column from `open_data.vw_permit`

wrridgeway:
This column is difficult to understand and rather dirty, it seems more confusing than helpful to include on the open data portal. Part of https://github.com/ccao-data/data-architecture/issues/776.

@ccao-jardine Jean and I feel it's best to just remove this column rather than try to explain it to the public since we're not entirely sure ourselves how it works, even after reaching out to field.

### DIFF
--- a/dbt/models/open_data/exposures.yml
+++ b/dbt/models/open_data/exposures.yml
@@ -177,7 +177,7 @@ exposures:
       name: Data Department
     meta:
       test_row_count: true
-      year_field: assessment_year
+      year_field: year
       asset_id: 6yjf-dfxs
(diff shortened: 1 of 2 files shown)


jeancochrane (approved):
One quick change -- I think we'll want to remove this line so that the open data asset test `year_field` falls back to `"year"`:

https://github.com/ccao-data/data-architecture/blob/25e87d40231bde9028ddb0153de4297be3a06181/dbt/models/open_data/exposures.yml#L180

Once that's done, this is good to merge!

PR #780, opened 2025-04-02
Decrease data pushed to socrata through api

wrridgeway:
This won't fix randomly getting denied by Socrata's API, but it will make what we upload smaller so we're less likely to run up against tokens expiring, runners timing out, etc.

- Only upload most recent year for each asset. Permits asset has a row from 2032, so if most recent year is > whatever the current year is, script will default to current year. Previously we were uploading current year and/or current year - 1 for every asset, depending on which years are available in the asset.
- Added dbt tag for parcel status asset so it won't get updated since it's private.

I ran the script with these updates in an environment that simulates the scheduled github action and it worked as intended.

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -56,6 +56,8 @@ def parse_assets(assets=None):
         "open_data.*",
         "--resource-types",
         "exposure",
+        "--exclude",
+        "tag:inactive",
         "--output",
         "json",
         "--output-keys",
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -146,7 +146,7 @@ def parse_years_list(athena_asset, years=None):
     elif not years and os.getenv("WORKFLOW_EVENT_NAME") == "schedule":
         # Update most recent year only on scheduled workflow
         years_list = (
-            cursor.execute("SELECT MAX(year) FROM " + athena_asset)
+            cursor.execute("SELECT MAX(year) AS year FROM " + athena_asset)
             .as_pandas()["year"]
             .to_list()
         )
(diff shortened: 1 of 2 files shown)


socrata/socrata_upload.py
```diff
@@ -145,11 +147,16 @@ def parse_years_list(athena_asset, years=None):
...
+        # Update most recent year only on scheduled workflow. In some
+        # cases the max year is incorrectly in the future, so we
+        # append the current year to the list and take the minimum.
+        years_list = (
+            cursor.execute("SELECT MAX(year) AS year FROM " + athena_asset)
+            .as_pandas()["year"]
+            .to_list()
+        )
+        years_list.append(str(datetime.now().year))
+        years_list = [min(years_list)]
```
wrridgeway, line 159: The other option here is to create a specific treatment for the permits asset, maybe always updating all years for that asset since it's small.

socrata/socrata_upload.py
```diff
@@ -57,6 +57,8 @@ def parse_assets(assets=None):
         "open_data.*",
         "--resource-types",
         "exposure",
+        "--exclude",
+        "tag:inactive",
```
jeancochrane, line 61: **[Praise]** Nice use of dbt tags and selection syntax here 👍🏻

socrata/socrata_upload.py
jeancochrane: **[Suggestion, non-blocking]** I'm OK with either of these options, no strong feelings from my end! From an interface perspective, I think we could merge these two ideas if we added an optional attribute like `schedule_most_recent_n_years` to the `meta` object in the exposures config, and updated the `socrata_upload` script to exhibit the following behavior:

* If `schedule_most_recent_n_years` is null or not present, upload all years
* Otherwise, parse the value of the attribute as an integer and filter for the most recent N years

But I'm fine with the current solution for now, since it seems like we're still trying to get at the root of the problem and a more broad solution seems reasonable in that context.

jeancochrane (approved):
Seems like a reasonable first step!

PR #782, opened 2025-04-02
Add "Model Group" column to daily IC reference file for commercial buildings

jeancochrane:
This PR implements a quick tweak to the IC reference files requested by the IC team: Adding a new "Model Group" column to the `Com_Bldg_Details` file that we export once per day.

A few additional small changes, to support this effort:

* A new iasWorld column `comdat.user29` powers this column, so in addition to configuring the column in the `Com_Bldg_Details` query, we also add a quick entry for it in the `comdat` documentation.
* In order to match the output of the Inquire query, we add a new seed `ccao.commercial_model_group` that maps codes to human-readable descriptions for the new field.


dbt/export/templates/qc.vw_ic_reference_all_towns_com_bldg_detail.xlsx
jeancochrane, line 1: The only change to this file is adding a new "Model Group" column in the same position as the Inquire query (second-to-last column).

wrridgeway (approved):
This is great! Out of curiosity I ran the qc query to see what we already have in there:

model_group | _col1
-- | --
  | 2663417
1:517s | 15935
4:Multifamily | 13029
5:Specials | 11367
3:Industrials | 11182
2:Condos | 4511
8:Gas Stations | 1060
7:Hotels | 472
6:Nursing | 190
13:Unmodeled | 7
9:Multifamily/517s | 2

PR #787, opened 2025-04-09
Fix permit year field in open data asset tests

wrridgeway:
This PR addresses an issue with the permits view in Athena/asset on the open data portal. Because this dataset is the first we've had that can have `NULL`  year values, handling its year field can lead to difficulties we haven't encountered in the past. Specifically, the Socrata API won't return a year field at all for grouped queries where the group is `year = NULL`, and python's `sort` function cannot handle `NULL` values.

Additionally, this PR makes sure that the test script successfully links years from our Athena db that are typed as `int` to year fields from Socrata, which its API delivers as `string` regardless of being numbers.

### DIFF
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -94,12 +94,12 @@ def main() -> None:
         # numbers. It's unclear why this is the case, but it seems to be a quirk
         # of the API. We need to convert them to int so that we can compare them
         # to Athena data.
-        asset_row_counts_by_year = [
-            int(year_count[asset_year_field])
-            if year_count[asset_year_field] is not None
-            else None
-            for year_count in asset_row_counts_by_year
-        ]
+
+        for year_count in asset_row_counts_by_year:
+            if year_count[asset_year_field] is not None:
+                year_count[asset_year_field] = int(
+                    year_count[asset_year_field]
+                )
 
         dbt_output = io.StringIO()
         with contextlib.redirect_stdout(dbt_output):
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -85,9 +85,9 @@ def main() -> None:
 
         # Socrata won't return a year column for rows with no year, so we need to add a None year value to the list of rows with no year. This is a bit of a hack, but it works for now.
         [
-            year.update({"year": None})
+            year.update({asset_year_field: None})
             for year in asset_row_counts_by_year
-            if "year" not in year
+            if asset_year_field not in year
         ]
 
         dbt_output = io.StringIO()
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -84,22 +84,19 @@ def main() -> None:
         asset_row_counts_by_year = response.json()
 
         # Socrata won't return a year column for rows with no year, so we need
-        # to add None as the year value to any rows with no year.
+        # to add None as the year value to any rows with no year. Also,
+        # Socrata returns year columns as strings even though they're typed as
+        # numbers (it's not clear why the API behaves this way), so we need to
+        # convert them to int in order to compare them to Athena data.
         asset_row_counts_by_year = [
-            {asset_year_field: None, **year_count}
+            {
+                **year_count,
+                asset_year_field: int(year_count[asset_year_field])
+                if year_count.get(asset_year_field) is not None
+                else None
+            }
             for year_count in asset_row_counts_by_year
         ]
-
-        # Socrata returns year columns as strings even though they're typed as
-        # numbers. It's unclear why this is the case, but it seems to be a quirk
-        # of the API. We need to convert them to int so that we can compare them
-        # to Athena data.
-
-        for year_count in asset_row_counts_by_year:
-            if year_count[asset_year_field] is not None:
-                year_count[asset_year_field] = int(
-                    year_count[asset_year_field]
-                )
 
         dbt_output = io.StringIO()
         with contextlib.redirect_stdout(dbt_output):
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -93,7 +93,7 @@ def main() -> None:
                 **year_count,
                 asset_year_field: int(year_count[asset_year_field])
                 if year_count.get(asset_year_field) is not None
-                else None
+                else None,
             }
             for year_count in asset_row_counts_by_year
         ]
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -191,8 +191,9 @@ def diff_row_counts(
     open_data_asset_row_counts = sorted(
         open_data_asset_row_counts,
         key=lambda x: (
-            [open_data_asset_year_field] == [None],
-            [open_data_asset_year_field],
+            x[open_data_asset_year_field]
+            if x[open_data_asset_year_field] is not None
+            else 0
         ),
     )
 
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -83,7 +83,8 @@ def main() -> None:
 
         asset_row_counts_by_year = response.json()
 
-        # Socrata won't return a year column for rows with no year, so we need to add a None year value to the list of rows with no year. This is a bit of a hack, but it works for now.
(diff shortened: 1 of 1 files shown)


wrridgeway:
@jeancochrane I think we've got another problem that is making this test fail - the athena assets have year as a string while the open data assets have it as a number:

The following view/asset pairs had mismatching row counts by year:

|  year | open_data.vw_permit | Permits |
| ----- | ------------------- | ------- |
|    22 |                   1 |         |
|   122 |                   2 |         |
|   201 |                   1 |         |
|   202 |                   7 |         |
|   204 |                   1 |         |
|   210 |                   3 |         |
|   211 |                   2 |         |
|   212 |                   3 |         |
|   213 |                   1 |         |
|   220 |                   3 |         |
|   221 |                   1 |         |
|   222 |                   2 |         |
|   223 |                   1 |         |
|   224 |                   9 |         |
|   232 |                   6 |         |
|   236 |                   1 |         |
|   245 |                   1 |         |
|   250 |                   1 |         |
|   421 |                   1 |         |
|   522 |                   1 |         |
|   722 |                   1 |         |
|   723 |                  11 |         |
|   821 |                   4 |         |
|   822 |                   2 |         |
|   823 |                   2 |         |
|   920 |                   1 |         |
| 2,020 |             108,477 | 108,489 |
| 2,021 |             117,966 | 117,974 |
| 2,022 |              99,958 |  99,980 |
| 2,023 |             103,725 | 103,463 |
| 2,025 |               4,564 |   4,363 |
|    22 |                     |       1 |
|   122 |                     |       2 |
|   201 |                     |       1 |
|   202 |                     |       7 |
|   204 |                     |       1 |
|   210 |                     |       3 |
|   211 |                     |       2 |
|   212 |                     |       3 |
|   213 |                     |       1 |
|   220 |                     |       3 |
|   221 |                     |       1 |
|   222 |                     |       2 |
|   223 |                     |       1 |
|   224 |                     |       9 |
|   232 |                     |       6 |
|   236 |                     |       1 |
|   245 |                     |       1 |
|   250 |                     |       1 |
|   421 |                     |       1 |
|   522 |                     |       1 |
|   722 |                     |       1 |
|   723 |                     |      11 |
|   821 |                     |       4 |
|   822 |                     |       2 |
|   823 |                     |       2 |
|   920 |                     |       1 |

.github/scripts/test_open_data_assets.py
```diff
@@ -163,10 +180,18 @@ def diff_row_counts(
     """
     # We expect these lists to already be sorted, but double-check anyway
     athena_model_row_counts = sorted(
-        athena_model_row_counts, key=lambda x: x[athena_model_year_field]
+        athena_model_row_counts,
+        key=lambda x: (
+            [athena_model_year_field] == [None],
```
wrridgeway, line 185: We need to handle values of None or they'll cause an error when the script attempts to sort them.

wrridgeway:
https://github.com/ccao-data/data-architecture/pull/788 did not resolve the behavior I documented above. Seems like the athena cursor and the Socrata API both have quirky behavior when it comes to the year field we need to work with. Socrata does not serve it as a number through its API even if it's typed as one. Regardless, I think the changes in 788 were right, and this script is now running as expected:

image

.github/scripts/test_open_data_assets.py
```diff
@@ -163,10 +181,18 @@ def diff_row_counts(
...
+            [athena_model_year_field],
+        ),
```
jeancochrane, line 188: **[Suggestion, required]** Hmm, something weird is going on with the sort key here. Don't we want to check `x[athena_model_year_field]`, which should resolve to an integer year (or None), rather than `[athena_model_year_field]`, which should always resolve to `["year"]`? In other words, I think the current sort key is equivalent to the following expression once the interpreter resolves the relevant `year_field` variable, but this expression doesn't actually make sense because it's not reading any data from the elements in `athena_model_row_counts`:

```python
open_data_asset_row_counts = sorted(
    open_data_asset_row_counts,
    key=lambda x: (["year"] == [None], ["year"]),
)
```

The code as written currently works, but I think this is just downstream of the fact that the input data are generally already sorted: If you call `reversed()` on the object before you call `sorted()` on it, you'll find that `sorted()` returns the same object without sorting it.

Here's a proposed alternative that I think is more correct, using `0` as a fallback for null values to sort them first:

```suggestion
        athena_model_row_counts,
        key=lambda x: (
            x[athena_model_year_field]
            if x[athena_model_year_field] is not None
            else 0
    ),
```

Does that seem right to you?

.github/scripts/test_open_data_assets.py
```diff
@@ -163,10 +181,18 @@ def diff_row_counts(
...
     )
     open_data_asset_row_counts = sorted(
-        open_data_asset_row_counts, key=lambda x: x[open_data_asset_year_field]
+        open_data_asset_row_counts,
+        key=lambda x: (
+            [open_data_asset_year_field] == [None],
+            [open_data_asset_year_field],
+        ),
```
jeancochrane, line 195: **[Suggestion, required]** Similar change here as above:

```suggestion
        open_data_asset_row_counts,
        key=lambda x: (
            x[open_data_asset_year_field]
            if x[open_data_asset_year_field] is not None
            else 0
        ),
```

jeancochrane (changes requested):
Thanks for tackling this! I think the sort key is not correct, so I left a suggestion for fixing it below. Also left one optional recommendation for cleaning up the null/integer transformations. Re-request me once you've resolved these to your liking and I'll approve.

wrridgeway:
@jeancochrane I made a [small change](https://github.com/ccao-data/data-architecture/pull/787/commits/e084c37711fae4271a0c5fa4ccd95a61f37b4965) to get this to fail rather than error out. It kind of undoes one of the changes you suggested, so perhaps you have a better code style suggestion.

.github/scripts/test_open_data_assets.py
```diff
@@ -83,6 +83,24 @@ def main() -> None:
...
+        # Socrata returns year columns as strings even though they're typed as
+        # numbers. It's unclear why this is the case, but it seems to be a quirk
+        # of the API. We need to convert them to int so that we can compare them
+        # to Athena data.
+
+        for year_count in asset_row_counts_by_year:
+            if year_count[asset_year_field] is not None:
+                year_count[asset_year_field] = int(
+                    year_count[asset_year_field]
+                )
```
jeancochrane, line 102: **[Suggestion, optional]** Ah, I see now the data type mistake I made in my earlier suggestion. Unfortunately, the new solution still mutates the list while iterating it, which is what I was trying to avoid in my earlier suggestion. What do you think about this alternative solution, that also combines both loops into one? I haven't tested it but I think it should work:

```suggestion
        # Socrata won't return a year column for rows with no year, so we need
        # to add None as the year value to any rows with no year. Also,
        # Socrata returns year columns as strings even though they're typed as
        # numbers (it's not clear why the API behaves this way), so we need to
        # convert them to int in order to compare them to Athena data.
        asset_row_counts_by_year = [
            {
                **year_count,
                asset_year_field: int(year_count[asset_year_field])
                if year_count.get(asset_year_field) is not None
                else None
            }
            for year_count in asset_row_counts_by_year
        ]
```

jeancochrane (approved):
One thought on an improved loop below, but if it doesn't work as I expect, you're welcome to merge as-is!

.github/scripts/test_open_data_assets.py
wrridgeway: This works! I have a question though - is this just a list comprehension? And if it is, is there something I should know about how they work that makes them different from a for loop other than syntactically?

jeancochrane: A list comprehension is indeed just syntactic sugar around a loop; the relevant difference here is that we're using the list comprehension to return a fresh list with copied dicts, rather than looping through the list and mutating the objects in place. We could accomplish the same goal without a list comprehension like so:

```python
        cleaned_asset_row_counts_by_year = []
        for year_count in asset_row_counts_by_year:
            cleaned_year_count = year_count.copy()
            if cleaned_year_count.get(asset_year_field) is not None:
                cleaned_year_count[asset_year_field] = int(cleaned_year_count[asset_year_field])
            else:
                cleaned_year_count[asset_year_field] = None
            cleaned_asset_row_counts_by_year.append(cleaned_year_count)
```

I just think the list comprehension makes this simpler and easier to read.

Does that make sense? Happy to talk it through more if not!

PR #788, opened 2025-04-09
Change open data year columns type to INT

wrridgeway:
We transform year columns from strings to numbers for each of the assets on the open data portal. Part of having the `open_data` Athena db is to have those views represent what's on the open data portal without any transformations being done there (Socrata). It stands to reason year fields in the Athena db should also be numbers.

Having year columns as strings in our Athena dbs is an artifact of partitioning by year, rather than any design choice on our part - I shouldn't have continued this practice in the `open_data` db when it isn't necessary there.

### DIFF
--- a/dbt/models/open_data/open_data.vw_appeal.sql
+++ b/dbt/models/open_data/open_data.vw_appeal.sql
@@ -6,7 +6,7 @@ SELECT
     pin,
     class,
     township_code,
-    year,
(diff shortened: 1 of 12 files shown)


jeancochrane (approved):
Thanks for the super clear explanation of this change! Does anything need to change on the Socrata config side once we deploy this to account for the fact that the input data type is changing?

PR #789, opened 2025-04-11
Update commercial minor subclass description for co-op ➡️ live/work

jeancochrane:
Valuations changed the description for the commercial minor subclass with code 45 from `MULTIFAMILY-CO-OP` to `MULTIFAMILY-LIVE/WORK`. Since we configure these descriptions in a seed, rather than pulling them directly from iasWorld, this PR updates our seed manually to reflect this change.


wrridgeway (approved):
Thanks Jean!

PR #793, opened 2025-04-16
Add pinval tables

wagnerlmichael:
This PR tackles https://github.com/ccao-data/pinval/issues/34, and is utilized by https://github.com/ccao-data/pinval/pull/35. This PR adds two new dbt models, with the end goal of speeding up pinval data manipulation. We add various joins (like school district data) to reduce overall query and report generation time.

> [!WARNING]
> PINVAL is an experimental work-in-progress and is not yet used in
> production. Reports generated using its code may not accurately reflect
> model behavior or CCAO policy.

### `pinval.assessment_card.sql`
This table contains the data used to represent characteristics about the subject property in pinval.

### `pinval.comp.sql`
This table contains the comparable sales data and their relevant characteristics.


dbt/models/default/default.pinval_assessment_card.sql
```diff
@@ -0,0 +1,285 @@
...
+        materialized='table'
+    )
+}}
+
+WITH run_ids_to_include AS (
+    SELECT run_id
+    FROM {{ source('model', 'metadata') }}
+    -- This will eventually grab all run_ids where
+    -- run_type == comps
+    WHERE run_id = '2025-02-11-charming-eric'
```
wagnerlmichael, line 12: Placeholder for testing

dbt/models/default/default.pinval_assessment_card.sql
```diff
@@ -0,0 +1,285 @@
...
+            SPLIT(LOWER(ap.loc_property_address), ' '),
+            x -> CONCAT(UPPER(SUBSTR(x, 1, 1)), SUBSTR(x, 2))
+        ),
+        ' '
+    ) AS property_address,
+
+    school.school_elementary_district_name
+        AS loc_school_elementary_district_name,
+    school.school_secondary_district_name
+        AS loc_school_secondary_district_name
```
wagnerlmichael, line 276: Assessment pin columns added here to factor out the old `assessment_pin_df` pull from the quarto doc, which speeds up the program generation

dbt/models/default/default.pinval_assessment_card.sql
```diff
@@ -0,0 +1,285 @@
...
+    WHERE run_id = '2025-02-11-charming-eric'
+),
+
+school_data AS (
+    SELECT
+        pin10 AS school_pin,
+        year,
+        school_elementary_district_name,
+        school_secondary_district_name
+    FROM location.school
```
wagnerlmichael, line 21: Still having trouble trying to use `source()` to call this. Do I understand correctly that if the model is materialized as a table, the `{{ source('', '') }}` call should work?

dbt/models/default/schema/default.pinval_comp.yml
```diff
@@ -0,0 +1,14 @@
...
+      - unique_combination_of_columns:
+          name: default_pinval_comp_unique_run_id_pin_card_comp_num
+          combination_of_columns:
+            - run_id
+            - pin
+            - card
+            - comp_num
+          allowed_duplicates: 0
+          config:
+            error_if: ">1"
```
wagnerlmichael, line 14: This [test fails](https://github.com/ccao-data/data-architecture/actions/runs/14580754916/job/40896744099), but I'm not sure these  [these](https://github.com/ccao-data/pinval/issues/34) should be the unique keys. If this were to hold, then the same pin/card combo could not be used twice as a comp for different properties, which seems like it could be a relatively common occurrence

dbt/models/default/docs.md
```diff
@@ -236,9 +252,9 @@ is the view you're looking for.
 - `spatial.parcel` typically lags behind `iasworld.pardat` by a year, so data
   for current year can be relatively sparse or missing. Parcel shapefiles
   typically become available to populate this view at the end of each year.
-- There are some parcels in `iasworld.pardat` that have *never* been present in
+- There are some parcels in `iasworld.pardat` that have _never_ been present in
   `spatial.parcel`. These parcels are missing spatial data since they cannot be
-  spatially joined *nor* filled forward. The number of discrepancies seems to
+  spatially joined _nor_ filled forward. The number of discrepancies seems to
```
wagnerlmichael, line 257: More autoformatting

dbt/models/default/docs.md
```diff
@@ -157,12 +173,12 @@ Sourced from `iasworld.sales`, which is sourced from
...
   interested in:
-    - If you exclude sales of multiple PINs ("multisales") by filtering where
+  - If you exclude sales of multiple PINs ("multisales") by filtering where
       `not is_multisale`, sales are unique by `doc_no`
-    - If you include multisales but filter where
+  - If you include multisales but filter where
       `not sale_filter_same_sale_within_365`, sales are unique by `pin`,
       `doc_no`, and `sale_price`
-        - The reason `sale_price` is necessary here is to handle some known
+    - The reason `sale_price` is necessary here is to handle some known
```
wagnerlmichael, line 181: Autoformatting upon save. Not sure if this is what we want. Lmk and I can revert

dbt/models/default/docs.md
```diff
@@ -1,3 +1,19 @@
...
+{% docs pinval_assessment_card %}
+Table that holds card level data for subject PINs for PINVAL
+
+**Primary Key**: `run_id`, `pin`, `meta_card_num`
+{% enddocs %}
+
+# pinval_comp
+
+{% docs pinval_comp %}
+Table that holds card level data for comparables for PINVAL
```
wagnerlmichael, line 12: I'm happy to beef these up, but didn't want to spend a bunch of time on it while drawing up the first draft

dbt/models/default/default.pinval_comp.sql
```diff
@@ -0,0 +1,247 @@
...
+    FROM {{ source('model', 'metadata') }}
+    -- This will eventually grab all run_ids where
+    -- run_type == comps
+    WHERE run_id = '2025-02-11-charming-eric'
+),
+
+raw_comp AS (
+    SELECT *
+    FROM model.comp
+    WHERE run_id IN (SELECT run_id FROM run_ids_to_include)
```
jeancochrane, line 18: **[Suggestion, optional]** Same here RE: joins vs. filters:

```suggestion
    SELECT *
    FROM model.comp AS comp
    INNER JOIN run_ids_to_include AS run
        ON comp.run_id = run.run_id
```

But also, is there a compelling reason to do this filtering here, rather than doing it in a join in `comp_with_training_chars`, to more closely match `default.pinval_assessment_card`?

dbt/models/default/default.pinval_comp.sql
```diff
@@ -0,0 +1,247 @@
...
+    SELECT
+        pin,
+        card,
+        5 AS comp_num,
+        comp_pin_5,
+        comp_score_5,
+        comp_document_num_5,
+        year,
+        run_id
+    FROM raw_comp
```
jeancochrane, line 81: **[Suggestion, optional]** A little bit of Jinja magic can simplify this block:

```suggestion
    {% for i in range(1, 6) %}
        SELECT
            pin,
            card,
            {{ i }} AS comp_num,
            comp_pin_{{ i }} AS comp_pin,
            comp_score_{{ i }} AS comp_score,
            comp_document_num_{{ i }} AS comp_document_num,
            year,
            run_id
        FROM raw_comp
        {% if not loop.last %}
        UNION ALL
        {% endif %}
    {% endfor %}
```

dbt/models/default/default.pinval_comp.sql
```diff
@@ -0,0 +1,247 @@
...
+        END AS property_label,
+        -- Fancy way to capitalize the first letter of each word
+        -- since `initcap` doesn't seem to work in athena
+        ARRAY_JOIN(
+            TRANSFORM(
+                SPLIT(LOWER(train.loc_property_address), ' '),
+                x -> CONCAT(UPPER(SUBSTR(x, 1, 1)), SUBSTR(x, 2))
+            ),
+            ' '
+        ) AS property_address,
```
jeancochrane, line 117: **[Nitpick, optional]** Same here as above:

```suggestion
        train.loc_property_address AS property_address,
```

dbt/models/default/default.pinval_comp.sql
```diff
@@ -0,0 +1,247 @@
...
+        train.meta_class AS char_class,
+        train.meta_nbhd_code,
+        train.char_yrblt,
+        train.char_bldg_sf,
+        train.char_land_sf,
+        train.char_beds,
+        train.char_fbath,
+        train.char_hbath,
```
jeancochrane, line 125: **[Nitpick, optional]** Same here as above, for features that we don't need to transform at all, I think it makes sense to just return everything from `train` and leave it up to the caller to filter for predictors rather than manually specify the predictors we're interested in.

dbt/models/default/default.pinval_comp.sql
```diff
@@ -0,0 +1,247 @@
...
+    FROM pivoted_comp AS pc
+    LEFT JOIN model.pinval_test_training_data AS train
+        ON pc.comp_pin = train.meta_pin
+        AND pc.comp_document_num = train.meta_sale_document_num
+    LEFT JOIN school_data AS school
+        ON SUBSTRING(pc.comp_pin, 1, 10) = school.school_pin
+        AND train.meta_year = school.year
+)
+
+SELECT * FROM comp_with_training_chars
```
jeancochrane, line 247: **[Nitpick, optional]** If we're just going to return everything from this subquery, do we need the subquery at all? Or can we just use the `SELECT` statement that currently defines `comp_with_training_chars` as the primary `SELECT` statement in this model?

dbt/models/default/docs.md
```diff
@@ -1,3 +1,19 @@
+# pinval_assessment_card
+
+{% docs pinval_assessment_card %}
+Table that holds card level data for subject PINs for PINVAL
```
jeancochrane, line 4: **[Nitpick, optional]** As a quick style note, we typically treat these descriptions as full sentences, including punctuation:

```suggestion
Table that holds card level data for subject PINs for PINVAL.
```

dbt/models/default/docs.md
jeancochrane: **[Nitpick, optional]** Reasonable! But let's add punctuation, just in case we don't end up expanding on them in the future:

```suggestion
Table that holds card level data for comparables for PINVAL.
```

dbt/models/default/docs.md
jeancochrane: **[Praise]** All good to leave in! Thanks for flagging.

dbt/models/default/docs.md
```diff
@@ -1,3 +1,19 @@
+# pinval_assessment_card
+
+{% docs pinval_assessment_card %}
```
jeancochrane, line 3: **[Nitpick, optional]** We usually follow the pattern of prepending model docs with `table_` or `view_`, depending on the materialization:

```suggestion
{% docs table_pinval_assessment_card %}
```

Note that you'll also have to make the corresponding change to the `schema.yml` files if you accept this suggestion.

dbt/models/default/docs.md
```diff
@@ -1,3 +1,19 @@
...
+Table that holds card level data for subject PINs for PINVAL
+
+**Primary Key**: `run_id`, `pin`, `meta_card_num`
+{% enddocs %}
+
+# pinval_comp
+
+{% docs pinval_comp %}
```
jeancochrane, line 11: **[Nitpick, optional]**

```suggestion
{% docs table_pinval_comp %}
```

dbt/models/default/schema/default.pinval_assessment_card.yml
```diff
@@ -0,0 +1,13 @@
...
+    data_tests:
+      - unique_combination_of_columns:
+          name: default_pinval_assessment_card_unique_run_id_pin_card
+          combination_of_columns:
+            - run_id
+            - pin
+            - meta_card_num
+          allowed_duplicates: 0
+          config:
+            error_if: ">1"
```
jeancochrane, line 13: **[Question, optional]** Huh! What's the one row that fails this test? I wouldn't expect there to be any, but maybe there's some data we need to clean up.

dbt/models/default/default.pinval_comp.sql
```diff
@@ -0,0 +1,247 @@
...
+        train.time_sale_day_of_month,
+        train.time_sale_day_of_week,
+        train.time_sale_post_covid,
+
+        school.school_elementary_district_name
+            AS loc_school_elementary_district_name,
+        school.school_secondary_district_name
+            AS loc_school_secondary_district_name
+    FROM pivoted_comp AS pc
+    LEFT JOIN model.pinval_test_training_data AS train
```
jeancochrane, line 239: **[Nitpick, optional]** For now, let's treat this as a `source()` for consistency:

```suggestion
    LEFT JOIN {{ source('model', 'pinval_test_training_data') }} AS train
```

jeancochrane (changes requested):
Awesome start here! Some suggestions below; none are severe enough to be blocking on their own, but there's enough that I suspect you'll generate a large diff and I'll want to take a second look before merging.

Also note that I haven't reviewed the corresponding PINVAL PR in detail yet, so I may come back with additional changes if there ends up being anything in there that I think we should move into the data layer.

jeancochrane (changes requested):
This is coming along nicely! A couple suggestions below, but nothing huge. Re-request my review once you resolve comments on https://github.com/ccao-data/pinval/pull/35 and here, and I'll take a second look.

Thanks for benchmarking the various components of Quarto render time. 3-5s of irreducible overhead roughly matches my experience, and feels like a dealbreaker to me, since my back-of-the-envelope math indicates that we'd need 29 days of compute time to render 500k reports:

```
(5 * 500000) / (60 * 60 * 24) ~= 29
```

I think we'll almost certainly need to switch to a static site generator to bring this time down to a reasonable amount, so I'll work on opening up a follow-up issue to implement that refactor.

dbt/models/default/default.pinval_assessment_card.sql
wagnerlmichael: This is very clear, thank you!

dbt/models/default/default.pinval_assessment_card.sql
```diff
@@ -0,0 +1,285 @@
...
+    ac.time_sale_roll_mean_nbhd_condo_t1_w3,
+    ac.time_sale_roll_mean_nbhd_condo_t2_w1,
+    ac.time_sale_roll_mean_nbhd_condo_t2_w2,
+    ac.time_sale_roll_mean_nbhd_condo_t2_w3,
+    ac.meta_bldg_roll_mean,
+    ac.meta_pin10_bldg_roll_mean,
+    ac.meta_pin10_bldg_roll_pct_sold,
+    ac.year,
+    ac.run_id,
+    ac.township_code,
```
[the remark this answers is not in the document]
wagnerlmichael: Yep, good catch. Sounds like a big improvement to me

dbt/models/default/default.pinval_comp.sql
wagnerlmichael: Talked about this out of band and came up with a solution for now. We just removed the year param from the pinval quarto params section since `run_id` is sufficient

dbt/models/default/schema/default.pinval_assessment_card.yml
wagnerlmichael: Oops, this was an oversight!

dbt/models/default/default.pinval_assessment_card.sql
```diff
@@ -0,0 +1,57 @@
+{{
+    config(
+        materialized='table'
+    )
+}}
```
jeancochrane, line 5: **[Nitpick, required]** Two thoughts on materialization:

1. If we decide to materialize these models as tables, we should tag them with the `daily` tag so that they get built by the `build-daily-dbt-models` workflow on a regular schedule.
2. _However_, if the main purpose of these models is to enable batch generation of large numbers of PINVAL reports, then we will only query the models once per (infrequent) batch generation job, and I think it makes sense to leave them as views instead of models.

What do you think? Personally I lean towards option 2 since it means less complexity in the build process, even though it means that we won't speed up the current Quarto doc by switching to these models (since these models take ~1min to execute, similar to the current query time). But now that I'm thinking about everything again, I don't even think we need to bother switching up the Quarto docs to use these models, since we're going to deprecate those docs as soon as we deploy the new batch generation anyway.

I know this is a departure from what we discussed earlier, so I'm happy to chat it out if it'd be helpful!

dbt/models/default/default.pinval_assessment_card.sql
jeancochrane, line 1: **[Nitpick, required]** Thinking about it again, I don't see why these models should live under the `default` schema, which is just a fallback for views with no categorization system; we're already using the `pinval_` prefix for the rest of the model names, which is a sign that there _is_ in fact an underlying categorization system. Are you down to move these to a new schema `pinval` that lives under the path `dbt/models/pinval/`? That would mean this path would become `dbt/models/pinval/pinval.assessment_card.sql`. Sort of an annoying change, but I think we'll be happy we did it down the line.

jeancochrane (commented):
Well, I was expecting to give this the green light with very few changes, but when I stopped to think about it again I realized that there were a few unresolved big-picture questions 😅 Sorry for the mixup! I don't feel super strongly about these new suggestions, so if you disagree, I'm happy to chat about them.

dbt/models/default/default.pinval_assessment_card.sql
wagnerlmichael: This makes a lot of sense to me. Initially I thought we would have to be making 1000s of queries, so the table would be a big help. But if we are only querying once per batch I think the trade off makes a lot of sense to keep as views

dbt/models/default/default.pinval_assessment_card.sql
wagnerlmichael: I like this better as well

dbt/models/pinval/docs.md
```diff
@@ -0,0 +1,16 @@
+# vw_assessment_card
+
+{% docs view_vw_pinval_assessment_card %}
```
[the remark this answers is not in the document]
jeancochrane: It's a bit of a confusing pattern, but the basic idea is that doc blocks are named like `{entity_type}_{entity_name}`. So in this case, the entity type is `view`, and the entity name (i.e. tablename, in this case) is `vw_pinval_assessment_card`. This ends up being a bit confusing because `view` and `vw` appear to be duplicative, but they are in fact describing two separate components of the standardized doc block name.

dbt/models/pinval/schema.yml
```diff
@@ -0,0 +1,26 @@
...
+    description: '{{ doc("vw_pinval_comp") }}'
+    data_tests:
+      - unique_combination_of_columns:
+          name: pinval_comp_unique_run_id_pin_card_comp_num
+          combination_of_columns:
+            - run_id
+            - pin
+            - card
+            - comp_num
+          allowed_duplicates: 0
```
jeancochrane, line 12: **[Nitpick, optional]** I think `allowed_duplicates` defaults to 0, right?

https://github.com/ccao-data/data-architecture/blob/4c91fa41df07b64ad1f77ac1bb2ef90562432b55/dbt/tests/generic/test_unique_combination_of_columns.sql#L12-L17

```suggestion
```

dbt/models/pinval/schema.yml
```diff
@@ -0,0 +1,26 @@
...
+          config:
+            error_if: ">0"
```
jeancochrane, line 14: **[Nitpick, optional]** Same here, tests are configured by default to error with >0 results.

```suggestion
```

dbt/models/pinval/schema.yml
```diff
@@ -0,0 +1,26 @@
...
+    data_tests:
+      - unique_combination_of_columns:
+          name: pinval_assessment_card_unique_run_id_pin_card
+          combination_of_columns:
+            - run_id
+            - meta_pin
+            - meta_card_num
+          allowed_duplicates: 0
+          config:
+            error_if: ">0"
```
jeancochrane, line 26: **[Nitpick, optional]** Same change to fall back to defaults here:

```suggestion
```

dbt/models/pinval/schema.yml
```diff
@@ -0,0 +1,26 @@
...
+          name: pinval_comp_unique_run_id_pin_card_comp_num
+          combination_of_columns:
+            - run_id
+            - pin
+            - card
+            - comp_num
+          allowed_duplicates: 0
+          config:
+            error_if: ">0"
+  - name: pinval.vw_pinval_assessment_card
```
jeancochrane, line 15: **[Nitpick, optional]** Super nitpicky -- we should have a formatter/linter for this, to be honest -- but we typically add a newline between model definitions. Random example:

https://github.com/ccao-data/data-architecture/blob/4c91fa41df07b64ad1f77ac1bb2ef90562432b55/dbt/models/location/schema.yml#L2-L6

```suggestion

  - name: pinval.vw_pinval_assessment_card
```

dbt/models/pinval/pinval.vw_assessment_card.sql
```diff
@@ -0,0 +1,51 @@
...
+        school_secondary_district_name
+    FROM {{ ref('location.school') }}
+    WHERE year > '2014'
+),
+
+final_model_run AS (
+    SELECT
+        year,
+        SUBSTRING(run_id, 1, 10) AS final_model_run_date
+    FROM model.final_model
```
jeancochrane, line 26: **[Nitpick, optional]** Small tweak to ensure we're providing dbt the information it needs to generate the directed graph of resources:

```suggestion
    FROM {{ source('model', 'final_model') }}
```

dbt/models/pinval/pinval.vw_comp.sql
```diff
@@ -0,0 +1,71 @@
...
+        model_predictor_all_name
+    FROM {{ source('model', 'metadata') }}
+    -- This will eventually grab all run_ids where
+    -- run_type == comps
+    WHERE run_id = '2025-02-11-charming-eric'
+),
+
+raw_comp AS (
+    SELECT comp.*
+    FROM model.comp AS comp
```
jeancochrane, line 13: **[Nitpick, optional]** Same here:

```suggestion
    FROM {{ source('model', 'comp') }} AS comp
```

dbt/models/pinval/pinval.vw_assessment_card.sql
```diff
@@ -0,0 +1,51 @@
...
+    ap.pred_pin_final_fmv_round,
+    ap.loc_property_address AS property_address,
+    school.school_elementary_district_name
+        AS loc_school_elementary_district_name,
+    school.school_secondary_district_name AS loc_school_secondary_district_name,
+    run.model_predictor_all_name,
+    final.final_model_run_date
+FROM runs_to_include AS run
+INNER JOIN model.assessment_card AS ac
+    ON run.run_id = ac.run_id
```
jeancochrane, line 43: **[Thought, non-blocking]** One thing I noticed while QCing this view: By pulling all rows for the given run ID from `model.assessment_card`, we end up populating this view with parcels that do not have comps. This is because `model.assessment_card` records all card assessments whether or not the parcel is up for reassessment in the current tri, but the model pipeline only computes comps for the current tri.

I don't think this quirk of the data model is a dealbreaker, it just means that any consumers of this table need to know to filter rows for the tri corresponding to the assessment year, which requires joining to external data sources. I think it would be easy enough for us to do in the context of the view if we just filtered `model.assessment_pin.meta_triad_code` based on `model.final_model.triad_code`, which would make life easier for downstream data consumers like our static site generator. We're already joining against those tables so it shouldn't make the query any more complex. What do you think?

jeancochrane (approved):
Nice work! A couple small optional changes below. If you decide to implement any of the changes, give me one final review request once you're done and I'll take a quick final pass before we merge.

dbt/models/pinval/pinval.vw_assessment_card.sql
wagnerlmichael: Switched this to `ref` since it is defined within DBT via a sql script.

dbt/models/pinval/pinval.vw_assessment_card.sql
wagnerlmichael: Good point, this change should be done

dbt/models/pinval/pinval.vw_comp.sql
wagnerlmichael: Had to add this as an entry in the schema file

dbt/models/model/docs.md
```diff
@@ -26,6 +26,16 @@ PIN-level predicted values, etc. Predicted values are aggregated from
...
+# comp
+
+{% docs table_comp %}
+
+Table containing comparable sales along with similarity scores
+for each comparable sale.
+
+**Primary Key**: `pin`, `card`, `run_id`
+{% enddocs %}
+
```
wagnerlmichael, line 38: Added this so we can use `source`. Is this primary key right?

jeancochrane (approved):
Awesome work here! One tiny suggestion for you before you merge, but no need to re-request me before you do so, since it's pretty simple. The dupes should also now be removed from `model.comp`, so I expect the build will pass when you push a commit to tweak the docs.

PR #677, opened 2024-12-10
Lint etl subdirectories

wrridgeway:
This PR applies `styler::style_dir()` to `scripts-ccao-data-raw-us-east-1` and `scripts-ccao-data-warehouse-us-east-1`.

PR #794, opened 2025-04-17
Add Sidwell sections to data lake

wrridgeway:


.pre-commit-config.yaml
wrridgeway, line 1: @jeancochrane is there a reason we don't lint R files in this repo that you know of?

jeancochrane: Not that I'm aware of! Seems like a mistake that we should fix. I assumed we configured it when we linted the `etl/` subdir in https://github.com/ccao-data/data-architecture/pull/677 but it looks like that PR didn't update the pre-commit config. I also assumed that superlinter would handle it even if pre-commit doesn't, but a quick look at the most recent [superlinter main branch logs](https://github.com/ccao-data/data-architecture/actions/runs/14497683784/job/40669474377#step:5:845) reveals no R linting, even though [our config](https://github.com/ccao-data/data-architecture/blob/d4a57a97056cc3d534a10141829d90058f191f37/.github/workflows/lint.yaml#L36-L51) and [the superlinter docs](https://github.com/super-linter/super-linter?tab=readme-ov-file#configure-super-linter) suggest that it _should_ lint R files. This is pretty mysterious, but ultimately I don't think it matters; the R code is a core part of this repo so we should just lint it using pre-commit.

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-tax.R
```diff
@@ -85,3 +85,33 @@ districts %>%
...
+# but with sidwell's numbering system rather than the county's) and are
+# identified by the first four digits of any PIN.
+tmp_file <- tempfile(fileext = ".geojson")
+aws.s3::save_object(
+  file.path(AWS_S3_RAW_BUCKET, "spatial/tax/sidwell_grid/sidwell_grid.geojson"),
+  file = tmp_file
+)
+st_read(tmp_file) %>%
+  # Make sure to perform union on projected rather than geographic coordinate
+  # system
```
jeancochrane, line 101: **[Question, non-blocking]** Is the union in question here the `summarise()` call? And are projected coordinates important so that the buffer units are correct, or for some other reason?

etl/scripts-ccao-data-warehouse-us-east-1/spatial/spatial-tax.R
```diff
@@ -85,3 +85,33 @@ districts %>%
...
+  group_by(section) %>%
+  summarise() %>%
+  # Buffering the polygons very slightly removes a lot of odd artifacts from
+  # within them
+  st_buffer(10) %>%
+  st_buffer(-10) %>%
+  st_transform(crs = 4326) %>%
+  mutate(
+    geometry_3435 = st_transform(geometry, 3435),
+    year = "2025"
```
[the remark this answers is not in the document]
jeancochrane: **[Suggestion, optional]** Sounds good! Mind if we persist this knowledge in a comment?

```suggestion
    # This year is arbitrary -- we don't know what year these data were
    # created, so we default to the year we received them
    year = "2025"
```

jeancochrane (approved):
Awesome work! One final request before we merge: Do you mind adding an entry to [`spatial/schema.yml`](https://github.com/ccao-data/data-architecture/blob/master/dbt/models/spatial/schema.yml) to document the table name and a brief description of the table?

PR #795, opened 2025-04-21
Add r linting

wrridgeway:
For whatever reason, we haven't implemented any R code linting in this repo. Perhaps we'll realize why once this is merged in, but for now it seems like pre-commit should prevent un-formatted code from being committed to the repo.

PR also includes some linting of files that needed it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 # All available hooks: https://pre-commit.com/hooks.html
 repos:
   - repo: https://github.com/sqlfluff/sqlfluff
-    rev: 3.4.0
+    rev: 3.1.0
     hooks:
     -   id: sqlfluff-lint
     -   id: sqlfluff-fix
   - repo: https://github.com/tconbeer/sqlfmt
-    rev: v0.26.0
+    rev: v0.19.2
     hooks:
       - id: sqlfmt
         entry: >-
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
     rev: v0.4.3.9009
     hooks:
       - id: style-files
-        args: [--style_pkg=styler, --style_fun=tidyverse_style, --cache-root=styler]
+        args: [--style_pkg=styler, --style_fun=tidyverse_style]
         require_serial: true
       - id: lintr
       - id: readme-rmd-rendered
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,6 +30,17 @@ repos:
           - --fix
       # Formatter
       - id: ruff-format
+  - repo: https://github.com/lorenzwalthert/precommit
+    rev: v0.4.3
+    hooks:
+      - id: style-files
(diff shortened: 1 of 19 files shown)


wrridgeway:
Well, that was a wild ride.

.pre-commit-config.yaml
wrridgeway, line 1: Probably wouldn't be a bad idea to actually bump these deps, but they were catching so many issues it seemed best to do in a separate PR.

dbt/scripts/run_iasworld_data_tests.py
```diff
@@ -1533,7 +1533,7 @@ def get_tablename_from_node(node: typing.Dict) -> str:
     raise ValueError(
         "Can't infer tablename: Failed to parse model name from "
         f'`test_metadata.kwargs.model` attribute "{model_getter_str}" '
-        f" for test \"{node['name']}\". Inspect the dbt manifest file "
+        f' for test "{node["name"]}". Inspect the dbt manifest file '
```
wrridgeway, line 1536: This is the only non-R linting change. It's worth a double-check.

jeancochrane: **[Nitpick, optional]** Looks good! Though I noticed while testing it that the pre-existing code has a stray whitespace, we might as well fix that while we're here:

```suggestion
        f'for test "{node["name"]}". Inspect the dbt manifest file '
```

etl/scripts-ccao-data-raw-us-east-1/spatial/spatial-kriging_surfaces.R
```diff
@@ -113,7 +113,7 @@ create_ap_bbox <- function(airport_points, mult) {
 
   bbox_final <- st_buffer(bbox_sf, 5280 * mult) # 2 mile buffer
 
-  return(bbox_final)
+  bbox_final
```
jeancochrane, line 116: **[Nitpick, required]** As a disgruntled Python programmer, I find the implicit return rule to be annoying: I'm fine with implicit return in some cases, like when the return value for a function is the result of a long pipeline, but generally I think explicit return is clearer for the reader. Instead of changing all of these files to enforce explicit return, can we just adjust the [linter config](https://github.com/ccao-data/data-architecture/blob/d5c8911fd8234ab8208cd9be22c79fb9b8ad5cc5/.lintr#L1-L5) to disable `return_linter`?

jeancochrane (commented):
Thanks for this chore! My one requested change is to disable the return linter and undo the changes to files with explicit returns. Let me know if you feel strongly about keeping this linter on, and I'm willing to reconsider.

etl/scripts-ccao-data-raw-us-east-1/spatial/spatial-kriging_surfaces.R
wrridgeway: I'm fine with this. I'll update it.

wrridgeway:
Alright, `return_linter` is disabled and I reverted those changes.

jeancochrane (approved):
Excellent work!

PR #796, opened 2025-04-22
Add two new descriptions to `ccao.commercial_minor_subclass` seed

jeancochrane:
Valuations flagged for us that there are two new descriptions for the `Sub Class 2` field on the `Com Bldgs` tab in iasWorld. Since we don't currently sync these descriptions from iasWorld automatically, we need to manually add them to our seed that serves as a crosswalk between minor subclass codes and descriptions. This PR does that.


wrridgeway (approved):
Thanks Jean!

PR #802, opened 2025-05-01
Address `year` field to `int` in `open_data` db

wrridgeway:
This PR addresses two issues:

- In https://github.com/ccao-data/data-architecture/pull/788 I changed all of the open data views' `year` fields to `int` from `string` without anticipating how it would affect this script. It did, and it [caused our monthly upload to fail](https://github.com/ccao-data/data-architecture/actions/runs/14770244541/job/41469029211).
- The 'all' option for the `years` input was essentially being ignore by the script and is now properly handled

Successful run with `all` option invoked and handling `year` fields as `int` [here](https://github.com/ccao-data/data-architecture/actions/runs/14782947275/job/41505671868).

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -155,7 +155,7 @@ def parse_years_list(athena_asset, years=None):
             .as_pandas()["year"]
             .to_list()
         )
-        years_list.append(str(datetime.now().year))
+        years_list.append(datetime.now().year)
         years_list = [min(years_list)]
 
     else:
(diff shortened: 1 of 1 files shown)


socrata/socrata_upload.py
```diff
@@ -116,12 +116,15 @@ def parse_years(years=None):
...
-        ]
+        if years == "all":
+            years = ["all"]
+        else:
+            # Only allow anticipated values
+            years = [
+                re.sub("[^0-9]", "", year)
+                for year in str(years).split(",")
+                if re.sub("[^0-9]", "", year)
+            ]
```
wrridgeway, line 127: An input value of "all" was being converted to `None` previously, when it should have produced output of `["all"]` for the `parse_years` function.

PR #804, opened 2025-05-05
Upload final model training data

Damonamajor:


etl/scripts-ccao-data-warehouse-us-east-1/model/model-training_data.R
```diff
@@ -0,0 +1,85 @@
...
+library(noctua)
+library(purrr)
+library(stringr)
+
+# Once original data has been uploaded,
+# we should only need to upload the current year data.
+run_year <- format(Sys.Date(), "%Y")
+
+# Connect to Athena
+noctua_options(cache_size = 10)
```
wrridgeway, line 15: Let's scrub the cache syntax and use the unload option instead.

etl/scripts-ccao-data-warehouse-us-east-1/model/model-training_data.R
```diff
@@ -0,0 +1,85 @@
...
+  if ("ccao_is_active_exe_homeowner" %in% names(df)) {
+    df <- df %>%
+      mutate(
+        ccao_is_active_exe_homeowner =
+          as.logical(ccao_is_active_exe_homeowner)
+      )
+  }
+
+  # Add run_id after cleaning types
+  df <- df %>%
```
wrridgeway, line 77:
```suggestion
df %>%
```

etl/scripts-ccao-data-warehouse-us-east-1/model/model-training_data.R
```diff
@@ -0,0 +1,85 @@
...
+    collect()
+
+  # Ensure known type mismatches are cast consistently
+  if ("ccao_is_active_exe_homeowner" %in% names(df)) {
+    df <- df %>%
+      mutate(
+        ccao_is_active_exe_homeowner =
+          as.logical(ccao_is_active_exe_homeowner)
+      )
+  }
```
wrridgeway, line 74:
```suggestion
  df <- df %>%
    mutate(across(matches('ccao_is_active_exe_homeowner'), as.logical))
```

etl/scripts-ccao-data-warehouse-us-east-1/model/model-training_data.R
```diff
@@ -0,0 +1,85 @@
...
+    FROM model.metadata
+    WHERE run_type = 'final'
+      AND year IN ({run_year*})
+    ",
+    .con = conn
+  )
+)
+
+# Iterate through each run
+for (i in seq_len(nrow(metadata))) {
```
wrridgeway, line 39: I would suggest using our usual

```
pwalk(metadata, \(...) {
  df <- tibble::tibble(...)
```

syntax here.

wrridgeway (changes requested):
Alright, I have one key concern about this - we have really only used our etl scripts for two things in the past: loading/cleaning/transforming the data that becomes the features for the model or open data.  One of my core assumptions about this folder in the past has been that that all the scripts in here will do when we trigger them every year.

Perhaps it's time this etl folder becomes more flexible, but I'm wondering if this is where we want this script to live or if it should be integrated into the `upload` scripts in the modeling pipelines where all the other model relevant product is uploaded to aws.

Damonamajor:
@jeancochrane @wrridgeway 
Billy and I think that DBT may not be able to reference S3 (outside the DAG). Is that possible?

jeancochrane:
@Damonamajor Do you mind sending me the error message you're seeing over Teams? I suspect we may need to adjust the permissions for the Spark executor to allow it to read from S3.

Damonamajor:
Test: Switch out year >= 2023 and <= 2023 to make sure it adds as intended with new values

    base_query = """
        SELECT
          CAST(run_id                  AS STRING) AS run_id,
          CAST(year                    AS STRING) AS year,
          CAST(dvc_md5_assessment_data AS STRING) AS dvc_md5_assessment_data
        FROM model.metadata
        WHERE run_type = 'final'
        and year >= 2023
    """

Query A: Row count = 6,229,115

    base_query = """
        SELECT
          CAST(run_id                  AS STRING) AS run_id,
          CAST(year                    AS STRING) AS year,
          CAST(dvc_md5_assessment_data AS STRING) AS dvc_md5_assessment_data
        FROM model.metadata
        WHERE run_type = 'final'
        and year < 2023
    """
Query B - Appended to A: Row count = 7,788,885

Removing all years matches Query B, as well as our currently uploaded model.training_data

dbt/models/model/model.training_data.py
```diff
@@ -0,0 +1,79 @@
...
+    for _, row in metadata.iterrows():
+        run_id = row["run_id"]
+        year = int(row["year"])
+        h = row["dvc_md5_assessment_data"]
+
+        prefix = "" if year <= 2023 else "files/md5/"
+        key = f"{prefix}{h[:2]}/{h[2:]}"
+        s3p = f"{bucket}/{key}"
+
+        print(f">>> reading all columns for run {run_id!r}")
```
Damonamajor, line 59: I included print statements, but they seem to not do anything, should we drop them? Are they worth anything?

Damonamajor:
This should make the following pull request irrelevant
https://github.com/ccao-data/model-res-avm/pull/377

jeancochrane (commented):
I still need to do some more extensive QC, but here's a couple suggestions to get you started in the meantime!

dbt/models/model/model.training_data.py
```diff
@@ -0,0 +1,78 @@
...
+
+from pyspark.sql.functions import lit
+
+
+def model(dbt, session):
+    # configure incremental insert-overwrite by run_id
+    dbt.config(
+        materialized="incremental",
+        incremental_strategy="insert_overwrite",
+        partitions_by=[{"field": "run_id", "data_type": "string"}],
```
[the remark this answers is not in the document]
Damonamajor: This seems to work, but the following is also in the docs

https://github.com/dbt-labs/dbt-adapters/tree/main/dbt-athena#incremental-models

> partitioned_by (default=none)
An array list of columns by which the table will be partitioned
Limited to creation of 100 partitions (currently)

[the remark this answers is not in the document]
jeancochrane: I don't quite get the connection to the original comment, are you pointing out that the docs say the parameter should be called `partitioned_by` instead of `partitions_by`?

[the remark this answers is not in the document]
Damonamajor: I read it as having 100 created partitions `year ` x `run_id` x `township_code`, but it may be 100 variables.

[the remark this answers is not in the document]
jeancochrane: Ahh, I see. Yeah, that's an interesting ambiguity in the docs. Glad to hear at least that it works as currently written.

jeancochrane (approved):
Finally done QCing this new model! With the changes I suggested below, everything works great.

I tested out the following types of runs:

- Confirmed that running the model with no pre-existing data creates the table with all expected run IDs
- Confirmed that deleting one run ID and rerunning the model with pre-existing data only creates data for the missing run ID
- Confirmed that rerunning the model with all final run IDs changes nothing

I also confirmed that row counts and partitioning look correct in all three cases.

I'm excited to have our first incremental model, this is very cool!

dbt/models/model/model.training_data.py
```diff
@@ -0,0 +1,80 @@
...
+    metadata_df = session.sql(base_query)
+
+    if dbt.is_incremental:
+        # anti-join out any run_ids already in the target
+        existing = (
+            session.table(f"{dbt.this.schema}.{dbt.this.identifier}")
+            .select("run_id")
+            .distinct()
+        )
+        metadata_df = metadata_df.join(existing, on="run_id", how="left_anti")
```
wrridgeway, line 33: Is the if statement just part of how these models are supposed to be built? Or do we expect it to not be true under some circumstance?

wrridgeway (approved):

PR #806, opened 2025-05-05
Remove deprecated columns from permits open data view

wrridgeway:
Two columns that were removed from the Permits open data asset are still in the athena view that feeds it. There should be 22 columns in the view (21 that will show up on the open data portal + a row_id columns). This PR leaves us with 22 columns rather than 24. Output from an API update performing as expected can be found here.


jeancochrane (approved):
Oops, good catch!

PR #808, opened 2025-05-05
Fix PINVAL table configs

jeancochrane:
I noticed while looking at our source freshness tests today that [dbt is throwing some warnings related to PINVAL tables](https://github.com/ccao-data/data-architecture/actions/runs/14837513125/job/41652085074#step:4:26). It looks like when we added these tables in https://github.com/ccao-data/data-architecture/pull/793, we used the incorrect table names and docs blocks names in the `schema.yml` configuration, which was causing dbt to not know about the tables. This PR fixes those errors to resolve the warnings and correct the DAG.

See [this workflow](https://github.com/ccao-data/data-architecture/actions/runs/14841942935/job/41666771867?pr=808#step:9:43) for evidence that the DAG builds correctly, with no warnings.

### DIFF
--- a/dbt/models/pinval/docs.md
+++ b/dbt/models/pinval/docs.md
@@ -1,7 +1,6 @@
 # vw_assessment_card
 
-{% docs view_vw_assessment_card %}
-
+{% docs view_pinval_vw_assessment_card %}
 Table that holds card level data for subject PINs for PINVAL.
 
 **Primary Key**: `run_id`, `pin`, `meta_card_num`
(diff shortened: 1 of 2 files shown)


dbt/models/pinval/docs.md
```diff
@@ -1,15 +1,14 @@
 # vw_assessment_card
 
-{% docs view_vw_assessment_card %}
-
+{% docs view_pinval_vw_assessment_card %}
```
jeancochrane, line 3: Even though this isn't exactly the same pattern that we usually use, since typically we leave the schema name out of the docs block variable name, I'm using it here since I'd like to start switching to this pattern eventually.

wagnerlmichael: Sounds good, so we want to move towards `tabletype_schemaname_tablename`?

wagnerlmichael (approved):
Thanks for catching this, looks good to me.

dbt/models/pinval/docs.md
jeancochrane: Yup, that's the goal!

wrridgeway (approved):
Thanks Jean. Definitely nice to have the schema name in there.

PR #814, opened 2025-05-12
Use comp run with fixed small multicards in `pinval.vw_comp` view

jeancochrane:
In https://github.com/ccao-data/model-res-avm/pull/360 we made some changes to the comps pipeline so that comps for 2-3 card parcels better reflect the way that we value these parcels as of https://github.com/ccao-data/model-res-avm/pull/328. This PR updates the `pinval.vw_comp` view to use the comp run from this PR as the basis for comps, so that we can start using fixed comps in PINVAL development.

Connects https://github.com/ccao-data/pinval/issues/41.

### DIFF
--- a/dbt/models/pinval/pinval.vw_comp.sql
+++ b/dbt/models/pinval/pinval.vw_comp.sql
@@ -5,7 +5,7 @@ WITH runs_to_include AS (
     FROM {{ source('model', 'metadata') }}
     -- This will eventually grab all run_ids where
     -- run_type == comps
-    WHERE run_id = '2025-02-11-charming-eric'
+    WHERE run_id = '2025-04-25-fancy-free-billy'
 ),
 
 raw_comp AS (


dbt/models/pinval/pinval.vw_comp.sql
```diff
@@ -5,7 +5,7 @@ WITH runs_to_include AS (
     FROM {{ source('model', 'metadata') }}
     -- This will eventually grab all run_ids where
     -- run_type == comps
-    WHERE run_id = '2025-02-11-charming-eric'
+    WHERE run_id = '2025-04-25-fancy-free-billy'
```
jeancochrane, line 8: Note that in https://github.com/ccao-data/pinval/issues/41 we mention one problem with swapping this model run in for the final 2025 model: A very small number of cards have different predicted values between this run and the 2025 run due to instability in tree structures between model runs (https://github.com/ccao-data/model-res-avm/issues/373), so using `2025-04-25-fancy-free-billy` for characteristics will not exactly match the 2025 final model.

While this problem remains unresolved, I don't believe it affects this change, because we're only using the fixed comp run for identifying comps, which we _do_ believe to be stable between model runs. This PR doesn't touch the view containing values, which still uses the 2025 final model run:

https://github.com/ccao-data/data-architecture/blob/8d869f2cee070c5e373e57cb8668de1cc6331956/dbt/models/pinval/pinval.vw_assessment_card.sql#L1-L10

wagnerlmichael: I think that's right. Thanks for putting this down on the PR though, it's nice to have it documented.

wagnerlmichael (approved):
Looks good to me, nice catch. I'll merge this into my data architecture dev branch for pinval development when this is merged into master.

wrridgeway (approved):
Okay, just so I'm sure I'm understanding:

This PR is swapping out a new run to use as the basis for comps because the new model run (`2025-04-25-fancy-free-billy`) included the updated comps pipeline which better aligns comps with predicted values, but because our models are not entirely deterministic we don't expect predicted values to align _exactly_ across the two models.

Do I have that right?

PR #818, opened 2025-05-13
Update open data asset test to accomodate `:deleted` rows

wrridgeway:


.github/scripts/test_open_data_assets.py
```diff
@@ -110,6 +110,7 @@ def main() -> None:
                         "{"
                         f"model: '{source_model_name}', "
                         "group_by: 'year', "
+                        """where: '":deleted" is null', """
```
wrridgeway, line 113: I'm sure there's a better way to do this, but it does work. If i set
`source_model_name='z_ci_817_update_open_data_asset_test_to_accomodate_deleted_rows_open_data.vw_appeal'`
and run these lines i get:
```
[{"COUNT": 156621, "year": 1999}, {"COUNT": 230578, "year": 2000}, {"COUNT": 198419, "year": 2001}, {"COUNT": 232716, "year": 2002}, {"COUNT": 318327, "year": 2003}, {"COUNT": 288567, "year": 2004}, {"COUNT": 232970, "year": 2005}, {"COUNT": 355976, "year": 2006}, {"COUNT": 336395, "year": 2007}, {"COUNT": 315669, "year": 2008}, {"COUNT": 449028, "year": 2009}, {"COUNT": 353612, "year": 2010}, {"COUNT": 317395, "year": 2011}, {"COUNT": 392624, "year": 2012}, {"COUNT": 412460, "year": 2013}, {"COUNT": 340849, "year": 2014}, {"COUNT": 522393, "year": 2015}, {"COUNT": 446840, "year": 2016}, {"COUNT": 370217, "year": 2017}, {"COUNT": 531406, "year": 2018}, {"COUNT": 389870, "year": 2019}, {"COUNT": 238034, "year": 2020}, {"COUNT": 373569, "year": 2021}, {"COUNT": 302327, "year": 2022}, {"COUNT": 243427, "year": 2023}, {"COUNT": 440467, "year": 2024}, {"COUNT": 19308, "year": 2025}]
```

dbt/models/open_data/open_data.vw_parcel_proximity.sql
```diff
@@ -73,5 +73,6 @@ SELECT
     nearest_neighbor_2_pin10,
     nearest_neighbor_2_dist_ft,
     nearest_neighbor_3_pin10,
-    nearest_neighbor_3_dist_ft
+    nearest_neighbor_3_dist_ft,
+    CAST(NULL AS BOOLEAN) AS ":deleted" --noqa: L057
```
wrridgeway, line 77: Need this for row count test to work.

dbt/models/open_data/open_data.vw_parcel_universe_current.sql
wrridgeway, line 1: Refactored to simplify code and mirror historical view.

dbt/macros/open_data_columns.sql
```diff
@@ -10,12 +10,20 @@ The only real complication here is that feeder views can have different columns
...
+            pin || coalesce(permit_number, '') || coalesce(date_issued, ''),
+            cast(deleted_rows.row_id as varchar)
+        ) as row_id,
+    {%- elif row_id_cols is not none %}
+        coalesce(
+            cast(
+                feeder.{{ row_id_cols | join(" as varchar) || cast(feeder.") }}
+                as varchar
+            ),
+            cast(deleted_rows.row_id as varchar)
```
wrridgeway, line 26: Just making this more flexible for `row_id` construction.

dbt/macros/open_data_join_rows_to_delete.sql
wrridgeway, line 1: This logic has gotten kind of wild. I'm wondering if there's a better way to do it, but given all of the different possibilities we need to accommodate, i'm not sure anything is going to look great.

wrridgeway, line 1: This logic has gotten kind of wild. I'm wondering if it might be a better idea just to have every addn_table be its own `if` statement, but as it is this does work.

dbt/macros/open_data_columns.sql
```diff
@@ -7,15 +7,23 @@ universe along with `pardat`) and also needs to be flagged for deletion from the
...
+{%- macro open_data_columns(row_id_cols=none) -%}
     coalesce(cast(feeder.year as int), cast(deleted_rows.year as int)) as year,
-    {%- if card == true %}
+    {%- if row_id_cols is not none and "permit_number" is in row_id_cols %}
         coalesce(
-            feeder.pin || cast(feeder.card as varchar) || feeder.year,
-            deleted_rows.row_id
+            pin || coalesce(permit_number, '') || coalesce(date_issued, ''),
+            cast(deleted_rows.row_id as varchar)
+        ) as row_id,
```
jeancochrane, line 19: **[Suggestion, optional]** My personal feeling is that when you find yourself hardcoding specific edge cases into a function using conditional branching, the function is not doing its job of providing a useful abstraction for that edge case 😅 The simplest way to handle this is just to not use the function for that particular edge case; here, that would mean a little bit of duplication between this macro and the `open_data.vw_permit` view, but I think that would be better than having a conditional branch that's devoted exclusively to handling `open_data.vw_permit`.

Alternatively, we could try to think of a way of consolidating the conditional branches into one set of shared logic that works regardless of the presence of the permit columns. This is a tricky and speculative proposal, but one idea would be to swap the position of the `deleted_rows` row ID and the `feeder` row ID in the `coalesce()` call, and then `coalesce()` the `feeder` row ID components to empty strings, as we currently do for non-PIN permit columns. Something like:

```sql
coalesce(
    cast(deleted_rows.row_id) as varchar,
    {% for col in row_id_cols %}
        coalesce(feeder.{{ col }}, ''){% if not loop.last %} ||{% endif %}
    {% endfor %}
)
```

I _think_ this should work if my expectation holds that for every row, either `deleted_rows.row_id` is null _or_ all the `feeder` columns are null, but both conditions cannot be true at the same time (i.e. `deleted_rows.row_id` is null XOR `feeder.pin` is null). If that's the case, then if `deleted_rows.row_id` is not null we can be confident that all the `feeder` columns are null, and we don't have to worry about the fact that by coalescing them to strings we're making it so that the `feeder` row ID will never be null.

One risk here is that if the assumption about the XOR null relationship between `feeder` and `deleted_rows` gets violated at some point in the future, we might end up deleting rows that we don't intend to delete. I think this would be a good opportunity for a data integrity test to ensure the XOR null relationship always holds, which we could add on each view that calls this macro.

dbt/models/default/default.vw_card_res_char.sql
```diff
@@ -132,21 +112,23 @@ SELECT
     dwel.user7 AS char_air,
     dwel.user5 AS char_tp_plan
 
-FROM {{ source('iasworld', 'dweldat') }} AS dwel
-LEFT JOIN {{ source('iasworld', 'pardat') }} AS pardat
-    ON dwel.parid = pardat.parid
-    AND dwel.taxyr = pardat.taxyr
+FROM {{ source('iasworld', 'pardat') }} AS par
+INNER JOIN {{ source('iasworld', 'dweldat') }} AS dwel
```
jeancochrane, line 116: **[Question, non-blocking]** Can you explain the motivation behind this change? I can imagine why we might want to remove cards that have no representation in `pardat` at all, but I'm curious if we're doing that now because such cards exist and they're causing us problems, or if this is cleanup that only affects potential cards in theory.

dbt/macros/open_data_join_rows_to_delete.sql
jeancochrane: **[Suggestion, optional]** I'm with you, I think this level of table-specific conditional branching is a code smell that this macro is not providing a particularly useful abstraction. I thought the previous version was on the edge, but it was simple enough that it tipped onto the "useful" side; here, the `addn_table` conditions here feel solidly on the "not useful" side.

Since the conditional branching here is substantially more complicated than the `open_data_columns()` case, what do you think about reverting to the previous version of the macro that worked for most of the views, and just hardcoding the logic while accepting a bit of duplication for the views that don't fit with the old logic?

jeancochrane (approved):
I worry I may have led us astray with the suggestion to encapsulate the join to `deleted_rows` in a macro 😅 But other than that component, I think this PR is actually pretty solid! Happy to talk through any of my bigger-picture suggestions related to the macros if it'd be useful.

PR #821, opened 2025-05-16
Resolve SSL errors and use Spark DataFrames for `model.training_data`

jeancochrane:
This PR makes some tweaks to the draft `model.training_data` model that's currently under construction in https://github.com/ccao-data/data-architecture/pull/804. The PR makes two main changes:

1. Use [`SparkSession.read.parquet()`](https://spark.apache.org/docs/latest/sql-data-sources-parquet.html#loading-data-programmatically) to read files from DVC instead of the pyarrow `S3FileSystem` interface, since the latter has issues with SSL cert verification in S3 ([see this SO thread for details](https://stackoverflow.com/questions/76924941/disable-ssl-validation-while-connecting-to-s3-using-pyarrow-fs-library-s3fs-lib))
2. Use the [native Spark DataFrame class](https://spark.apache.org/docs/latest/api/python/getting_started/quickstart_df.html) for dataframe operations, since it's faster and pandas was raising errors when trying to load all the data into memory

Once this PR is merged into the branch that forms the basis of #804, I think the final step is to implement the incremental filtering logic as outlined in https://github.com/ccao-data/data-architecture/pull/804#issuecomment-2877107415.

### DIFF
--- a/dbt/models/model/model.training_data.py
+++ b/dbt/models/model/model.training_data.py
@@ -1,6 +1,6 @@
-import pandas as pd
-import pyarrow.dataset as ds
-import pyarrow.fs as fs
+import functools
+
+from pyspark.sql.functions import lit
 
 
 def model(dbt, session):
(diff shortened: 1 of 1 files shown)


jeancochrane:
Merged, thanks @Damonamajor!

PR #807, opened 2025-05-05
Test open data row deletion fix

wrridgeway:
High-level summary: This PR grabs all rows that have been deactivated or re-classed to 999 (parcel status view allows 999s since they are explicitly indicated in the view) in pardat (and occasionally other tables that define a view's universe) and outer joins them to `open_data` views so that uploads to socrata will include all rows that should not be present on the open data portal.

Note: this update will break our current open data testing syntax. we will have to make sure to count only rows from open_data athena views `where ":deleted" is null`

Pay special attention to the sf/mf view. sf/mf now removes 999s and deactivated `pardat` parcels at the `default` view level which were sneaking in previously, so there is a slight disagreement with the old open data view. For this view I've compared the new open data view to the new `default.vw_card_res_char`, instead of the old open data view. [Otherwise, the open data test workflow is returning what we'd hope for it to return.](https://github.com/ccao-data/data-architecture/actions/runs/15000276948/job/42147705471)

<details>

<summary>check row_id uniqueness</summary>

```
select 'vw_assessed_value' as view_name, row_id, count(*)
from z_ci_fix_open_data_row_deletion_open_data.vw_assessed_value
group by row_id
having count(*) > 1

union all

select 'vw_parcel_address' as view_name, row_id, count(*)
from z_ci_fix_open_data_row_deletion_open_data.vw_parcel_address
group by row_id
having count(*) > 1

union all

select 'w_parcel_status' as view_name, row_id, count(*)
from z_ci_fix_open_data_row_deletion_open_data.vw_parcel_status
group by row_id
having count(*) > 1

union all

select 'vw_parcel_universe_current' as view_name, row_id, count(*)
from z_ci_fix_open_data_row_deletion_open_data.vw_parcel_universe_current
group by row_id
having count(*) > 1

union all

select 'vw_parcel_universe_historical' as view_name, row_id, count(*)
from z_ci_fix_open_data_row_deletion_open_data.vw_parcel_universe_historical
group by row_id
having count(*) > 1

union all

select 'vw_property_tax_exempt_parcel' as view_name, row_id, count(*)
from z_ci_fix_open_data_row_deletion_open_data.vw_property_tax_exempt_parcel
group by row_id
having count(*) > 1

union all

select 'vw_res_condo_unit_char' as view_name, row_id, count(*)
from z_ci_fix_open_data_row_deletion_open_data.vw_res_condo_unit_char
group by row_id
having count(*) > 1

union all

[...]

PR #827, opened 2025-05-28
Simplify `open_data_rows_join_rows_to_delete` macro

wrridgeway:
The purpose here is to remove some of the complexity of the `open_data_rows_join_rows_to_delete` by performing the joins necessary to populate our `open_data` views with `:deleted` rows directly in those view scripts rather trying to abstract away every single one of them. Most views still employ the macro, but 4 now do not since their joins are rather complex and it's both useful to see the code explicitly and have the macro not be a minefield of conditional logic.  All of the queries below produce 0 rows, meaning the code changes have not altered the content of the views.

I more or less copy and pasted code from `dbt compile` into the views that have been changed, so there isn't a ton to see in the views. The macro has been gutted.

## Views adjusted directly

<details>

<summary>vw_appeal</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_appeal
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_appeal
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_parcel_sale</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_parcel_sale
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_parcel_sale
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_permit</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_permit
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_permit
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_sf_mf_improvement_char</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_sf_mf_improvement_char
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_sf_mf_improvement_char
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

## Views adjusted indirectly

<details>

<summary>vw_assessed_value</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_assessed_value
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_assessed_value
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_parcel_status</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_parcel_status
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_parcel_status
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_parcel_address</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_parcel_address
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_parcel_address
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_parcel_universe_historical</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_parcel_universe_historical
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_parcel_universe_historical
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_property_tax_exempt_parcel</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_property_tax_exempt_parcel
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_property_tax_exempt_parcel
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

<details>

<summary>vw_res_condo_unit_char</summary>

```
with new as (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as new
	from z_ci_826_simplify_open_data_rows_join_rows_to_delete_macro_open_data.vw_res_condo_unit_char
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
old AS (
	select case
			when ":deleted" then true else false
		end as ":deleted",
		case
			when year is null then '' else cast(year as varchar)
		end as year,
		count(*) as old
	from open_data.vw_res_condo_unit_char
	group by case
			when year is null then '' else cast(year as varchar)
		end,
		":deleted"
),
matches as (
	select new.":deleted",
		new.year,
		new.new,
		old.old,
		new.new = old.old as match
	from new
		full outer join old on new.year = old.year
		and new.":deleted" = old.":deleted"
	order by year desc
)
select *
from matches
where not match
	or match is null
```

</details>

### DIFF
--- a/dbt/macros/open_data_join_rows_to_delete.sql
+++ b/dbt/macros/open_data_join_rows_to_delete.sql
@@ -5,11 +5,9 @@ Macro that can selectively add:
 - non-condo class
 rows to the open data views, as well as rows from tables other than pardat so
 that a ":deleted" flag associated with their row_id can be sent to the open data
-portal.
+portal. Currently this can only be "owndat".
 
-There are multiple complications here:
-- Feeder views can have different columns that define row_id.
-- The universe of parcels that might need to be purged from the open data assets
+The universe of parcels that might need to be purged from the open data assets
 is different for different feeder views. The macro takes arguments to specify
 how to construct the approriate universe of rows to purge.
 */
--- a/dbt/macros/open_data_join_rows_to_delete.sql
+++ b/dbt/macros/open_data_join_rows_to_delete.sql
@@ -17,31 +17,21 @@ how to construct the approriate universe of rows to purge.
     full outer join
         (
             select
-                {% if addn_table == "dweldat" %}
-                    pdat.parid || cast(addndat.card as varchar) || pdat.taxyr as row_id,
-                {% else %} pdat.parid || pdat.taxyr as row_id,
-                {% endif %}
+                pdat.parid || pdat.taxyr as row_id,
                 pdat.taxyr as year,
                 true as ":deleted"
             from {{ source("iasworld", "pardat") }} as pdat
-            {% if addn_table is not none %}
-                {% if addn_table == "dweldat" %} inner join
-                {% elif addn_table == "owndat" %} left join
-                {% endif %}
+            {% if addn_table == "owndat" %}
+                left join
                     {{ source("iasworld", addn_table) }} as addndat
                     on pdat.parid = addndat.parid
                     and pdat.taxyr = addndat.taxyr
             {% endif %}
             where
                 pdat.deactivat is not null
-                {% if addn_table == "dweldat" %} or addndat.deactivat is not null
-                {% elif addn_table == "owndat" %} or addndat.ownnum is null
-                {% endif %}
+                {% if addn_table == "owndat" %} or addndat.ownnum is null {% endif %}
                 {% if condo == true %} or pdat.class not in ('299', '399') {% endif %}
                 {% if allow_999 == false %} or pdat.class = '999' {% endif %}
         ) as deleted_rows
-        {% if addn_table == "dweldat" %}
-            on feeder.pin || cast(feeder.card as varchar) || feeder.year
-        {% else %} on feeder.pin || feeder.year
-        {% endif %} = deleted_rows.row_id
+        on feeder.pin || feeder.year = deleted_rows.row_id
 {% endmacro %}
--- a/dbt/macros/open_data_join_rows_to_delete.sql
+++ b/dbt/macros/open_data_join_rows_to_delete.sql
@@ -26,7 +26,7 @@ how to construct the approriate universe of rows to purge.
             from {{ source("iasworld", "pardat") }} as pdat
             {% if addn_table is not none %}
                 {% if addn_table == "dweldat" %} inner join
-                {% else %} left join
+                {% elif addn_table == "owndat" %} left join
                 {% endif %}
                     {{ source("iasworld", addn_table) }} as addndat
                     on pdat.parid = addndat.parid
@@ -34,10 +34,9 @@ how to construct the approriate universe of rows to purge.
             {% endif %}
             where
                 pdat.deactivat is not null
-                {% if addn_table == "dweldat" %}
-                    or addndat.deactivat is not null
+                {% if addn_table == "dweldat" %} or addndat.deactivat is not null
+                {% elif addn_table == "owndat" %} or addndat.ownnum is null
                 {% endif %}
-                {% if addn_table == "owndat" %} or addndat.ownnum is null {% endif %}
                 {% if condo == true %} or pdat.class not in ('299', '399') {% endif %}
                 {% if allow_999 == false %} or pdat.class = '999' {% endif %}
         ) as deleted_rows
--- a/dbt/macros/open_data_join_rows_to_delete.sql
+++ b/dbt/macros/open_data_join_rows_to_delete.sql
@@ -16,35 +16,7 @@ how to construct the approriate universe of rows to purge.
 {% macro open_data_join_rows_to_delete(allow_999=false, condo=false, addn_table=none) %}
     full outer join
         (
-            {% if addn_table == "htpar" %}
-                -- htpar is by far the most complex here since row_id's can show
-                -- up multiple times, sometimes being deactivated and sometimes
-                -- not.
-                with
-                    entire_case as (
-                        select
-                            addndat.parid || addndat.taxyr || addndat.caseno as row_id,
-                            addndat.taxyr as year,
-                            true as ":deleted",
-                            avg(
-                                case
-                                    when addndat.deactivat is not null then 1 else 0
-                                end
-                            ) as deactivat
-                        from {{ source("iasworld", addn_table) }} as addndat
-                        left join
-                            {{ source("iasworld", "pardat") }} as pdat
-                            on addndat.parid = pdat.parid
-                            and addndat.taxyr = pdat.taxyr
-                        where addndat.caseno is not null or pdat.deactivat is not null
-                        group by
-                            addndat.parid || addndat.taxyr || addndat.caseno,
-                            addndat.taxyr
-                    )
-                select row_id, year, ":deleted"
-                from entire_case
-                where deactivat = 1
-            {% elif addn_table == "sales" %}
+            {% if addn_table == "sales" %}
                 select
                     salekey as row_id, substr(saledt, 1, 4) as year, true as ":deleted"
                 from {{ source("iasworld", addn_table) }}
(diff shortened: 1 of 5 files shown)


jeancochrane:
Thanks for tackling this change! Overall I think this approach is indeed simpler than what we added in #807. However, now that I'm seeing the full solution in all its glory, I'm starting to have second thoughts about whether we're moving in the right direction more broadly.

My biggest concern is that our current approach to determining which rows need to be deleted in an upsert relies on our ability to decide which rows in iasWorld tables have been deactivated. As this PR demonstrates, that decision is actually quite complicated, and it varies depending on which iasWorld asset we're looking to upsert. An important consequence of this complexity is that anyone who needs to maintain this pipeline in the future will need to have as confident of an understanding of the iasWorld deactivation data model as we do. I'm not even all that confident in my own understanding of this particular data model, and I suspect you might not be either; I definitely wouldn't trust someone with less context than either of us to make any changes to it.

All of this leads me to wonder: **Is there a way we can figure out which rows to delete with significantly less complexity?** My ideal solution would be for Socrata to offer an API that allows us to fully replace one year of data in a single request, but if [the docs](https://dev.socrata.com/publishers/soda-producer/soda-producer-basics) are complete, it doesn't seem like they offer that capability. Still, I'm thinking that perhaps we could approximate it.

Here's one half-baked idea: Instead of determining which rows to delete by using the iasWorld data model to find deactivated rows, what if we were to anti-join the current state of the latest year of each Socrata asset to the current state of the corresponding Athena table? Then, we could add a `:deleted` element to the request payload for any records that are in Socrata but not Athena, and push it along with the full state of the Athena table to Socrata.

I haven't sketched out how this might work given the current state of the Socrata upload script, so I think it's possible that this proposal would be even more complex than the approach we're currently using in this PR. Maybe we think that complexity is prohibitive, but maybe not! At the end of the day, I think it's better to add complexity to the upload script than to the data model, since I think a new staff member could read and understand a Python script much faster than they could understand the iasWorld deactivation data model. As long as that complexity is reasonable, of course.

Let me know if you want to talk this through in person, I expect that we'll need to clarify a bunch of edge cases before we decide to jump into a different approach.

wrridgeway:
I really like this idea. Right off the bat my instinct is to make our `row_id` columns in socrata visible so that we can grab them using the api. This would allow us to avoid having to construct them ourselves for the socrata data by pulling multiple columns and combining them in ways that are asset specific. It would be a minor change to the assets that would save us a lot of complexity and pain, but I'm open to the `row_id`s staying hidden if we think that's important. I would move them to the end of the assets and make sure to describe the columns as not for end-users but for maintenance.

We can sit on this PR for now and I can spin up a new one testing this proposed solution.

wrridgeway:
Deprecated via https://github.com/ccao-data/data-architecture/pull/829

PR #830, opened 2025-06-04
Remove `:deleted` conditional from `test_open_data_assets.py`

wrridgeway:
We no longer use this column in athena. It now causes the test to error.


jeancochrane (approved):
Nice catch!

PR #836, opened 2025-06-11
Switch test bucket to prod training data

wagnerlmichael:
Draft
Likely will be picked up in https://github.com/ccao-data/data-architecture/issues/855


jeancochrane:
Resolved in #855.

PR #840, opened 2025-06-11
Add 2019 and 2020 model values to data lake

wrridgeway:


etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-dtbl_modelvals.R
jeancochrane, line 1: **[Question, non-blocking]** Why not store this table under the `model` schema instead of `ccao`? I.e. `model.dtbl_modelvals` instead of `ccao.dbtl_modelvals`?

jeancochrane (changes requested):
Nice and simple! Before we merge, would you mind taking a stab at some docs for this table? I'm thinking:

- A simple table description explaining the provenance and time period for the data, along with the primary key that defines unique rows
- A quick `unique_combinations_of_columns` test to ensure uniqueness by the primary key
- A brief mention in the [`model.assessment_pin` table description](https://github.com/ccao-data/data-architecture/blob/7e3a8bc5bfaff535d82816ec21a5621616451af2/dbt/models/model/docs.md?plain=1#L17-L27) noting that 2019 and 2020 model values are available in this table
- (Optional, if you're feeling ambitious) Brief descriptions of each column
    - I expect you may not know what each column means, but I also suspect you may have the most knowledge of any of us and preserving that knowledge could be useful 

Also, I'm curious if you have thoughts about the distribution of counts across years? I'm surprised there are so many more values for 2019 compared to 2020.

```sql
select tax_year, count(*) from ccao.dtbl_modelvals group by 1
```

| tax_year |   count |
| --------- | -------- |
| 2019      | 425736 |
| 2020      | 663471 |

jeancochrane (approved):
Nice work, thanks for the clear docs!

PR #842, opened 2025-06-12
Mask sensitive AWS identifiers in GitHub workflow logs

jeancochrane:
This PR updates the `setup_dbt` composite action to add a [masks](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#masking-a-value-in-a-log) for sensitive AWS identifiers. This is a best practice to ensure that these sensitive identifiers never get accidentally printed to logs in the course of a dbt command or direct AWS CLI operation. All of our workflows that interact with AWS resources use the `setup_dbt` composite action, so by masking identifiers in the context of the action, we should get good coverage for protecting our logs.

See here for a log line showing the mask: https://github.com/ccao-data/data-architecture/actions/runs/15620684692/job/44004516960?pr=842#step:3:219

### DIFF
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -62,6 +62,7 @@ jobs:
         uses: ./.github/actions/setup_dbt
         with:
           role-to-assume: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
+          aws-account-id: ${{ secrets.AWS_ACCOUNT_ID }}
           role-duration-seconds: 14400
 
       # Set these env vars as secrets so they get masked in the GitHub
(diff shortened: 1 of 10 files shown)


.github/workflows/socrata_upload.yaml
```diff
@@ -73,8 +74,6 @@ jobs:
 
       - name: Upload to Socrata
         env:
-          AWS_IAM_ROLE_TO_ASSUME_ARN: ${{ secrets.AWS_IAM_ROLE_TO_ASSUME_ARN }}
-          AWS_ACCOUNT_ID: ${{ secrets.AWS_ACCOUNT_ID }}
```
jeancochrane, line 77: I noticed that these env vars were being passed into the script but not used, so I removed them.

PR #843, opened 2025-06-17
Fix school district geoid

wrridgeway:
@jeancochrane found two school districts with the same census geoids in `spatial.school`. Turns out I screwed up manually assigning geoids due to an incomplete join in the `spatial-school.R` warehouse script. Got the new geoid from tidycensus:
```
tidycensus::get_acs(
  geography = "school district (elementary)",
  table = "B01001",
  survey = "acs5",
  output = "wide",
  state = "IL",
  year = 2023,
  cache_table = TRUE
)
```
image

Also found some random errors with the raw cps attendance data in 2011 that we were failing to address - one school number and one school name.

```
select geoid,
	year,
	count(*)
from spatial.school_district
where geoid is not null
group by geoid,
	year
having count(*) > 1
```

shows we're good now.


jeancochrane (approved):
Thanks for getting to the bottom of this!

PR #848, opened 2025-06-24
Add pred_per_sqft col

wagnerlmichael:
This PR adds `pred_card_initial_fmv_per_sqft` to `pinval.wv_assessment_card`, which makes building the front matter for pinval simpler

### DIFF
--- a/dbt/models/pinval/pinval.vw_assessment_card.sql
+++ b/dbt/models/pinval/pinval.vw_assessment_card.sql
@@ -31,6 +31,8 @@ school_districts AS (
 SELECT
     ac.*,
     ap.pred_pin_final_fmv_round,
+    ac.pred_card_initial_fmv / ac.char_bldg_sf
+        AS pred_card_initial_fmv_per_sqft,
     ap.loc_property_address AS property_address,
     tw.township_name,
     CONCAT(CAST(ac.char_class AS VARCHAR), ': ', cd.class_desc)


dbt/models/pinval/pinval.vw_assessment_card.sql
```diff
@@ -31,6 +31,8 @@ school_districts AS (
 SELECT
     ac.*,
     ap.pred_pin_final_fmv_round,
+    CAST(ROUND(ac.pred_card_initial_fmv / ac.char_bldg_sf, 0) AS INTEGER)
+        AS pred_card_initial_fmv_per_sqft,
```
jeancochrane, line 35: **[Suggestion, required]** A bit more defensiveness will help guard against division-by-zero errors, which I'm currently getting when I try to query this column from the staging view:

```suggestion
    CASE
        WHEN ac.char_bldg_sf = 0 THEN NULL
        ELSE CAST(ROUND(ac.pred_card_initial_fmv / ac.char_bldg_sf, 0) AS INTEGER)
    END AS pred_card_initial_fmv_per_sqft,
```

Query I used to surface the error:

```sql
select pred_card_initial_fmv_per_sqft
from z_ci_add_pred_price_per_sqft_to_vw_assessment_card_pinval.vw_assessment_card
where char_bldg_sf = 0
```

```
NUMERIC_VALUE_OUT_OF_RANGE: Out of range for integer: Infinity
```

jeancochrane (changes requested):
One small concern about division-by-zero errors! Re-request me once you've resolved it and I'll do final QC 👍🏻

dbt/models/pinval/pinval.vw_assessment_card.sql
jeancochrane: Or actually, maybe we can just match the way the `pinval.vw_comp` view does it using `NULLIF()`?

https://github.com/ccao-data/data-architecture/blob/1058c826d3849bbe64124fc37c47c2c62ed152e1/dbt/models/pinval/pinval.vw_comp.sql#L69-L70

wagnerlmichael: I went with the `NULLIF` route.

jeancochrane (approved):
Nice work! We'll have to keep in mind that this variable can be null in some rare cases when we build it into the report.

PR #850, opened 2025-06-25
Add zoning to data lake

Damonamajor:


wrridgeway:
Current next steps:

- Ingest [municipal shapefile](https://hub-cookcountyil.opendata.arcgis.com/datasets/534226c6b1034985aca1e14a2eb234af_2/explore) into qc report
- use it to coalesce(tax muni, spatial join muni) for all pins (inlcuding new pins), leaving anything that is still null as unincorporated
- plot that outcome against the unincorporated shapefile
- create a map comparing the municipal shapefile and the unincorporated shapefile in order to determine how much or little they agree
- try to figure out a way to compare differences between municipalities per pin based on tax munis and shapefile munis - this may be a little tricky since they probably won't have the same names and the tax muni data isn't polygons so we can't just do an spatial intersection, union or something like that

wrridgeway:
@Damonamajor this looks great so far, we can sit on it for a bit while we wait for expected data.

Damonamajor:
@wrridgeway Let's review this during this sprint. Doesn't seem to be a fixed timeframe on getting the other geographies.

PR #851, opened 2025-06-27
Tweak `pinval.vw_assessment_card` to support PINVAL error pages

jeancochrane:
This PR supports the changes in https://github.com/ccao-data/pinval/pull/79, allowing us to deploy error pages for non-tri and non-regression-class PINs.

The PR makes the following changes to the `pinval.vw_assessment_card` view:

- Makes `default.vw_pin_universe` the base of the query so that PINs that are not in `model.assessment_card` still get reports
- Adds new columns to indicate report eligibility and explain ineligible PINs
  - `is_report_eligible` (boolean) 
  - `reason_report_ineligible` (string)
- Adds new columns to provide extra context when explaining ineligible PINs
  - `parcel_township_code` (string)
  - `parcel_township_name` (string)
  - `parcel_triad_name` (string)
  - `parcel_class` (string)
  - `parcel_class_description` (string)

It also adds a bunch of data integrity tests to `pinval.vw_assesment_card` to verify our assumptions about the data model.

Connects https://github.com/ccao-data/pinval/issues/51.


dbt/models/model/schema.yml
```diff
@@ -6,6 +6,14 @@ sources:
...
         tags:
           - load_auto
+        columns:
+          - name: meta_card_num
+            data_tests:
+              - not_null:
+                  name: model_assessment_card_meta_card_num_not_null
+                  config:
+                    where: meta_class != '299'
+                    error_if:  ">560"
```
jeancochrane, line 16: I noticed this tiny data error in the process of QCing the report eligibility reasons for this PR. Not totally sure where it comes from yet, but I'm adding this data integrity test while I'm here so that we remember to fix it when it crops up in future modeling seasons.

dbt/models/pinval/columns.md
```diff
@@ -0,0 +1,121 @@
...
+
+## model_run_id
+
+{% docs column_pinval_model_run_id %}
+Run ID for the model run associated with this card and its values.
+
+Prefer this to `run_id`, which comes from `model.assessment_card`,
+because `run_id` will be null if the parcel is ineligible for a report
+for this model run. In contrast, this column will never be null.
+{% enddocs %}
```
wagnerlmichael, line 33: **[Suggestion, non-blocking]:** It took me a good chunk of time to wrap my head around a component of the data model regarding `model_run_id`. 

The big point of confusion for me was that a row of a condo PIN for example would have `model_run_id` attached to a res model `run_id`. 

I now understand that this join helps us facilitate the inclusion of `vw_pin_universe`, and helps us build out a logic for the `is_report_eligible` structure.

```sql
INNER JOIN runs_to_include AS run
    ON uni.year = run.assessment_data_year
```

This strikes me as potentially confusing. I think either extended docs here or a column name change could potentially be useful.

Some ideas:
- Something that would tell the developer that this column is constructed essentially as a static set for a given pinval year interface.
- That this column allows the user a full set of all of the PINs for the downstream user to access helpful error pages.
- Maybe an explicit mention that the `model_run_id` doesn't link a PIN to a model run within this table, but the `run_id` does. 

What do you think?

dbt/models/pinval/schema.yml
```diff
@@ -12,10 +12,88 @@ models:
...
+              name: pinval_vw_assessment_card_reason_report_ineligible_not_non_tri_for_tri
+              values:
+                - non_tri
+              config:
+                where: LOWER(parcel_triad_name) = LOWER(assessment_triad_name)
+          - not_accepted_values:
+              name: pinval_vw_assessment_card_reason_report_ineligible_not_unknown
+              values:
+                - unknown
+
```
wagnerlmichael, line 92: **[Praise]:** Great thinking on all of these tests

wagnerlmichael (approved):
A few questions and a suggestion, but otherwise this looks good. Great work! I agree with your sentiment about the pesky `ac*` call, thanks for making an issue for that.

dbt/models/pinval/columns.md
jeancochrane: Good points all around! I agree that it's super confusing. I beefed up the docs in 438d47fea65ad782f6372b72a1f68c320c1b6e74 to explain some of the issues you pointed out above.

Since we're going to refactor all of this in https://github.com/ccao-data/data-architecture/issues/855 anyway, I'm not inclined to put a lot more effort into clarifying the column name. I've bumped the priority of #855 to make sure we actually get the refactor done in time to avoid confusion for more team members. Does that seem like a reasonable path forward to you?

dbt/models/pinval/columns.md
```diff
@@ -0,0 +1,121 @@
...
+
+## parcel_township_code
+
+{% docs column_pinval_parcel_township_code %}
+Township code for the card's parcel.
+
+See `parcel_class` for details on the difference between parcel classes and
+card classes in the context of this view.
+{% enddocs %}
```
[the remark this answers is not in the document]
jeancochrane: Yup, that's it! We'll be able to get rid of these confusing column names in #855, but in the meantime, you've got it right.

> And this is necessary because the PIN's class and a card's class within that PIN could be different, right?

That's true, although our motivation is slightly different: We mainly want to be able to access a class for the parcel in order to show eligibility errors, even if that parcel is ineligible in a run and so is missing from `model.assessment_card`. Without `parcel_class`, we could only rely on `meta_class` or `char_class`, both of which would be missing for ineligible parcels. We'll change this up in #855 to something like `select coalesce(pin_universe.class, assessment_card.char_class) as char_class` so that we don't need a dedicated `parcel_class` column.

dbt/models/pinval/pinval.vw_assessment_card.sql
```diff
@@ -3,6 +3,7 @@ WITH runs_to_include AS (
         run_id,
         model_predictor_all_name,
         assessment_year,
+        assessment_data_year,
```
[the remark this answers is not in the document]
jeancochrane: That's right! We want to display e.g. 2024 characteristics for the 2025 model.

PR #853, opened 2025-06-30
Add new frankencard related columns to pinval.vw_assessment_card

wagnerlmichael:
Adds the following columns to make pinval frontmatter generation simpler
### `pinval.vw_assessment_card`
- `is_parcel_small_multicard`
- `is_frankencard`
- `combined_bldg_sf`
### `pinval.vw_comp`
- `combined_bldg_sf`

Connects https://github.com/ccao-data/pinval/pull/78


dbt/models/pinval/pinval.vw_assessment_card.sql
```diff
@@ -38,9 +38,29 @@ SELECT
     )
         AS pred_card_initial_fmv_per_sqft,
     ap.loc_property_address AS property_address,
+    CAST(ap.meta_pin_num_cards AS INTEGER) AS ap_meta_pin_num_cards,
```
wagnerlmichael, line 41: Interestingly, calling `ac.meta_pin_num_cards` here gives us an NA value. Upon brief inspection, there is a large amount of `meta_pin_num_cards` data missing in `model.assessment_card`, it seems like it might be stratified by `run_id`. I'm not sure why this is. The `model.assessment_pin` data seems to have this data, so we are temporarily using that, although this is not the best solution.

Let me know if there is a different strategy you would use to grab the `ap.meta_pin_num_cards`.

Perhaps we should also make an issue about the missingness in `model.assessment_card`, unless @jeancochrane or @wrridgeway know why this is off the top of their heads

wagnerlmichael: Interesting, it must have been because the card_num was actually used since we probably incorporated multi-card PINs in the training data in that run. That plan sounds excellent, I have no problem with deleting it.

jeancochrane: > it must have been because the card_num was actually used since we probably incorporated multi-card PINs in the training data in that run

Is there any chance we'd want to preserve that run for the sake of reminding ourselves how that run performed compared to the final approach we used?

jeancochrane: Hmm, sounds like the table might indeed be useful. Let's table this for now.

jeancochrane (approved):
Looks good to me! We'll refine this further when we pick up the deeper refactor after https://github.com/ccao-data/pinval/pull/78 lands.

PR #857, opened 2025-07-14
Add `not_null` test for PINVAL card num field to support better char table UX

jeancochrane:
In https://github.com/ccao-data/pinval/pull/93, we're changing the PINVAL reports based on the assumption that report-eligible PINs will always have not-null card numbers (see in particular [this comment](https://github.com/ccao-data/pinval/pull/93/files#r2208624634)). This PR adds a data test to enforce that assumption so we will be notified if we ever break it.


wagnerlmichael (approved):
Good idea to add this!

wrridgeway (approved):

PR #858, opened 2025-07-15
Add `loc_property_city` to `pinval.vw_assessment_card`

wagnerlmichael:
Add a municipality column ref for downstream pinval outputs.

Connects https://github.com/ccao-data/pinval/pull/92


jeancochrane (approved):
Awesome 👍🏻

PR #860, opened 2025-07-22
Add remote logging and error reporting to daily IC reference file export script

wrridgeway:
Logging in Python is clearer than R, and we had previous examples to pull from. I borrowed heavily from the [spark ingest script](https://github.com/ccao-data/service-spark-iasworld/blob/main/src/submit_jobs.py) and associated [utils](https://github.com/ccao-data/service-spark-iasworld/tree/main/src/utils). I tried to keep it as clean and simple as possible, which meant stripping out a bunch of functionality that exists in the spark repo that we won't need here.

AWS Notes:

- Created [ic_reference_file_export](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fccao$252Fjobs$252Fic_reference_file_export) log group and a dynamically named log stream for CloudWatch, and added a [metric filter](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fccao$252Fjobs$252Fic_reference_file_export/edit-metric-filter/ICReferenceFileExportDailyWriteError) for errors.
- Added SNS alarms [ccao-ic_reference_file_export-job-error](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#alarmsV2:alarm/ccao-ic_reference_file_export-job-error) and [ccao-ic_reference_file_export-job-not-run](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#alarmsV2:alarm/ccao-ic_reference_file_export-job-not-run)

I tested `ccao-ic_reference_file_export-job-error` with a deliberate error and it was triggered as expected.

### DIFF
--- a/dbt/scripts/export_models.py
+++ b/dbt/scripts/export_models.py
@@ -92,9 +92,9 @@ if __name__ == "__main__":
         log_group_name=args.log_to_cloudwatch_group,
     )
 
-    logger.info("Starting model export")
-
     try:
+        logger.info("Starting model export")
+
         export_models(
             args.target,
             args.select,
@@ -107,12 +107,8 @@ if __name__ == "__main__":
         logger.info("Export completed successfully.")
 
     except Exception as e:
-        # The CloudWatch log handler does not include "ERROR - " at the
-        # beginning of log lines, so we need to preface the message with
-        # "ERROR - " to trigger alarms.
-        logger.error(f"ERROR - {e}")
-
-        # Suppress the error if we're uploading to Cloudwatch so that we can
-        # perform cleanup below, but otherwise, raise the error
-        if args.log_to_cloudwatch_group is None:
-            raise
+        logger.error(
+            str(e) or f"{type(e).__name__} raised with no message",
+            exc_info=True,
+        )
+        raise
--- a/dbt/scripts/utils/helpers.py
+++ b/dbt/scripts/utils/helpers.py
@@ -28,14 +28,25 @@ def create_logger(
     if log_file_path is not None and os.path.exists(log_file_path):
         os.remove(log_file_path)
 
-    # Create and start the default stream logger
-    logging.basicConfig(
-        level=logging.INFO,
-        format="%(asctime)s - %(levelname)s - %(message)s",
-        datefmt="%Y-%m-%d_%H:%M:%S.%f",
-        force=True,
+    formatter = logging.Formatter(
+        fmt="%(asctime)s.%(msecs)03d - %(levelname)s - %(message)s",
+        datefmt="%Y-%m-%d_%H:%M:%S",
     )
+
+    # Create logger and add custom handlers
     logger = logging.getLogger(name)
+    logger.setLevel(logging.INFO)
+    # Clear handlers in case we accidentally call this function multiple times
+    logger.handlers.clear()
+
+    # Always log to the console
+    stream_handler = logging.StreamHandler()
+    stream_handler.setFormatter(formatter)
+    logger.addHandler(stream_handler)
+
+    # Turn off log propagation to the root logger to avoid duplication with
+    # the stream handler
+    logger.propagate = False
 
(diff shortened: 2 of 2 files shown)


dbt/scripts/utils/aws.py
```diff
@@ -0,0 +1,49 @@
...
+            logEvents=log_events,
+        )
+        print("Successfully uploaded log file to CloudWatch")
+
+        # Remove the log file after successful upload
+        os.remove(log_file_path)
+        print(f"Successfully removed log file: {log_file_path}")
+
+    except Exception as e:
+        print(f"Failed to upload log file to CloudWatch: {e}")
```
jeancochrane, line 49: **[Suggestion, optional]** It would be cool if we could somehow write these log lines to a file, since they're most likely to run as part of an automated job and so will get lost if we just print them to the console. But that's a little bit tricky, since we intend to remove the logfile upon successful execution of this function.

Maybe we configure this function to log to `log_file_path`, remove the log line "Successfully removed log file: [...]" which would otherwise be writing to a deleted file, and accept that the logs will A) not be in CloudWatch and B) usually not be accessible since the logfile will get deleted? While A and B might be slightly confusing behavior for future users/maintainers, I suspect that this confusion will be offset by the benefit we'll get if we ever run into a situation where the CloudWatch logs fail to upload, in which case `os.remove(log_file_path)` will not get called and we'll still have the local copy of the logfile to inspect to see what went wrong.

I also think in this particular case we should `raise e` after logging the line so that the script exits on failure, which is how I would expect a script to behave when it encounters a critical failure like this.

jeancochrane (changes requested):
I like this new approach using watchtower @wrridgeway! Much simpler. Two big-picture requests before I do a more detailed review:

- This interface works well for automated jobs that call `export_models`, but I think it comes at the expense of the UX when a user is running it manually in a terminal (which we do occasionally). A user running the script in the terminal almost certainly won't want to log to CloudWatch, and probably won't want to log to a file either; they'll just want to see logs print to the console for the duration of the script run. Here's my vision for how we can align the automated UX with the manual UX:
    -  Add two new script flags to configure logging, both disabled by default:
        - `--log-to-file` (string): When present, registers a [`FileHandler`](https://docs.python.org/3/library/logging.handlers.html#logging.FileHandler) to log to the file path as represented by the argument value
        - `--log-to-cloudwatch-group` (string): When present, registers a `CloudWatchLogHandler` to log to the group name as represented by the argument value
    - Regardless of the presence or absence of the new logging flags, ensure the script logs to the console
        - I expect this should Just Work without any extra configuration if you register the `FileHandler` and `CloudWatchLogHandler` instances using [`Logger.addHandler()`](https://docs.python.org/3/library/logging.html#logging.Logger.addHandler), but if it doesn't, you may have to manually register a [`StreamHandler`](https://docs.python.org/3/library/logging.html#logging.Logger.addHandler) to get the logger to continue logging to stdout even in the presence of a file handler and/or CloudWatch handler
- Let's make sure we add `watchtower` as a dependency to the [`dbt_tests` optional dependency in `dbt/pyproject.toml`](https://github.com/ccao-data/data-architecture/blob/ae10574e4dca78ace73ac255085b789a31bb0160/dbt/pyproject.toml#L12) and update `dbt/uv.lock` so the dependency gets persisted
    - I think the correct incantation to install the dependency and simultaneously update the lockfile is `uv add --optional dbt_tests watchtower`, but I'm not totally sure -- refer to [the docs](https://docs.astral.sh/uv/concepts/projects/dependencies/#adding-dependencies) if that doesn't work

jeancochrane (approved):
OK, I remember my in person comment! But on second thought, I think it's unnecessarily complicated. I'm happy merging what we have now and exploring an enhancement later on if we decide we need one.

PR #868, opened 2025-08-04
Fix open data asset test year field conversion

wrridgeway:


.github/scripts/test_open_data_assets.py
```diff
@@ -91,7 +91,10 @@ def main() -> None:
         asset_row_counts_by_year = [
             {
                 **year_count,
-                asset_year_field: int(year_count[asset_year_field])
+                # Convert year field to float first to handle cases where
+                # Socrata passes a string representation of a float, e.g.
```
jeancochrane, line 95: **[Nitpick, optional]** I think this would be slightly clearer:

```suggestion
                # Socrata passes a float representation of an integer, e.g.
```

jeancochrane (approved):
Thanks for tracking this down! Super strange. I double-checked [a raw API response](https://datacatalog.cookcountyil.gov/resource/y282-6ig3.json?$query=SELECT%20*%20WHERE%20pin%20=%20%2704091050480000%27%20and%20year%20=%202025) and confirmed that the weird floats appear in the API response regardless of whether you issue the query in Python or not.

I think the root problem here is that the Socrata [Number datatype](https://dev.socrata.com/docs/datatypes/number.html) does not distinguish between floats and integers, so presumably all these years are stored in their database as floats (or some other form of decimal) and then their API serialization code is responsible for truncating trailing zeros when returning a JSON response. It's weird that some years get returned as integers and some as floats with a trailing zero, however; I can't think of a reason why they would be different.

Would you be open to raising a support ticket with Socrata to ask whether this is a bug or expected behavior? It doesn't need to block this PR, but I think it's worth checking to see if they can fix this upstream. I suspect they'll tell us that it's expected behavior due to the lack of a dedicated integer type, but if there's any chance it's a bug on their side, it would be awesome to get it fixed. I suspect this limitation of the API will also be frustrating for the end users of our data and an upstream fix would be the best way to help them, too.

wrridgeway:
Sure thing. I'll open a ticket with them, and we can revert this PR if they make any sort of changes that address our issue.

PR #869, opened 2025-08-06
Fix formatting problem for yamllint

### LINKED ISSUE #835: Fix YAML linting for dbt config files
We have yamllint configured on this repo via superlinter, but it doesn't seem to be catching indentation problems in dbt config files. You can test this out by introducing extraneous indents and confirming that superlinter doesn't flag them. Let's see if we can figure out why not, and potentially move it into pre-commit so we can catch problems faster.

wagnerlmichael:
Here is one situation where it seems like an explicit declaration of indentation rules catches a spacing problem where it previously didn't, confirmed by CI checks:
- Error that wasn't flagged by super linter: https://github.com/ccao-data/data-architecture/pull/869/commits/530a27e666b7bcdbb8c37a895eb691df84628f92
- Error flagged upon rule implementation: https://github.com/ccao-data/data-architecture/pull/869/commits/683c83cb01e504f48ecafcfd8d7d3c63d0a70f43

More formatting problem tests that seem to be already getting caught: https://github.com/ccao-data/data-architecture/pull/869/commits/07b115b1efbc6ee30a148829a94ec339c0fc7f58

Closes #835

### DIFF
--- a/.yamllint.yml
+++ b/.yamllint.yml
@@ -1,2 +1,4 @@
 rules:
   document-start: disable
+  indentation:
+    spaces: 2
--- a/dbt/models/default/schema/default.vw_pin_condo_char.yml
+++ b/dbt/models/default/schema/default.vw_pin_condo_char.yml
@@ -31,7 +31,7 @@ models:
       - name: char_unit_sf
         description: '{{ doc("shared_column_char_unit_sf") }}'
       - name: char_yrblt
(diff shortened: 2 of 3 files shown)


wagnerlmichael:
Let me know if the overlooked formatting problems that motivated the creation of https://github.com/ccao-data/data-architecture/issues/835 are still happening/what they are and I can keep looking into those.

PR #874, opened 2025-08-07
Simplify IC reference file logging config

jeancochrane:
This PR adds a few suggestions on top of https://github.com/ccao-data/data-architecture/pull/860 to improve remote logging for the IC reference file export script. Python logging is annoying and confusing but I think with these changes we'll be in good shape.

See the log stream for today for the three example scenarios I tested (script completes successfully, script encounters an error with a message, script encounters an error without a message).

### DIFF
--- a/dbt/scripts/export_models.py
+++ b/dbt/scripts/export_models.py
@@ -92,9 +92,9 @@ if __name__ == "__main__":
         log_group_name=args.log_to_cloudwatch_group,
     )
 
-    logger.info("Starting model export")
-
     try:
+        logger.info("Starting model export")
+
         export_models(
             args.target,
             args.select,
@@ -107,12 +107,8 @@ if __name__ == "__main__":
         logger.info("Export completed successfully.")
 
     except Exception as e:
-        # The CloudWatch log handler does not include "ERROR - " at the
-        # beginning of log lines, so we need to preface the message with
-        # "ERROR - " to trigger alarms.
-        logger.error(f"ERROR - {e}")
-
-        # Suppress the error if we're uploading to Cloudwatch so that we can
-        # perform cleanup below, but otherwise, raise the error
-        if args.log_to_cloudwatch_group is None:
-            raise
+        logger.error(
+            str(e) or f"{type(e).__name__} raised with no message",
+            exc_info=True,
+        )
+        raise
--- a/dbt/scripts/utils/helpers.py
+++ b/dbt/scripts/utils/helpers.py
@@ -28,14 +28,25 @@ def create_logger(
     if log_file_path is not None and os.path.exists(log_file_path):
         os.remove(log_file_path)
 
-    # Create and start the default stream logger
-    logging.basicConfig(
-        level=logging.INFO,
-        format="%(asctime)s - %(levelname)s - %(message)s",
-        datefmt="%Y-%m-%d_%H:%M:%S.%f",
-        force=True,
+    formatter = logging.Formatter(
+        fmt="%(asctime)s.%(msecs)03d - %(levelname)s - %(message)s",
+        datefmt="%Y-%m-%d_%H:%M:%S",
     )
+
+    # Create logger and add custom handlers
     logger = logging.getLogger(name)
+    logger.setLevel(logging.INFO)
+    # Clear handlers in case we accidentally call this function multiple times
+    logger.handlers.clear()
+
+    # Always log to the console
+    stream_handler = logging.StreamHandler()
+    stream_handler.setFormatter(formatter)
+    logger.addHandler(stream_handler)
+
+    # Turn off log propagation to the root logger to avoid duplication with
+    # the stream handler
+    logger.propagate = False
 
     # Log to CloudWatch if desired
     if log_group_name is not None:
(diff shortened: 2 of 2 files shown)


dbt/scripts/export_models.py
```diff
@@ -92,9 +92,9 @@ def parse_args():
         log_group_name=args.log_to_cloudwatch_group,
     )
 
-    logger.info("Starting model export")
-
     try:
+        logger.info("Starting model export")
```
jeancochrane, line 96: I figure we might as well try to include as much as possible in the `try`/`except` block so we can catch and log as many errors as we can.

dbt/scripts/export_models.py
```diff
@@ -107,12 +107,8 @@ def parse_args():
         logger.info("Export completed successfully.")
 
     except Exception as e:
-        # The CloudWatch log handler does not include "ERROR - " at the
-        # beginning of log lines, so we need to preface the message with
-        # "ERROR - " to trigger alarms.
-        logger.error(f"ERROR - {e}")
```
jeancochrane, line 113: Now that we're pushing `levelname` to CloudWatch as part of all log messages (see comments below), everything that `logger.error()` pushes will include the string `ERROR`, so we no longer need to add it manually.

dbt/scripts/export_models.py
```diff
@@ -107,12 +107,8 @@ def parse_args():
...
-
-        # Suppress the error if we're uploading to Cloudwatch so that we can
-        # perform cleanup below, but otherwise, raise the error
-        if args.log_to_cloudwatch_group is None:
-            raise
+        logger.error(
+            str(e) or f"{type(e).__name__} raised with no message",
```
jeancochrane, line 111: The CloudWatch handler won't push a log line if it doesn't include a message, so we need to catch the case where there is no message to make sure all errors get pushed to CloudWatch. In the real world (i.e. outside of my testing) we should never encounter an unhandled exception with no error, but we might as well be extra defensive while we're here.

dbt/scripts/export_models.py
```diff
@@ -107,12 +107,8 @@ def parse_args():
...
+            exc_info=True,
```
jeancochrane, line 112: Setting [`exc_info=True`](https://docs.python.org/3/library/logging.html#logging.Logger.debug) ensures that we push tracebacks to our logs, which will help for debugging errors in the future.

dbt/scripts/export_models.py
```diff
@@ -107,12 +107,8 @@ def parse_args():
...
     except Exception as e:
-        # The CloudWatch log handler does not include "ERROR - " at the
-        # beginning of log lines, so we need to preface the message with
-        # "ERROR - " to trigger alarms.
-        logger.error(f"ERROR - {e}")
-
-        # Suppress the error if we're uploading to Cloudwatch so that we can
-        # perform cleanup below, but otherwise, raise the error
-        if args.log_to_cloudwatch_group is None:
-            raise
```
jeancochrane, line 118: We're no longer doing any special cleanup, so we don't need this conditional branch anymore, we can just ensure all callers `raise`.

dbt/scripts/utils/helpers.py
```diff
@@ -28,14 +28,25 @@ def create_logger(
     if log_file_path is not None and os.path.exists(log_file_path):
         os.remove(log_file_path)
 
-    # Create and start the default stream logger
-    logging.basicConfig(
-        level=logging.INFO,
-        format="%(asctime)s - %(levelname)s - %(message)s",
-        datefmt="%Y-%m-%d_%H:%M:%S.%f",
-        force=True,
```
jeancochrane, line 36: Something I learned while trying to figure out why the log `levelname` wasn't being included in CloudWatch: This `basicConfig` only applies to the root (default) handler, not to any handlers that you add to the logger later on. I decided to refactor to use an explicit `StreamHandler` just to make this clearer.

dbt/scripts/utils/helpers.py
```diff
@@ -28,14 +28,25 @@ def create_logger(
...
+    formatter = logging.Formatter(
+        fmt="%(asctime)s.%(msecs)03d - %(levelname)s - %(message)s",
+        datefmt="%Y-%m-%d_%H:%M:%S",
```
jeancochrane, line 33: Note a slight difference in the datetime format here: Python's logging module uses `strftime` under the hood, which does not support the `%f` microseconds directive. I pulled this alternative from [this SO answer](https://stackoverflow.com/a/31335874).

dbt/scripts/utils/helpers.py
```diff
@@ -45,10 +56,18 @@ def create_logger(
             log_group_name=log_group_name,
             stream_name=stream_name,
         )
+
+        # Don't push timestamps to CloudWatch, since CloudWatch adds its own
+        # timestamps
+        cw_handler.setFormatter(
+            logging.Formatter(fmt="%(levelname)s - %(message)s")
+        )
```
jeancochrane, line 64: Previously we had not been formatting the logs that we push to CloudWatch (or to the local file), which meant the log level name (like `INFO` or `ERROR) was not being included in the CloudWatch logs. This meant that we had to add the string `Error` manually during our error handling in order to trigger our metrics filter alarm. By adding this formatter to the CloudWatch log handler, we can ensure we always push the log level name, and avoid having to add a custom string to trigger the metrics filter alarm.

Note that we define a custom formatter here and skip the timestamp that we use in the stream and file handlers because CloudWatch adds its own timestamps to log lines.

wrridgeway (approved):
This is great. If I'm understanding correctly, other than even further simplifying the whole logging scripting, the primary issue this addresses in my PR is how the actual logs were being formatted and how level names were being excluded?

PR #875, opened 2025-08-07
Add SHAPs to `pinval.vw_assessment_card` so that we can compute feature weights

jeancochrane:
This PR adds SHAP values for all predictors to `pinval.vw_assessment_card` so that we can use them to compute feature weights to use for ranking features in the final reports.

In my first attempt at this PR, I computed the feature weights and rankings in SQL, which works but causes the query to increase from ~20s to ~10m in execution time. The SQL code was also quite confusing due to Athena's lack of first-class support for pivot/unpivot operations. Instead, I decided to refactor things so that the view just returns SHAPs, and leave it up to some simple pandas code in the report generation Python script in https://github.com/ccao-data/pinval/pull/113 to compute the weights and rankings.

Connects https://github.com/ccao-data/pinval/issues/86.

### DIFF
--- a/dbt/macros/all_predictors.sql
+++ b/dbt/macros/all_predictors.sql
@@ -10,7 +10,7 @@
 --
 -- If the `alias_prefix` parameter is set to a string, it will alias each
 -- predictor with the specified prefix prepended to the predictor name.
--- For example, `prefix='shap'` will return column aliases following the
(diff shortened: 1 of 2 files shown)


wagnerlmichael (approved):
Nice work, just left a couple questions/thoughts.

PR #879, opened 2025-08-18
ModVal data model tweaks ahead of full deploy

jeancochrane:
Companion PR to https://github.com/ccao-data/pinval/pull/121.

Complete list of changes:

- Computes `pinval.vw_assessment_card.combined_bldg_sf` for all PINs, regardless of multicard status
- Adds new field `pinval.vw_assessment_card.pred_pin_final_fmv_round_per_sqft` based on `combined_bldg_sf`, to use in displaying final estimates in units of $/SF
- Adds an exception to `pinval.vw_assessment_card.is_report_eligible` for one specific 2024 PIN that got values but no comps
- Ensures that `pinval.vw_assessment_card.meta_nbhd_code` is never null by coalescing it with `default.vw_pin_universe.nbhd_code` and adding a data integrity test
- Fixes a couple of errors in the `pinval.vars_dict` seed:
    - Adds missing row for `loc_env_airport_noise_dnl`
    - Removes trailing empty column
    - Removes extra set of double-quotes in `loc_env_airport_noise_dnl` and `prox_airport_dnl_total`
    - Removes description for `char_tp_dsgn`, which was not adding any information
    - Removed stray space in the string "OpenStreet Map" in the description for `prox_nearest_secondary_road_dist_ft`
- Adds new data test to confirm that the `pinval.vars_dict` seed documents all of the predictors in every instance of `pinval.vw_assessment_card.model_predictor_all_name`


dbt/models/pinval/schema.yml
```diff
@@ -126,7 +127,7 @@ models:
                 # There are a handful of known PINs that had the wrong class
                 # at modeling time, and so got a model value even though it
                 # should not have
-                error_if: ">29"
+                error_if: ">28"
```
jeancochrane, line 130: Previously this test was capturing the 2024 PIN that we are now manually setting to `is_report_eligible = false` (see above), so now we can drop the threshold to reflect that that particular PIN has been manually fixed.

wrridgeway (approved):
I don't see anything extraneous here.

Issue #862, opened 2025-07-25
re-add spatial.municipality data


Damonamajor:
# Data Onboarding

After filling out this issue template: onboard the data, close the issue, and add the necessary information to the department's [data catalog](https://gitlab.com/groups/ccao-data-science---modeling/-/wikis/Data/data_catalog.xlsx).

### Description

As found in https://github.com/ccao-data/enterprise-intelligence/issues/297, we are going to revert cook_municipality data to ensure that municipal file is accurate

### Origin

- [ ] Internal

spatial.municipality

PR #880, opened 2025-08-18
Add combined_municipality to location_fill

Damonamajor:
Follow-up for https://github.com/ccao-data/data-architecture/issues/862

### DIFF
--- a/dbt/models/location/location.vw_pin10_location.sql
+++ b/dbt/models/location/location.vw_pin10_location.sql
@@ -92,7 +92,7 @@ SELECT
             THEN tax.tax_municipality_name
 
         -- tax is NULL; if cook contains UNINCORPORATED -> empty array
-        WHEN political.cook_municipality_name IS NOT NULL
+        WHEN tax.tax_municipality_name IS NULL
             AND political.cook_municipality_name = 'UNINCORPORATED'
             THEN CAST(ARRAY[] AS ARRAY<VARCHAR>)
 


wrridgeway (approved):
Great work. I did some tests that suggest everything is working as intended:

```
select pin10,
	year
from z_ci_add_combined_municipality_to_pin10_fill_location.vw_pin10_location_fill
where combined_municipality_name != tax_municipality_name
	and array_join(combined_municipality_name, ', ') != cook_municipality_name
```

```
select pin10,
	year
from z_ci_add_combined_municipality_to_pin10_fill_location.vw_pin10_location_fill
where combined_municipality_name is null
	and (
		tax_municipality_name is not null
		or cook_municipality_name is not null
	)
```

We may want to look into whether we need to join on the crosswalk for each `cook_` column rather than just do it once.

dbt/models/location/location.vw_pin10_location.sql
```diff
@@ -93,8 +93,7 @@ SELECT
 
         -- tax is NULL; if cook contains UNINCORPORATED -> empty array
         WHEN political.cook_municipality_name IS NOT NULL
```
Damonamajor, line 95: Do we need to explicetely state that tax is NULL here?

wrridgeway: I don't think so, no. Go ahead and update it and we'll double-check everything is still working as intended.

Damonamajor: I think it's just for code clarity if we want it. I added it in and ran the following with no differences
```

SELECT 
    COALESCE(a.pin10, b.pin10)    AS pin10,
    COALESCE(a.year, b.year)      AS year,
    a.combined_municipality_name  AS combined_muni_dev,
    b.combined_municipality_name  AS combined_muni_loc
FROM z_dev_damajor_location.vw_pin10_location a
FULL OUTER JOIN location.vw_pin10_location b
    ON a.pin10 = b.pin10
   AND a.year  = b.year
WHERE a.combined_municipality_name IS DISTINCT FROM b.combined_municipality_name
ORDER BY pin10, year;

```

Issue #878, opened 2025-08-15
Remove negative `asmt` values test for unused column on `asmt` table (`val48`)


jeancochrane:
Res Val says they don't need QC tests or workbooks for negative values in `asmt.tot48` anymore anymore, since they refer to fields that are not used in CCAO processes. As such, we can strip out the following models and tests:

- Delete test `qc_vw_neg_asmt_value_val48_gte_0`:

https://github.com/ccao-data/data-architecture/blob/8ba5d15a724cabfb423a44af34919e5845e6a7da/dbt/models/qc/schema.yml#L860-L867

- Get rid of `val48` from `qc.vw_report_town_close_neg_asmt_value` model:

https://github.com/ccao-data/data-architecture/blob/8ba5d15a724cabfb423a44af34919e5845e6a7da/dbt/models/qc/qc.vw_report_town_close_neg_asmt_value.sql#L6

PR #883, opened 2025-08-20
Remove `val48` tests from negative `asmt` value QC workbooks

jeancochrane:
This PR edits a few QC views and removes a test to reflect the fact that Res Val doesn't want to test the `asmt.val48` field anymore because it is not used by the CCAO.

Closes https://github.com/ccao-data/data-architecture/issues/878.


wrridgeway (approved):
RIP val48.

PR #887, opened 2025-08-25
Bump AWS R deps in etl

wrridgeway:
Just bumping `aws.s3`, `paws` and a few other minor package versions to avoid the annoying `get_bucket_df` error we've gotten in the past.

### DIFF
--- a/etl/renv.lock
+++ b/etl/renv.lock
@@ -1,6 +1,6 @@
 {
   "R": {
-    "Version": "4.4.1",
+    "Version": "4.4.2",
     "Repositories": [
       {
         "Name": "CRAN",
@@ -291,7 +291,7 @@
     },
     "aws.s3": {
       "Package": "aws.s3",
-      "Version": "0.3.21",
+      "Version": "0.3.22",
       "Source": "Repository",
       "Repository": "CRAN",
       "Requirements": [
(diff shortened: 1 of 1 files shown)


etl/renv.lock
```diff
@@ -2056,156 +2056,157 @@
...
-      "Hash": "4a768547ec143a4a0742b33929010da4"
+      "Hash": "6cde06aeae9f2a6f193886a7756afa7f"
     },
     "paws.common": {
       "Package": "paws.common",
-      "Version": "0.7.2",
+      "Version": "0.8.5",
       "Source": "Repository",
-      "Repository": "CRAN",
+      "Repository": "RSPM",
```
jeancochrane, line 2085: **[Thought, non-blocking]** I wonder if we should also switch to the `https://paws-r.r-universe.dev` repo here, both for consistency and so that we can pick up [the latest version](https://paws-r.r-universe.dev/paws.common)? I don't have strong feelings about it, just curious what you think.

etl/renv.lock
```diff
@@ -1420,9 +1420,9 @@
     },
     "httr2": {
       "Package": "httr2",
-      "Version": "1.0.1",
+      "Version": "1.2.1",
       "Source": "Repository",
-      "Repository": "CRAN",
+      "Repository": "RSPM",
```
jeancochrane, line 1425: **[Question, non-blocking]** Any reason to switch to RSPM here from CRAN? I don't really understand the difference between the two, so I don't have strong feelings, but I'm curious if this was an intentional choice or a byproduct of automated `renv` behavior.

jeancochrane (approved):
Looks great, thanks for picking up this chore!

etl/renv.lock
wrridgeway: Just automated `renv` behavior.

PR #894, opened 2025-09-03
Anticipate different year values between Athena and Socrata assets

wrridgeway:
Addressing an issue where we get new years for data in Athena which don't already exist in Socrata, or years are deleted from Athena and the corresponding rows need to be removed from Socrata. Before, this would error out trying to find a `row_id` column in an empty dataframe if the year value wasn't present on Socrata, or would leave rows undeleted on Socrata if they had a year value that had been removed from Athena.

This hasn't caused us an issue yet since we've almost exclusively been using the script to update the present year (2025) of data for each asset, and when I tested the script updating all years in the past enough time hadn't passed between runs for there to be differences in year values.

[This run of the workflow](https://github.com/ccao-data/data-architecture/actions/runs/17444977323/job/49537092859) successfully handled `year = 205` which was causing the error [here](https://github.com/ccao-data/data-architecture/actions/runs/17441950840/job/49526909163) (205 had been added to athena but wasn't present in socrata), but ended up failing when athena got nuked.

[Here's a run](https://github.com/ccao-data/data-architecture/actions/runs/17476185532/job/49636504132) I initialized after solving the data is on athena but not on socrata issue. There was still a one row difference between the athena and socrata assets due to a year value that no longer existed in athena. The run found the row on socrata and removed it.

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -21,7 +21,7 @@ pd.set_option("display.max_rows", None)
 
 # Create a session object so HTTP requests can be pooled
 session = requests.Session()
-session.headers.update({"X-App-Token": os.getenv("SOCRATA_APP_TOKEN")})
+session.headers.update({"X-App-Token": str(os.getenv("SOCRATA_APP_TOKEN"))})
 session.auth = (
     str(os.getenv("SOCRATA_USERNAME")),
     str(os.getenv("SOCRATA_PASSWORD")),


socrata/socrata_upload.py
```diff
@@ -275,12 +275,15 @@ def check_deleted(input_data, asset_id, app_token):
...
-    input_data = input_data.drop(columns=["_merge"])
+    if len(socrata_rows) > 0:
+        input_data = input_data.merge(
+            socrata_rows, on="row_id", how="outer", indicator=True
+        )
+        input_data[":deleted"] = None
+        input_data.loc[input_data["_merge"] == "right_only", ":deleted"] = True
+        input_data = input_data.drop(columns=["_merge"])
+    else:
+        input_data[":deleted"] = None
```
wrridgeway, line 286: Sometimes we get new years in Athena and they don't exist in Socrata. Before this would error out trying to find a `row_id` in an empty dataframe.

socrata/socrata_upload.py
```diff
@@ -207,7 +207,6 @@ def build_query_dict(athena_asset, asset_id, years=None):
         .split(",")
     )
     # row id won't show up here since it's hidden on the open data portal assets
-    asset_columns += ["row_id"]
```
wrridgeway, line 210: This is no longer needed since we exposed `row_id` on the open data assets.

wrridgeway:
@jeancochrane I think we both forgot about this guy

socrata/socrata_upload.py
```diff
@@ -314,7 +367,14 @@ def upload(asset_id, sql_query, overwrite):
...
         # Ensure rows that need to be deleted from Socrata are marked as such
-        input_data = check_deleted(input_data, asset_id, app_token)
+        if not overwrite:
+            input_data = check_deleted(input_data, asset_id, app_token)
+            # If there are years present on Socrata that are not in the current
+            # upload, add them to the final input data so they can be deleted
+            if year == list(sql_query.keys())[-1]:
+                input_data = pd.concat(
+                    [input_data, missing_years], ignore_index=True
+                )
```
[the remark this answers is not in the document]
jeancochrane: I opened up an issue for this in https://github.com/ccao-data/data-architecture/issues/922.

PR #912, opened 2025-10-20
Update ETL folder R dependencies

wrridgeway:
Bumping R dependencies so they'll build with R 4.5.1 and Rtools 4.5 on windows. I was able to successfully build this environment locally on windows. I paid special attention to packages with `RemoteRepo` entries, particularly `geoarrow` and `tabulizer` as we are using deprecated versions of those packages that should not be updated in the lock file.

Also worth noting renv will want to change this lockfile on windows unless `options(renv.lockfile.version = 1)` is added to the user-level `.RProfile` file. 

EDIT - this user-level .rprofile fix doesn't seem to work for rstudio or positron. The presence of project-level .Rprofile files will lead to user-level files [not being sourced unless the project-level file includes `source("~/.Rprofile")`](https://docs.posit.co/ide/user/ide/guide/environments/r/managing-r.html#rprofile).

I'm not sure where this leaves us. It seems rstudio server is more respectful of higher-level .Rprofile code. For local development we need to either include the `source("~/.Rprofile")` command in the project-level .Rprofile, or the renv options themselves.

### DIFF
--- a/etl/renv.lock
+++ b/etl/renv.lock
@@ -2095,9 +2095,12 @@
     },
     "paws": {
       "Package": "paws",
-      "Version": "0.9.0",
+      "Version": "0.10.0",
       "Source": "Repository",
-      "Repository": "CRAN",
+      "Repository": "https://paws-r.r-universe.dev",
+      "RemoteUrl": "https://github.com/paws-r/paws",
+      "RemoteSha": "ba99337a65925ced77d6aa510ed09b9a041c9640",
+      "RemoteSubdir": "cran/paws",
       "Requirements": [
         "paws.analytics",
         "paws.application.integration",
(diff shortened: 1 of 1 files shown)


etl/renv.lock
```diff
@@ -2012,34 +2073,31 @@
...
         "utils",
         "xml2"
       ],
-      "Hash": "484fb25b6267ae0a885511dbac53f193"
+      "Hash": "bb7044059aa3b8bfcd6076897c726114"
     },
     "paws": {
       "Package": "paws",
-      "Version": "0.10.0",
+      "Version": "0.9.0",
```
jeancochrane: **[Question, non-blocking]** Seems like we're downgrading here so as to align with the latest version on CRAN, is that right? I'm fine with it if so, just want to make sure I understand the reasoning. I vaguely recall some issues with older versions, but if you feel confident that we'll be fine on the latest CRAN versions, I'm down to keep it.

jeancochrane (approved):
Thanks for picking up this chore! Boy, do we have a lot of dependencies for our ETL scripts 😅 

We had an out-of-band discussion of the issue of local RStudio/Positron failing to load user-level `.Rprofile` files if a project-level file exists, so to formalize our discussion here -- We decided to add a line like the following one to any project-level `.Rprofile` files that we maintain:

```r
if (file.exists("~/.Rprofile")) source("~/.Rprofile")
```

We'll do this as it comes up, instead of proactively updating `.Rprofile` files across all of our repos, since local RStudio/Positron is not widely adopted on our team yet.

PR #914, opened 2025-10-23
Remove legacy COE data from appeals view

wrridgeway:
Quick PR to implement filtering discussed [here](https://teams.microsoft.com/l/message/19:d8e8e8e6a11e4d379305856c2ab2eaf5@thread.v2/1761231546245?context=%7B%22contextType%22%3A%22chat%22%7D).


wrridgeway:
It's not actually the filter leading to the majority of this drop in appeals, it's updates to htpar itself.

PR #931, opened 2025-11-17
Increase `default_vw_pin_appeal_row_count` threshold

wrridgeway:
We didn't update this [row count test threshold](https://github.com/ccao-data/data-architecture/actions/runs/19427347068/job/55577852576) after the second loss of appeals, so it needs to be lowered again.


dbt/models/default/schema/default.vw_pin_appeal.yml
```diff
@@ -68,7 +68,7 @@ models:
       - row_count:
           name: default_vw_pin_appeal_row_count
           # The exact row count is too complex to compute, so approximate it
-          above: 2800000
+          above: 2400000
```
wrridgeway: Current view row count is 2,464,787.

PR #934, opened 2025-12-02
Increase retry attemps for Socrata upload script

wrridgeway:
The [upload-open-data-assets workflow](https://github.com/ccao-data/data-architecture/actions/workflows/socrata_upload.yaml) [failed](https://github.com/ccao-data/data-architecture/actions/runs/19820375254) repeatedly while trying to run its automated first-of-the-month update.

After investigating, it seems that the workflow was failing due to connection resets on Socrata's end, not due to any scripting errors. [Jean suggested increasing the number of retries](https://stackoverflow.com/a/35504626) our script makes during post and put requests before throwing an error and failing. This seems to have solved our issue, as the script continued running even when socrata broke a connection:

image

***
Successful runs with updated retries:

https://github.com/ccao-data/data-architecture/actions/runs/19901190761
https://github.com/ccao-data/data-architecture/actions/runs/19899213552 (failed, but only because I specified year 2025, which doesn't exist in the parcel proximity asset, which doesn't need to be updated monthly anyways)
https://github.com/ccao-data/data-architecture/actions/runs/19875920083
https://github.com/ccao-data/data-architecture/actions/runs/19871057037

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -40,7 +40,7 @@ retries = Retry(
     status_forcelist=[500, 502, 503, 504],
     # Set default allowed methods for retries to include POST, since the Socrata API
     # uses idempotent POST requests
-    allowed_methods=Retry.DEFAULT_ALLOWED_METHODS | {"POST"}
+    allowed_methods=Retry.DEFAULT_ALLOWED_METHODS | {"POST"},
 )
 
 session.mount(


socrata/socrata_upload.py
```diff
@@ -27,6 +29,14 @@
     str(os.getenv("SOCRATA_PASSWORD")),
 )
 
+retries = Retry(
+    total=5, backoff_factor=0.1, status_forcelist=[500, 502, 503, 504]
+)
```
jeancochrane: **[Suggestion, optional]** Based on [the `Retry` docs](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry), it seems like [the set of HTTP methods that will trigger a retry](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry.DEFAULT_ALLOWED_METHODS) does not include `POST` by default. This is because most APIs don't consider `POST` requests to be idempotent, but the Socrata Producer API treats `POST` as an [upsert](https://dev.socrata.com/publishers/soda-producer/upsert), so it is idempotent in our case. We do sometimes use the `POST` method in the `upload()` function, so I think we should add it to the list of methods that will trigger this retry.

According to the docs, you can enable retries for all HTTP methods by setting the `allowed_methods` parameter to `None`. I think this is _probably_ safe in our case, but we could be even safer by extending `urllib3.util.Retry.DEFAULT_ALLOWED_METHODS` to add `"post"` and using that for the `allowed_methods` parameter instead. I have a slight preference for the latter choice but am down for the former too if you're finding extending `DEFAULT_ALLOWED_METHODS` to be too annoying.

wrridgeway:
Successful [test-open-data-assets run](https://github.com/ccao-data/data-architecture/actions/runs/19904131959/job/57055761313) (it failed, but as we'd expect and 2025 counts look good!)

jeancochrane (approved):
One small suggested tweak, but otherwise this looks great!

socrata/socrata_upload.py
```diff
@@ -27,6 +29,22 @@
...
 )
 
+# Set default allowed methods for retries to include POST, since the Socrata API
+# uses idempotent POST requests
+Retry.DEFAULT_ALLOWED_METHODS = frozenset(
+    ["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE", "POST"]
+)
+
+# Configure retries for requests to the open data portal API since it has a
+# tendency to return 500 errors under load
```
jeancochrane: **[Suggestion, optional]** Same comment as above:


```suggestion
# Configure retries for requests to the open data portal API since it has a
# tendency to return 500 errors under load. We need to do this via a custom
# `Retry` instance because `requests` does not retry connection errors or
# requests where data has made it to the server unless you use a custom
# `Retry` instance to override that behavior
```

socrata/socrata_upload.py
```diff
@@ -27,6 +29,22 @@
...
+retries = Retry(
+    total=5, backoff_factor=0.1, status_forcelist=[500, 502, 503, 504]
+)
```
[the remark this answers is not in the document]
wrridgeway: Oh, rad, thanks Jean.

PR #945, opened 2025-12-11
Use new `procname` for final eav in `default.vw_pin_tax_roll`

wrridgeway:
Apparently `tot51` should only be used when `procname = 'FINALEAV'`. This PR removes our per-stage "final eav" values and switches over to a single value of `final_eav` for the correct procname.

### DIFF
--- a/dbt/models/default/schema/default.vw_pin_tax_roll.yml
+++ b/dbt/models/default/schema/default.vw_pin_tax_roll.yml
@@ -49,7 +49,7 @@ models:
               FROM {{ source('iasworld', 'asmt_all') }}
               WHERE rolltype != 'RR'
                 AND deactivat IS NULL
-                AND procname IN ('CCAOVALUE', 'CCAOFINAL', 'BORVALUE')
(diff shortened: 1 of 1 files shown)


jeancochrane (approved):
Looks right to me! This change introduces four new rows, all of which have nothing in `asmt` except for a `FINALEAV` procname. Kinda weird, especially since they all happen to be 0, but I think it's correct to pull them in.

```sql
select dev.*
from z_ci_use_new_procname_for_finaleav_default.vw_pin_tax_roll dev
left join default.vw_pin_tax_roll prod
    on dev.pin = prod.pin
    and dev.year = prod.year
where prod.pin is null
```

This change is making me wonder whether we also need to make any edits to `default.vw_pin_value`, too? In particular, now that I know it's possible for a PIN to have no procnames except for `FINALEAV`, I wonder if our logic for inferring provisional values is not correct. Not a big deal I don't think, since we only use those values for preliminary reporting for towns that are getting ready to close, but I think it's worth thinking through the implications with @ccao-jardine.

Issue #946, opened 2025-12-11
Add ELT for sales val analyst overrides


jeancochrane:
## Background

We are expecting we will soon receive a batch of a few hundred South tri sales that Res Val analysts have reviewed in order to determine whether they are arm's-length open-market transactions and whether they require any characteristic correction. We should bring these analyst determinations into the data lake so that we can use them for modeling, since analyst determinations should override the determinations that our sales val model makes.

## Data model

* Add a new table `sale.flag_override` with (at minimum) the following columns
    * `doc_no`
    * `is_arms_length`
* Use `sale.flag_override.is_arms_length` in `default.vw_pin_sale` to produce a new column `is_arms_length` that coalesces `flag_override.is_arms_length` and `flag.sv_is_outlier`

We may want to add more new columns from the data that Res Val sends us, depending on what it contains. Add these columns to `sale.flag_override` and use the prefix `override_` to add them to `default.vw_pin_sale`, if we think they would be useful in `default.vw_pin_sale`.

## ELT

Check in with Billy about these steps:

* Draft a raw script that uploads the raw data to our raw S3 bucket
    * Should we upload raw Excel, or format as Parquet? I would lean towards raw Excel but will defer to Billy
    * Design the file locations in the raw bucket such that it will be possible for us to upload multiple workbooks
* Draft a warehouse script that pulls the raw files from the raw bucket, combines them into a single dataframe, extracts only the relevant columns, and saves the output to `sale.flag_override`  

## Diagram

![Image](https://github.com/user-attachments/assets/d7554778-e7df-4b76-930d-2be145df11ed)

wrridgeway:
As far as loading the raw data into S3, if it's something that they attach to an email or share from onedrive then we should probably download it and place it into a folder in `O:\CCAODATA\data` and have the raw script reference that location. I'd also leave a comment in the raw script about how the file got where it ended up. I would keep the original format.

Make sure to either use [one of our functions](https://github.com/ccao-data/data-architecture/blob/132769d94a2dd1f5c8ba6b9951aacf17090ce4eb/etl/utils.R#L60) to upload the files to the warehouse folder after cleaning, or to make sure to include a `loaded_at` column formatted as expected.

PR #949, opened 2025-12-22
Add manual sale override integration to existing sales validation 

wagnerlmichael:
This PR adds a new table `sale.flag_override` and adjusts `default.vw_pin_sale`.

`sale.flag_override` is a table built off of excel workbooks filled in from valuations. These workbook contain sales that valuations are manually reviewing. Based on their review, we determine whether or not the sale is valid for modeling. We use the valuations review to overwrite our existing sale outlier classification that was algorithmically produced by our sales val pipeline.

We tweak `default.vw_pin_sale` to add a new column `is_outlier` which adds a preference for the override values from analysts but defaults to the sales val determination in the case that the `doc_no` does not have a analyst determination. We include the relevant columns that determine this `is_outlier` column in `vw_pin_sale` as well.

Closes https://github.com/ccao-data/data-architecture/issues/946

### DIFF
--- a/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
@@ -166,7 +166,7 @@ dfs_ready_to_write <- purrr::imap(
     source_file
   ) %>%
     mutate(
-      loaded_at = Sys.time()
+      loaded_at = as.character(Sys.time())
     )
 )
 


etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,44 @@
...
+library(dplyr)
+library(arrow)
+library(aws.s3)
+source("utils.R")
+
+df <-
+  read_excel(
+    "O:/CCAODATA/data/sale/res_characteristic_potentials_reviewed.xlsx" # nolint
+  ) %>%
+  janitor::clean_names() %>%
```
wagnerlmichael: Handled this ingest with [this comment](https://github.com/ccao-data/data-architecture/issues/946#issuecomment-3657549627) in mind. I believe this path will work out of the box on everyone's local machine, but I am not sure.

dbt/models/sale/docs.md
```diff
@@ -9,6 +9,17 @@ should be possible to reconstruct using the other sales validation tables:
 `sale.group_mean`, `sale.parameter`, and `sale.metadata`.
 
 **Primary Key**: `meta_sale_document_number`, `run_id`, `version`
+{% enddocs %}
+
+        description: '{{ doc("flag_override") }}'
```
jeancochrane: **[Nitpick, optional]** I bet you'll catch this later on anyway, but as long as I'm here, a small copy/paste error:

```suggestion
```

jeancochrane (commented):
Nice start! Some high level thoughts on the design below.

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,114 @@
...
+  prefix = "sale/flag_override/"
+)
+
+tmp_dir <- tempdir()
+
+dfs <- list()
+
+for (obj in objs) {
+  key <- obj[["Key"]]
+  if (is.null(key) || grepl("/$", key)) next # skip directory placeholders
```
wagnerlmichael: I believe this line is needed because `get_bucket` also returns prefix objects that end in `/`

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -333,9 +340,16 @@ SELECT
     sales_val.sv_outlier_reason2,
     sales_val.sv_outlier_reason3,
     sales_val.sv_run_id,
-    sales_val.sv_version
+    sales_val.sv_version,
+    flag_override.is_valid_for_modeling AS override_is_valid_for_modeling,
+    COALESCE(
+        NOT flag_override.is_valid_for_modeling,
+        NOT sales_val.sv_is_outlier
+    ) AS use_in_model
```
wagnerlmichael: Maybe we can improve on `use_in_model` here? We could do something like put `analyst_is_valid_for_modeling` in `sale.flag_override` and then put just `is_valid_for_modeling` alias to `pin_sale`? Any other ideas?

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,114 @@
...
+      "dont_use_char_col", .y,
+      ignore.case = TRUE
+    )
+  )
+)
+
+# Bind all processed dfs together and write to S3 as parquet
+data_to_write <- purrr::imap_dfr(
+  dfs_processed,
+  ~ .x %>% select(doc_no, is_valid_for_modeling),
```
wagnerlmichael: Only writing the MVP columns, for now. Eventually we could incorporate flips, class changes, etc

dbt/models/default/default.vw_pin_sale.sql
wagnerlmichael: This seems like a much better approach! I've got some follow up questions.

Regarding the `loaded_at` column. My initial plan was to build the script such that every time it is run, a loop manipulates and writes every file's data. But if I implemented that idea, the `loaded_at` value would always be the same.

I could do something like build the manipulation function, and then for each new file we get, add a PR and a line every time like `new_data %>% modify_df %>% write_to_s3()`. I could also build the logic to keep the loop and look in the s3 bucket prior and only write data that isn't already in there, maybe a non-overwrite option. 

Regarding the `has_characteristic_change` column. This column should be major, minor, or none. The complication here is that we would only use this as determining a valid sale if we have the major value. Your suggested column is a boolean. Should I change this to a string for those 3 values? I could also only count major changes and we could have the data as `has_major_characteristic_change`, I could also add a `has_minor_charactersitic_change`.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -333,9 +334,37 @@ SELECT
...
+            OR flag_override.has_class_change = TRUE
+            OR flag_override.has_characteristic_change = 'yes_major'
+            OR flag_override.requires_field_check = TRUE
+        )
+        -- If there is no override, default to sv_is_outlier
+        WHEN sales_val.sv_is_outlier IS NOT NULL
+        THEN NOT sales_val.sv_is_outlier
+        -- If neither override nor sv_is_outlier is populated, leave null
+        ELSE NULL
+    END AS is_valid_for_modeling
```
wagnerlmichael: Thoughts on factoring this out into a CTE?

dbt/models/sale/schema.yml
```diff
@@ -50,6 +50,20 @@ sources:
               meta:
                 description: flag is unique by meta_sale_document_num and version
 
+      - name: flag_override
+        description: '{{ doc("flag_override") }}'
+        columns:
```
jeancochrane: **[Suggestion, optional]** Now seems as good a time as any to document the new fields we've added since my last review!

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,161 @@
...
+    mutate( # nolint
+      is_arms_length =
+        coalesce(sale_is_arms_length == "YES", FALSE),
+      is_flip =
+        coalesce(flip == "YES", FALSE),
+      has_class_change =
+        coalesce(
+          grepl("YES", class_change, ignore.case = TRUE),
+          FALSE
+        ),
```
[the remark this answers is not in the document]
jeancochrane: **[Suggestion, optional]** What would you think about performing a quick validation of the columns we're going to parse to make sure they don't contain any unexpected values before we parse them? For example, if `is_arms_length` is anything other than `"YES"`, `"NO"`, or null, the script would raise an error so that we can be aware of the unexpected value and update our script (or the raw file) to handle it accordingly. That way we wouldn't have to remember to do `count()` calls every time we process a new batch of sales. Does that seem reasonable to you?

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,170 @@
...
+dfs$valuations_sale_review_2025.12.16 <-
+  dfs$valuations_sale_review_2025.12.16 %>%
+  janitor::clean_names() %>%
+  rename(
+    sale_is_arms_length = sale_is_arm_s_length,
+    doc_no = sale_doc_no
+  ) %>%
+  mutate(
+    is_arms_length = coalesce(
+      sale_is_arms_length == "YES", FALSE
```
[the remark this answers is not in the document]
wagnerlmichael: https://github.com/ccao-data/data-architecture/pull/949/changes/773a96f6d9bf9a52a7abda8f266f383cca99925d

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -333,9 +334,36 @@ SELECT
...
+                flag_override.is_arms_length = FALSE
+                OR flag_override.is_flip = TRUE
+                OR flag_override.has_class_change = TRUE
+                OR flag_override.has_characteristic_change = 'yes_major'
+                OR flag_override.requires_field_check = TRUE
+            )
+        -- If there is no override, default to sv_is_outlier
+        WHEN sales_val.sv_is_outlier IS NOT NULL
+            THEN NOT sales_val.sv_is_outlier
+    END AS is_valid_for_modeling
```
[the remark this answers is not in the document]
jeancochrane: I hear you, it's not ideal! Though with proper docs we can mitigate the confusion a bit. I would be down to revisit this holistically at a later date -- I've been thinking that it would be nice to have all of our exclusion criteria implemented in the `default.vw_pin_sale` view anyway, so that it's reproducible across use cases (not just in the model ingest stage), and I think in that context a column name like `is_valid_for_modeling` or `exclude_from_training_set` could make sense and could resolve the ambiguity.

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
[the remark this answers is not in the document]
wagnerlmichael: I left a comment at the top of the script.

dbt/models/default/default.vw_pin_sale.sql
```diff
@@ -333,9 +333,36 @@ SELECT
...
+                flag_override.is_arms_length = FALSE
+                OR flag_override.is_flip = TRUE
+                OR flag_override.has_class_change = TRUE
+                OR flag_override.has_characteristic_change = 'yes_major'
+                OR flag_override.requires_field_check = TRUE
+            )
+        -- If there is no override, default to sv_is_outlier
+        WHEN sales_val.sv_is_outlier IS NOT NULL
+            THEN NOT sales_val.sv_is_outlier
+    END AS is_outlier
```
jeancochrane: **[Nitpick, blocking]** Hmm, do we need to flip the `NOT` logic here in order to change the meaning to "is outlier"? Or maybe I'm misunderstanding?

```suggestion
    CASE
        -- If there is an override, use override logic
        -- If neither override nor sv_is_outlier is populated, leave null
        WHEN
            flag_override.is_arms_length IS NOT NULL
            OR flag_override.is_flip IS NOT NULL
            OR flag_override.has_class_change IS NOT NULL
            OR flag_override.has_characteristic_change IS NOT NULL
            OR flag_override.requires_field_check IS NOT NULL
            THEN (
                flag_override.is_arms_length = FALSE
                OR flag_override.is_flip = TRUE
                OR flag_override.has_class_change = TRUE
                OR flag_override.has_characteristic_change = 'yes_major'
                OR flag_override.requires_field_check = TRUE
            )
        -- If there is no override, default to sv_is_outlier
        WHEN sales_val.sv_is_outlier IS NOT NULL
            THEN sales_val.sv_is_outlier
    END AS is_outlier
```

jeancochrane (commented):
Looking great! The below comments are all that remains from me. Tomorrow I'm going to QC this more thoroughly by comparing the parsed data to the raw sheets, but I don't want that to block the couple of changes I suggested below.

@wrridgeway Could you take a quick look at the ETL portion of this and approve if it looks good to you?

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,198 @@
...
+    arrow::write_parquet(
+      .x,
+      tmp_file,
+      compression = "snappy"
+    )
+
+    save_local_to_s3(
+      s3_uri = out_uri,
+      path = tmp_file,
+      overwrite = FALSE
```
wagnerlmichael: Should I toggle `overwrite` to `TRUE` here? Seems like the right move, otherwise we would have to manually delete the S3 data every time we refresh.

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,198 @@
...
+library(arrow)
+library(aws.s3)
+library(dplyr)
+library(janitor)
+library(readxl)
+
+source("utils.R")
+
+s3_dir <- "s3://ccao-data-raw-us-east-1/sale/flag_override/"
+
```
jeancochrane: **[Suggestion, optional]** Are we using this at all? It seems like we use `objs` defined below instead:


```suggestion
```

wrridgeway (approved):

wrridgeway (approved):

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -0,0 +1,185 @@
...
+    is_arms_length,
+    is_flip,
+    has_class_change,
+    has_characteristic_change,
+    requires_field_check,
+    work_drawer,
+    source_file
+  ) %>%
+    mutate(
+      loaded_at = Sys.time()
```
wrridgeway: Sorry, one last thing: `sys.time()` has to be wrapped in `as.character()` for it to be brought into athena correctly formatted.

wrridgeway (approved):

jeancochrane (approved):
Awesome work here @wagnerlmichael. Let's merge it!

PR #950, opened 2025-12-23
Update Commercial Open Data asset with 2025 public model

wrridgeway:
If I exclude the columns `loaded_at` from the old data (which only gets added during upload), `gross_building_area` from the new data (which is a new column), and `taxpayer` from both, the 2021-2024 data match exactly using `dplyr::all_equal()`. Taxpayer is the only column that no longer matches, because I added it to the declared `char_cols` so it doesn't get reduced only to the numbers it contains through `parse_number()`. This is an improvement, though it's a shame I hadn't been handling it correctly in the past.

I have not uploaded the new data yet to S3. My primary concern was overwriting the old data and changing it, but I'm confident now that the old data has been accurately reproduced.

### DIFF
--- a/etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-commercial_valuation.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-commercial_valuation.R
@@ -227,5 +227,5 @@ output <- list.files(
   distinct()
 
 output %>%
-  group_by(year) %>% skim()
+  group_by(year) %>%
   write_partitions_to_s3(output_bucket, is_spatial = FALSE, overwrite = TRUE)


etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-commercial_valuation.R
```diff
@@ -50,6 +51,7 @@ int_cols <- c(
   "4brunits",
   "aprx_comm_sf",
   "bldgsf",
+  "gross_building_area",
```
wrridgeway: Only new column that will be added to commercial valuation data.

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-commercial_valuation.R
```diff
@@ -67,7 +69,9 @@ remove_cols <- c(
   "locrating",
   "mobilehomepads",
   "tot_apts",
-  "usecode"
+  "usecode",
+  "imprname",
+  "units/beds"
```
wrridgeway: New 2025 columns that are removed.

etl/scripts-ccao-data-warehouse-us-east-1/ccao/ccao-commercial_valuation.R
```diff
@@ -196,7 +212,8 @@ list.files(
       TRUE ~ str_replace_all(str_squish(.x), " ", ", ")
     )),
     across(int_cols, as.integer),
-    across(char_cols, as.character)
+    across(char_cols, as.character),
+    across(where(is.character), ~ gsub("\\r", "", .x))
```
wrridgeway: Seems like some different new line characters have made their way into the data while it's ingested. This removes them.

jeancochrane (approved):
Annoying that the underlying data format continues to change, but this looks good to me!

PR #955, opened 2025-12-29
Add multisale uniqueness constraint for `sale.mydec`

wrridgeway:
We can see [from below](https://github.com/ccao-data/data-architecture/actions/runs/20580773681/job/59107612533?pr=955) that `sale.mydec` is passing the new test.


jeancochrane (approved):
The new test will give me some peace of mind, thanks!

PR #961, opened 2026-01-02
Fix first day of new year issue

wrridgeway:
The open data upload script [failed](https://github.com/ccao-data/data-architecture/actions/runs/20637432586/job/59264337703) (😒) while trying to upload our permits asset. It failed because we make the script choose the minimum value between the current year and the maximum year in the asset - when the clock rolled over to 2026, the script had to compare the maximum value in the permits asset, `2032`, and the current year, `2026`. Because there are no values in the permit asset for 2026, the script returned an empty dataframe, which it doesn't expect.

This fix ensures the script can't query an asset for a year that isn't in that asset and chooses the maximum year from an asset less than or equal to the current year.


jeancochrane (approved):
Nice catch! I think the new solution is simpler and easier to understand anyway.

PR #969, opened 2026-01-15
Use combined municipality for reporting views

wrridgeway:
Just switching over to `combined_municipality_name` for our reporting views to decrease missingness and the difference in pin count between tri = city and muni = city of chicago.

```
select false as new,
	count(*) as pins_missing_muni
from reporting.vw_pin_township_class
where triad_name = 'City'
	and year = '2025'
	and municipality_name is null
union all
select true as new,
	count(*) as pins_missing_muni
from z_ci_switch_over_to_combined_muni_name_for_reporting_feeder_view_reporting.vw_pin_township_class
where triad_name = 'City'
	and year = '2025'
	and municipality_name is null
```

| new   | pins_missing_muni |
|-------|-----------------|
| true  |            2714 |
| false |            4175 |

```
select false as new,
	stage_name,
	num_pin_total
from reporting.assessment_progress
where geo_id = 'CITY OF CHICAGO'
	and year = '2025'
union all
select true as new,
	stage_name,
	num_pin_total
from z_ci_switch_over_to_combined_muni_name_for_reporting_feeder_view_reporting.assessment_progress
where geo_id = 'CITY OF CHICAGO'
	and year = '2025'
```

| new   | stage_name         | num_pin_total |
|-------|--------------------|---------------|
| true  | MAILED             |        880027 |
| true  | BOR CERTIFIED      |        880027 |
| true  | ASSESSOR CERTIFIED |        880027 |
| false | MAILED             |        878566 |
| false | BOR CERTIFIED      |        878566 |
| false | ASSESSOR CERTIFIED |        878566 |

and thankfully 880,027 - 878,566 = 4,175 - 2,714 = 1,461

### DIFF
--- a/dbt/models/reporting/reporting.vw_pin_township_class.sql
+++ b/dbt/models/reporting/reporting.vw_pin_township_class.sql
@@ -58,7 +58,7 @@ LEFT JOIN {{ source('spatial', 'township') }} AS town
 -- Exclude classes without a reporting class
 INNER JOIN {{ ref('ccao.class_dict') }} AS groups
     ON correct.class = groups.class_code
--- Tax municipality data lags iasWorld data by a year or two at any given time
+-- Location data in our lake lags iasWorld data by a year or two
 LEFT JOIN {{ ref('location.vw_pin10_location') }} AS vpl
     ON SUBSTR(correct.parid, 1, 10) = vpl.pin10
     AND CASE


dbt/models/reporting/reporting.vw_pin_township_class.sql
```diff
@@ -57,14 +59,15 @@ LEFT JOIN {{ source('spatial', 'township') }} AS town
 INNER JOIN {{ ref('ccao.class_dict') }} AS groups
     ON correct.class = groups.class_code
 -- Tax municipality data lags iasWorld data by a year or two at any given time
```
jeancochrane: **[Suggestion, optional]** Not a big deal, and this may still be technically correct, but it strikes me that this comment seems to have been referencing the `location.tax` table which is no longer the direct target of this join.

PR #970, opened 2026-01-16
Integrate third analyst sale review excel file

wagnerlmichael:
This PR ingests another excel file we received from valuations. It requires us to make a number of adjustments to the script.

### DIFF
--- a/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
@@ -46,8 +46,6 @@ for (obj in objs) {
   dfs[[df_name]] <- df
 }
 
-# TODO: For this file there is a high incidence of field_check filled in as
-# 'yes', so we are going to need to decide how strict to be here
 dfs$valuations_sale_review_2026_01_16 <-
   dfs$valuations_sale_review_2026_01_16 %>%
   filter(WHOM == "LYDIA") %>%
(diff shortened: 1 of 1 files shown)


etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -60,9 +74,40 @@ clean_columns_and_whitespace <- function(df) {
...
+required_raw_cols <- c(
+  "sale_is_arms_length",
+  "flip",
+  "class_change",
+  "characteristic_change",
+  "field_check",
+  "work_drawer"
+)
+
+assert_required_cols <- function(df, df_name = "<unnamed>") {
```
wagnerlmichael: Adding this to fail loudly if the column names are different from different excel files

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_override.R
```diff
@@ -46,6 +47,23 @@ for (obj in objs) {
   dfs[[df_name]] <- df
 }
 
+# TODO: For this file there is a high incidence of field_check filled in as
+# 'yes', so we are going to need to decide how strict to be here
```
wagnerlmichael: I'm really not sure about this
```
> dfs_processed$valuations_sale_review_2026_01_16 %>%
+     count(requires_field_check)
# A tibble: 2 × 2
  requires_field_check     n
  <lgl>                <int>
1 FALSE                  540
2 TRUE                   147
```
This is a lot of true values, and they are basically all _potential_ SF outliers. I wonder if we should err on the side of letting more sales in and not use this field to flag an outlier.

jeancochrane: **[Suggestion, optional]** We're going to discuss this further in a cross-team meeting. In the meantime, what do you think about interpreting this file slightly differently, such that we only consider `requires_field_check` to be true if the sale is also an outlier according to our model? Would that help de-emphasize the importance of the field check for this particular file, or are there so few outliers in these data anyway that it would remove too much information?

PR #971, opened 2026-01-18
Add new groups for south tri sales val in `location.neighborhood_group`

wagnerlmichael:
After some EDA, iteration, and collaboration with res valuations, we settled on new geographic groupings for south tri sales val. From what I can tell, this is the simplest change that allows us to integrate these new groupings into sales val.


jeancochrane:
Nice, I opened https://github.com/ccao-data/reporting/issues/11 to track the map work!

PR #975, opened 2026-01-22
Add `default.vw_pin_exe_long` view with CofE columns

### LINKED ISSUE #962: Distinguish exemptions added via CofE in `default.vw_pin_exe`
As far as I can tell, our exemption view (`default.vw_pin_exe`) currently does not provide the ability to filter exemptions that were added via Certificate of Error (CofE). This is important for a number of different downstream use cases, including [PTAXSIM](https://github.com/ccao-data/ptaxsim/), which needs to exclude CofE exemptions when generating the `tax.pin` table.

These are the relevant columns:

- `exadmn.user126`: CofE indicator
- `exadmn.udate9`: CofE date (Not reliably filled out prior to November 2025, but we may as well start including it)

jeancochrane:
## Overview

This PR extracts a view `default.vw_pin_exe_long` from the existing `default.vw_pin_exe` view so that we can add columns with information about CofEs. New CofE columns include `is_cofe` (boolean, whether the exemption was applied via CofE) and `cofe_date` (timestamp reflecting the date of the CofE).

You can test out the data in the staging table `z_ci_jeancochrane_962_distinguish_exemptions_added_via_cofe_in_defaultvw_pin_exe_default.vw_pin_exe_long`.

Tagging in @wrridgeway for code review and @ccao-jardine for feedback about whether this schema seems useful for responding to the types of CofE requests we've gotten in the past.

Closes #962. 

## Alternatives considered

My first instinct was to add these new CofE columns directly to the `default.vw_pin_exe` view. However, that view is pivoted wide such that each exemption type gets its own column; since CofEs apply to individual exemptions, the wide nature of the view means that we would need 2*N additional CofE columns (where N is the number of exemption types) in order to support CofE information in `default.vw_pin_exe`.

Since `default.vw_pin_exe` already relied on a long CTE, and since the long view only requires two additional CofE columns, I decided to extract the long CTE to its own view and add the columns there. I'm definitely open to other suggestions for the data model, though.

### DIFF
--- a/dbt/models/iasworld/columns.md
+++ b/dbt/models/iasworld/columns.md
@@ -245,14 +245,6 @@ Possible values for this variable are:
 City name
 {% enddocs %}
 
-## cofe_date
-
-{% docs column_cofe_date %}
-Certificate of Error (CofE) date.
-
-This column has only been in use since November 2025.
-{% enddocs %}
-
 ## cofe_indicator
 
 {% docs column_cofe_indicator %}
--- a/dbt/models/default/schema/default.vw_pin_exe_long.yml
+++ b/dbt/models/default/schema/default.vw_pin_exe_long.yml
@@ -4,7 +4,7 @@ models:
 
     columns:
       - name: cofe_date
-        description: '{{ doc("column_cofe_date") }}'
+        description: '{{ doc("shared_column_cofe_date") }}'
       - name: exemption_amount
         description: Exemption amount
       - name: exemption_type
--- a/dbt/models/iasworld/schema/iasworld.exadmn.yml
+++ b/dbt/models/iasworld/schema/iasworld.exadmn.yml
@@ -129,7 +129,7 @@ sources:
           - name: trans_id
             description: '{{ doc("column_trans_id") }}'
           - name: udate9
-            description: '{{ doc("column_cofe_date") }}'
+            description: '{{ doc("shared_column_cofe_date") }}'
           - name: unit
             description: '{{ doc("column_unit") }}'
           - name: upd_status
--- a/dbt/models/shared_columns.md
+++ b/dbt/models/shared_columns.md
@@ -1395,6 +1395,16 @@ are calculated. Tax bills are paid in arrears, so an assessment from TY2023
 will be paid in calendar year 2024
 {% enddocs %}
 
+# Exemptions
+
+## cofe_date
+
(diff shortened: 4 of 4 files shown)


dbt/models/default/schema/default.vw_pin_exe.yml
```diff
@@ -62,7 +62,5 @@ models:
           combination_of_columns:
             - pin
             - year
-          config:
-            error_if: ">10"
```
jeancochrane: Not sure what happened here, but we aren't seeing these duplicates anymore -- I assume they fixed the underlying ias data.

dbt/models/iasworld/columns.md
```diff
@@ -245,6 +245,26 @@ Possible values for this variable are:
...
+## cofe_indicator
+
+{% docs column_cofe_indicator %}
+Certificate of Error (CofE) indicator.
+
+Possible values include:
+
+- `Y`: Exemption added via CofE
+- `N`: Exemption **not** added via CofE
+- Null: Unknown
```
jeancochrane: TODO: This is different for the raw table.

dbt/models/default/default.vw_pin_exe_long.sql
```diff
@@ -0,0 +1,99 @@
...
+        year,
+        exemption_type,
+        exemption_amount,
+        is_cofe,
+        cofe_date,
+        ROW_NUMBER() OVER (
+            PARTITION BY pin, year, exemption_type
+            ORDER BY
+                -- Prioritize CofEs first, because they represent corrections
+                is_cofe DESC NULLS LAST,
```
wrridgeway: what the heck `NULLS LAST` is brand new tech for me, how have I never seen this

jeancochrane:
Sorry for the sneaky commit @wrridgeway, but I noticed that `cofe_date` should be a shared column while giving this a final once-over, so I committed that change in 7384d56.

Also, I missed this question:

> My only high-level question is whether this new view should have a _long suffix or the old view should have a _wide suffix since it's fed by this new one. Ultimately indifferent.

My feeling is that `*_long` is more appropriate here than `*_wide` would be on the other view because in general `default.pin_*` views are unique by PIN and year. This new view breaks that general pattern, so it feels more natural to me to include a pointer to the deviation in the model name. I similarly don't feel super strongly about it, though.

PR #977, opened 2026-01-23
Refactor outlier columns in `default.vw_pin_sale` to incorporate human review

jeancochrane:
## Overview

This PR builds off of https://github.com/ccao-data/data-architecture/pull/967/, suggesting a refactored data model that factors out the logic that combines our algorithmic sale flags with information from our human reviewers in order to produce a final outlier determination with clear reasons.

I went a little bit further with the data model than I expected -- I hope it's not too confusing! Though big refactors can be risky, this one feels appropriate because this data model is still very new. I'd be happy to walk through the changes on a call if it's too much to review from scratch.

See https://github.com/ccao-data/model-res-avm/pull/423 for the corresponding res model PR.

## Data model changes

New models that this PR introduces:

- `sale.vw_flag`: View that pulls the most recent version of each algorithmically flagged sale from `sale.flag`
- `sale.vw_outlier`: View that combines algorithmic sale flags with human-reviewed sale attributes in order to produce a final outlier determination and corresponding reasons

Changes to existing models:

- Renamed `sale.flag_override` to `sale.flag_review` for clarity (we are not directly using the information to "override" sales val flags, we are just incorporating that information into our final decision, so "review" seems more neutral)
- Added new column `default.vw_pin_sale.is_outlier` reflecting the final decision on outlier status based on `sale.vw_outlier` (which pulls from `sale.flag` and `sale.flag_review`)
- Added new column `default.vw_pin_sale.outlier_reason` reflecting a human-readable string with the reason behind the sale's outlier status (also pulled from `sale.vw_outlier`)
- Renamed audit trail columns in `default.vw_pin_sale` that come directly from `sale.flag` and `sale.flag_review` to use the prefixes `flag_*` and `review_*`, to make the provenance of each column more obvious, e.g. `is_arms_length` ➡️ `review_is_arms_length`
    - Thanks to Michael and Tim for this idea!

## Open questions

- Should we update `default.vw_pin_sale_combined`? I haven't done so yet because I'm not sure of the status of it. Are we still using it? I almost wonder if we should remove that view to reduce complexity, now that we are getting new sales on a more regular schedule.

### DIFF
--- a/dbt/models/default/default.vw_pin_sale.sql
+++ b/dbt/models/default/default.vw_pin_sale.sql
@@ -263,7 +263,7 @@ SELECT
     unique_sales.sale_filter_same_sale_within_365,
     unique_sales.sale_filter_less_than_10k,
     unique_sales.sale_filter_deed_type,
-    outlier.is_outlier AS sale_filter_is_outlier,
+    COALESCE(outlier.is_outlier, FALSE) AS sale_filter_is_outlier,
     mydec_sales.mydec_deed_type,
     mydec_sales.sale_filter_ptax_flag,
     mydec_sales.mydec_property_advertised,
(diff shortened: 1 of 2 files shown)


wrridgeway (approved):
From my end everything looks pretty solid here, I can't find anything that sticks out as wrong. I'm assuming all the unit tests for `sale.vw_outlier` would catch any unexpected results produced by the conditional logic.

PR #980, opened 2026-01-27
Add maildat columns to `default.vw_pin_address`

wrridgeway:
This PR adds `iasworld.maildat` to `default.vw_pin_address` as the new source of `mail_` columns. We keep columns from `iasworld.owndat`, but we rename these columns with an `owner_` prefix. This allows us to avoid committing entirely to one table or the other and provide as much information to users as possible while being clear about the limitations of both sources of address data.

The only real nuance here is that `maildat` is not unique by pin and year when `cur = 'Y' and deactivat is null` like most iasworld tables. Instead we also need to depend on `mailseq` and choosing which value of `mailseq` to privilege in order to make `maildat` unique by pin and year (we take the max value) boiled down to trying to match up with what's displayed on the treasurer's website. Two example pins are below:

| parid          | taxyr | mailseq | maddr1               |
|----------------|-------|---------|----------------------|
| 01124000010000 |  2024 |       0 | 665 SW 8TH STREET    |
| 01124000010000 |  2024 |       1 | 1475 S BARRINGTON RD |
| 02221170250000 |  2024 |       0 | 935 W. KENILWORTH    |
| 02221170250000 |  2024 |       1 | 493 W HELEN RD       |

Updates have not changed row count, as expected:

```
with old as (
	select year,
		count(*) as old
	from default.vw_pin_address
	group by year
),
new as (
	select year,
		count(*) as new
	from z_ci_941_owndat_vs_maildat_default.vw_pin_address
	group by year
)
select old.year,
	old.old,
	new.new,
	old.old - new.new as diff
from old
where old.old != new.new
```

returns nothing. and `mail_address_full` is largely non-null:

```
select year,
	count(*) as new
from z_ci_941_owndat_vs_maildat_default.vw_pin_address
where mail_address_full is not null
group by year
```

| year | new     |
|------|---------|
| 2026 | 1749105 |
| 2025 | 1749120 |
| 2024 | 1750955 |
| 2023 | 1766525 |
| 2022 | 1767234 |
| 2021 | 1766333 |
| 2020 | 1768219 |
| 2019 | 1767675 |
| 2018 | 1768100 |
| 2017 | 1767909 |
| 2016 | 1767301 |
| 2015 | 1765763 |
| 2014 | 1765389 |
| 2013 | 1765993 |
| 2012 | 1766926 |
| 2011 | 1766397 |
| 2010 | 1764226 |
| 2009 | 1756264 |
| 2008 | 1739854 |
| 2007 | 1711776 |
| 2006 | 1678523 |
| 2005 | 1648889 |
| 2004 | 1625030 |
| 2003 | 1603501 |
| 2002 | 1586382 |
| 2001 | 1571199 |
| 2000 | 1555196 |
| 1999 | 1540553 |

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.owndat.yml
+++ b/dbt/models/iasworld/schema/iasworld.owndat.yml
@@ -215,6 +215,8 @@ sources:
             description: '{{ doc("column_skip_addr_validation") }}'
           - name: splitno
             description: '{{ doc("column_splitno") }}'
+          - name: ssn
+            description: '{{ doc("column_ssn") }}'
           - name: statecode
             description: '{{ doc("shared_column_owner_address_state") }}'
           - name: stateid


jeancochrane (commented):
Looking great! A few small suggestions, plus one request to clean up the docs. I'll give this a final look once that's done.

dbt/models/iasworld/schema/iasworld.maildat.yml
```diff
@@ -14,194 +14,231 @@ sources:
...
-            description: iasWorld transaction number
-          - name : upd_status
-            description: iasWorld update status for this row
-          - name : vldrsc
+          - name: trans_id
+            description: '{{ doc("column_trans_id") }}'
+          - name: upd_status
+            description: '{{ doc("column_upd_status") }}'
+          - name: vldrsc
             description: No Validation Reason Code
```
jeancochrane: **[Nitpick, optional]** Looks like this is shared with `iasworld.cname` and could be factored out:

https://github.com/ccao-data/data-architecture/blob/abc72dc80a4ba1fa411b38ffa30dd272ef817132/dbt/models/iasworld/schema/iasworld.cname.yml#L144-L145

jeancochrane (approved):
Thanks so much for the docs cleanup! A few tiny suggested tweaks, but otherwise this is good to go.

PR #985, opened 2026-02-06
Update `pinval.model_run` with 2026 model runs

jeancochrane:
This PR updates the seed that powers https://github.com/ccao-data/homeval/ so that we can perform HomeVal deployments with 2026 model runs.

I've already run two different HomeVal deployments off of this PR, using different model runs in the `pinval.model_run` seed. See the latest one here: https://github.com/ccao-data/homeval/actions/runs/22119406898


dbt/models/pinval/pinval.assessment_card.sql
```diff
@@ -294,7 +294,7 @@ SELECT
         WHEN
             pin_cd.class_code IS NULL  -- Class is not in our class dict
             OR NOT pin_cd.regression_class
-            OR (pin_cd.modeling_group NOT IN ('SF', 'MF'))
+            OR (pin_cd.modeling_group NOT IN ('SF', 'MF', 'BB'))
```
jeancochrane: We do include B&Bs in the training and assessment sets for the model, even though they usually get modeled by hand. That means that a B&B will wind up with `is_report_eligible == TRUE` and `reason_report_ineligible == 'non_regression_class'` unless we allow the `'BB'` modeling group here. This is not really a huge deal, since it doesn't affect the HomeVal reports that we generate for these PINs, but it means that all B&Bs fail the `pinval_assessment_card_reason_report_ineligible_is_null_when_is_report_eligible` data test, which adds unnecessary noise to that test.

jeancochrane:
Tagging in @wagnerlmichael as a domain expert, and @wrridgeway as a codeowner.

PR #986, opened 2026-02-09
Add year filter to resolve `model_assessment_card_meta_card_num_not_null` test failures

wrridgeway:
This test [perpetually fails](https://github.com/ccao-data/data-architecture/actions/runs/21822893887/job/62960464408) and adds noise to our dbt failure alarms. Maybe there's a more permanent solution that's preferred, but while we're modeling it would be nice to know an alarm means something new and unexpected has happened.

### DIFF
--- a/dbt/models/model/schema.yml
+++ b/dbt/models/model/schema.yml
@@ -12,8 +12,9 @@ sources:
               - not_null:
                   name: model_assessment_card_meta_card_num_not_null
                   config:
-                    where: meta_class != '299'
-                    error_if:  ">667"
+                    where: >
+                      meta_class != '299'
+                      AND year >= '2026'
         data_tests:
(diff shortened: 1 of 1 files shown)


jeancochrane:
I actually think this test is useful for current and future model runs, since it's a gut check that ensures we're not missing critical data in the assessment set. I think the test is annoying right now because we're including past assessment years with known problems that we can't fix.

What do you think about tweaking the `config.where` filter condition for the test so that we only include rows where `year >= '2026'`? A quick check confirms there's only one junk run for this year that violates the condition -- we could investigate to see where those null card numbers came from, or since it's an old junk run, we can just delete it to ensure the check passes on all eligible model runs:

```sql
select *
from model.assessment_card
where year >= '2026'
and meta_card_num is null
and meta_class != '299'
```

PR #990, opened 2026-02-25
Finish 2026 south tri ingest

wagnerlmichael:
Finish up ingesting the final round of south tri 2026 sales from valuations' 2026 capacity.

Total number of reviewed sales increased from 4103 to 4737.

The main complication with this ingest was that some data had hidden rows with different reviewers in the `WHOM` field that this script was picking up due the way we detect which rows to keep (NA checking), which took me a while to figure out. I basically ignore these fields, on the assumption that since they were hidden they weren't meant to be the ones we were looking at.

Also fixed some other duplication that arose upon replacing the updated files.


wagnerlmichael:
Note: I have not ran this script all the way through the upload. It will likely change some prod sales val classifications, and I wanted to double-check to make sure there isn't any sort of lingering model clean up work that would cause us to want to wait.

jeancochrane (approved):
Thanks so much for this! Feel free to run the script and output to the prod database, I don't expect it should interfere with model finalization for this year.

PR #1001, opened 2026-03-19
Add previous year to years updated monthly for assessed value open data asset

### LINKED ISSUE #999: Open Data Updates
Below are our updates for Open Data! Some are actionable now and can be done this sprint, others later. 

@wrridgeway and cc'ing @jeancochrane 

### Update [Address](https://datacatalog.cookcountyil.gov/Property-Taxation/Assessor-Parcel-Addresses/3723-97qp/about_data) Data
We're frequently asked for owner data, so we've [teed up](https://github.com/ccao-data/data-architecture/pull/980) an update of the view that feeds this asset accordingly. However, Socrata issues have been ongoing. We hope to resolve by Friday March 20. 

### More frequent updates of  [Assessed Values](https://datacatalog.cookcountyil.gov/Property-Taxation/Assessor-Assessed-Values/uzyt-m557)
Let's test out scheduling Assessed Values to update on the 1st and 15th of every month and evaluate how this goes, given our history of occasional issues updating Socrata assets.

### New data: Land SF per PIN
We'd like to publish land sf per PIN, and could use some user guidance about whether to add it to an existing asset (like PIN Universe) or create a standalone new asset. @ccao-jardine will reach out to some of our colleagues outside the CCAO to solicit feedback for our April sprint.

### Add sales validation to [Parcel Sales](https://datacatalog.cookcountyil.gov/Property-Taxation/Assessor-Parcel-Sales/wvhk-k5uv)
We've had our [sales val pipeline](https://github.com/ccao-data/model-sales-val) for a while to flag statistical outliers, and now we also have analyst review. The goal of updating Parcel Sales is to make the outcomes of these processes transparent without being too complicated to understand. @ccao-jardine will solicit some ideas about how to accomplish this clearly and transparently for our April sprint.

wrridgeway:
We'd like to update the "Assessed Values" asset on the 1st and 15th of every month rather than just the 1st. We can keep our old monthly cadence for all the other active assets.

### DIFF
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -84,5 +84,7 @@ jobs:
           OVERWRITE: ${{ inputs.overwrite }}
           YEARS: ${{ inputs.years }}
           WORKFLOW_EVENT_NAME: ${{ github.event_name }}
-          UPLOAD_SCHEDULE: ${{ github.event.schedule == '0 11 1 * *' && 'monthly' || github.event.schedule == '0 11 15 * *' && 'semimonthly' || '' }}
+          UPLOAD_SCHEDULE: >
+            ${{ github.event.schedule == '0 11 1 * *' && 'monthly' ||
+            github.event.schedule == '0 11 15 * *' && 'semimonthly' || '' }}
         run: python ./socrata/socrata_upload.py
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -26,6 +26,8 @@ on:
   schedule:
   # First of every month at 11am UTC (6am UTC-5)
     - cron: '0 11 1 * *'
+  # Fifteenth of every month at 11am UTC (6am UTC-5)
+    - cron: '0 11 15 * *'
 
 
 env:
@@ -82,4 +84,8 @@ jobs:
           OVERWRITE: ${{ inputs.overwrite }}
           YEARS: ${{ inputs.years }}
           WORKFLOW_EVENT_NAME: ${{ github.event_name }}
+          if: github.event.schedule == '0 11 1 * *'
+          UPLOAD_SCHEDULE: 'monthly'
+          if: github.event.schedule == '0 11 15 * *'
+          UPLOAD_SCHEDULE: 'semimonthly'
         run: python ./socrata/socrata_upload.py
--- a/.github/workflows/socrata_upload.yaml
+++ b/.github/workflows/socrata_upload.yaml
@@ -24,9 +24,13 @@ on:
         description: Overwrite socrata asset
         required: false
   schedule:
-  # First of every month at 11am UTC (6am UTC-5)
+  # We use these schedules to determine the value of the `UPLOAD_SCHEDULE`
+  # env var below, so make sure to change that env var definition if you
+  # change any of these schedules.
+  # 
+  # Run a monthly job on the first of every month at 11am UTC (6am UTC-5).
     - cron: '0 11 1 * *'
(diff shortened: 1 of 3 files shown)


socrata/socrata_upload.py
```diff
@@ -71,6 +71,14 @@ def parse_assets(assets=None):
 
     os.chdir("./dbt")
 
+    # If the upload is running on a semi-monthly schedule, we want to exclude
+    # any assets tagged with "tag:monthly" since those are only meant to be
+    # updated once per month. If the upload is running on a monthly schedule,
+    # we want to include those assets.
+    monthly_tag = (
+        "tag:monthly" if os.getenv("UPLOAD_SCHEDULE") == "semi-monthly" else ""
```
jeancochrane: **[Nitpick, blocking]** Looks like the spelling of the "semi-monthly" tag here is mismatching its value in the workflow. I'm fine with either spelling, but we should pick one:


```suggestion
        "tag:monthly" if os.getenv("UPLOAD_SCHEDULE") == "semimonthly" else ""
```

.github/workflows/socrata_upload.yaml
```diff
@@ -26,6 +26,8 @@ on:
   schedule:
   # First of every month at 11am UTC (6am UTC-5)
     - cron: '0 11 1 * *'
+  # Fifteenth of every month at 11am UTC (6am UTC-5)
+    - cron: '0 11 15 * *'
```
jeancochrane: **[Suggestion, optional]** Since the conditionals in the `UPLOAD_SCHEDULE` env var depend on the precise formatting of these schedules, let's make sure to add a comment warning that any change should be duplicated across the two instances of the schedule definitions. (It would be great if we could factor out a variable to centralize these definitions, but I'm pretty sure that's not possible in the context of a workflow schedule.) We also might as well add a little bit more context explaining the double schedules, as long as we're here:

```suggestion
  # We use these schedules to determine the value of the `UPLOAD_SCHEDULE`
  # env var below, so make sure to change that env var definition if you
  # change any of these schedules.
  # 
  # Run a monthly job on the first of every month at 11am UTC (6am UTC-5).
    - cron: '0 11 1 * *'
  # Run a semi-monthly job on the fifteenth of every month at 11am UTC (6am UTC-5).
    - cron: '0 11 15 * *'
```

.github/workflows/socrata_upload.yaml
```diff
@@ -82,4 +84,7 @@ jobs:
           OVERWRITE: ${{ inputs.overwrite }}
           YEARS: ${{ inputs.years }}
           WORKFLOW_EVENT_NAME: ${{ github.event_name }}
+          UPLOAD_SCHEDULE: >
+            ${{ github.event.schedule == '0 11 1 * *' && 'monthly' ||
+            github.event.schedule == '0 11 15 * *' && 'semimonthly' || '' }}
```
jeancochrane, line 97: **[Suggestion, optional]** Similar comment here as above:

```suggestion
          # Determine if we're running on the monthly or semi-monthly schedule.
          # These cron expressions are shared by the `on.schedule` workflow
          # attribute defined above, so make sure to change that attribute
          # definition alongside any changes you make here
          UPLOAD_SCHEDULE: >
            ${{ github.event.schedule == '0 11 1 * *' && 'monthly' ||
            github.event.schedule == '0 11 15 * *' && 'semimonthly' || '' }}
```

jeancochrane (approved):
I'm not 100% sure if the expression in the `UPLOAD_SCHEDULE` variable definition will evaluate correctly, but I can't think of a great way to test it, so I'm fine merging this and seeing what happens on the next scheduled run dates.

PR #1006, opened 2026-03-25
Remove embedded nulls in address view

wrridgeway:
I was getting errors in R when querying `default.vw_pin_address`: 

```
Error in `rbindlist()`:
! embedded nul in string: '\0'; to strip nuls when converting from Arrow to R, set options(arrow.skip_nul = TRUE)
```

In response I've made sure `chr(0)` gets replaced in the problem columns from `iasworld.maildat`. Other than the `chr(0)` replacement, the transformation for these columns is the same. I've moved this code to a macro since it was pretty convoluted and needed to be repeated, and so we can apply it anywhere else we might need to.

Running:

```
library(arrow)
library(DBI)
library(noctua)
library(readr)
library(stringr)
library(tibble)

options(scipen = 999)

noctua_options(unload = TRUE)
AWS_ATHENA_CONN_NOCTUA <- dbConnect(noctua::athena(), rstudio_conn_tab = FALSE)

DBI::dbGetQuery(
  AWS_ATHENA_CONN_NOCTUA, "select * from default.vw_pin_address where year in ('2023', '2024')"
)

DBI::dbGetQuery(
  AWS_ATHENA_CONN_NOCTUA, "select * from z_ci_remove_embedded_nulls_in_address_view_default.vw_pin_address where year in ('2023', '2024')"
)
```

shows the error is not a problem after these changes.

Running

```
with new as (
	select pin,
		year,
		mail_address_name,
		mail_address_full
	from z_ci_remove_embedded_nulls_in_address_view_default.vw_pin_address
),
old as (
	select pin,
		year,
		mail_address_name,
		mail_address_full
	from default.vw_pin_address
)
select new.pin,
	new.year,
	new.mail_address_name as mail_address_name_new,
	old.mail_address_name as mail_address_name_old,
	new.mail_address_full as mail_address_full_new,
	old.mail_address_full as mail_address_full_old
from new
	left join old on new.pin = old.pin
	and new.year = old.year
where new.mail_address_name != old.mail_address_name
	or new.mail_address_full != old.mail_address_full
```

returns one row, the problem row:

| pin            | year | mail_address_name_new | mail_address_name_old | mail_address_full_new | mail_address_full_old |
|----------------|------|-----------------------|-----------------------|-----------------------|-----------------------|
| 27164010150000 | 2021 |                      |                      | SWC 156TH LLC         | SWC 156TH LLC         |


dbt/macros/clean_address.sql
```diff
@@ -0,0 +1,12 @@
+-- Macro that cleans and concatenates address columns by replacing multiple
+-- spaces with a single space, removing null characters, and returning null
+-- if the result is an empty string.
+{% macro clean_address(address_columns) %}
```
jeancochrane: **[Nitipick, optional]** I think the behavior of this macro might be more intuitive if it had a name like `concat_address`, to make it clear that it expects an array input and will combine the results.

dbt/macros/clean_address.sql
```diff
@@ -0,0 +1,12 @@
...
+    {% set column_name = address_columns | join(", ") %}
+    nullif(
+        replace(
+            regexp_replace(concat_ws(' ', {{ column_name }}), '\s+', ' '), chr(0), ''
+        ),
+        ''
+    )
+{% endmacro %}
```
jeancochrane: **[Nitpick, required]** It's an annoying chore, but do you mind adding some unit tests for this macro to match [our other macro unit tests](https://github.com/ccao-data/data-architecture/tree/master/dbt/macros/tests)?

jeancochrane (changes requested):
Nice find!

I'm a little bit confused about your QC query that checks for mismatches. For one thing, the result for the row you shared make it seem like there is still an embedded null in `mail_address_name_new`. What's up with that? Also, I tweaked the query to check for mismatches between types of nulls, and it surfaced a few extra rows that switched from empty strings to nulls in `mail_address_name`. Do you have a sense of why the previous version of the query wasn't coercing these empty strings in its own `NULLIF()` expression?

```sql
with new as (
	select pin,
		year,
		mail_address_name,
		mail_address_full
	from z_ci_remove_embedded_nulls_in_address_view_default.vw_pin_address
),
old as (
	select pin,
		year,
		mail_address_name,
		mail_address_full
	from default.vw_pin_address
)
select new.pin,
	new.year,
	new.mail_address_name as mail_address_name_new,
	old.mail_address_name as mail_address_name_old,
	new.mail_address_full as mail_address_full_new,
	old.mail_address_full as mail_address_full_old
from new
	left join old on new.pin = old.pin
	and new.year = old.year
where new.mail_address_name != old.mail_address_name
    or (new.mail_address_name is null and old.mail_address_name is not null)
    or (new.mail_address_name is not null and old.mail_address_name is null)
	or new.mail_address_full != old.mail_address_full
    or (new.mail_address_full is null and old.mail_address_full is not null)
    or (new.mail_address_full is not null and old.mail_address_full is null)
```

CSV output, which shows the empty strings in constrast to the nulls:

```
"pin","year","mail_address_name_new","mail_address_name_old","mail_address_full_new","mail_address_full_old"
"20324030160000","2025",,"","1335 S PRAIRIE AVE1808","1335 S PRAIRIE AVE1808"
"20111010210000","2025",,"","904 E 48TH ST","904 E 48TH ST"
"20111010210000","2026",,"","904 E 48TH ST","904 E 48TH ST"
"20324030160000","2026",,"","1335 S PRAIRIE AVE1808","1335 S PRAIRIE AVE1808"
"20111010210000","2024",,"","904 E 48TH ST","904 E 48TH ST"
"27164010150000","2021","   ","   ","SWC 156TH LLC","SWC 156TH LLC"
"20324030160000","2024",,"","1335 S PRAIRIE AVE1808","1335 S PRAIRIE AVE1808"
```

wrridgeway:
Can we pretend my query didn't show essentially the opposite of what I said it did

wrridgeway:
Alright, sorry about how sloppy that was. I think if you inspect the output from your query again you'll see the new code handles all control characters and re-ordering the `regexp_replace` clauses makes sure there aren't strings of multiple spaces left over. I added a few unit tests as well.

As for the old syntax leaving empty strings rather than nulls, they aren't actually empty strings, they have control characters in them. For pin `20324030160000` for example, the old value wasn't "" it was 
image

dbt/macros/tests/test_concat_address.sql
jeancochrane, line 1: **[Thought, non-blocking]** It's annoying, but the macro test design is kind of hacked together, and one consequence is that you have to make sure to call any new macro tests in the main [`test_all.sql`](https://github.com/ccao-data/data-architecture/blob/master/dbt/macros/tests/test_all.sql) file or else they won't get picked up by the `build-and-test-dbt` workflow:

https://github.com/ccao-data/data-architecture/blob/e048a95f1c40984aebb8684258eea7f10caa83e9/.github/workflows/build_and_test_dbt.yaml#L65-L68

However, see my thought below about how a dbt unit test might be more appropriate than a macro test in this case. I'm leaving this comment anyway for the sake of knowledge transfer.

dbt/macros/tests/test_concat_address.sql
```diff
@@ -0,0 +1,35 @@
...
+
+{% macro test_concat_address_replaces_control_characters() %}
+    {{
+        assert_equals(
+            "test_concat_address_replaces_control_characters",
+            concat_address(["'foo'", "'bar'", "chr(15)"]),
+            "foo bar",
+        )
+    }}
+{% endmacro %}
```
jeancochrane: **[Thought, blocking]** My bad for not thinking about this earlier, but upon reflection, I think macro tests are actually not the appropriate way of testing this change. The failure for this test hints at the reason why:

```
$ dbt run-operation test_concat_address
...
    test_concat_address_replaces_control_characters - FAIL:

        nullif(
            regexp_replace(
                regexp_replace(concat_ws(' ', 'foo', 'bar', chr(15)), '[[:cntrl:]]', ''),
                '\s+',
                ' '
            ),
            ''
        )
     != foo bar
```

The trick here is that calling a dbt macro doesn't actually execute any SQL, it just templates the text according to the macro logic. So the output of this test is the text of the `nullif()` function, not the final value `"foo bar"`.

Since the macro logic is so simple, and it's really the SQL logic that is doing the heavy lifting, I wonder if this might be a good opportunity for a dbt unit test instead of macro tests? We could set up `maildat` fixtures that have the various combinations of whitespace and control characters that you include in these macro tests, and make sure that the resulting SQL produces the `mail_address_name` and `mail_address_full` values that we expect. [See here](https://github.com/ccao-data/data-architecture/blob/e048a95f1c40984aebb8684258eea7f10caa83e9/dbt/models/default/schema/default.vw_pin_address.yml#L83-L104) for an example of an existing `default.vw_pin_address` unit test that you could build off of.

However, I also acknowledge that my unit testing fervor has already derailed this relatively simple PR, so I'm fine moving ahead without tests knowing that you've tested it by hand already. I leave the decision up to you, depending on how excited you would be to write a dbt unit test.

jeancochrane (approved):
I led you astray with those macro tests... my bad! I think we should remove them, but I leave it up to you to decide if you want to replace them with a dbt unit test, per my comment below.

dbt/macros/tests/test_concat_address.sql
wrridgeway: I'm happy to add a unit test, these tests save me from myself all the time.

wrridgeway: Unfortunately we can't test control characters using unit test:

image

wrridgeway: Unit test for removing spaces working:

image

PR #1009, opened 2026-03-30
Revert "Move HIE counts to res model view"

wrridgeway:
Reverts ccao-data/data-architecture#995

@jeancochrane If we don't need HIE flags, there's no reason I can think of to add these columns to the model view, and they create [extra work in the model pipeline](https://github.com/ccao-data/model-res-avm/pull/455/changes#diff-6b1854802e961990f3d40a59c31191eb3d9a78b65a65c1658f16e6c081bf95f9R262) since they only apply to the assessment set. We can move any information cataloged here to the issue I have to record HIE knowledge.

### DIFF
--- a/dbt/models/default/default.vw_card_res_char.sql
+++ b/dbt/models/default/default.vw_card_res_char.sql
@@ -12,58 +12,6 @@ WITH multicodes AS (
     GROUP BY parid, taxyr
 ),
 
--- Gather HIE start and end years from both ADDN and OBY tables. We need to
--- stack them to properly count the number of active and expiring HIEs for each
--- PIN-year combination below. We define unique start and end dates within cards
--- as separate HIEs.
-all_hies AS (
-    SELECT DISTINCT
-        parid AS pin,
-        card,
-        taxyr AS year,
-        CAST(taxyr AS INT) AS year_int,
-        userval1 AS hie_start,
-        userval2 AS hie_end
-    FROM {{ source('iasworld', 'addn') }}
-    WHERE lline > 0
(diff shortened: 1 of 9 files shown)


dbt/models/default/default.vw_card_res_char.sql
```diff
@@ -12,58 +12,6 @@ WITH multicodes AS (
...
-                -- end years, so we use a between statement here.
-                year_int BETWEEN hie_start AND hie_end AS INT
-            )
-        ) AS hie_num_active,
-        SUM(CAST(year_int = hie_end AS INT)) AS hie_num_expiring
-    FROM all_hies
-    GROUP BY
-        pin,
-        year
-),
```
jeancochrane, line 65: **[Question, non-blocking]** I wonder if it's worth continuing to preserve this information in `default.vw_card_res_char` or elsewhere, so that we can use it for analysis later? I agree that it makes sense to remove it from the modeling view but I'm curious if you think there's somewhere else where it would make more sense.

jeancochrane (approved):
Thanks for your diligence here!

dbt/models/default/default.vw_card_res_char.sql
wrridgeway: I think we can add it to this [issue](https://github.com/orgs/ccao-data/projects/22/views/9?filterQuery=assignee%3A%40me+sprint%3A%40current&layout=table&groupedBy%5BcolumnId%5D=Status&pane=issue&itemId=156087315).

PR #1013, opened 2026-04-10
Ingest abnormal 2025 city tri sale review

wagnerlmichael:
We received a file with updated analyst review classifications for the most non-standard 2025 sales in city-tri, according to algorithmic sales val. We make a small set of modifications to capture non-standard data.

Data is updated in S3/prod.

Counts of new data can be viewed with this query -

```sql
SELECT column_name, value, count
FROM (
    SELECT 'is_arms_length' AS column_name, CAST(is_arms_length AS VARCHAR) AS value, COUNT(*) AS count
    FROM sale.flag_review
    WHERE source_file = 'valuations_sale_review_03_05_2026.xlsx'
    GROUP BY is_arms_length

    UNION ALL

    SELECT 'is_flip', CAST(is_flip AS VARCHAR), COUNT(*)
    FROM sale.flag_review
    WHERE source_file = 'valuations_sale_review_03_05_2026.xlsx'
    GROUP BY is_flip

    UNION ALL

    SELECT 'has_class_change', CAST(has_class_change AS VARCHAR), COUNT(*)
    FROM sale.flag_review
    WHERE source_file = 'valuations_sale_review_03_05_2026.xlsx'
    GROUP BY has_class_change

    UNION ALL

    SELECT 'has_characteristic_change', CAST(has_characteristic_change AS VARCHAR), COUNT(*)
    FROM sale.flag_review
    WHERE source_file = 'valuations_sale_review_03_05_2026.xlsx'
    GROUP BY has_characteristic_change

    UNION ALL

    SELECT 'requires_field_check', CAST(requires_field_check AS VARCHAR), COUNT(*)
    FROM sale.flag_review
    WHERE source_file = 'valuations_sale_review_03_05_2026.xlsx'
    GROUP BY requires_field_check

    UNION ALL

    SELECT 'work_drawer', CAST(work_drawer AS VARCHAR), COUNT(*)
    FROM sale.flag_review
    WHERE source_file = 'valuations_sale_review_03_05_2026.xlsx'
    GROUP BY work_drawer
) sub
ORDER BY column_name, value
```

### DIFF
--- a/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
@@ -12,6 +12,7 @@ library(aws.s3)
 library(dplyr)
 library(janitor)
 library(readxl)
+library(stringr)
 
 source("utils.R")
 
--- a/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
@@ -225,12 +225,12 @@ dfs$valuations_sale_review_01_30_2026_peter <-
     )
   )
 
-# Two rows in Jon's 03_05_2026 workbook use free-text values our regexes
+# Two rows in the 03_05_2026 workbook use free-text values our regexes
 # don't catch, so we rewrite them here before transform_columns() runs:
 # - doc 2516820281: Characteristic Change "YES-FIRE DAMAGE" -> "YES-MAJOR"
 # - doc 2529320200: Class Change "I CHANGED TO 2-10" -> "YES"
-dfs$valuations_sale_review_03_05_2026_jon <-
-  dfs$valuations_sale_review_03_05_2026_jon %>%
+dfs$valuations_sale_review_03_05_2026 <-
+  dfs$valuations_sale_review_03_05_2026 %>%
   mutate(
     `Characteristic Change` = str_replace(`Characteristic Change`, "YES-FIRE DAMAGE", "YES-MAJOR") ,
     `Class Change` = str_replace(`Class Change`, "I CHANGED TO 2-10", "YES")
--- a/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
+++ b/etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
@@ -230,6 +230,25 @@ dfs$valuations_sale_review_01_30_2026_peter <-
     )
   )
 
+# Two rows in Jon's 03_05_2026 workbook use free-text values our regexes
+# don't catch, so we rewrite them here before transform_columns() runs:
+# - doc 2516820281: Characteristic Change "YES-FIRE DAMAGE" -> "YES-MAJOR"
+# - doc 2529320200: Class Change "I CHANGED TO 2-10" -> "YES"
+dfs$valuations_sale_review_03_05_2026_jon <-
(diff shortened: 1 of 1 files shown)


etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
wrridgeway:
```suggestion
objs <- get_bucket_df(
  bucket = "ccao-data-raw-us-east-1",
  prefix = "sale/flag_review/"
) %>%
  filter(grepl("xlsx", Key)) %>%
  pull(Key)
```

can I make this suggestion? `get_bucket_df` is a lot easier to work with than `get_bucket`, and if we use grepl we won't have to worry about skip logic in the for loop below.

wrridgeway: I'm sorry, every time I commit a suggestion that involves unchanged lines of code github is pretty squirrely about it. I'd be careful using `Commit suggestion` here.

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
```diff
@@ -230,6 +230,25 @@ dfs$valuations_sale_review_01_30_2026_peter <-
...
+    `Characteristic Change` = if_else(
+      `Sale Doc No` == "2516820281",
+      "YES-MAJOR",
+      `Characteristic Change`
+    ),
+    `Class Change` = if_else(
+      `Sale Doc No` == "2529320200",
+      "YES",
+      `Class Change`
+    )
```
wrridgeway: Rather than depend on the sale doc, could we trigger changes when we detect the string pattern we want to change? It feels a little indirect using doc number to change a string when you already know what the string is. You won't even have to document what you're changing to what in your comment if you make the code this literal. You'll have to use `stringr` for this syntax:

```suggestion
  mutate(
    `Characteristic Change` = str_replace(`Characteristic Change`, "YES-FIRE DAMAGE", "YES-MAJOR") ,
    `Class Change` = str_replace(`Class Change`, "I CHANGED TO 2-10", "YES")
  )
```

wrridgeway (approved):
Thanks for working on this!

wrridgeway:
Sorry for being annoying and changing that ingest syntax, I just wanted to make it a bit more tidyverse.

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
wagnerlmichael: Nice, this is much nicer

etl/scripts-ccao-data-warehouse-us-east-1/sale/sale-flag_review.R
```diff
@@ -12,39 +12,35 @@ library(aws.s3)
...
-
-for (obj in objs) {
-  key <- obj[["Key"]]
-  # I believe this line is needed because get_bucket
-  # also returns prefix objects that end in /
-  if (is.null(key) || grepl("/$", key)) next
-
-  s3_uri <- paste0("s3://ccao-data-raw-us-east-1/", key)
+dfs <- objs %>%
+  set_names(tools::file_path_sans_ext(objs)) %>%
```
wagnerlmichael, line 30: we need this otherwise the files get written as xlsx.parquet

jeancochrane (approved):
You nailed my one requested change, so let's just make sure Billy is happy with the R code and then we're good to go!

jeancochrane:
This is waiting on a final review from @wrridgeway.

wrridgeway (approved):
Awesome work, looks good to me!

PR #1020, opened 2026-04-22
Fix semimonthly socrata upload

### LINKED ISSUE #1018: Twice-a-month Socrata refresh updated all assets
Looks like our twice-a-month refresh that ran today on the 15th [updated all assets](https://github.com/ccao-data/data-architecture/actions/runs/24451602639/job/71441557522#step:7:37):

```
Assets that will be updated:
0                                              Appeals
1                                      Assessed Values
2                                     Parcel Addresses
3                                         Parcel Sales
4                  Parcel Universe (Current Year Only)
5                           Parcel Universe (Historic)
6                                              Permits
7                          Property Tax-Exempt Parcels
8         Residential Condominium Unit Characteristics
9    Single and Multi-Family Improvement Characteri...
```

I don't think this was expected behavior -- it should have only updated assessed values. Let's see if we can figure out why it did this.

wrridgeway:
I've reverted some of the changes to `socrata_upload.yaml`, but you can see what they were like when I successfully triggered a semimonthly scheduled workflow [here](https://github.com/ccao-data/data-architecture/blob/5a6a2d3bc79fb1497a940455e3678a4c9ba9ebad/.github/workflows/socrata_upload.yaml) (disabled requirement for user to define assets, `workflow_dispatch` event returns `semimonthly` for `UPLOAD_SCHEDULE`).

This let me print what was being passed to `UPLOAD_SCHEDULE` and it seems like it was `semimonthly` but with a newline at the end. Using a folded block scalar with strip chomping `>-` for `UPLOAD_SCHEDULE` [sorts this out](https://github.com/ccao-data/data-architecture/actions/runs/24796610729/job/72567997220).

You can see a workflow with the unwanted newline in `UPLOAD_SCHEDULE` [here](https://github.com/ccao-data/data-architecture/actions/runs/24795524548/job/72564126704).

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -76,9 +76,7 @@ def parse_assets(assets=None):
     # updated once per month. If the upload is running on a monthly schedule,
     # we want to include those assets.
     monthly_tag = (
-        "tag:monthly"
-        if str(os.getenv("UPLOAD_SCHEDULE")).strip() == "semimonthly"
-        else ""
+        "tag:monthly" if os.getenv("UPLOAD_SCHEDULE") == "semimonthly" else ""
     )
 
     DBT = dbtRunner()
@@ -496,10 +494,7 @@ if __name__ == "__main__":
     event_type = " ".join(
         filter(
             None,
-            [
-                str(os.getenv("UPLOAD_SCHEDULE")).strip(),
-                os.getenv("WORKFLOW_EVENT_NAME"),
-            ],
+            [os.getenv("UPLOAD_SCHEDULE"), os.getenv("WORKFLOW_EVENT_NAME")],
         )
     )
     print(f"Running upload for event type: {event_type}")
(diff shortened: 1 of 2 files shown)


socrata/socrata_upload.py
```diff
@@ -491,7 +491,13 @@ def socrata_upload(asset_info, overwrite=False, years=None):
...
 
 if __name__ == "__main__":
-    print(f"Running upload for event type: {os.getenv('WORKFLOW_EVENT_NAME')}")
+    event_type = " ".join(
+        filter(
+            None,
+            [os.getenv("UPLOAD_SCHEDULE"), os.getenv("WORKFLOW_EVENT_NAME")],
+        )
+    )
+    print(f"Running upload for event type: {event_type}")
```
wrridgeway, line 500: Just adding more printing here so it's easier to see what the script is being passed.

jeancochrane (approved):
Such a devious little bug. Nice catch!

PR #1022, opened 2026-04-27
Create view from cmap data EI issue

### LINKED ISSUE #1021: Create view from CMAP data issue
The data generated by [this query](https://github.com/ccao-data/enterprise-intelligence/blob/main/ei-issue-0364/ei-issue-0364.sql) will likely need to be exported every year for CMAP. We can create a permanent view to make this task easier to replicate.

wrridgeway:
This view should make a yearly export to cmap more manageable. It is essentially the same as the query from this [EI issue](https://github.com/ccao-data/enterprise-intelligence/pull/412). I created a new `export` schema for this since it didn't seem to fit well into anything we already have, but I'm happy to move it elsewhere if we're not into it.


dbt/models/external/schema/external.vw_cmap.yml
```diff
@@ -0,0 +1,139 @@
...
+      - name: mail_address_name
+        description: '{{ doc("shared_column_mail_address_name") }}'
+      - name: mail_address_state
+        description: '{{ doc("shared_column_mail_address_state") }}'
+      - name: mail_address_zipcode_1
+        description: '{{ doc("shared_column_mail_address_zipcode_1") }}'
+      - name: multi_imp_num
+        description: '{{ doc("shared_column_pin_num_cards") }}'
+      - name: municipality
+        description: Municipality name
```
wrridgeway: We could move this to a shared column description if we feel the need.

dbt/models/external/external.vw_cmap.sql
```diff
@@ -0,0 +1,224 @@
...
+    -- town, muni, tax code
+    vpu.township_name AS township,
+    vpu.nbhd_code AS neighborhood,
+    -- Combined municipality name is an array. Empty arrays indicate
+    -- unincorporated parcels. Null values indicate missing data.
+    CASE
+        WHEN CARDINALITY(vpu.combined_municipality_name) = 0
+            THEN 'UNINCORPORATED'
+        ELSE ARRAY_JOIN(vpu.combined_municipality_name, ', ')
+    END AS municipality,
```
wrridgeway, line 147: Updated from the EI issue.

dbt/models/external/schema/external.vw_cmap.yml
```diff
@@ -0,0 +1,139 @@
...
+      - name: building_av
+        description: '{{ doc("shared_column_board_bldg") }}'
+      - name: building_sf
+        description: '{{ doc("shared_column_char_bldg_sf") }}'
+      - name: card
+        description: '{{ doc("shared_column_card") }}'
+      - name: class_code
+        description: '{{ doc("shared_column_class") }}'
+      - name: class_description
+        description: 'A short description of the class code'
```
wrridgeway, line 25: We could move this to a shared column description if we feel the need, but the only other place I saw a column like this was in pinval and I'm not sure if that column is exactly the same as what we have here.

jeancochrane: **[Thought, non-blocking]** Yeah, the HomeVal class description is slightly different. It seems like this column is just coming directly from the `ccao.class_dict` seed though, so potentially those two columns could share docs? But since [we're not currently documenting the columns on `ccao.class_dict`](https://github.com/ccao-data/data-architecture/blob/eececaf8bfbe65431907eb2300c773d07e3dc216/dbt/seeds/ccao/schema.yml#L33-L47), I don't feel strongly about it.

dbt/models/export/schema/export.vw_cmap.yml
```diff
@@ -0,0 +1,139 @@
...
+      - name: total_av
+        description: '{{ doc("shared_column_board_tot") }}'
+      - name: township
+        description: '{{ doc("shared_column_township_name") }}'
+      - name: wall_construction
+        description: '{{ doc("shared_column_char_ext_wall") }}'
+      - name: year
+        description: '{{ doc("shared_column_year") }}'
+      - name: year_built
+        description: Year built
```
jeancochrane: **[Nitpick, optional]** Any reason not to use [the shared column description](https://github.com/ccao-data/data-architecture/blob/master/dbt/models/shared_columns.md#char_yrblt) here?


```suggestion
      - name: year_built
        description: '{{ doc("shared_column_char_yrblt") }}'
```

dbt/models/export/schema/export.vw_cmap.yml
```diff
@@ -0,0 +1,139 @@
...
+      - row_count:
+          name: export_vw_cmap_row_count
+          equals: >
+            (
+              SELECT COUNT(*)
+              FROM (
+                SELECT DISTINCT parid, taxyr
+              FROM {{ source("iasworld", "asmt_all") }}
+              WHERE procname = 'BORVALUE'
+                AND valclass IS NULL -- Class 999 are test pins
```
jeancochrane: **[Nitpick, optional]** Technically this comment belongs to the following line:


```suggestion
                AND valclass IS NULL
                -- Class 999 are test pins
```

dbt/models/export/export.vw_cmap.sql
```diff
@@ -0,0 +1,224 @@
...
+        taxyr,
+        ARBITRARY(class) AS class,
+        ARBITRARY(valasm1) AS land,
+        ARBITRARY(valasm2) AS bldg,
+        ARBITRARY(valasm3) AS tot
+    FROM {{ source("iasworld", "asmt_all") }}
+    WHERE deactivat IS NULL
+        AND procname = 'BORVALUE'
+        AND valclass IS NULL -- Class 999 are test pins
+        AND class NOT IN ('999')
```
jeancochrane: **[Nitpick, optional]** Same here as above:


```suggestion
        AND valclass IS NULL
        -- Class 999 are test pins
        AND class NOT IN ('999')
```

dbt/models/export/export.vw_cmap.sql
```diff
@@ -0,0 +1,224 @@
...
+        AND taxyr IN ('2023', '2024')
```
jeancochrane: **[Question, non-blocking]** What's our plan for updating this in subsequent years? Do we just plan to put up a PR every year incrementing these year values? I could also add it to my draft list of yearly update chores, if we want to institutionalize it.

dbt/models/external/external.vw_cmap.sql
```diff
@@ -0,0 +1,224 @@
...
+        card,
+        pin_num_cards AS multi_imp_num,
+        char_bldg_sf AS building_sf,
+        char_beds AS beds,
+        char_rooms AS rooms,
+        char_fbath AS full_bath,
+        char_hbath AS half_bath,
+        CASE WHEN char_frpl = 1 THEN 'Full'
+            WHEN char_frpl = 2 THEN 'Partial'
+            WHEN char_frpl = 3 THEN 'None'
```
jeancochrane, line 83: **[Thought, non-blocking]** No action now, but it would be really cool if we could centralize this kind of SQL char conersion logic in the same way we do for R/Python code in the `ccao` package. Something to think about for a future improvement.

jeancochrane (approved):
I only took a quick skim of the query since it was already reviewed, but overall it looks good to me! This is a very cool view that aggregates a lot of disparate tables and I bet other folks outside CMAP would be interested it. I assume you and @ccao-jardine have talked about this already, but it would be interesting to think about what this could look like as an open data asset, which would also remove the need for scheduling and completing a yearly export.

Regarding the schema name: We could also consider `external`, which I think would be slightly more descriptive than `export`. But ultimately I'm fine with either.

dbt/models/export/export.vw_cmap.sql
wrridgeway: oh gosh, i think i should just remove these years. i dont' think there's any reason someone can't just add their own where clause when pulling the data.

ccao-jardine:
> I only took a quick skim of the query since it was already reviewed, but overall it looks good to me! This is a very cool view that aggregates a lot of disparate tables and I bet other folks outside CMAP would be interested it. I assume you and @ccao-jardine have talked about this already, but it would be interesting to think about what this could look like as an open data asset, which would also remove the need for scheduling and completing a yearly export.
> 
> Regarding the schema name: We could also consider `external`, which I think would be slightly more descriptive than `export`. But ultimately I'm fine with either.

Re: potential Open Data Asset, agreed, @jeancochrane, I think this has strong potential! We didn't go for it quite yet because of the nuance of working with this bespoke data model. Rows are distinct to PIN and card (and year), but we provide PIN AVs -- such that if a PIN has two cards, the AV is duped in two rows. Users without CMAP-level knowledge could draw the wrong conclusions by, for example, summing up all AVs of all rows, rather than summing AVs per distinct PIN. 

But I'm open to being convinced 😎

PR #1028, opened 2026-05-05
Add type hinting to socrata upload script and improve commenting

wrridgeway:
Thank you, copilot. PR to make this script easier to understand w/o context.

See a successful upload from this branch [here](https://github.com/ccao-data/data-architecture/actions/runs/25395284065/job/74480294516).

### DIFF
--- a/socrata/socrata_upload.py
+++ b/socrata/socrata_upload.py
@@ -58,8 +58,18 @@ cursor = connect(
 
 def parse_assets(assets: str | None = None) -> dict[str, dict[str, Any]]:
     """
-    Retrieve metadata for all socrata assets and filter that data based
-    on parsed input assets names.
+    Retrieve metadata for all Socrata assets and filter that data based
+    on parsed input asset names.
+
+    Args:
+        assets: Comma-separated string of asset label names to filter to.
+            If None or empty, metadata for all active assets is returned.
+
+    Returns:
+        A dict keyed by asset label, where each value is a dict containing
+        the keys ``asset_id`` (Socrata dataset ID), ``years_active`` (number
+        of years that should be kept current), and ``athena_asset`` (fully
+        qualified Athena view name).
     """
 
     if assets == "":
@@ -146,7 +156,17 @@ def parse_assets(assets: str | None = None) -> dict[str, dict[str, Any]]:
 
 def parse_years(years: str | None = None) -> list[str] | None:
     """
-    Make sure the years environmental variable is formatted correctly.
+    Parse and validate the years environment variable.
+
+    Args:
+        years: A comma-separated string of four-digit year values, the
+            special string ``"all"`` to select every available year, or
+            None/empty string to indicate no year filter (use the default
+            scheduled-run behaviour instead).
+
+    Returns:
+        A list of year strings (e.g. ``["2022", "2023"]``), the single-
+        element list ``["all"]``, or None if no years were provided.
     """
 
     if years == "":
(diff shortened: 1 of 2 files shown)


socrata/socrata_upload.py
```diff
@@ -143,31 +154,63 @@ def parse_assets(assets=None):
...
             # Only allow anticipated values
-            years = [
+            years_parsed = [
                 re.sub("[^0-9]", "", year)
                 for year in str(years).split(",")
                 if re.sub("[^0-9]", "", year)
             ]
 
-    return years
+    return years_parsed
```
wrridgeway, line 190: Appeasing mypy.

socrata/socrata_upload.py
```diff
@@ -287,7 +363,7 @@ def build_query_dict(athena_asset, asset_id, years=None):
 
     # Build a dictionary with queries for each year requested, or no years
     if not years:
-        query_dict = {None: query}
+        query_dict: dict[str | None, str] = {None: query}
```
wrridgeway: Appeasing mypy.

socrata/socrata_upload.py
```diff
@@ -302,32 +378,46 @@ def build_query_dict(athena_asset, asset_id, years=None):
...
-        where.append(f"year IN ({years})")
-    where = " or ".join(where)
+    where_list = []
+    if "nan" in years_list:
+        years_list.remove("nan")
+        where_list.append("year is NULL")
+    if any(year != "nan" for year in years_list):
+        years_str = ", ".join(years_list)
+        where_list.append(f"year IN ({years_str})")
+    where_str = " or ".join(where_list)
```
wrridgeway, line 420: Appeasing mypy.

socrata/socrata_upload.py
```diff
@@ -403,12 +508,12 @@ def check_missing_years(athena_asset, asset_id):
...
         if missing_years:
-            missing_years = ", ".join(list(missing_years))
-            where = f"year in ({missing_years})"
+            missing_years_str = ", ".join(list(missing_years))
+            where = f"year in ({missing_years_str})"
         if socrata_nulls:
             where = f"{socrata_nulls}"
         if missing_years and socrata_nulls:
-            where = f"year in ({missing_years}) or {socrata_nulls}"
+            where = f"year in ({missing_years_str}) or {socrata_nulls}"
```
wrridgeway, line 522: Appeasing mypy.

socrata/socrata_upload.py
```diff
@@ -424,10 +529,33 @@ def check_missing_years(athena_asset, asset_id):
     return years_to_remove
 
 
-def upload(asset_id, sql_query, overwrite, missing_years):
+def upload(
+    asset_id: str,
+    sql_query: dict[Any, str],
```
jeancochrane: **[Question, non-blocking]** If this arg expects a "mapping of year keys to SQL query strings", as the "Args" section of the docstring suggests, why do we use `Any` for the type hint for keys? Can we be more specific?

socrata/socrata_upload.py
```diff
@@ -231,11 +287,31 @@ def check_overwrite(overwrite=None):
     return overwrite
 
 
-def build_query_dict(athena_asset, asset_id, years=None):
+def build_query_dict(
+    athena_asset: str,
+    asset_id: str,
+    years: list[Any] | None = None,
```
jeancochrane: **[Question, optional]** Why are the list elements `Any` here? Why not `list[str]`?

jeancochrane (approved):
Awesome work! These type hints and docstrings are super helpful.

socrata/socrata_upload.py
```diff
@@ -143,31 +160,64 @@ def parse_assets(assets=None):
...
+        years_parsed = None
+    elif years == "all":
+        years_parsed = ["all"]
+    else:
+        # Only allow anticipated values
+        years_parsed = [
+            re.sub("[^0-9]", "", year)
+            for year in str(years).split(",")
+            if re.sub("[^0-9]", "", year)
+        ]
```
wrridgeway, line 188: This is just better logic that accomplishes the same thing as before.

socrata/socrata_upload.py
wrridgeway: Alright, I worked a bit more on this and getting rid of `Any` led to me having to work through the type hints a bit more.  Long story short, there are a number of different output types `parse_years_list` and `build_query_dict` can return, and a number of input types they have to be able to handle. Seeing as how the script has been working as expected lately, I don't want to screw with that in this PR and have such sensitive changes be buried in so many other. So if we'd like to tighten up what `parse_years_list` can return I'm happy to put that into another PR and do the testing.

jeancochrane: Thanks for looking into it! All good by me, I agree that it's not worth the extra effort to have rock solid type hints.

jeancochrane (approved):
Thanks for tightening this up!

PR #1032, opened 2026-05-14
Update workflows, actions, and docs to ensure we install Python dependencies from lockfiles

jeancochrane:
## Background

As we found in https://github.com/ccao-data/homeval/pull/158 and https://github.com/ccao-data/homeval/pull/162, the way we use `uv pip install` in our workflows, actions, and docs in this repo will cause uv to ignore `uv.lock` when installing dependencies; instead, it will decide which dependency versions to install based on `pyproject.toml`. This PR updates all workflows, actions, and docs to instead use uv install and run commands that will ensure we use the exact versions of packages pinned in our lockfiles.

## Testing

I tested the new docs instructions by using them to recreate my dev environment:

```bash
cd dbt
rm -rf .venv
uv sync --frozen --all-extras
uv run --frozen dbt deps
source .venv/bin/activate
dbt compile
```

I ran all changed workflows to confirm that they continue to work:

- [x] `build-and-test-dbt`
    - Workflow dispatch, to test `Running build on manually selected resources` branch: https://github.com/ccao-data/data-architecture/actions/runs/25877837759/job/76049632055
    - Automatically run on model change, to test `Running build on modified/new resources only` branch: https://github.com/ccao-data/data-architecture/actions/runs/25878423397/job/76051648843
    - I skipped the `Running build on modified/new resources and their children` branch because it requires deploying to prod, and I also skipped the `Running build on all resources` branch to save time
- [x] `build-daily-dbt-models`: https://github.com/ccao-data/data-architecture/actions/runs/25878689974/job/76052579087
- [x] `deploy-dbt-docs`: https://github.com/ccao-data/data-architecture/actions/runs/25878767394/job/76052852761
- [x] `socrata-upload`: https://github.com/ccao-data/data-architecture/actions/runs/25881072615/job/76060913562
- [x] `test-dbt-models`: https://github.com/ccao-data/data-architecture/actions/runs/25878871492/job/76053222128
- [x] `test-dbt-source-freshness`: https://github.com/ccao-data/data-architecture/actions/runs/25878937860/job/76053449725
- [x] `test-iasworld-data`: https://github.com/ccao-data/data-architecture/actions/runs/25879218101/job/76054453092
- [x] `test-open-data-assets`: https://github.com/ccao-data/data-architecture/actions/runs/25879381612/job/76055018772

### DIFF
--- a/.github/actions/install_dbt_requirements/action.yaml
+++ b/.github/actions/install_dbt_requirements/action.yaml
@@ -24,9 +24,7 @@ runs:
       - name: Install Python dependencies
         working-directory: ${{ inputs.dbt_project_dir }}
         shell: bash
-        run: uv pip install .
-        env:
-          UV_SYSTEM_PYTHON: 1
+        run: uv sync --frozen
 
       - name: Cache dbt requirements
         uses: actions/cache@v4
@@ -37,4 +35,4 @@ runs:
       - name: Install dbt requirements
         working-directory: ${{ inputs.dbt_project_dir }}
         shell: bash
-        run: dbt deps
+        run: uv run --frozen dbt deps
--- a/.github/actions/install_dbt_requirements/action.yaml
+++ b/.github/actions/install_dbt_requirements/action.yaml
@@ -8,31 +8,31 @@ inputs:
 runs:
   using: composite
   steps:
-      - name: Install uv
-        uses: astral-sh/setup-uv@v3
-        with:
-          enable-cache: true
-          cache-dependency-glob: |
-            ${{ inputs.dbt_project_dir }}/pyproject.toml
-            ${{ inputs.dbt_project_dir }}/uv.lock
+    - name: Install uv
+      uses: astral-sh/setup-uv@v3
+      with:
+        enable-cache: true
+        cache-dependency-glob: |
+          ${{ inputs.dbt_project_dir }}/pyproject.toml
+          ${{ inputs.dbt_project_dir }}/uv.lock
 
-      - name: Install Python
-        uses: actions/setup-python@v5
-        with:
-          python-version-file: ${{ inputs.dbt_project_dir }}/.python-version
+    - name: Install Python
+      uses: actions/setup-python@v5
+      with:
+        python-version-file: ${{ inputs.dbt_project_dir }}/.python-version
 
-      - name: Install Python dependencies
-        working-directory: ${{ inputs.dbt_project_dir }}
-        shell: bash
-        run: uv sync --frozen
+    - name: Install Python dependencies
+      working-directory: ${{ inputs.dbt_project_dir }}
+      shell: bash
+      run: uv sync --frozen
 
-      - name: Cache dbt requirements
-        uses: actions/cache@v4
-        with:
-          path: ${{ inputs.dbt_project_dir }}/dbt_packages
-          key: dbt-${{ hashFiles(format('{0}/packages.yml', inputs.dbt_project_dir)) }}
+    - name: Cache dbt requirements
+      uses: actions/cache@v4
+      with:
+        path: ${{ inputs.dbt_project_dir }}/dbt_packages
+        key: dbt-${{ hashFiles(format('{0}/packages.yml', inputs.dbt_project_dir)) }}
 
-      - name: Install dbt requirements
-        working-directory: ${{ inputs.dbt_project_dir }}
-        shell: bash
-        run: uv run --frozen dbt deps
+    - name: Install dbt requirements
+      working-directory: ${{ inputs.dbt_project_dir }}
+      shell: bash
+      run: uv run --frozen dbt deps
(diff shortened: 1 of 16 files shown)


.github/actions/install_dbt_requirements/action.yaml
```diff
@@ -8,33 +8,31 @@ inputs:
...
-      - name: Install Python dependencies
-        working-directory: ${{ inputs.dbt_project_dir }}
-        shell: bash
-        run: uv pip install .
-        env:
-          UV_SYSTEM_PYTHON: 1
+    - name: Install Python dependencies
+      working-directory: ${{ inputs.dbt_project_dir }}
+      shell: bash
+      run: uv sync --frozen
```
jeancochrane, line 27: In the workflows below, we use the equivalent [`UV_FROZEN` environment variable](https://docs.astral.sh/uv/reference/cli/#uv-run--frozen) to ensure that none of our uv commands update the lockfile without having to pass the `--frozen` flag to each command. However, there's no guarantee that this action will run in the context of a workflow that has `UV_FROZEN` set in this way, and I generally want to leave it up to the workflow to decide whether that's the right choice, so we fall back to the CLI flag in the context of this action so as to ensure these commands skip updating the lockfile without relying on any assumptions about the environment.

.github/scripts/deploy_dbt_model_dependencies.sh
```diff
@@ -43,7 +43,7 @@ echo "Parsing dbt Python model dependencies for target '$target'"
 s3_dependency_dir="s3://ccao-athena-dependencies-us-east-1"
 
 # Compile the DAG so that we have up-to-date info on dependencies
-dbt compile --quiet --target "$target"
+uv run --frozen dbt compile --quiet --target "$target"
```
jeancochrane, line 46: Same here as above: This script shouldn't rely on the assumption that `UV_FROZEN` is set in its environment, so we specify the `--frozen` flag to be safe.

.github/workflows/build_and_test_dbt.yaml
```diff
@@ -77,7 +81,13 @@ jobs:
...
             if [[ $MANUALLY_DISPATCHED == 'true' ]]; then
               echo "Running build on manually selected resources"
-              dbt build -t "$TARGET" -s ${{ inputs.models }} --defer --state "$STATE_DIR" --resource-types model seed
+              uv run \
+                dbt build \
+                --target "$TARGET" \
+                --select ${{ inputs.models }} \
+                --defer \
+                --state "$STATE_DIR" \
+                --resource-types model seed
```
jeancochrane, line 90: Two unrelated formatting changes I'm making to all of these calls for the sake of readability, since I already have to touch these lines anyway:

1. Splitting long commands onto multiple lines
2. Spelling out the full names of options (e.g. `--target` instead of `-t`)

.github/workflows/socrata_upload.yaml
```diff
@@ -95,4 +95,4 @@ jobs:
           UPLOAD_SCHEDULE: >-
             ${{ github.event.schedule == '0 11 1 * *' && 'monthly' ||
             github.event.schedule == '0 11 15 * *' && 'semimonthly' || '' }}
-        run: python ./socrata/socrata_upload.py
+        run: uv run --project socrata python3 ./socrata/socrata_upload.py
```
jeancochrane, line 98: We need to pass [the `--project` option](https://docs.astral.sh/uv/reference/cli/#uv-run--project) into this call so that uv will use the right environment. That's because we're running the command outside of the `socrata/` subdir, but we want to use the virtual environment in the `socrata/` subdir during execution.

dbt/pyproject.toml
```diff
@@ -20,9 +20,6 @@ dbt_tests = [
   "simplejson~=3.19.2",
   "watchtower>=3.4.0",
 ]
-ci_checks = [
-  "pyyaml==6.0.2",
-]
```
jeancochrane, line 25: I did a quick search to confirm that [we're not using this set of optional dependencies anywhere](https://github.com/search?q=repo%3Accao-data%2Fdata-architecture%20ci_checks&type=code). `pyyaml` is also [a dependency of `dbt-core`](https://github.com/dbt-labs/dbt-core/blob/cffe8a3136c7df10948c06014663df52226b53e3/core/setup.py#L80) and likely always will be, since YAML is a core part of dbt, so we don't need to include it explicitly anyway.

socrata/uv.lock
```diff
@@ -1,10 +1,56 @@
 version = 1
+revision = 3
```
jeancochrane, line 2: I only made two changes to this lockfile, but they produced a large diff:

1. Installed `dbt-athena-community` and `dbt-core` along with all of their dependencies, which are required to run the `socrata/socrata_upload.py` script
2. Regenerated the lockfile with a newer version of the dbt CLI than the one we first used to generate the lockfile, which added extra metadata to all the packages

Sorry for the large diff!

.github/workflows/socrata_upload.yaml
wrridgeway: This is rad, I didn't know about the project option.

wrridgeway (approved):
Thanks for handling all this Jean. I've looked through it as best I can, but I doubt I'd be able to catch anything you didn't catch yourself.

PR #1037, opened 2026-05-20
Improve land view commenting

wrridgeway:
Quicky to make this aggregation logic easier to parse. This is like 95% copilot with me editing.

### DIFF
--- a/dbt/models/default/default.vw_pin_land.sql
+++ b/dbt/models/default/default.vw_pin_land.sql
@@ -13,7 +13,11 @@
 --      more than one land line carries an influence/allocation factor AND
 --      all lines share the same sf, the repeated sf values represent the
 --      same land area measured multiple times. We take sf_top to avoid
---      double-counting.
+--      double-counting. This logic attempts to address data errors such as
+--      parid 02-02-200-010-1007 in 2025 (duplicate land lines) and
+--      non-erroneous cases such as parid 01-35-200-031-1001 in 2025 (different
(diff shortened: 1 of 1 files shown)


dbt/models/default/default.vw_pin_land.sql
```diff
@@ -1,32 +1,69 @@
...
+--
+--   1. SPLIT CLASS (infl1 = '35'): These parcels have their land value
+--      split across multiple lines, each representing a share of the same
+--      total area. We take the top line sf only (sf_top), not the sum.
+--
+--   2. MULTIPLE NON-NULL INFLU/ALLOCPCT WITH IDENTICAL sf VALUES: When
+--      more than one land line carries an influence/allocation factor AND
+--      all lines share the same sf, the repeated sf values represent the
+--      same land area measured multiple times. We take sf_top to avoid
+--      double-counting.
```
jeancochrane: **[Question, non-blocking]** Does this case usually represent a data error that we are patching, or is it expected behavior for the `iasworld.land` data model?

jeancochrane (approved):
Thanks Copilot!

For real though, these docs are great.

dbt/models/default/default.vw_pin_land.sql
wrridgeway: It can be both 😓. I just walked through two different examples with Brian for `01-35-200-031-1001` (not an error) and `02-02-200-010-1007` (an error). I'll add this to the comment.

PR #902, opened 2025-09-09
Add TY2024 exemption data

wrridgeway:
Two new views here:

### `default.vw_pin_exe`
Parcel-level exemption amounts in EAV. PINs will only have a row in this view if they have exemptions for a given year.

### `default.vw_pin_tax_roll`
Parcel-level exemption amounts in EAV, taxable AV, and mailed EAV for all current, active PINs in `iasworld.pardat` for 2021 - 2024.

The views themselves are relatively simple and straightforward, thankfully, and the rest is just documentation.

PR #1039, opened 2026-05-27
Fix exemption and HIE tests, and handle new exemptions in `default.vw_pin_exe_long`

jeancochrane:
While fulfilling an iasWorld test workbook request for Valuations, I noticed that a couple of data integrity tests were leaking into the workbook without the proper context. On further inspection, this leak revealed a few incorrect test choices:

- We added some tests to `iasworld.excode` in https://github.com/ccao-data/data-architecture/pull/902 that are defined on iasWorld sources and so run as part of the iasWorld test suite, but our iasWorld test stakeholders cannot take action on them. Since there is no resolution path for these tests, I removed them, and instead buttoned up our views that read from exemption tables (`default.vw_pin_exe` and `default.vw_pin_exe_long`) to make sure we're set up to alert ourselves if unexpected exemption codes make their way into ias.
    - As part of this change, I added filtering to `default.vw_pin_exe_long` to remove exemption codes starting with `I-` (which our data stakeholder confirmed are for incentives, not exemptions), and I added handling for the IDOR 100% disabled veteran exemption (code `DV5`). Since the IDOR 100% exemption is applied at the AV level rather than the EAV level, I left it out of `default.vw_pin_exe`, which only works for EAV exemptions. After adding this handling, I was able to add a new passing data integrity test on `default.vw_pin_exe_long` that confirms `exemption_type` is never null, which would be an indicator that there is an exemption code in `exdet` that we're not handling.
- One iasWorld test (`iasworld_asmt_all_tot30_lte_7500`) neglected to select the additional columns that are necessary for inclusion in the iasWorld test workbook, so I configured the `additional_select_columns` key on that test.

## Testing

I ran this query to make sure the PR doesn't change the data drastically:

```sql
select
    dev.pin as dev_pin,
    prod.pin as prod_pin,
    dev.year as dev_year,
    prod.year as prod_year,
    dev.exemption_type as dev_exemption_type,
    prod.exemption_type as prod_exemption_type,
    case
        when dev.pin is null then 'missing_in_dev'
        when prod.pin is null then 'missing_in_prod'
        when dev.exemption_amount != prod.exemption_amount then 'amount_mismatch'
        when dev.is_cofe != prod.is_cofe then 'is_cofe_mismatch'
        when dev.cofe_date != prod.cofe_date then 'cofe_date_mismatch'
    end AS error_type
from z_ci_jeancochrane_move_some_iasworld_tests_to_data_integrity_tests_default.vw_pin_exe_long dev
full outer join default.vw_pin_exe_long prod
    on dev.pin = prod.pin
    and dev.year = prod.year
    and (
        dev.exemption_type = prod.exemption_type
        or (dev.exemption_type = 'exe_vet_dis_100_idor' and prod.exemption_type is null)
    )
where dev.pin is null
    or prod.pin is null
    or dev.exemption_amount != prod.exemption_amount
    or dev.is_cofe != prod.is_cofe
    or dev.cofe_date != prod.cofe_date
```

This only results in one row, which is `missing_in_dev` because it has an `I-*` exemption code.


dbt/models/default/schema/default.vw_pin_exe_long.yml
```diff
@@ -7,8 +7,20 @@ models:
         description: '{{ doc("shared_column_cofe_date") }}'
       - name: exemption_amount
         description: Exemption amount
+      - name: exemption_code
```
jeancochrane, line 10: If the `default_vw_pin_exe_long_exemption_type_not_null` test ever fails, this field will be useful for debugging that failure (I used it to find the `I-*` and `DV5` excodes that we were previously neglecting to handle). Although the distinction between this column and `exemption_type` might confuse new users, I figure it's fine because this view is primarily an intermediate table used to populate `default.vw_pin_exe`, so it can handle some extraneous columns.

dbt/models/default/schema/default.vw_pin_exe_long.yml
```diff
@@ -7,8 +7,20 @@ models:
...
+          Most consumers should prefer `exemption_type`, which is cleaned and
+          validated.
+        data_tests:
+          - not_null:
+              name: default_vw_pin_exe_long_exemption_code_not_null
       - name: exemption_type
         description: Exemption type
+        data_tests:
+          - not_null:
+              name: default_vw_pin_exe_long_exemption_type_not_null
```
jeancochrane, line 23: Since we are now handling all expected exemption codes in `default.vw_pin_exe_long`, and the `CASE` statement that parses `exemption_type` has no `ELSE` fallback, this not-null test should alert us if `iasworld.exdet` ever contains unexpected exemption codes in the future.

dbt/models/iasworld/schema/iasworld.excode.yml
wrridgeway, line 1: These were some of my favorite tests. RIP.

wrridgeway (approved):
Looks good to me, thanks for taking care of this.

PR #1040, opened 2026-05-27
Remove mailing names and address associated with suppresion requests from address view

### LINKED ISSUE #1041: Hotfix for Public Act 104-0443
[Public Act 104-0443](https://www.ilga.gov/documents/legislation/PublicActs/104/104-0443.htm) requires government agencies to hide/remove private information within five business days upon receiving the request of a public official. This requests affects some internal assets as well as our Parcel Address asset on Open Data.

Initially we thought we could use ias's `hidename` column to administer this, but right now that is not the case.

Our legal team has provided a list, and advises us to hide the name and mailing address associated with the PIN. For now we'll use this list to update `vw_pin_address` and downstream internal and external assets to comply with this law.

wrridgeway:
This PR creates a new table `ccao.hidename` which is fed by a list of PINs stored on the O drive. Any PIN from this list will not have any data from `iasworld.maildat` surfaced in `default.vw_pin_address`.

```sql
select mail_address_name is null as "mailing address is null",
	count() as rows,
	'new' as source
from z_ci_hidenames_default.vw_pin_address
group by mail_address_name is null
union all
select mail_address_name is null as "mailing address is null",
	count() as rows,
	'old' as source
from default.vw_pin_address
group by mail_address_name is null
```

| mailing_address_is_null | rows     | source |
|-------------------------|----------|--------|
| false                   | 48003634 | old    |
| true                    |  2667468 | old    |
| true                    |  2668055 | new    |
| false                   | 48003047 | new    |

the difference in null rows is 587, which is exactly what we expect:

```sql
select count()
from default.vw_pin_address vpa
	inner join ccao.hidename hide on vpa.pin = hide.pin
```

gives us 587.

```sql
with pins as (
	select pin
	from ccao.hidename
)
select vpa.*
from z_ci_hidenames_default.vw_pin_address as vpa
	inner join pins on vpa.pin = pins.pin
where mail_address_name is not null
```

returns no rows.

### DIFF
--- a/dbt/models/ccao/schema.yml
+++ b/dbt/models/ccao/schema.yml
@@ -78,6 +78,11 @@ sources:
 
       - name: hidename
         description: '{{ doc("table_hidename") }}'
+        data_tests:
+          - unique_combination_of_columns:
+              name: ccao_hidename_unique_by_pin
+              combination_of_columns:
+                - pin
 
       - name: hie
         description: '{{ doc("table_hie") }}'
--- a/dbt/models/ccao/schema.yml
+++ b/dbt/models/ccao/schema.yml
@@ -76,6 +76,9 @@ sources:
           - name: version
             description: In the past, models were not assigned alphanumeric IDs but rather PIN and year-level output was consecutively "versioned" by model run.
 
+      - name: hidename
+        description: '{{ doc("table_hidename") }}'
+
       - name: hie
         description: '{{ doc("table_hie") }}'
         tags:
--- a/dbt/models/default/default.vw_pin_address.sql
+++ b/dbt/models/default/default.vw_pin_address.sql
@@ -7,18 +7,24 @@
 -- on the Treasurer's website.
 WITH mail AS (
     SELECT
-        parid,
-        taxyr,
-        {{ concat_address(['mail1', 'mail2']) }} AS mail_address_name,
-        {{ concat_address(['maddr1', 'maddr2']) }} AS mail_address_full,
-        mcityname AS mail_address_city_name,
-        mstatecode AS mail_address_state,
-        NULLIF(mzip1, '00000') AS mail_address_zipcode_1,
-        NULLIF(mzip2, '0000') AS mail_address_zipcode_2,
-        mailseq = MAX(mailseq) OVER (PARTITION BY parid, taxyr) AS newest
-    FROM {{ source('iasworld', 'maildat') }}
-    WHERE cur = 'Y'
-        AND deactivat IS NULL
+        mail.parid,
+        mail.taxyr,
+        {{ concat_address(['mail.mail1', 'mail.mail2']) }}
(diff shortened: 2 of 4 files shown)


dbt/models/ccao/schema.yml
```diff
@@ -76,6 +76,9 @@ sources:
           - name: version
             description: In the past, models were not assigned alphanumeric IDs but rather PIN and year-level output was consecutively "versioned" by model run.
 
+      - name: hidename
+        description: '{{ doc("table_hidename") }}'
```
jeancochrane, line 80: **[Suggestion, optional]** Should we add a quick data integrity test to make sure this table is always unique by PIN? It shouldn't make a huge difference but seems reasonable and easy for cleanliness purposes.


```suggestion
      - name: hidename
        description: '{{ doc("table_hidename") }}'
        data_tests:
          - unique_combination_of_columns:
              name: ccao_hidename_unique_by_pin
              combination_of_columns:
                - pin
```

dbt/models/default/default.vw_pin_address.sql
```diff
@@ -7,18 +7,24 @@
...
+        NULLIF(mail.mzip1, '00000') AS mail_address_zipcode_1,
+        NULLIF(mail.mzip2, '0000') AS mail_address_zipcode_2,
+        mail.mailseq = MAX(mail.mailseq)
+            OVER (PARTITION BY mail.parid, mail.taxyr) AS newest,
+        hide.pin AS hide_pin
+    FROM {{ source('iasworld', 'maildat') }} AS mail
+    LEFT JOIN {{ source('ccao', 'hidename') }} AS hide
+        ON mail.parid = hide.pin
+    WHERE mail.cur = 'Y'
+        AND mail.deactivat IS NULL
```
jeancochrane, line 27: **[Question, non-blocking]** Just confirming that this behavior will remove the entire row for every year for every PIN in the source list. Is that the behavior we want @ccao-jardine? Or do we want to preserve the row and replace the non-PIN/year data with indicator values, or nulls?

jeancochrane (approved):
Thanks for picking this up so quickly! I have one question about high-level behavior, but otherwise this looks fine to me.

dbt/models/default/default.vw_pin_address.sql
wrridgeway: The PIN will still have a row in the view, just NULLs for columns from `maildat`. Could totally change the NULLs to a filler value.

jeancochrane: Oops, that's right, my bad! The current behavior seems like the best option to me, then.

ccao-jardine: Thanks for this bandaid and thorough review! Confirming that I do want the PIN to have a row, replacing the contents of the `MAILDAT` columns with nulls.

PR #1043, opened 2026-06-04
Fix divide-by-zero error in `qc.vw_report_town_close_res_edit.sql`

jeancochrane:
The percent change fields in our town close Res Edit view are currently subject to divide-by-zero errors, which fail silently in Athena and produce the values `Infinity`. This was causing Tableau to render all cells in the Res Edit workbook using scientific notation, and also causedExcel to raise file corruption errors when opening the workbook.

This PR updates all percent change fields in the Res Edit view to coerce them to nulls in cases where we would otherwise divide by zero.


wrridgeway (approved):
Nice job figuring this out 😓

PR #1042, opened 2026-06-03
Ensure views are consistent with active PINs in `iasworld.pardat`

wrridgeway:
We recently found out that some iasWorld tables will start to have some deactivated parcels reactivated so that the Treasurer can use them for tax bills. Thankfully this won't be the case for `iasworld.pardat` and we can depend on that table as a consistent universe of active parcels. We need to review our views and tables in `default` and `reporting` and make sure they are based on pardat or inner joined to that table. I've created a shared excel workbook to track our progress [here](https://cookcounty.sharepoint.com/:x:/s/Data-Assessor/IQCC5uAx4XW2R6npaLptaa0hAYwk3socyCcdSSQEjXyjZok?e=LxQBuI).

There is unfortunately 1 reactivated parcel in pardat that shouldn't be there (13122300590000) causing 3 test failures for `default.vw_pin_value`. We detect 311 rows with missing values for certain columns rather than 310. This is a known issue and is theoretically being dealt with by Mirella.

I also updated tests where needed. Differences between new views:

<details>
<summary>Click to expand code</summary>

```r
library(DBI)
library(dplyr)
library(glue)
library(noctua)
library(purrr)

noctua_options(unload = TRUE)
AWS_ATHENA_CONN_NOCTUA <- dbConnect(noctua::athena(), rstudio_conn_tab = FALSE)

views <- c('vw_pin_exe_long',
'vw_pin_land',
'vw_pin_sale',
'vw_pin_tax_roll',
'vw_pin_value')

query <- "
  select sum(cast(old.pin is not null as int)) as old,
	sum(cast(new.pin is not null as int)) as new,
	sum(cast(old.pin is not null as int)) - sum(cast(new.pin is not null as int)) as change
from default.{view} as old
	full outer join z_ci_1033_ensure_views_are_consistent_with_active_pins_in_iasworldpardat_default.{view} as new on old.pin = new.pin
	and old.year = new.year
  "

diffs <- views %>%
  set_names() %>%
  map(\(view) {
    dbGetQuery(conn = AWS_ATHENA_CONN_NOCTUA, glue(query, view = view))
  }) %>%
  bind_rows(.id = 'view')

query <- "
  with pardat as (
	select parid,
		max(deactivat) as deactivat
	from iasworld.pardat
	group by parid
)
select sum(cast(par.deactivat is not null as int)) as pins_deactivated_in_pardat
from default.{view} as old
[...]

PR #1046, opened 2026-06-08
Fix failing data integrity tests

jeancochrane:
This PR fixes two data integrity tests that [started failing this week](https://github.com/ccao-data/data-architecture/actions/runs/27137666289/job/80094133759):

- We forgot to update `default_vw_pin_exe_matches_iasworld_exdet` so that the source table filters would match two recent changes we made to the view:
    - Excluding iasWorld exemptions from our exemption views when they actually correspond to incentives (https://github.com/ccao-data/data-architecture/pull/1039)
    - Filtering our exemption views to remove PINs that are no longer active in `PARDAT` (https://github.com/ccao-data/data-architecture/pull/1042)
-  More PINs have gotten caught in the test `default_vw_pin_sale_reasonable_number_of_sales_per_year` since we last tweaked the threshold in https://github.com/ccao-data/data-architecture/pull/675. We're breaching the threshold again, so we bump it once more.
    - I dug in to the failures a bit, and there appear to be at least two factors that make it hard to debug these failures:
        1. For at least some of the PINs with >6 sales in a year, the sales are valid, and simply reflect the fact that the PIN has been divided in the real world but the division is not recorded in our system in the year of sale; see for example PIN 17181020610000 in 2025.
        2. When the test starts breaching the threshold, it's very hard to filter the sales to know which ones are the newest ones that are breaching the threshold, since new sales are not added chronologically in relation to their sale date.
    - Bumping this threshold is the easiest way to resolve these errors in the short term, but in the long term, we should rethink this test to decide what we want it to accomplish. As it is currently defined, it doesn't provide much information to us, outside of the exceptional case where we modify the `default.vw_pin_sale` view and accidentally introduce a bunch of new duplicates.

I fixed the failing `model_assessment_card_meta_card_num_not_null ` test by [deleting the model runs that introduced null cards](https://github.com/ccao-data/model-res-avm/actions/runs/27167189808/job/80197333050). This is only a temporary band-aid, since the root cause of the test failures is five PINs that represent one property whose class is up for some debate right now. We need to fix the root cause of this issue at some point, but I'm considering that out of scope for now.

### DIFF
--- a/dbt/models/default/schema/default.vw_pin_exe.yml
+++ b/dbt/models/default/schema/default.vw_pin_exe.yml
@@ -52,8 +52,14 @@ models:
                 AND admn.cur = 'Y'
                 AND admn.deactivat IS NULL
                 AND admn.exstat = 'A'
+              INNER JOIN {{ source('iasworld', 'pardat') }} AS par
+                ON det.parid = par.parid
+                AND det.taxyr = par.taxyr
+                AND par.cur = 'Y'
+                AND par.deactivat IS NULL
               WHERE det.cur = 'Y'
                 AND det.deactivat IS NULL
+                AND NOT SUBSTR(det.excode, 1, 2) = 'I-'
             )
           meta:
             description: check that vw_pin_exe row count is equal to constituent iasworld tables after pivot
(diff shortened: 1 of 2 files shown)


wrridgeway (approved):
What a bunch of annoying cases. Thanks for finding the missing test that needed to be updated with a pardat join.

> outside of the exceptional case where we modify the default.vw_pin_sale view and accidentally introduce a bunch of new duplicates

"exceptional" is a bold choice to describe the chances of this happening.

PR #1050, opened 2026-06-16
Revisit condo parking space hueristics

### LINKED ISSUE #1034: Revisit condo parking space hueristics
Our [condo parking space heuristics](https://github.com/ccao-data/data-architecture/blob/4bd2d19b9f096ffbe1550440145669fd84753672/dbt/models/default/default.vw_pin_condo_char.sql#L317-L336) are pretty old and incorporate rules we've outgrown at this point. Here's an example PIN:

| pin            | year | is_parking_space | parking_space_flag_reason | oneyr_pri_board_tot |
|----------------|------|------------------|---------------------------|---------------------|
| 10361200031008 | 2026 | false            |                           |               17892 |
| 10361200031008 | 2025 | false            |                           |               17892 |
| 10361200031008 | 2024 | true             | prior value               |                 500 |
| 10361200031008 | 2023 | true             | prior value               |                 500 |
| 10361200031008 | 2022 | true             | prior value               |                 500 |
| 10361200031008 | 2021 | false            |                           |                   3 |
| 10361200031008 | 2020 | false            |                           |                   3 |
| 10361200031008 | 2019 | false            |                           |                   3 |
| 10361200031008 | 2018 | false            |                           |                   3 |
| 10361200031008 | 2017 | false            |                           |                   3 |
| 10361200031008 | 2016 | false            |                           |                   3 |
| 10361200031008 | 2015 | false            |                           |                   3 |
| 10361200031008 | 2014 | false            |                           |                   3 |
| 10361200031008 | 2013 | false            |                           |                   3 |
| 10361200031008 | 2012 | false            |                           |                   3 |
| 10361200031008 | 2011 | false            |                           |                   3 |
| 10361200031008 | 2010 | false            |                           |               13670 |
| 10361200031008 | 2009 | false            |                           |                   4 |
| 10361200031008 | 2008 | false            |                           |                   4 |
| 10361200031008 | 2007 | false            |                           |                   4 |
| 10361200031008 | 2006 | false            |                           |                   4 |
| 10361200031008 | 2005 | false            |                           |                   4 |
| 10361200031008 | 2004 | false            |                           |               13758 |
| 10361200031008 | 2003 | false            |                           |                   4 |
| 10361200031008 | 2002 | false            |                           |                   4 |
| 10361200031008 | 2001 | false            |                           |                   4 |
| 10361200031008 | 2000 | false            |                           |               11578 |
| 10361200031008 | 1999 | false            |                           |                     |

This is a product of us looking at prior-year board total av to detect parking spaces.

- First off, I'd like to check to make sure the heuristic is even working as expected. I'm not sure why this PIN isn't flagged as a parking space for 2021.
- Once we've figured out if it's working or what fixes it needs to work as intended, it'd be worth adding a good description to the [shared docs](https://github.com/ccao-data/data-architecture/blob/4bd2d19b9f096ffbe1550440145669fd84753672/dbt/models/shared_columns.md?plain=1#L1266) outlining exactly how the heuristic actually works.
- Last let's take a look at the `prior value` heuristic specifically and see how often it leads to changes like the one above. Perhaps it just needs to be fixed, but if it's causing pins to flip and flop after being fixed we should probably just kill it.

wrridgeway:
Thanks to @Damonamajor we know the prior year AV parking space heuristics were working as intended, despite [some wonky outcomes](https://github.com/ccao-data/data-architecture/issues/1034#issue-4448619997). While it's nice to know that's the case, I think we've moved past the point where we it's wise to depend on these backward-looking rules to identify parking spaces.
- We no longer mail parking space values that align with this logic. We now determine parking space value entirely as a function of the value of the livable condo units they share a 10 digit pin with.
- This was one of the only ways for us to detect possible parking spaces and common areas back around 2018, but valuations has since gone through almost all the condo buildings in the county and updated CDU columns with 'GR' where appropriate. We do still use other rules to detect parking spaces that are more durable than value-based rules.
- This method has a habit of switching back and forth between `TRUE` and `FALSE` for `is_parking_space` as AVs change over time, clearly not a desired outcome.

<details>

<summary>Check to see which condo units are no longer marked as parking spaces</summary>

```r
changes <- dbGetQuery(conn = AWS_ATHENA_CONN_NOCTUA, "
select old.*
from z_ci_1034_revisit_condo_parking_space_hueristics_default.vw_pin_condo_char as new
	full outer join default.vw_pin_condo_char as old on new.pin = old.pin
	and new.year = old.year
where new.is_parking_space != old.is_parking_space
")

table(changes$year, changes$parking_space_flag_reason)
```

| year    | declaration percent prior | value |
|---------|---------------------------|-------|
|    2000 |                       186 |   144 |
|    2001 |                       181 |   115 |
|    2002 |                       181 |   129 |
|    2003 |                       170 |   117 |
|    2004 |                       181 |   187 |
|    2005 |                       167 |   304 |
|    2006 |                       160 |   126 |
|    2007 |                       160 |    74 |
|    2008 |                       168 |    26 |
|    2009 |                       185 |   153 |
|    2010 |                       203 |   188 |
|    2011 |                       202 |   130 |
|    2012 |                       200 |   145 |
|    2013 |                       221 |   137 |
|    2014 |                       199 |   226 |
|    2015 |                       200 |   347 |
|    2016 |                       159 |   133 |
|    2017 |                       156 |    78 |
|    2018 |                       133 |    49 |
|    2019 |                       120 |    19 |
|    2020 |                        99 |   117 |
|    2021 |                        96 |    20 |
|    2022 |                        97 |   325 |
|    2023 |                        89 |   326 |
|    2024 |                        83 |   347 |
|    2025 |                        53 |    94 |
|    2026 |                        76 |    62 |

</details>

<details>

<summary>Check to see why condo units are still being identified as parking spaces in new view</summary>

```sql
select distinct parking_space_flag_reason
from z_ci_1034_revisit_condo_parking_space_hueristics_default.vw_pin_condo_char
```

| parking_space_flag_reason            |
|--------------------------------------|
|                                      |
| model predicted negative value       |
| identified by valuations as non-unit |
| unit number                          |
| cdu                                  |

</details>

### DIFF
--- a/dbt/models/default/default.vw_pin_condo_char.sql
+++ b/dbt/models/default/default.vw_pin_condo_char.sql
@@ -325,10 +325,11 @@ SELECT DISTINCT
         -- If a unit's percent of the declaration is less than half of
         -- what it would be if all units had an equal share, AV limited
         OR (
+            -- We use 50 rather than .5 since iasworld stores ownership as
+            -- percentages rather than in decimal form
             filled.tiebldgpct < (50 / filled.building_pins)
             AND vph.oneyr_pri_board_tot BETWEEN 10 AND 5000
         )
-        OR vph.oneyr_pri_board_tot BETWEEN 10 AND 1000
         OR nonlivable.flag = 'negative pred'
     )
     AND (nonlivable.flag != 'questionable' OR nonlivable.flag IS NULL),
@@ -367,6 +368,8 @@ SELECT DISTINCT
         -- suggesting this is a low-value non-livable unit like a parking space.
         WHEN
             (
+                -- We use 50 rather than .5 since iasworld stores ownership as
+                -- percentages rather than in decimal form
                 filled.tiebldgpct < (50 / filled.building_pins)
                 AND vph.oneyr_pri_board_tot BETWEEN 10 AND 5000
             )
--- a/dbt/models/default/default.vw_pin_condo_char.sql
+++ b/dbt/models/default/default.vw_pin_condo_char.sql
@@ -291,12 +291,6 @@ SELECT DISTINCT
             OR SUBSTR(filled.unitno, 1, 3) = 'GAR'
             OR filled.note = 'PARKING/STORAGE/COMMON UNIT'
             OR filled.parking_pin = TRUE
-            -- If a unit's percent of the declaration is less than half of
-            -- what it would be if all units had an equal share, AV limited
-            OR (
-                filled.tiebldgpct < (50 / filled.building_pins)
-                AND vph.oneyr_pri_board_tot BETWEEN 10 AND 5000
-            )
             OR nonlivable.flag = 'negative pred'
         )
         AND (nonlivable.flag != 'questionable' OR nonlivable.flag IS NULL)
@@ -324,12 +318,6 @@ SELECT DISTINCT
         OR filled.parking_pin = TRUE
         -- If a unit's percent of the declaration is less than half of
         -- what it would be if all units had an equal share, AV limited
-        OR (
-            -- We use 50 rather than .5 since iasworld stores ownership as
-            -- percentages rather than in decimal form
-            filled.tiebldgpct < (50 / filled.building_pins)
-            AND vph.oneyr_pri_board_tot BETWEEN 10 AND 5000
-        )
         OR nonlivable.flag = 'negative pred'
     )
     AND (nonlivable.flag != 'questionable' OR nonlivable.flag IS NULL),
@@ -362,18 +350,6 @@ SELECT DISTINCT
             AND SUBSTR(filled.unitno, 1, 2) != 'PH'
         )
         OR SUBSTR(filled.unitno, 1, 3) = 'GAR' THEN 'unit number'
-        -- Unit's share of the building declaration (tiebldgpct) is less
-        -- than half of the equal-share threshold (50 / building_pins),
-        -- and the building has a plausible AV (between 10 and 5000),
-        -- suggesting this is a low-value non-livable unit like a parking space.
-        WHEN
-            (
-                -- We use 50 rather than .5 since iasworld stores ownership as
-                -- percentages rather than in decimal form
-                filled.tiebldgpct < (50 / filled.building_pins)
-                AND vph.oneyr_pri_board_tot BETWEEN 10 AND 5000
-            )
-            THEN 'declaration percent'
     END AS parking_space_flag_reason,
     COALESCE(vph.oneyr_pri_board_tot < 10, FALSE) AS is_common_area,
     COALESCE(nonlivable.flag = 'questionable', FALSE)
--- a/dbt/models/default/default.vw_pin_condo_char.sql
+++ b/dbt/models/default/default.vw_pin_condo_char.sql
@@ -316,8 +316,6 @@ SELECT DISTINCT
         OR SUBSTR(filled.unitno, 1, 3) = 'GAR'
         OR filled.note = 'PARKING/STORAGE/COMMON UNIT'
         OR filled.parking_pin = TRUE
-        -- If a unit's percent of the declaration is less than half of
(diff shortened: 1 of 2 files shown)


dbt/models/default/default.vw_pin_condo_char.sql
```diff
@@ -325,40 +318,38 @@ SELECT DISTINCT
...
-                AND vph.oneyr_pri_board_tot BETWEEN 10 AND 5000
-            )
-            THEN 'declaration percent'
+        -- Unit received a negative predicted AV in a prior model run,
+        -- indicating it was misclassified as livable.
         WHEN
-            vph.oneyr_pri_board_tot BETWEEN 10 AND 1000
-            THEN 'prior value'
+            nonlivable.flag = 'negative pred'
+            THEN 'model predicted negative value'
```
wrridgeway, line 350: I reordered things here to respect the clearest signals the most, though it doesn't change whether a unit is determined to be a parking space. There is a case to be made that we should include all the possible reasons for detection rather than just the first.

Damonamajor:
This is outside the scope of this issue, but we have so many flag indicators. What do people think about having a single indicator which pulls together all the flags into a single binary indicator that something is flagged as abnormal?

I guess we have this as nonliveable in the condo pipeline, so this may be more relevant for the res pipeline.

Damonamajor (commented):
I think the actual code restructuring is the right move. My one thought is that we still use the prior_board_value [here](https://github.com/ccao-data/data-architecture/blob/1034-revisit-condo-parking-space-hueristics/dbt/models/default/default.vw_pin_condo_char.sql#L354). Seeing this makes the previous `between of 10 and 5000` make sense, but also raises the question of if we want to raise that threshold since we are removing this indicator.

dbt/models/default/default.vw_pin_condo_char.sql
```diff
@@ -325,40 +318,38 @@ SELECT DISTINCT
         OR filled.parking_pin = TRUE
         -- If a unit's percent of the declaration is less than half of
         -- what it would be if all units had an equal share, AV limited
```
jeancochrane: **[Question, non-blocking]** Does this comment still apply now that we've removed the logic that used to follow it?

dbt/models/shared_columns.md
```diff
@@ -1268,7 +1269,13 @@ Corner lot indicator
...
 Deeded parking/garage space or storage unit.
 
-Detected either by valuations, CDU, or unit number/proration rate heuristics.
+Determined by an order-sensitive heuristic that uses the following sources:
+1. `ccao.pin_nonlivable` for "questionable" status
+2. CDU columns
+3. Note columns and a list of parking pins provided by valuations
+4. Address unit numbers
+
+either by valuations, CDU, or unit number/proration rate heuristics.
```
jeancochrane: **[Nitpick, optional]** I think these last two lines are leftover from the edit:


```suggestion
```

jeancochrane (approved):
Makes sense, and ends up being a lot simpler than I was expecting. Thanks!

dbt/models/default/default.vw_pin_condo_char.sql
wrridgeway: Nope.

wrridgeway:
@Damonamajor i left the prior board value definition for common areas since

1. there's no other way to detect them
2. we don't actually use common areas for anything anymore, so it's not affecting anything anywhere

Damonamajor:
> @Damonamajor i left the prior board value definition for common areas since
> 
> 1. there's no other way to detect them
> 2. we don't actually use common areas for anything anymore, so it's not affecting anything anywhere

Looks good to me then. If we don't use them anymore, do we want to drop it from the code?

wrridgeway:
I'm gonna keep them just because they crop up occasionally in conversations with valuations.

PR #1065, opened 2026-07-08
Increase open data refresh cadence

wrridgeway:
This PR moves all open data assets that were previously on a monthly update cadence to a semimonthly cadence, per an offline conversation. I will update the metadata for the open data assets once this PR is merged.


jeancochrane (approved):
Looks great to me, though let's make sure @ccao-jardine is on board before we merge.

ccao-jardine (approved):
Let's roll and see how it goes!

PR #1067, opened 2026-07-08
Add new fields to PIN-level data in IC reference files

jeancochrane:
I received a request from the IC team to add two new fields `aprval.ecf` and `aprval.exmppct` to the PIN-level reference file output (`AllNonRes_PINLevel.Query.xlsx`). This PR tweaks the underlying view that produces that output (`qc.vw_ic_reference_all_non_res_pin_level_data`) along with its template in order to produce output with the new columns that exactly matches an iasWorld Inquire query.

I tested this change by finding a PIN that has values for both columns and generating output to ensure it exactly matches what the new `AllNonRes_PINLevelData` Inquire query produces:

```
python3 scripts/export_models.py --select qc.vw_ic_reference_all_non_res_pin_level_data --where "taxyr = '2026' and parid = '07272120200000'"
```


dbt/export/templates/qc.vw_ic_reference_all_non_res_pin_level_data.xlsx
jeancochrane, line 1: The only changes to this template file are adding the two new columns.

dbt/models/qc/schema.yml
```diff
@@ -307,6 +307,12 @@ models:
           - name: DWELVAL
             index: K
             data_type: int
+          - name: ECF
+            index: L
+            data_type: int
+          - name: Exemption
```
wrridgeway, line 313: I don't know if it's worth naming this in some way to indicate the `pct` or that it can only have values between 0-100.

wrridgeway (approved):
Command works for me and output looks good.

dbt/models/qc/schema.yml
jeancochrane: In general that would be a good idea, but my goal with the format of this file is to exactly replicate the Inquire query output, and Inquire uses the confusing name "Exemption" to label this column.

PR #1070, opened 2026-07-09
Implement res land qc view

### LINKED ISSUE #1051: Implement res land QC view
## Background

Valuations would like help putting together a dashboard to make their process for QCing land rates more efficient and more reproducible. They shared their QC process over email; message me if you would like to see the raw email and I can forward it along to you.

Their current process involves running an Inquire query in iasWorld to produce a big spreadsheet containing information about all land lines in a township. Then, they run through a series of checks to make sure land attributes are sensible. They do this when a town opens for editing, and again when the town is ready to close and mail.

Our goal is to produce a Tableau dashboard that will reproduce the spreadsheet that they currently query by hand. In this dashboard, we will also add a column flagging each row for the data errors and checks in their QC checklist. With this dashboard, they will be able to load the query whenever they want, with automated flags pointing them to the rows that are most likely to need review and correction.

## Requirements

- Create a new dbt model `qc.vw_report_res_land`, with a corresponding entry in `qc/schema.yml`
    - View should be based on the Inquire query `Current vs Prior BOR Values – Change QC -Land`; if you don't have access to this in ias, I will work on sharing it with you
        - Try your best to exactly reproduce the logic from the Inquire query, including joins and filters on columns that are not selected into the view. Let me know if you've never worked with an Inquire query before and we can do a tutorial before you start
    - View should also filter for only parcel classes between 100 and 299, which the Inquire query does not currently do
    - See [View columns](#view-columns) for a list of the columns that should be in this view
- Work with Data Analytics to pull data from `qc.vw_report_res_land` into a Tableau workbook called "Res Land QC Reports"
    - We are tracking this work in https://github.com/ccao-data/reporting/issues/22 
    - Workbook should be configured to refresh once per day at 6am 
    - Workbook should provide a "Select Township" filter
    - Workbook header should include a couple notes:
        - Data refreshes every weekday at 6AM.
        - **❌ Errors** represent situations that we know for sure are impossible, and need to be fixed. These are the highest priority fixes, and they require the least amount of capacity to fix, since there is very little ambiguity that requires investigation to resolve.
        - **🔍 Checks** represent situations that seem fishy, but that we aren't sure need to be fixed. Checks require more capacity to resolve than errors, because there is ambiguity that requires investigation.

### View columns

| Athena column name | Tableau column name | Notes |
| --- | --- | --- |
| `pin` | PIN | |
| `tests` | Tests | Comma-separated string of all tests that are failing for this row; compute using `CASE`/`WHEN` statement based on [Tests](#tests) detailed below |
| `tieback` | Parcel Tieback | |
| `township_code` | Township | |
| `class` | Current Class | |
| `class_prev` | Prior Class | |
| `owner_name` | Owner | |
| `board_land_av_prev` | Prior BoR Land AV | |
| `board_bldg_av_prev` | Prior BoR Bldg AV | |
| `board_tot_av_prev` | Prior BoR Total AV | |
| `asmt_all_wen` | WEN (ASMT) | |
| `asmt_all_wencalc` | WENCALC (ASMT) | |
| `asmt_all_who` | WHO (ASMT) | |
| `asmt_all_whocalc` | WHOCALC (ASMT) | |
| `land_av` | Current Land AV | |
| `bldg_av` | Current Bldg AV | |
| `tot_av` | Current Total AV | |
| `hie_incentive_amount` | HIE/Incentive Amount | |
| `hie_incentive_year` | HIE/Incentive Amount | |
| `land_av_diff` | Land AV Difference | Subtract `board_land_av_prev` from `land_av` |
| `land_av_pct_diff` | Land AV % Difference | Subtract `board_land_av_prev` from `land_av`, divide by `board_land_av_prev` (handling divide-by-zero errors) | 
| `land_av_pct_diff_sd` | Land AV % Difference Std Dev | Compute at the township level. _TK: How to compute this in an efficient way?_ | 
| `tot_av_diff` | Total AV Difference | Subtract `board_tot_av_prev` from `tot_av` |
| `reascd` | REASCD | |
| `nbhd_code` | Neighborhood Number | |
| `taxdist` | Tax District | |
| `aprval_who` | WHO (APRVAL) | | 
| `aprval_wen` | WEN (APRVAL) | | 
| `aprval_wencalc` | WENCALC (APRVAL) | | 
| `aprval_whocalc` | WHOCALC (APRVAL) | | 
| `spcflg` | Special Processing Flag | |
| `lline` | Land Line | |
| `ltype` | Land Type | |
| `code` | Land Code | |
| `land_class` | Land Class | |
| `land_sf` | Land SF | |
| `land_sf_sd` | Land SF Std Dev | Compute at the township level. _TK: How to compute this in an efficient way?_ |
| `land_brate` | Land Base Rate | |
| `nbhd_land_rate` | Neighborhood Land Rate | Source from `ccao.land_nbhd_rate`, joined by `nbhd_code` |

## Tests

Apply the filters from Errors and Checks below to run these tests on each PIN in the view. Each PIN should have an array in the `tests` column containing all of the failing tests that apply to that PIN.

### ❌ Errors

Errors represent situations that we know for sure are impossible, and need to be fixed. These are the highest priority fixes, and they require the least amount of capacity to fix, since there is very little ambiguity that requires investigation to resolve.

When formatting an error label, prefix it with the string "❌ **Error:**"

- **Farmland (class 239 or 224) must have land type A (Acre)**
    - Filter: `class IN ('239, '224') AND land_type != 'A'` 
- **Farmland (class 239 or 224) must have land code 900**
    - Filter: `class IN ('239, '224') AND land_code != '900'` 
- **Land type A (Acre) must be farmland (class 239 or 224)**
    - Filter: `land_type = 'A' AND class NOT IN ('239, '224')` 
- **Land code 900 must be farmland (class 239 or 224)**
    - Filter: `land_code = '900' AND class NOT IN ('239, '224')` 
- **Land type G (Exempt) must have land class EX or RR**
    - Filter: `land_type = 'G' AND class NOT IN ('EX', 'RR')`  
- **Land code 0 should not exist**
    - Filter: `land_code = '0'` 
- **Land code 21 (Common Area) must have total AV <= 1**
    - Filter: `land_code = '21' AND land_av > 1` 
- **Land code 56 (Unbuildable) must not be common area (total AV <= 1)**
    - Filter: `land_code = '56' AND land_av <= 1` 
- **Land class must be 100, 200, 239, 241, or 500**
    - Filter: `land_class NOT IN ('100', '200', '239', '241', '500')`
- ~**Land rate must match neighborhood land rate**~
    - ~Filter: `land_brate != nbhd_land_rate`~
    - Turns out this check is complicated and might not be possible to perform in an automated fashion. See [Open questions](#open-questions).

### 🔍 Checks

Checks represent situations that seem fishy, but that we aren't sure need to be fixed. Checks require more capacity to resolve than errors, because there is ambiguity that requires investigation.

When formatting a check label, prefix it with the string "🔍 **Check:**"

- **Review class 200 homestead land on farmland PINs (class 224 or 239)**
    - Filter: `class IN ('239, '224') AND land_class = '200'` 
- **Send industrial land (class 500) on farmland PINs (class 224 or 239) to Manager of Industrial**
    - Filter: `class IN ('239, '224') AND land_class = '500'` 
- **Review land codes 500, 600, and EX**
    - Filter: `land_code IN ('500', '600', 'EX')`
- **Review big increases in land AV**
    - Filter: `land_av_pct_diff_sd > 1`
    - Open question: Is SD the best way to define "big increases"? Or is there a simpler/more intuitive approach?
- **Review values for PINs with largest land in the township**
    - Filter: `land_sf_sd > 1`
    - Open question: Is SD the best way to define "largest land"? Or is there a simpler/more intuitive approach? 
- **Review land value that is flat or decreasing**
    - Filter: `(land_av - board_land_av_prev) <= 0`

### Open questions

- How can we define/compute "big increases" and "largest land" in the most efficient and intuitive fashion?
- What's the most efficient way to compute the "Tests" column value, given that we want to run all checks on all rows in the input data?
    - Reference PR that implements a similar system: https://github.com/ccao-data/data-architecture/pull/1057 
- Question for Res Val: Is there an automated way to check whether the land rate is correct? Perhaps we could make sure neighborhood codes match the township, and/or land codes match some other piece of info about the PIN? 
- Question for Res Val: Does the land reviewer need look at some other attributes of the PIN to confirm that a land line is underwater (land code 26)? Or is this note just provided for context, and we don't actually need to worry about QCing land lines with code 26?
- Question for Res Val: Will farmland parcels with industrial on them show up as PARDAT class 224/239 and LAND class 500? Or is there some other indication of a farmland PIN having industrial?
- Question for Res Val: How should a reviewer make sure 1 acre looks to be 1 acre? By viewing the PIN in CookViewer/Nearmap, or some other way?
- Question for Res-Val: Do we want to calculate biggest decreases and smallest lands in the township?

Damonamajor:
Output can be found here:
`select * from z_dev_damajor_qc.vw_report_res_land`

Counts can be found here:

```
SELECT
    test,
    taxyr,
    COUNT(*) AS pin_count
FROM z_dev_damajor_qc.vw_report_res_land
CROSS JOIN UNNEST(tests) AS t(test)
GROUP BY test, taxyr
ORDER BY test, taxyr
```

High level takeaway

- There's some discrepencies betwen ccao.land_nbhd_rate and iasworld.land which we may want to investigate more. I presume that they change the values after we run our etl script.


dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,213 @@
+WITH base AS (
+    SELECT
+        aprval.reascd AS reascd,
```
Damonamajor: Ordered by table here so that it's easier to look through the iasworld query.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,213 @@
...
+        aprval.revcode AS revcode,
```
Damonamajor: `revcode` was in the inquire query, but not in the documented columns

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        aprval.whocalc AS aprval_whocalc,
+        aprval.spcflg,
+        asmt.valasm1 AS land_av,
+        asmt.valasm2 AS bldg_av,
+        asmt.valasm3 AS tot_av,
+        asmt.tot30 AS hie_incentive_amount,
+        asmt_prev.wen AS asmt_all_wen,
+        asmt_prev.wencalc AS asmt_all_wencalc,
+        asmt_prev.who AS asmt_all_who,
+        asmt_prev.whocalc AS asmt_all_whocalc,
```
Damonamajor: Do we want to name these with the same prefix? Asmt_all vs asmt_prev

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+                *,
+                ROW_NUMBER()
+                    OVER (PARTITION BY parid, taxyr ORDER BY parid)
+                    AS _rn
+            FROM {{ source('iasworld', 'asmt_all') }}
+            WHERE cur = 'Y'
+                AND valclass IS NULL
+        )
+        WHERE _rn = 1
+    ) AS asmt
```
Damonamajor, line 70: Billy said asmt and asmt_all have duplication due to errors in encoding. I removed the rn= 1 and it created errors

Image

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+    LEFT JOIN (
+        SELECT *
+        FROM (
+            SELECT
+                *,
+                ROW_NUMBER()
+                    OVER (PARTITION BY parid, taxyr ORDER BY parid)
+                    AS _rn
+            FROM {{ source('iasworld', 'asmt_all') }}
+            WHERE cur = 'Y'
```
Damonamajor: cur = Y and valclass = NULL seems to be how they identify the most recent valuation stage.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+                CASE WHEN ABS(land_av_pct_diff - land_av_pct_diff_mean)
+                        > 2 * land_av_pct_diff_sd
+                        THEN '🔍 Check: Review big increases in land AV'
+                END,
+                CASE WHEN ABS(land_sf - land_sf_mean) > 2 * land_sf_sd
+                        THEN '🔍 Check: Review values for PINs with'
+                        || ' largest land in the township'
+                END,
+                CASE WHEN land_av_diff < 0
+                        THEN '🔍 Check: Review land value that is decreasing'
```
Damonamajor: seems pretty normal that land would stay the same, especially since it may be off tri, so I modified it to decreasing.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        land_sf,
+        land_sf_mean,
+        land_sf_sd,
+        land_brate
+    FROM base_with_sd
+)
+
+SELECT *
+FROM base_with_tests
+WHERE CARDINALITY(tests) > 0
```
Damonamajor: Only select observations which have at least one flag / error.

dbt/models/qc/schema.yml
```diff
@@ -1049,78 +1043,32 @@ models:
...
+    data_tests:
+      - unique_combination_of_columns:
+          name: qc_vw_report_res_land_unique_by_pin_lline_and_taxyr
+          combination_of_columns:
+            - pin
+            - lline
+            - taxyr
+          meta:
+            description: res land report should be unique by pin, lline, and taxyr
+    columns:
```
Damonamajor: Don't know if this is needed, (probably not), just saw it in another test.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        asmt_prev.wen AS asmt_all_wen,
+        asmt_prev.wencalc AS asmt_all_wencalc,
+        asmt_prev.who AS asmt_all_who,
+        asmt_prev.whocalc AS asmt_all_whocalc,
+        asmt_prev.valasm1 AS board_land_av_prev,
+        asmt_prev.valasm2 AS board_bldg_av_prev,
+        asmt_prev.valasm3 AS board_tot_av_prev,
+        asmt.valasm1 - asmt_prev.valasm1 AS land_av_diff,
+        CASE
+            WHEN asmt_prev.valasm1 = 0 OR asmt_prev.valasm1 IS NULL THEN NULL
```
Damonamajor: Can't divide by 0 so make NULL

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+                    OVER (PARTITION BY parid, taxyr ORDER BY parid)
+                    AS _rn
+            FROM {{ source('iasworld', 'asmt_all') }}
+            WHERE cur = 'Y'
+                AND valclass IS NULL
+        )
+        WHERE _rn = 1
+    ) AS asmt_prev
+        ON pardat.parid = asmt_prev.parid
+        AND CAST(pardat.taxyr AS INT) = CAST(asmt_prev.taxyr AS INT) + 1
```
Damonamajor, line 114: Join previous year asmt_all to current year of pardat

dbt/models/qc/qc.vw_report_res_land.sql
jeancochrane: **[Suggestion, optional]** I would vote for `asmt_prev`, so that we can preserve the indication that these fields refer to the prior-year values.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        asmt_prev.who AS asmt_all_who,
+        asmt_prev.whocalc AS asmt_all_whocalc,
+        asmt_prev.valasm1 AS board_land_av_prev,
+        asmt_prev.valasm2 AS board_bldg_av_prev,
+        asmt_prev.valasm3 AS board_tot_av_prev,
+        asmt.valasm1 - asmt_prev.valasm1 AS land_av_diff,
+        CASE
+            WHEN asmt_prev.valasm1 = 0 OR asmt_prev.valasm1 IS NULL THEN NULL
+            ELSE (asmt.valasm1 - asmt_prev.valasm1) / asmt_prev.valasm1
+        END AS land_av_pct_diff,
```
jeancochrane: **[Suggestion, optional]** We can simplify this conditional using [`NULLIF()`](https://docs.aws.amazon.com/redshift/latest/dg/r_NULLIF_function.html): 

```suggestion
        (asmt.valasm1 - asmt_prev.valasm1) /
            NULLIF(asmt_prev.valasm1, 0) AS land_av_pct_diff,
```

Note that we don't need to explicitly handle the case where `asmt_prev.valasm1 IS NULL`, because division by null will result in null value.

This expression is functionally equivalent to yours, so I leave it up to you to decide which one you like better.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        pardat.parid AS pin,
+        pardat.taxyr,
+        pardat.tieback,
+        pardat.class,
+        pardat.nbhd AS nbhd_code,
+        pardat_prev.class AS class_prev
+    FROM {{ source('iasworld', 'pardat') }} AS pardat
+    LEFT JOIN {{ source('iasworld', 'legdat') }} AS legdat
+        ON pardat.parid = legdat.parid
+        AND pardat.taxyr = legdat.taxyr
```
jeancochrane, line 45: **[Nitpick, required]** We should make sure to filter for `cur` whenever we join to a new table:


```suggestion
        ON pardat.parid = legdat.parid
        AND pardat.taxyr = legdat.taxyr
        AND legdat.cur = 'Y'
```

dbt/models/qc/qc.vw_report_res_land.sql
jeancochrane: **[Suggestion, optional]** Hm, I'm not sure I follow @wrridgeway's logic here. We don't have direct access to `ASMT`, but we do have access to `ASMT_ALL` and the history table `ASMT_HIST`, so our typical practice for querying from `ASMT` is to use the set difference of `ASMT_ALL` and `ASMT_HIST`. Here's an example how how we do that in another QC view:

https://github.com/ccao-data/data-architecture/blob/a2b93c20a4b40936c252073be5db41117eb5af9b/dbt/models/qc/qc.vw_neg_asmt_value.sql#L78-L87

I might be misunderstanding the duplicate/encoding issue you mention though, so we can discuss further if I'm speaking to the wrong problem.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        SELECT *
+        FROM (
+            SELECT
+                *,
+                ROW_NUMBER()
+                    OVER (PARTITION BY parid, taxyr ORDER BY parid)
+                    AS _rn
+            FROM {{ source('iasworld', 'aprval') }}
+        )
+        WHERE _rn = 1
```
jeancochrane, line 69: **[Question, blocking]** Is the goal with this ranking operation to pull the most recent year from `aprval`? I don't understand why would do this instead of just joining `aprval` to `asmt` by `taxyr`.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+    ) AS aprval
+        ON pardat.parid = aprval.parid
+        AND pardat.taxyr = aprval.taxyr
+    LEFT JOIN {{ source('iasworld', 'land') }} AS land
+        ON pardat.parid = land.parid
+        AND pardat.taxyr = land.taxyr
+    LEFT JOIN {{ source('iasworld', 'owndat') }} AS owndat
+        ON pardat.parid = owndat.parid
+        AND pardat.taxyr = owndat.taxyr
```
jeancochrane, line 83: **[Nitpick, required]** Same here as above, we should filter all of these tables by `cur = 'Y'` when we join to them.

dbt/models/qc/qc.vw_report_res_land.sql
jeancochrane: **[Nitpick, required]** Same as above, I think we can eliminate the ranking subquery on `asmt_all`, which will help us simplify this join (i.e. we can just join directly to `asmt_all` by `parid` and `taxyr`). I also think we need to filter `asmt_prev` by `rolltype = 'RP'` in addition to `cur` and `valclass`.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+                *,
+                ROW_NUMBER()
+                    OVER (PARTITION BY parid, taxyr ORDER BY parid)
+                    AS _rn
+            FROM {{ source('iasworld', 'pardat') }}
+        )
+        WHERE _rn = 1
+    ) AS pardat_prev
+        ON pardat.parid = pardat_prev.parid
+        AND CAST(pardat.taxyr AS INT) = CAST(pardat_prev.taxyr AS INT) + 1
```
jeancochrane: **[Nitpick, required]** Here again, I think we can eliminate the ranking subquery, and we should make sure to filter by `pardat_prev.cur = 'Y'`. Though per my comment above, I think we can get rid of this join and replace `pardat_prev.class` with `asmt_prev.class`.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+    WHERE pardat.cur = 'Y'
+        AND pardat.deactivat IS NULL
+        AND TRY_CAST(pardat.class AS INT) BETWEEN 100 AND 299
```
jeancochrane: **[Nitpick, optional]** Slightly more robust, since some classes (all non-res, I believe) contain alphabetic characters:


```suggestion
        AND SUBSTR(pardat.class, 1, 1) IN ('1', '2')
```

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+                    OVER (PARTITION BY parid, taxyr ORDER BY parid)
+                    AS _rn
+            FROM {{ source('iasworld', 'asmt_all') }}
+            WHERE cur = 'Y'
+                AND valclass IS NULL
+        )
+        WHERE _rn = 1
+    ) AS asmt
+        ON pardat.parid = asmt.parid
+        AND pardat.taxyr = asmt.taxyr
```
jeancochrane, line 72: **[Nitpick, required]** We're missing a couple filters here compared to the Inquire query:

```suggestion
        ON pardat.parid = asmt.parid
        AND pardat.taxyr = asmt.taxyr
        AND asmt.rolltype = 'RP'
        AND asmt.cur = 'Y'
        AND asmt.valclass IS NULL
```

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        aprval.revcode,
+        aprval.who AS aprval_who,
+        aprval.wen AS aprval_wen,
+        aprval.wencalc AS aprval_wencalc,
+        aprval.whocalc AS aprval_whocalc,
+        aprval.spcflg,
+        asmt.valasm1 AS land_av,
+        asmt.valasm2 AS bldg_av,
+        asmt.valasm3 AS tot_av,
+        asmt.tot30 AS hie_incentive_amount,
```
jeancochrane: **[Question, blocking]** The Inquire query seems to use `val30` here, is there a reason to use `tot30` instead?

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        legdat.user1 AS township_code,
+        legdat.taxdist,
+        oby.user10 AS hie_incentive_year,
+        owndat.own1 AS owner_name,
+        pardat.parid AS pin,
+        pardat.taxyr,
+        pardat.tieback,
+        pardat.class,
+        pardat.nbhd AS nbhd_code,
+        pardat_prev.class AS class_prev
```
jeancochrane: **[Nitpick, optional]** Based on the Inquire query, I think the prior year class is actually coming from `asmt_prev` (which is displayed as `ASMT_ALL` in the Inquire interface):

Image

I don't have an intuition regarding whether this is an important distinction, but my instinct would be to prefer whatever field they're using for the current version of the query.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        ON pardat.parid = land.parid
+        AND pardat.taxyr = land.taxyr
+    LEFT JOIN {{ source('iasworld', 'owndat') }} AS owndat
+        ON pardat.parid = owndat.parid
+        AND pardat.taxyr = owndat.taxyr
+    LEFT JOIN (
+        SELECT
+            parid,
+            taxyr,
+            ARRAY_JOIN(ARRAY_AGG(DISTINCT user10), ', ') AS user10
```
jeancochrane: **[Question, blocking]** I'm curious to get @wrridgeway's take on whether this is the right way to join `land` to `oby` in order to retrieve the year (`user10`) for an HIE recorded in `asmt`. (The Inquire query that this view is based on obfuscates joins, so we don't have access to the source join conditions.)

Two specific questions:

1. Since our view output is intended to be unique by `parid`, `taxyr`, and `land.lline`, should we include `oby.lline` in our joins and our grouping?
2. How do we handle the fact that `card` is also part of the `oby` primary key? Should we aggregate `user10` within the group `parid, taxyr, lline`? Or is it more appropriate to just take the largest (latest) value of `user10` within the group?

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,248 @@
...
+        land.code,
+        land.class AS land_class,
+        land.sf AS land_sf,
+        land.brate AS land_brate,
+        legdat.user1 AS township_code,
+        legdat.taxdist,
+        oby.user10 AS hie_incentive_year,
+        owndat.own1 AS owner_name,
+        pardat.parid AS pin,
+        pardat.taxyr,
```
jeancochrane: **[Nitpick, optional]** If we're going to use the name `pin` for the PIN column, I think we should use `year` for the tax year column, which is the typical way we name these columns in our views.

dbt/models/qc/qc.vw_report_res_land.sql
jeancochrane: **[Nitpick, optional]** I do think we should include flat values in this check, since our stakeholder included them as part of spec:

> Add a column of % change in land value.  [...] review anything with a zero-percentage change

If you feel strongly about removing it, I'm also happy to add it to our list of open questions so we can clear it with our stakeholder.

dbt/models/qc/qc.vw_report_res_land.sql
jeancochrane: **[Nitpick, optional]** Since our stakeholders have at least one check that we can't yet formally encapsulate in code (land rates), I think we should remove this filter.

dbt/models/qc/schema.yml
jeancochrane: **[Praise]** Seems like a good idea to me!

dbt/models/qc/schema.yml
```diff
@@ -1049,78 +1043,32 @@ models:
...
+      - name: pin
+        data_tests:
+          - not_null:
+              name: qc_vw_report_res_land_pin_not_null
```
jeancochrane: **[Nitpick, optional]** I don't personally think we need this test, but I don't feel strongly about it. I can't think of a situation where we would accidentally modify this view to null out PINs.

dbt/models/qc/schema.yml
```diff
@@ -1049,78 +1043,32 @@ models:
...
     description: '{{ doc("view_vw_pin_sale_high_number_of_sales_per_year") }}'
 
   - name: qc.vw_pin_value_null_values
     description: '{{ doc("view_vw_pin_value_null_values") }}'
 
+  - name: qc.vw_report_res_land
+    description: '{{ doc("view_vw_report_res_land") }}'
+    config:
+      tags:
+        - qc_report_res_land
```
jeancochrane: **[Nitpick, optional]** Since this QC report only includes one view, I don't think we actually need this tag. Tags help us group views together, but this view doesn't yet need to be grouped with anything else.

```suggestion
```

jeancochrane (changes requested):
Very nice work here! I left a bunch of comments below, but most of it corresponds to small issues that we can fix easily. I think this is pretty close to a state where we can stand up a simple Tableau view and show it to our stakeholders.

Tagging in @wrridgeway in to answer a few specific data model questions; don't feel like you need to review the entire issue and PR, though!

dbt/models/qc/qc.vw_report_res_land.sql
Damonamajor: I do feel semi-strongly in the sense that I'm not sure land values are updated in non-assessment years.

jeancochrane: I suspect our stakeholders may only perform this QC on towns that are part of the current reassessment tri, though that's just a guess.

dbt/models/qc/qc.vw_report_res_land.sql
Damonamajor: NVM misread

dbt/models/qc/qc.vw_report_res_land.sql
Damonamajor: The actual code wasn't Billys suggestion so I"ll try to implement this.

dbt/models/qc/qc.vw_report_res_land.sql
Damonamajor: image
Not pointing fingers, but we may want to make sure we have our understandings correct.

Damonamajor: They seem pretty similar
image

But there are some differences
```
select taxyr, tot30, val30
from iasworld.asmt_all
where tot30 != val30
limit 100
```

image

Damonamajor: It's also what we have in our docs
val30 doesn't have any description

Image

Damonamajor:
@jeancochrane @wrridgeway 
I made most of the changes which didn't require a discussion. Do the three of us want to talk through the remaining changes today. I also removed the deduplication so we can see that it fails the unique pin-year-lline test.

Damonamajor:
subquery on OBY to retreive pin - year - lline and the minimum HIE year within that group. Then join that to land by pin - year - lline. That would be getting the earliest HIE year which is available for that LLINE.

Damonamajor:
We also decided to revert to val30, which is what we use for our exemption values, and de-dupe asmt tables based on the most recent wen.

dbt/models/qc/qc.vw_report_res_land.sql
```diff
@@ -0,0 +1,232 @@
...
+        asmt_prev.class AS class_prev
+    FROM {{ source('iasworld', 'pardat') }} AS pardat
+    LEFT JOIN {{ source('iasworld', 'legdat') }} AS legdat
+        ON pardat.parid = legdat.parid
+        AND pardat.taxyr = legdat.taxyr
+        AND legdat.cur = 'Y'
+    INNER JOIN (
+        SELECT *
+        FROM (
+            SELECT
```
jeancochrane, line 52: **[Suggestion, optional]** Quick comment to remind ourselves why we did it this way:


```suggestion
            -- There are some dupes in asmt_all that we can't entirely account
            -- for, so we take the most recent row for each PIN/year
            SELECT
```

dbt/models/iasworld/schema/iasworld.oby.yml
```diff
@@ -448,6 +448,11 @@ sources:
             description: Number of identical units
           - name: upd_status
             description: '{{ doc("column_upd_status") }}'
+          - name: user10
+            description: Year in which the home improvement exemption (HIE) was
+              granted.
+          - name: user14
+            description: Year in which the home improvement exemption (HIE) expires.
```
jeancochrane: **[Nitpick, optional]** I would go even simpler here, to make the docs table easier to skim:

```suggestion
            description: Home improvement exemption (HIE) start year
          - name: user14
            description: Home improvement exemption (HIE) end year
```

jeancochrane (approved):
Awesome work! I'm very excited to demo this for our stakeholders.

Damonamajor:
Aside from the blocking github issues there are a couple updates which I uncovered while using tableau. 

The standard deviation was switched to Z-score. Zscore is just the standard deviations away from the norm the value is (standard deviation is fixed per grouping).

The % change was rounded from .46 -> 0,  .54 -> 1 , 1.54 -> 2 etc. Now it shows the actual percent value.

Once the github issues are resolved this should be good to merge.

@jeancochrane

jeancochrane (approved):
Good to go!

Damonamajor:
I'm merging this in. I think that it's going to break the actual tableau dashboard since we are using the dev view for it. I'm not sure I have permission to update it, but I"m looking into it. And, we should make sure that we pull data from the most recent year not 2026 for that table so it won't need to be updated in the future.

Damonamajor:
And I think this is ready to send along, but I think I have 3 questions for our stakeholder as they work through it.

- Can they provide the order of the tableau columns that they would like to use?
- For the % Increase, we use a z-score, but that is skewed by very some very high outlers $2 of land value -> $16,000 or things along these lines. Is there a % increase which he views as a cutoff. This was present in some townships, but not others. Maybe all parcels which had an AV increase of 10x, which seemed decently common.
- For large sf properties, there are some which are prorated where we could divide the total sf by the proration rate. For example PIN `18071090381089` (new in tax year 2026).

Damonamajor:
Just messaged these questions in teams.

PR #1071, opened 2026-07-13
Add new view `model.vw_model_run_type` to support junk model runs

### LINKED ISSUE #1066: Add new view for model.vw_metadata
Companion issue to https://github.com/ccao-data/model-res-avm/issues/522.

This will union model.metadata and z_dev_model.metadata so that we have a composite view for both `junk` and `non-junk` runs.

Damonamajor:
This adds a necessary view to have model runs for junk runs which are stored in z_dev_model.metadata and all other runs stored in model.metadata. We use just the three necessary columns since we are not concerned about divergences between metadata columns which may come as we add new columns based on our appending structure in Athena.

It also adds a shared column description for run_type.

I also had a previous test:
```
        description: '{{ doc("table_z_dev_model_metadata") }}'
        data_tests:
              name: z_dev_model_metadata_unique_year_run
              combination_of_columns:
                - year
                - run_id
```
which had the following error which I think may be related to permissions for dev tables but I wasn't sure
https://github.com/ccao-data/data-architecture/actions/runs/29341355279/job/87113632811

### DIFF
--- a/dbt/models/model/docs.md
+++ b/dbt/models/model/docs.md
@@ -218,6 +218,15 @@ you want to evaluate training set performance for any model run.
 **Primary Key**: `run_id`, `meta_card_num`, `meta_sale_document_num`
 {% enddocs %}
 
+# vw_metadata
+
+{% docs view_vw_metadata %}
+View combining metadata from both the production `model` database and the
+development `z_dev_model` database.
+
+**Primary Key**: `year`, `run_id`
+{% enddocs %}
+
 # vw_card_res_input
 
 {% docs view_vw_card_res_input %}
--- a/dbt/models/model/docs.md
+++ b/dbt/models/model/docs.md
@@ -79,14 +79,6 @@ hashes of the model code, notes, etc.
 **Primary Key**: `year`, `run_id`
 {% enddocs %}
 
-# z_dev_model.metadata
-
-{% docs table_z_dev_model_metadata %}
-Developmental counterpart to `model.metadata`. Used for junk runs that are not intended for stored indefinetely. Contents are deleted after 30 days.
-
-**Primary Key**: `year`, `run_id`
-{% enddocs %}
-
 # parameter_final
 
 {% docs table_parameter_final %}
@@ -274,4 +266,12 @@ View to compile PIN-level model inputs shared between the residential
 (`model.vw_card_res_input`) and condo (`model.vw_pin_condo_input`) model views.
 
 **Primary Key**: `year`, `meta_pin`
+{% enddocs %}
+
+# z_dev_model.metadata
+
+{% docs table_z_dev_model_metadata %}
+Developmental counterpart to `model.metadata`. Used for junk runs that are not intended for stored indefinetely. Contents are deleted after 30 days.
+
+**Primary Key**: `year`, `run_id`
 {% enddocs %}
--- a/dbt/models/model/docs.md
+++ b/dbt/models/model/docs.md
@@ -243,7 +243,7 @@ for more information.
 View combining metadata from both the production `model` database and the
 development `z_dev_model` database.
 
-**Primary Key**: `year`, `run_id`
+**Primary Key**: `run_id`
 {% enddocs %}
 
 # vw_pin_condo_input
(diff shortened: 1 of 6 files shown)


dbt/models/model/docs.md
```diff
@@ -237,6 +237,15 @@ for more information.
 **Primary Key**: `year`, `meta_pin`, `meta_card_num`
 {% enddocs %}
 
+# vw_model_run_type
+
+{% docs view_vw_model_run_type %}
+View combining metadata from both the production `model` database and the
+development `z_dev_model` database.
+
+**Primary Key**: `year`, `run_id`
```
Damonamajor: This is just because model.metadata is keyed to these, but shouldn't it just be run_id?

jeancochrane: **[Suggestion, optional]** Yeah, I believe that's right. No harm in correcting it for this table.


```suggestion
**Primary Key**: `run_id`
```

dbt/models/model/docs.md
```diff
@@ -258,3 +267,11 @@ View to compile PIN-level model inputs shared between the residential
 
 **Primary Key**: `year`, `meta_pin`
 {% enddocs %}
+
+# z_dev_model.metadata
+
+{% docs table_z_dev_model_metadata %}
+Developmental counterpart to `model.metadata`. Used for junk runs that are not intended for stored indefinetely. Contents are deleted after 30 days.
```
jeancochrane: **[Nitpick, optional]** A couple typos:


```suggestion
Development counterpart to `model.metadata`. Used for junk runs that are not intended to be stored indefinitely. Contents are deleted after 30 days.
```

dbt/models/model/docs.md
```diff
@@ -258,3 +267,11 @@ View to compile PIN-level model inputs shared between the residential
...
+
+**Primary Key**: `year`, `run_id`
```
jeancochrane: **[Suggestion, optional]** Same here as above:


```suggestion
**Primary Key**: `run_id`
```

dbt/models/shared_columns.md
```diff
@@ -1515,6 +1515,20 @@ when the run finished. Names are sourced from current and former employees
...
+{% docs shared_column_run_type %}
+Type of valuation model run. Possible values are:
+
+- `junk`
+- `rejected`
+- `test`
+- `baseline`
+- `candidate`
+- `final`
+- `comps`
```
jeancochrane: **[Suggestion, optional]** Good call documenting this column! Here's a little more detail for each run type:

```suggestion
- `junk`: Temporary test model run that will not be preserved.
  Junk model runs auto-delete after 30 days.
- `rejected`: Model run that we evaluated as a candidate for a final model,
  but ultimately rejected. See `baseline`, `candidate`, and `final` for
  more details.
- `test`: Test model run that should be preserved. In contrast to `junk`
  model runs, `test` model runs do not auto-delete.
- `baseline`: A model run that we use to compare alternative model runs
  (i.e. `candidate` runs) in a given assessment year. The baseline model run
  usually has the same features and parameters as the prior assessment year,
  but with updated data for the current assessment year. 
- `candidate`: A model run that we are evaluating as a potential `final` run
  for an assessment year. We compare this type of model run against the
  `baseline` model run to find one winning model. Winners get retagged as
  `final` while losers get retagged as `rejected` once we choose final models.
- `final`: The final model for a given property type (i.e. residential or
  condo) in a given an assessment year. Models tagged with this run type
  produce values that get sent to desk review and used as the basis for
  mailed values. Note that there can be more than one final model for a given
  property type in a given year, if we used a different final model for
  different townships.
- `comps`: A model run that we used to compute significant sales. Often
  the `final` model includes significant sales, but we use this run type if we
  need to make changes to the significant sales algorithm after we have already
  chosen a `final` model.
```

dbt/models/model/schema.yml
```diff
@@ -207,7 +207,23 @@ sources:
...
 models:
+  - name: model.vw_model_run_type
+    description: '{{ doc("view_vw_model_run_type") }}'
+    columns:
+      - name: run_id
+        description: '{{ doc("shared_column_run_id") }}'
+      - name: run_type
+        description: '{{ doc("shared_column_run_type") }}'
+      - name: year
+        description: '{{ doc("shared_column_year") }}'
```
jeancochrane, line 234: **[Suggestion, required]** Let's add a test to each of these tables to ensure they remain unique by `run_id`.

jeancochrane (approved):
Nice work!

dbt/models/model/schema.yml
Damonamajor: See the PR notes.

PR #1072, opened 2026-07-15
Remove `test-iasworld-data` workflow

### LINKED ISSUE #1063: Remove daily iasWorld data tests workflow
Let's remove the workflow that produces `qc.test_run_result`, as it is unused.

As part of this work, make sure to remove the `--run-github-workflow` option from [`service-spark-iasworld`](https://github.com/ccao-data/service-spark-iasworld/), which is the primary consumer of this workflow.

jeancochrane:
We haven't used any data generated by the `test-iasworld-data` workflow in a long time. To reduce the maintenance burden of maintaining a workflow that we don't use, we decided to remove it. This PR removes the workflow, which is safe to do since https://github.com/ccao-data/service-spark-iasworld/pull/42 updated its only consumer and stopped calling it.

Closes https://github.com/ccao-data/data-architecture/issues/1063.


wrridgeway (approved):
I'm gonna miss this guy.

PR #1073, opened 2026-07-16
Fix permits open data asset date filtering

### LINKED ISSUE #1061: Fix permits open data asset date filtering
The scheduled job that updates our open data assets is configured to only update the most recent year of data by default when it runs every month. This works fine for most of our assets, but it doesn't work well with permits, whose `year` is derived from `date_issued` (i.e. the date the municipality issued the permit) rather than `date_submitted` (i.e. the date that the municipality submitted the permit to the CCAO). See the [source code](https://github.com/ccao-data/data-architecture/blob/5b610576f036791bf51e4bea98075ce285d81da2/dbt/models/default/default.vw_pin_permit.sql#L28) for evidence.

Broadly speaking, I think there are two approaches we could take here:

1. Derive the year that we use for Socrata upload filtering from `date_submitted` instead of `date_issued`
2. Remove the year filtering for the permits asset, and simply upload all years of data

I leave it up to Billy to decide which of these approaches is simpler, though I have a slight preference for 2 if it's not too complicated.

wrridgeway:
We currently only refresh the most current year of the permits asset on the open data portal, defined as the year the permit was issued. Using this field to determine which rows need to be refreshed or uploaded can lead to permits that were issued for prior years but submitted more recently to be missed. This PR adjusts our upload script to look back through permits with a date issued within the last 5 years (rather than 1) to help alleviate this issue.

I can't really test this workflow with the new value since the `years_active` field is only used for scheduled runs, but it's been working without issue for a while now and we can wait till 8/1 to see if this upload goes as planned (I _could_ force it locally, but that wouldn't really be testing the same thing).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,7 @@ repos:
     hooks:
       - id: check-sort-dbt-yaml-files
         name: check-sort-dbt-yaml-files
-        entry: python3 dbt/scripts/check_sort_dbt_yaml_files.py
+        entry: python dbt/scripts/check_sort_dbt_yaml_files.py
         language: system
         types_or: [yaml, markdown]
         files: ^dbt/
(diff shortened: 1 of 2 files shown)


jeancochrane (approved):
Let's do it! I created a calendar invite for us on Monday 8/3 to remind ourselves to check in on this.

PR #1075, opened 2026-07-28
Upgrade R pre-commit hooks

jeancochrane:
Now that our server and laptops are running R 4.6, we need to update our R pre-commit hooks to the latest version.

I'm also taking this opportunity to start pinning our pre-commit dependencies to immutable releases or commit hashes, for better security.


wrridgeway (approved):
Thank you!!

PR #1078, opened 2026-07-29
Update ccao.adjective.csv

### LINKED ISSUE #1052: Revise adjective list in `ccao.adjective.csv`
Because we use `ccao.adjective.csv` in conjunction with staff and intern names to name our models, let's remove some of the adjectives with negative valence. (And add any other fun ones!)

ccao-jardine:
Removed some adjectives, added more!


wrridgeway (approved):
You may have removed boring, but that doesn't mean these changes aren't exactly that.

PR #1084, opened 2026-08-05
Add `land.user7` to IC reference query and fix unexpected formula bug

jeancochrane:
This PR makes two small tweaks to the IC reference file export:

1. Adds the `land.user7` column to the land details query, per stakeholder request[^1]
2. Fixes a small bug in the `export_models.py` script that I noticed while working on adding the new column

The bug in question relates to string values that start with an equals sign (`=`): `openpyxl` interprets these values to indicate formulae ([docs](https://openpyxl.readthedocs.io/en/stable/simple_formulae.html#using-formulae)), so it writes them to the workbook using the formula data type. The land details query for 2026 was pulling in an address that starts with an equals sign due to a typo, so `openpyxl` was writing that cell as a formula, causing Excel to prompt me to repair the workbook due to a nonsense formula when I opened it.

[^1]: From our stakeholder: "We found another field that would be helpful to have in the mix. When you get a chance, could you add USER7 to the Land Details file? It's the Incentive Control Number and should be useful to have available when needed."


dbt/export/templates/qc.vw_ic_reference_all_non_res_land_details.xlsx
jeancochrane, line 1: The only change to this template is adding a new column for `Incentive Control Number` as the last column of the workbook. I ran the Inquire query manually to confirm that this is the correct position for this new column.

wrridgeway (approved):
What a weird issue to have to anticipate.

PR #1087, opened 2026-08-12
Address docs drift by removing some stale columns

### LINKED ISSUE #1086: Documentation drift: 11 documented columns absent from the catalog (all 11 hand-checked)
Hi — we surveyed public dbt projects, comparing each one's `manifest.json` (what the YAML documents) against its `catalog.json` (what the warehouse reported at `dbt docs generate` time). Your published docs site was one we could reach.

The comparison flagged **11 documented columns across 7 models** that do not appear in the catalog.

Two things make this list unusually solid, so it is probably worth a look:

- **We checked all 11 by hand against your model SQL, not a sample.** All 11 are real disagreements between the two files.
- Your project builds on a single adapter (Athena), which rules out the main way a finding like this can be a false positive — a column that only exists under a different adapter.

### How to reproduce

```bash
dbt docs generate
npx clarilayer dbt-check
```

It reads `target/manifest.json` and `target/catalog.json` locally, sends nothing anywhere, and needs no account. Full disclosure: we build that tool, and this issue is the result of running it across public dbt projects.

### Caveats worth reading before anyone spends time

- These are disagreements between two files, not bugs we have confirmed in your pipeline.
- `dbt docs generate` does not rebuild models. If your docs job runs without a preceding `dbt run`, the catalog can describe an older deployment, in which case some of these reflect undeployed changes rather than documentation errors. Our check cannot tell those apart — a known blind spot on our side.
- Some may be intentional: a column deliberately not propagated, a model you know is not built in the docs target.

### Data and method

Corpus, per-project results, and the full false-positive audit: https://github.com/clarilayer/clarilayer/tree/main/survey — your project's record is in `per_project.json`. Write-up: https://clarilayer.com/blog/dbt-docs-warehouse-drift-survey

No action needed on our account. If any of these are wrong, that is the more useful outcome for us and we would genuinely like to hear it.

jeancochrane:
The schema definitions in our data catalog include a few "stale" columns that were removed from the model or source but never updated in the docs. This PR cleans out those columns to keep our docs up-to-date.

While I was auditing our docs, I also took a moment to clean up the docs for the `sale.parameter` table, which was missing a number of columns.

Full list of columns that I removed:

| Node                                  | Columns missing from DB                                            |
|---------------------------------------|--------------------------------------------------------------------|
| `model.default.vw_pin_sale`           | Removed `requires_field_check`, renamed `mydec_line_9_other_description` ➡️ `mydec_line_9_other_change_description`           |
| `model.default.vw_pin_sale_combined`  | Renamed renamed `mydec_line_9_other_description` ➡️ `mydec_line_9_other_change_description`                                   |
| `model.model.final_model`             | Renamed `triad_name` ➡️ `triad_code`                                                       |
| `model.model.vw_card_res_input`       | Renamed `other_ari` ➡️ `other_affordability_risk_index`, `other_dci` ➡️ `other_distressed_communities_index`                                           |
| `model.model.vw_pin_condo_input`      | Renamed `other_ari` ➡️ `other_affordability_risk_index`, `other_dci` ➡️ `other_distressed_communities_index`                                             |
| `model.model.vw_pin_shared_input`     | Renamed `other_ari` ➡️ `other_affordability_risk_index`, `other_dci` ➡️ `other_distressed_communities_index`                                             |
| `source.iasworld.htpar`               | Removed `name1` (defined on `htagnt`, not `htpar`)                                                             |
| `source.sale.flag`                    | Renamed `meta_sale_document_number` ➡️ `meta_sale_document_num`                                        |
| `source.sale.group_mean`              | Renamed `mean_price` ➡️ `group_mean`, removed `mean_price_per_sqft`                                |
| `source.sale.parameter`               | Removed/renamed `condo_stat_groups`, `dev_bounds`, `ptax_sd`, and `res_stat_groups`    |

Closes #1086.

### DIFF
--- a/dbt/models/iasworld/schema/iasworld.htagnt.yml
+++ b/dbt/models/iasworld/schema/iasworld.htagnt.yml
@@ -68,7 +68,7 @@ sources:
           - name: loaded_at
             description: '{{ doc("shared_column_loaded_at") }}'
           - name: name1
-            description: Name line 1
+            description: '{{ doc("shared_column_agent_name") }}'
           - name: name1_companyname
             description: Agent 1 company name
           - name: name1_firstname
--- a/dbt/models/sale/columns.md
+++ b/dbt/models/sale/columns.md
@@ -1,4 +1,16 @@
+## earliest_data_ingest
+
+{% docs column_earliest_data_ingest %}
+Date of earliest sale used in validation.
+
+This range is inclusive of the rolling window period used for calculating
+statistical groups. In other words, if the earliest sale to-be-flagged is
+2013-12-01 and the rolling window period is 9 months, then the earliest sale
+*used* (i.e. the `earliest_data_ingest`) would be 2013-03-01.
+{% enddocs %}
+
 ## group_id
+
 {% docs column_group_id %}
 Group string used as a unique identifier.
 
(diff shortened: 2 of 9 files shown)


dbt/models/iasworld/schema/iasworld.htagnt.yml
```diff
@@ -68,7 +68,7 @@ sources:
           - name: loaded_at
             description: '{{ doc("shared_column_loaded_at") }}'
           - name: name1
-            description: Name line 1
+            description: '{{ doc("shared_column_agent_name") }}'
```
jeancochrane, line 71: Switching to the shared column definition while I'm here, for centralization. Note that there is also a `name2` column on this model, so the shared column might not distinguish well between those two; we don't seem to use `name2` anywhere, however, so I assume it's fine.

dbt/models/sale/columns.md
```diff
@@ -1,4 +1,16 @@
+## earliest_data_ingest
```
jeancochrane, line 1: The diff here is larger than it might otherwise be, because I moved a bunch of column definitions from the `sale.parameter` table into this doc for cleanliness. So some of these docs blocks are truly new, while others have just been moved. It's probably best for Michael to give them all a once-over regardless, since this table has drifted quite a lot from the time when we originally wrote these docs.

jeancochrane:
Tagging in @wagnerlmichael to make sure I got my sales val docs right, and @wrridgeway for the rest.

dbt/models/shared_columns.md
```diff
@@ -1550,22 +1550,6 @@ Type of valuation model run. Possible values are:
...
-It is unclear of IHDA's data timeline. Year currently refers to final year of census data used.
-{% enddocs %}
-
-## dci
-
-{% docs shared_column_dci %}
-Distressed Communities Index from the Economic Innovation Group.
-
-Unit of observations is ZIP Code. Year refers to year of final year of census data used.
-{% enddocs %}
```
wrridgeway, line 1567: We do still have these columns in `model.vw_pin_shared_input` and there are `dci` and `ari` tables in the `other` db. Should we remove them from the shared input view and move this documentation to `schema.yml` in `other`?

wrridgeway (approved):
I looked at everything that wasn't in the sale db.

dbt/models/sale/columns.md
```diff
@@ -9,23 +21,128 @@ Typically a combination of a:
 {% enddocs %}
 
 ## group_size
+
 {% docs column_group_size %}
 Count of sales/properties in the group.
-This is the population within which price statistics (e.g., standard deviation) are calculated.
+
+This is the population within which price statistics (e.g., standard deviation)
```
wagnerlmichael:
```suggestion
This is the number of observations in a population within which price statistics (e.g., standard deviation)
```

dbt/models/sale/columns.md
```diff
@@ -9,23 +21,128 @@ Typically a combination of a:
...
+## iso_forest_cols
+
+{% docs column_iso_forest_cols %}
+Columns used as features in the isolation forest model
+{% enddocs %}
+
+## latest_data_ingest
+
+{% docs column_latest_data_ingest %}
+Date of latest sale used in validation
```
wagnerlmichael:
```suggestion
Date of the latest sale pulled by the ingest query
```

dbt/models/sale/columns.md
```diff
@@ -9,23 +21,128 @@ Typically a combination of a:
...
+(set to “Not outlier”).
+
+**Caveat:** Sales with certain PTAX-203 line-10 items marked may still be
+classified as outliers.
+{% enddocs %}
+
+## raw_price_threshold
+
+{% docs column_raw_price_threshold %}
+Upper bound price beyond which all sales are flagged as outliers
```
wagnerlmichael:
```suggestion
Upper bound price beyond which all sales are flagged as outliers, independent of any standard deviation or group size mechanics
```

dbt/models/sale/columns.md
```diff
@@ -9,23 +21,128 @@ Typically a combination of a:
...
+
+Any sales that match `run_filter` but not this filter will still get flagged,
+but those flags will not get written to the database.
+{% enddocs %}
+
+## short_term_owner_threshold
+
+{% docs column_short_term_owner_threshold %}
+Properties with a significant price change and multiple sales within this time
+duration (in days) are flagged
```
wagnerlmichael:
```suggestion
Number of days since a property's previous sale under which the seller is
considered a "short-term owner". Sales under this threshold
receive the "Short-term owner" characteristic reason in the
`sv_outlier_reason` fields
```

wagnerlmichael (approved):
Thank you for taking on these stale changes! I've added suggestions on a few of the column docs.

dbt/models/shared_columns.md
jeancochrane: Oops, good catch! I missed that the `model` columns have verbose names (i.e. `affordability_risk_index` and `distressed_community_index`). I reintroduced these docs in 984930b563708ba61d7105087f34b737e2e902a7, fixed the `model` column names, and added some quick column docs for the score columns on `other.ari` and `other.dci`, too.

dbt/models/sale/columns.md
jeancochrane: Thanks! Fixed in a batch in 625c80cd68b96a508c83285a4401186ae46e2d2a.

dbt/models/sale/columns.md
jeancochrane: Fixed in a batch in 625c80cd68b96a508c83285a4401186ae46e2d2a.

dbt/models/sale/columns.md
jeancochrane: Fixed in a batch in 625c80cd68b96a508c83285a4401186ae46e2d2a.

dbt/models/sale/columns.md
jeancochrane: Fixed in a batch in 625c80cd68b96a508c83285a4401186ae46e2d2a.

PR #1090, opened 2026-08-14
Remove unused `model.vw_model_run_type` view

jeancochrane:
This PR reverts the primary change from #1071 (the `model.vw_model_run_type` view) since we are no longer using it after pivoting to a new junk model management approach in https://github.com/ccao-data/model-res-avm/pull/531. See https://github.com/ccao-data/model-res-avm/issues/522 for background.

There are a few components of #1071 that I didn't revert, since I think they will be helpful going forward:

- I retained the change that tweaked the uniqueness test on `model.metadata` so that it is unique by `run_id`, rather than `(run_id, year)`
- I retained the new docs for the `run_type` column; even though these docs are not being used anywhere right now, I expect they may be useful in the future

Note that I've already deleted the following resources in AWS, so merging this PR won't actually make any changes to our resources:

- `model.vw_model_run_type` Glue table
- Dev model results S3 bucket
- Dev model results Glue crawler
- `z_dev_model` Glue database


wrridgeway (approved):
Looks good to me, thanks for handling this.

PR #1093, opened 2026-08-18
Create tableau dashboard based on iasworld tests

### LINKED ISSUE #1088: Create Tableau dashboard based on iasWorld tests
## Background

We currently maintain a set of iasWorld tests that we define as [dbt data tests](https://docs.getdbt.com/docs/build/data-tests) and run via the script [`run_iasworld_tests.py`](https://github.com/ccao-data/data-architecture/blob/master/dbt/scripts/run_iasworld_data_tests.py). For some docs explaining how we maintain and run these tests, see [the relevant section in our data architecture docs](https://github.com/ccao-data/data-architecture/tree/master/dbt#iasworld-tests). We run these tests whenever our Res Val QC stakeholders ask for them, which is typically at the open and close of every town.

We have recently had some success building out QC products using Athena views that we feed into Tableau dashboards: The Valuations team is now using a Tableau dashboard for town close QC, and soon they will test drive a new dashboard for res land QC. If we could move our iasWorld tests into Tableau, we could align this QC products with our emerging standard (Tableau), while offering all of the other benefits of a Tableau dashboard -- namely, that it will auto-update once per day, and our stakeholders will be able to access it on-demand without needing to contact us.

## Implementation plan

Moving our iasWorld tests into Tableau will require more work than we had to do when moving town close QC reports into Tableau. This is because our town close QC reports have always been based on individual Athena views, whereas we define our iasWorld tests as dbt data tests and split them across many different iasWorld source schema definitions. Hence, to create a Tableau dashboard based on our iasWorld tests, we'll need to recreate our data tests in some new Athena views that Tableau can pull from. This approach will be similar to the res land QC dashboard, just with a greater number of underlying Athena views.

Here is a rough list of the tasks that I think this issue will require:

- Create a new view for each iasWorld source called `qc.vw_report_iasworld_test_<tablename>` that recreates the tests defined in that iasWorld source's `schema.yml` file
    - For example, the view `qc.vw_report_iasworld_test_pardat` should include the following tests:
        - `iasworld_pardat_adrno_length_lte_5`
        - `iasworld_pardat_class_equals_luc`
        - `iasworld_pardat_cur_in_accepted_values`
        - `iasworld_pardat_nbhd_matches_legdat_township`
        - `iasworld_pardat_nbhd_matches_spatial_town_nbhd`
        - `iasworld_pardat_seq_all_sequential_exist`
        - `iasworld_pardat_unique_by_parid_taxyr`
    - Each view should select the following columns, with one row for each PIN/card/lline that is failing a particular test[^1]:
        - `test_name` (string): The name slug for the test, derived from the `name` field in the dbt test definition 
        - `test_description` (string): The human-readable description for the test, derived from the `meta.description` field in the dbt test definition
        - `test_category` (string): The slug category for the test, derived from the `meta.category` field in the dbt test definition
        - `taxyr`
        - `parid`
        - `card` (null if the underlying table doesn't include cards)
        - `lline` (null if the underlying table doesn't include land lines)
        - `township_code`
        - `class`
        - `who`
        - `wen`
        - `additional_fields` (type TBD): Key-value pairs for all of the additional contextual fields that are important for debugging this test failure
            - I'm not totally sure how best to format this column, so I'm open to feedback! In the current version of the test workbook, we include each field as its own column, which means that some test categories have a bunch of additional fields that are blank for most tests. It seems like an [array map](https://docs.aws.amazon.com/athena/latest/ug/arrays-create-maps.html) might be an easier way to manage this, but I don't have a good sense of how we would display that in Tableau. We should discuss with Kyra and decide on a path forward.
    - Note that there are a few scattered QC views that have iasWorld tests defined on their schema, namely `qc.vw_iasworld_sales_*`, `qc.vw_report_town_close_res_parcels_not_set_to_cost_approach`, `qc.vw_neg_asmt_value`, and `reporting.vw_pin_value_long`. It would be best to try to move these tests into the `qc.vw_report_iasworld_test_*` view for whichever table is most important in the test, but I haven't looked at these tests in detail, so I don't have a sense of how hard that will be. In the worst case scenario, it's fine to continue using these views as the basis for the new table definitions.
- Create a new view `qc.vw_report_iasworld_test` that unions all of the individual views
    - This view should add a `township_name` column by joining to `spatial.township` 
- Create a Tableau dashboard based on `qc.vw_report_iasworld_test` that replicates the structure of the existing test workbook
    - Unlike the workbook, the dashboard should not have tabs for each test category; instead, it should include the category as a column, and should have a top-level category filter
    - The dashboard should include a township filter

> [!WARNING]  
> As we implement these views, make sure to use `iasworld.pardat` as the basis for all views, to make sure we are excluding any deactivated parcels from the query set.

[^1]: Note that this is a different table design than the res land QC view, which is unique by PIN + lline rather than PIN + lline + test.

Damonamajor:
This is the query that I have been using to test. It seems like it gets updated pretty frequently, so although it passessed yesterday, it is failing today. Here's an example of a PIN which fails in this query, but has updated info for matching class_equals_luc: Parid `25343080220000`

image

```
WITH view_rows AS (
    FROM z_ci_1088_create_tableau_dashboard_based_on_iasworld_tests_qc.vw_report_iasworld_test_pardat
),
test_failure_rows AS (
    SELECT 'iasworld_pardat_adrno_length_lte_5' AS test_name, parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_pardat_adrno_length_lte_5
    UNION ALL
    SELECT 'iasworld_pardat_class_equals_luc', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_pardat_class_equals_luc
    UNION ALL
    SELECT 'iasworld_pardat_cur_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_pardat_cur_in_accepted_values
    UNION ALL
    SELECT 'iasworld_pardat_nbhd_matches_legdat_township', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_pardat_nbhd_matches_legdat_township
    UNION ALL
    SELECT 'iasworld_pardat_nbhd_matches_spatial_town_nbhd', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_pardat_nbhd_matches_spatial_town_nbhd
    UNION ALL
    SELECT 'iasworld_pardat_seq_all_sequential_exist', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_pardat_seq_all_sequential_exist
    UNION ALL
    SELECT 'iasworld_pardat_unique_by_parid_taxyr', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_pardat_unique_by_parid_taxyr
)
SELECT
    COALESCE(v.parid, f.parid) AS parid,
    COALESCE(v.taxyr, f.taxyr) AS taxyr,
    CASE
        WHEN v.parid IS NULL THEN 'only in old failures'
        WHEN f.parid IS NULL THEN 'only in new failures'
    END AS source
FROM view_rows v
    ON v.test_name = f.test_name
    AND v.parid = f.parid
    AND v.taxyr = f.taxyr
WHERE v.parid IS NULL OR f.parid IS NULL
ORDER BY parid, test_name

```

owndat only produces errors for whitespaces

```
WITH view_rows AS (
    FROM z_ci_1088_create_tableau_dashboard_based_on_iasworld_tests_qc.vw_report_iasworld_test_owndat
    WHERE test_name NOT IN ('iasworld_owndat_address_columns_no_extra_whitespace')
),
test_failure_rows AS (
    SELECT 'iasworld_owndat_cur_in_accepted_values' AS test_name, parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_owndat_cur_in_accepted_values
    UNION ALL
    SELECT 'iasworld_owndat_parid_in_pardat_parid', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_owndat_parid_in_pardat_parid
    UNION ALL
    SELECT 'iasworld_owndat_parid_not_null', CAST(NULL AS VARCHAR) AS parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_owndat_parid_not_null
    UNION ALL
    SELECT 'iasworld_owndat_seq_all_sequential_exist', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_owndat_seq_all_sequential_exist
    UNION ALL
    SELECT 'iasworld_owndat_unique_by_parid_taxyr', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_owndat_unique_by_parid_taxyr
)
SELECT
    COALESCE(v.parid, f.parid) AS parid,
    COALESCE(v.taxyr, f.taxyr) AS taxyr,
    CASE
        WHEN v.test_name IS NULL THEN 'only in old failures'
        WHEN f.test_name IS NULL THEN 'only in new failures'
    END AS source
FROM view_rows v
    ON v.test_name = f.test_name
    AND v.parid IS NOT DISTINCT FROM f.parid
    AND v.taxyr = f.taxyr
WHERE v.test_name IS NULL OR f.test_name IS NULL
ORDER BY parid, test_name
```

It then unions both tables into the view `dbt/models/qc/qc.vw_report_iasworld_test_all.sql`

### DIFF
--- a/dbt/macros/generate_iasworld_qc_test_view.sql
+++ b/dbt/macros/generate_iasworld_qc_test_view.sql
@@ -68,17 +68,20 @@
             '{{ test.description }}' as test_description,
             '{{ test.category }}' as test_category,
             {% if test.additional_select_columns -%}
-                map(
-                    array[
-                        {%- for col_name in test.additional_select_columns -%}
-                            '{{ col_name }}'{{ ", " if not loop.last }}
-                        {%- endfor %}
-                    ],
-                    array[
-                        {%- for col_name in test.additional_select_columns -%}
-                            cast({{ col_name }} as varchar) {{- ", " if not loop.last }}
-                        {%- endfor %}
-                    ]
+                map_filter(
+                    map(
+                        array[
+                            {%- for col_name in test.additional_select_columns -%}
+                                '{{ col_name }}'{{ ", " if not loop.last }}
+                            {%- endfor %}
+                        ],
+                        array[
+                            {%- for col_name in test.additional_select_columns -%}
+                                cast({{ col_name }} as varchar) {{- ", " if not loop.last }}
+                            {%- endfor %}
+                        ]
+                    ),
+                    (k, v) -> v IS NOT NULL
                 ) as additional_columns
             {%- else -%} cast(null as map(varchar, varchar)) as additional_columns
             {%- endif %}
--- a/dbt/macros/generate_iasworld_qc_test_view.sql
+++ b/dbt/macros/generate_iasworld_qc_test_view.sql
@@ -1,0 +1,140 @@
+{#-
+    Generate a QC report view that runs a set of boolean "tests" against a
+    base query and returns one row per (record, failing test) pair. We use
+    the resulting views to power iasWorld QC.
+
+    Args:
+        base_query: A SQL SELECT statement (as a string) that returns the
+            identifying columns plus any additional columns referenced by
+            the `condition` or `additional_select_columns` attributes of the
+            `tests` argument. Source tables that don't include a required
+            identifying column (usually `card` or `lline`) should return
+            `CAST(NULL AS <type>) AS <column_name>` for that column to conform
+            to the expected shape. Required identifying columns include:
+                - parid: varchar
+                - taxyr: varchar
+                - card: decimal
+                - lline: decimal
+                - township_code: varchar
+                - class: varchar
+                - who: varchar
+                - wen: varchar
+        tests: A list of dicts, each with keys:
+            - name: A unique, descriptive slug for the test
+            - description: A human-readable description of what the test checks
+            - category: A category slug used to group related tests
+            - condition: A SQL boolean expression, evaluated against
+                `base_query`, that is TRUE when the record passes the test
+                and FALSE when it fails
+            - additional_select_columns (optional): A list of column names
+                from `base_query` to include in the output as a map of column
+                name -> stringified value for records that fail the test
+
+    Returns:
+        A query that selects one row per record per failing test.
+-#}
+{% macro generate_iasworld_qc_test_view(base_query, tests) %}
+    {% do _validate_iasworld_qc_tests(tests, exceptions.raise_compiler_error) %}
+    with
+        base as ({{ base_query }}),
+
+        test_result as (
+            select
+                *,
+                -- noqa: disable=layout.indent
+                {% for test in tests %}
+                    not ({{ test.condition }}) as {{ test.name }}
+                    {{- "," if not loop.last }}
+                {% endfor %}
+            -- noqa: enable=layout.indent
+            from base
+            where
+                taxyr
+                between '{{ var("data_test_iasworld_year_start") }}'
+                and '{{ var("data_test_iasworld_year_end") }}'
+        )
+
+    {% for test in tests %}
+        select
+            parid,
+            taxyr,
+            card,
+            lline,
+            township_code,
+            class,
+            who,
+            wen,
+            '{{ test.name }}' as test_name,
+            '{{ test.description }}' as test_description,
+            '{{ test.category }}' as test_category,
+            {% if test.additional_select_columns -%}
+                map(
+                    array[
+                        {%- for col_name in test.additional_select_columns -%}
+                            '{{ col_name }}'{{ ", " if not loop.last }}
+                        {%- endfor %}
+                    ],
+                    array[
+                        {%- for col_name in test.additional_select_columns -%}
+                            cast({{ col_name }} as varchar) {{- ", " if not loop.last }}
+                        {%- endfor %}
+                    ]
+                ) as additional_columns
+            {%- else -%} cast(null as map(varchar, varchar)) as additional_columns
+            {%- endif %}
+        from test_result
+        where {{ test.name }} {{ "UNION ALL" if not loop.last }}
+    {% endfor %}
+{% endmacro %}
+
+{#-
+    Validate that `tests` is a list of dicts, each containing the keys
+    required by `generate_iasworld_qc_test_view`. Raises a compiler error
+    (via `raise_error_func`) on the first validation failure it finds.
+
+    Args:
+        tests: The `tests` argument passed to `generate_iasworld_qc_test_view`
+        raise_error_func: A function to call with an error message when
+            validation fails. Takes `exceptions.raise_compiler_error` in
+            production, and a mock in unit tests so that the error can be
+            returned for equality comparison instead of raised
+-#}
+{% macro _validate_iasworld_qc_tests(tests, raise_error_func) %}
+    {%- set required_keys = ["name", "description", "category", "condition"] -%}
+    {%- if tests is not iterable or tests is mapping or tests is string -%}
+        {{-
+            return(
+                raise_error_func('"tests" argument must be a list, got: ' ~ tests)
+            )
+        -}}
+    {%- endif -%}
+    {%- for test in tests -%}
+        {%- if test is not mapping -%}
+            {{-
+                return(
+                    raise_error_func(
+                        'Each element of "tests" must be an object/dict,'
+                        ~ " got: "
+                        ~ test
+                    )
+                )
+            -}}
+        {%- endif -%}
+        {%- for key in required_keys -%}
+            {%- if key not in test -%}
+                {{-
+                    return(
+                        raise_error_func(
+                            'Missing required "'
+                            ~ key
+                            ~ '" key in test'
+                            ~ " config: "
+                            ~ test
+                        )
+                    )
+                -}}
+            {%- endif -%}
+        {%- endfor -%}
+    {%- endfor -%}
+    {{ return(none) }}
+{% endmacro %}
--- a/dbt/macros/generate_iasworld_qc_test_view.sql
+++ b/dbt/macros/generate_iasworld_qc_test_view.sql
@@ -77,11 +77,13 @@
                         ],
                         array[
                             {%- for col_name in test.additional_select_columns -%}
-                                cast({{ col_name }} as varchar) {{- ", " if not loop.last }}
+                                cast(
+                                    {{ col_name }} as varchar
+                                ) {{- ", " if not loop.last }}
                             {%- endfor %}
                         ]
                     ),
-                    (k, v) -> v IS NOT NULL
+                    (k, v) -> v is not null
                 ) as additional_columns
             {%- else -%} cast(null as map(varchar, varchar)) as additional_columns
             {%- endif %}
--- a/dbt/macros/tests/test_all.sql
+++ b/dbt/macros/tests/test_all.sql
@@ -6,4 +6,5 @@
     {% do test_generate_alias_name() %}
     {% do test_format_additional_select_columns() %}
     {% do test_insert_hyphens() %}
+    {% do test_generate_iasworld_qc_test_view() %}
 {% endmacro %}
--- a/dbt/macros/tests/test_all.sql
+++ b/dbt/macros/tests/test_all.sql
@@ -6,5 +6,4 @@
     {% do test_generate_alias_name() %}
     {% do test_format_additional_select_columns() %}
     {% do test_insert_hyphens() %}
-    {% do test_generate_iasworld_qc_test_view() %}
 {% endmacro %}
--- a/dbt/models/qc/docs.md
+++ b/dbt/models/qc/docs.md
@@ -205,8 +205,8 @@ This view is not currently in use in any QC workflows.
 # vw_report_iasworld_test_pardat
 
 {% docs view_vw_report_iasworld_test_pardat %}
-Aggregates iasworld.pardat data tests as a queryable view, with one row
-per PIN failing a given test.
+Aggregates test failures, stemming from iasworld pardat. Each observation is a failure
+based on unique PIN, year, and failing test combinations.
 {% enddocs %}
 
 # vw_report_res_land
--- a/dbt/models/qc/docs.md
+++ b/dbt/models/qc/docs.md
@@ -207,6 +207,9 @@ This view is not currently in use in any QC workflows.
 {% docs view_vw_report_iasworld_test_all %}
 Combines test failures from all iasworld test views. Each observation is a failure
 based on unique PIN, year, and failing test combinations.
+
+**Primary Key**: `parid`, `taxyr`, `test_name`
+
 {% enddocs %}
 
 # vw_report_iasworld_test_owndat
@@ -214,6 +217,9 @@ based on unique PIN, year, and failing test combinations.
 {% docs view_vw_report_iasworld_test_owndat %}
 Aggregates test failures, stemming from iasworld owndat. Each observation is a failure
 based on unique PIN, year, and failing test combinations.
+
+**Primary Key**: `parid`, `taxyr`, `test_name`
+
 {% enddocs %}
 
 # vw_report_iasworld_test_pardat
@@ -221,6 +227,9 @@ based on unique PIN, year, and failing test combinations.
 {% docs view_vw_report_iasworld_test_pardat %}
 Aggregates test failures, stemming from iasworld pardat. Each observation is a failure
 based on unique PIN, year, and failing test combinations.
+
+**Primary Key**: `parid`, `taxyr`, `test_name`
+
 {% enddocs %}
 
 # vw_report_res_land
--- a/dbt/models/qc/docs.md
+++ b/dbt/models/qc/docs.md
@@ -202,6 +202,13 @@ classes.
 This view is not currently in use in any QC workflows.
 {% enddocs %}
 
+# vw_report_iasworld_test_pardat
+
+{% docs view_vw_report_iasworld_test_pardat %}
+Aggregates iasworld.pardat data tests as a queryable view, with one row
+per PIN failing a given test.
+{% enddocs %}
+
 # vw_report_res_land
 
 {% docs view_vw_report_res_land %}
--- a/dbt/models/qc/docs.md
+++ b/dbt/models/qc/docs.md
@@ -202,6 +202,20 @@ classes.
 This view is not currently in use in any QC workflows.
 {% enddocs %}
 
+# vw_report_iasworld_test_all
+
+{% docs view_vw_report_iasworld_test_all %}
+Combines test failures from all iasworld test views. Each observation is a failure
+based on unique PIN, year, and failing test combinations.
+{% enddocs %}
+
+# vw_report_iasworld_test_owndat
+
+{% docs view_vw_report_iasworld_test_owndat %}
+Aggregates test failures, stemming from iasworld owndat. Each observation is a failure
+based on unique PIN, year, and failing test combinations.
+{% enddocs %}
+
 # vw_report_iasworld_test_pardat
 
 {% docs view_vw_report_iasworld_test_pardat %}
--- a/dbt/models/qc/qc.vw_report_iasworld_test_all.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_all.sql
@@ -1,3 +1,9 @@
-SELECT *, 'pardat' AS source_table FROM {{ ref('qc.vw_report_iasworld_test_pardat') }}
+SELECT
+    *,
+    'pardat' AS source_table
+FROM {{ ref('qc.vw_report_iasworld_test_pardat') }}
 UNION ALL
-SELECT *, 'owndat' AS source_table FROM {{ ref('qc.vw_report_iasworld_test_owndat') }}
+SELECT
+    *,
+    'owndat' AS source_table
+FROM {{ ref('qc.vw_report_iasworld_test_owndat') }}
--- a/dbt/models/qc/qc.vw_report_iasworld_test_all.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_all.sql
@@ -1,0 +1,3 @@
+SELECT * FROM {{ ref('qc.vw_report_iasworld_test_pardat') }}
+UNION ALL
+SELECT * FROM {{ ref('qc.vw_report_iasworld_test_owndat') }}
--- a/dbt/models/qc/qc.vw_report_iasworld_test_all.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_all.sql
@@ -1,3 +1,3 @@
-SELECT * FROM {{ ref('qc.vw_report_iasworld_test_pardat') }}
+SELECT *, 'pardat' AS source_table FROM {{ ref('qc.vw_report_iasworld_test_pardat') }}
 UNION ALL
-SELECT * FROM {{ ref('qc.vw_report_iasworld_test_owndat') }}
+SELECT *, 'owndat' AS source_table FROM {{ ref('qc.vw_report_iasworld_test_owndat') }}
--- a/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
@@ -40,7 +40,7 @@
     },
     {
         "name": "iasworld_owndat_address_columns_no_extra_whitespace",
-        "description": "own1, own2, addr1-5, and other address columns should have no leading or trailing whitespace",
+        "description": "address columns should not have whitespace",
         "category": "column_values",
         "condition": "NOT (" ~ whitespace_parts | join(" OR ") ~ ")",
         "additional_select_columns": whitespace_cols
@@ -72,24 +72,10 @@
         LAG(owndat.seq)
             OVER (PARTITION BY owndat.parid, owndat.taxyr ORDER BY owndat.seq)
             AS prev_seq,
-        owndat.own1,
-        owndat.own2,
-        owndat.addr1,
-        owndat.addr2,
-        owndat.addr3,
-        owndat.addr4,
-        owndat.addr5,
-        owndat.adrpre,
-        owndat.adrdir,
-        owndat.adrstr,
-        owndat.adrsuf,
-        owndat.unitdesc,
-        owndat.unitno,
-        owndat.cityname,
-        owndat.statecode,
-        owndat.zip1,
-        owndat.zip2,
-        owndat.user27,
+        {% for col in whitespace_cols %}
+        CASE WHEN owndat.{{ col }} LIKE '% ' OR owndat.{{ col }} LIKE ' %'
+            THEN owndat.{{ col }} END AS {{ col }},
+        {% endfor %}
         COUNT(*)
             OVER (PARTITION BY owndat.parid, owndat.taxyr)
             AS num_duplicates
--- a/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
@@ -42,7 +42,7 @@
     {
         "name": "iasworld_owndat_unique_by_parid_taxyr",
         "description": "owndat should be unique by parid and taxyr",
-        "category": "duplicate_rows",
+        "category": "duplicate_records",
         "condition": "num_duplicates = 1",
         "additional_select_columns": ["num_duplicates"]
     }
--- a/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
@@ -72,7 +72,7 @@
         COUNT(*)
             OVER (PARTITION BY owndat.parid, owndat.taxyr)
             AS num_duplicates
-    FROM iasworld.owndat AS owndat
+    FROM {{ source('iasworld', 'owndat') }} AS owndat
     LEFT JOIN iasworld.legdat AS legdat
         ON owndat.parid = legdat.parid
         AND owndat.taxyr = legdat.taxyr
--- a/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
@@ -22,7 +22,7 @@
     {
         "name": "iasworld_owndat_parid_not_null",
         "description": "parid should not be null",
-        "category": "incorrect_values",
+        "category": "missing_values",
         "condition": "parid IS NOT NULL"
     },
     {
--- a/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
@@ -49,7 +49,7 @@
             OVER (PARTITION BY owndat.parid, owndat.taxyr)
             AS num_duplicates
     FROM {{ source('iasworld', 'owndat') }} AS owndat
-    LEFT JOIN iasworld.legdat AS legdat
+    LEFT JOIN {{ source('iasworld', 'legdat') }} AS legdat
         ON owndat.parid = legdat.parid
         AND owndat.taxyr = legdat.taxyr
         AND legdat.cur = 'Y'
--- a/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
@@ -78,14 +78,11 @@
         AND owndat.taxyr = legdat.taxyr
         AND legdat.cur = 'Y'
         AND legdat.deactivat IS NULL
-    INNER JOIN (
-        SELECT DISTINCT parid, taxyr
-        FROM iasworld.pardat
-        WHERE cur = 'Y'
-            AND deactivat IS NULL
-    ) AS pardat_parids
-        ON owndat.parid = pardat_parids.parid
-        AND owndat.taxyr = pardat_parids.taxyr
+    INNER JOIN {{ source('iasworld', 'pardat') }} AS pardat
+        ON owndat.parid = pardat.parid
+        AND owndat.taxyr = pardat.taxyr
+        AND pardat.cur = 'Y'
+        AND pardat.deactivat IS NULL
     WHERE owndat.cur = 'Y'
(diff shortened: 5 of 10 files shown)


Damonamajor:
I'm not sure this is exactly the structure you wanted. It doesn't create an individual view for each of the subqueries, but that also feels like a lot to mantain, document, etc.. For me a single view makes the most sense, but happy to reconfigure it based on your feedback. 

I also left out this section until we get a solid grasp on what we want

> Note that there are a few scattered QC views that have iasWorld tests defined on their schema, namely qc.vw_iasworld_sales_*, qc.vw_report_town_close_res_parcels_not_set_to_cost_approach, qc.vw_neg_asmt_value, and reporting.vw_pin_value_long. It would be best to try to move these tests into the qc.vw_report_iasworld_test_* view for whichever table is most important in the test, but I haven't looked at these tests in detail, so I don't have a sense of how hard that will be. In the worst case scenario, it's fine to continue using these views as the basis for the new table definitions.

It also doesn't union all iasworld views since we are only testing a single one first.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,263 @@
...
+    WHERE cur = 'Y'
+        AND deactivat IS NULL
+),
+
+distinct_town_nbhd AS (
+    SELECT DISTINCT town_nbhd
+    FROM {{ source('spatial', 'neighborhood') }}
+),
+
+pardat_seq AS (
```
Damonamajor: named like this due to sequential tests.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,263 @@
...
+        CAST(NULL AS INTEGER) AS lline,
+        legdat.township_code,
+        pardat.class,
+        pardat.who,
+        pardat.wen,
+        MAP(
+            ARRAY['adrno'],
+            ARRAY[CAST(pardat.adrno AS VARCHAR)]
+        ) AS additional_fields
+    FROM {{ source('iasworld', 'pardat') }} AS pardat
```
Damonamajor: We can pull the `from pardat` into it's own object and remove this from each query as well
```
        AND pardat.cur = 'Y'
        AND pardat.deactivat IS NULL
```

Damonamajor: I just did this, I think it makes it easier to follow.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,238 @@
...
+iasworld_pardat_adrno_length_lte_5 AS (
+    SELECT
+        'iasworld_pardat_adrno_length_lte_5' AS test_name,
+        'adrno should be <= 5 characters long' AS test_description,
+        'column_length' AS test_category,
+        pardat.taxyr,
+        pardat.parid,
+        CAST(NULL AS INTEGER) AS card,
+        CAST(NULL AS INTEGER) AS lline,
+        legdat.township_code,
```
Damonamajor: We often only need the join for township code, which should be deriveable from class, but we are also getting some errors for tests based on this. Let me know if we want this join for each. We can also move it up to the CTE.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,238 @@
...
+            ARRAY[pardat.nbhd]
+        ) AS additional_fields
+    FROM pardat
+    LEFT JOIN legdat_townships AS legdat
+        ON pardat.parid = legdat.parid
+        AND pardat.taxyr = legdat.taxyr
+    LEFT JOIN distinct_town_nbhd
+        ON pardat.nbhd = distinct_town_nbhd.town_nbhd
+    WHERE distinct_town_nbhd.town_nbhd IS NULL
+        AND pardat.nbhd NOT LIKE '%999'
```
Damonamajor: Do we want this in both nbhd queries, it seems only in this [one ](https://github.com/ccao-data/data-architecture/blob/16cfa4bdcb1668d14da56d611f3370aef0a40898/dbt/models/iasworld/schema/iasworld.pardat.yml#L358-L372)and not the previous [one](https://github.com/ccao-data/data-architecture/blob/16cfa4bdcb1668d14da56d611f3370aef0a40898/dbt/models/iasworld/schema/iasworld.pardat.yml#L171-L191)

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,238 @@
...
+        user1 AS township_code
+    FROM {{ source('iasworld', 'legdat') }}
+    WHERE cur = 'Y'
+        AND deactivat IS NULL
+),
+
+distinct_town_nbhd AS (
+    SELECT DISTINCT town_nbhd
+    FROM {{ source('spatial', 'neighborhood') }}
+),
```
jeancochrane: **[Suggestion, optional]** `spatial.neighborhood` is unique by `(town_nbhd, year)`, so I don't think we should need this subquery if instead we just join `PARDAT` to `spatial.neighborhood` by `taxyr = year` and `nbhd = town_nbhd`. This is slightly different than how the existing `iasworld_pardat_nbhd_matches_spatial_town_nbhd` test works, but I also think it's more correct; the previous definition of the test was limited by the dbt test syntax, if I recall correctly.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,238 @@
...
+        pardat.taxyr,
+        pardat.parid,
+        CAST(NULL AS INTEGER) AS card,
+        CAST(NULL AS INTEGER) AS lline,
+        legdat.township_code,
+        pardat.class,
+        pardat.who,
+        pardat.wen,
+        MAP(
+            ARRAY['luc'],
```
Damonamajor: We already select class as one of our base columns. Do we want it as additional too? This goes for class or township values.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,238 @@
...
+        ON pardat.parid = legdat.parid
+        AND pardat.taxyr = legdat.taxyr
+    WHERE NOT (pardat.class = pardat.luc)
+),
+
+iasworld_pardat_cur_in_accepted_values AS (
+    SELECT
+        'iasworld_pardat_cur_in_accepted_values' AS test_name,
+        'cur should be ''Y'' or ''D''' AS test_description,
+        CAST(NULL AS VARCHAR) AS test_category,
```
jeancochrane: **[Nitpick, optional]** Looks like we missed this in the dbt test definition, but I would put this test in the `incorrect_values` category.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,238 @@
...
+        ON pardat.nbhd = distinct_town_nbhd.town_nbhd
+    WHERE distinct_town_nbhd.town_nbhd IS NULL
+        AND pardat.nbhd NOT LIKE '%999'
+),
+
+iasworld_pardat_seq_all_sequential_exist AS (
+    SELECT
+        'iasworld_pardat_seq_all_sequential_exist' AS test_name,
+        'seq should be sequential' AS test_description,
+        CAST(NULL AS VARCHAR) AS test_category,
```
jeancochrane: **[Nitpick, optional]** I would also categorize this test as `incorrect_values`.

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,238 @@
...
+        ON pardat_seq.parid = legdat.parid
+        AND pardat_seq.taxyr = legdat.taxyr
+    WHERE NOT (pardat_seq.seq = pardat_seq.prev_seq + 1)
+),
+
+iasworld_pardat_unique_by_parid_taxyr AS (
+    SELECT
+        'iasworld_pardat_unique_by_parid_taxyr' AS test_name,
+        'pardat should be unique by parid and taxyr' AS test_description,
+        CAST(NULL AS VARCHAR) AS test_category,
```
jeancochrane: **[Suggestion, optional]** These uniqueness tests are pretty common, so maybe we should give them their own category. Perhaps `duplicate_rows`?

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
Damonamajor: Not all of the tests have a value for `meta.category`

jeancochrane (commented):
Very nice start here! I think the fundamentals of this approach are correct, and the output looks good.

That being said, it feels to me like this implementation might end up requiring a lot of repetitive boilerplate that could become difficult to maintain, i.e. a separate subquery for each individual test that needs to select the same set of columns in the same exact way. I thought it might be interesting to factor out as much of the boilerplate as possible into a shared macro, and I ended up producing a follow-up PR implementing that approach (https://github.com/ccao-data/data-architecture/pull/1095). Take a look at that PR and let me know if you think it's prohibitively confusing; if the macro stuff is impossible to understand, I'm down to continue forward with the more boilerplate-heavy approach we have here, and only refactor to a macro approach if/when it becomes unwieldy.

Damonamajor:
> Very nice start here! I think the fundamentals of this approach are correct, and the output looks good.
> 
> That being said, it feels to me like this implementation might end up requiring a lot of repetitive boilerplate that could become difficult to maintain, i.e. a separate subquery for each individual test that needs to select the same set of columns in the same exact way. I thought it might be interesting to factor out as much of the boilerplate as possible into a shared macro, and I ended up producing a follow-up PR implementing that approach (#1095). Take a look at that PR and let me know if you think it's prohibitively confusing; if the macro stuff is impossible to understand, I'm down to continue forward with the more boilerplate-heavy approach we have here, and only refactor to a macro approach if/when it becomes unwieldy.

I'll take a look through it tomorrow, but it seems like a more efficient system.

dbt/models/qc/qc.vw_report_iasworld_test_all.sql
```diff
@@ -0,0 +1,3 @@
+SELECT * FROM {{ ref('qc.vw_report_iasworld_test_pardat') }}
+UNION ALL
```
Damonamajor, line 5: Do we want a column based on which table it comes from?

dbt/macros/tests/test_all.sql
```diff
@@ -6,4 +6,5 @@
     {% do test_generate_alias_name() %}
     {% do test_format_additional_select_columns() %}
     {% do test_insert_hyphens() %}
+    {% do test_generate_iasworld_qc_test_view() %}
```
Damonamajor, line 9: Any particular reason you included this here?

dbt/macros/generate_iasworld_qc_test_view.sql
```diff
@@ -0,0 +1,143 @@
...
+                                '{{ col_name }}'{{ ", " if not loop.last }}
+                            {%- endfor %}
+                        ],
+                        array[
+                            {%- for col_name in test.additional_select_columns -%}
+                                cast({{ col_name }} as varchar) {{- ", " if not loop.last }}
+                            {%- endfor %}
+                        ]
+                    ),
+                    (k, v) -> v IS NOT NULL
```
Damonamajor: This filters to only show the columns which have whitespaces

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -0,0 +1,97 @@
...
+            AS num_duplicates
+    FROM iasworld.owndat AS owndat
+    LEFT JOIN iasworld.legdat AS legdat
+        ON owndat.parid = legdat.parid
+        AND owndat.taxyr = legdat.taxyr
+        AND legdat.cur = 'Y'
+        AND legdat.deactivat IS NULL
+    LEFT JOIN (
+        SELECT DISTINCT parid
+        FROM iasworld.pardat
```
Damonamajor: Do we want taxyr for this join?

dbt/macros/tests/test_all.sql
jeancochrane: **[Thought, non-blocking]** The way our macro unit tests work is that we define tests for each macro in a dedicated file (in this case, `dbt/macros/tests/test_generate_iasworld_qc_test_view.sql`), then in that file we define a top-level test that runs all the macro tests (in this case, `test_generate_iasworld_qc_test_view`), and then we call that top-level test in the context of the `test_all` test here so that it will run all of the unit tests for the macro when we run `dbt run-operation test_all`.

Does that make sense? It's a bit of a convoluted system, but we wrote it that way to work around the fact that dbt does not have a built-in macro unit testing system that implements test discovery. If we were writing Python unit tests in pytest or R unit tests in testthat, we would just be able to use a file naming pattern like `tests/test_*.py` to mark a file as including tests, and the test framework would pick it up. But here we have to do all of that work ourselves.

dbt/models/qc/docs.md
```diff
@@ -202,6 +202,27 @@ classes.
 This view is not currently in use in any QC workflows.
 {% enddocs %}
 
+# vw_report_iasworld_test_all
+
+{% docs view_vw_report_iasworld_test_all %}
+Combines test failures from all iasworld test views. Each observation is a failure
+based on unique PIN, year, and failing test combinations.
+{% enddocs %}
```
jeancochrane, line 213: **[Nitpick, optional]** Let's make sure this and the rest of the docs blocks for our new views include the "primary key" attribute that we list in docs for all of our tables. You're already explaining the primary key fields in the body, which is good, but I think the attribute is also helpful for skimming.

```suggestion

**Primary Key**: `parid`, `taxyr`, `test_name`
{% enddocs %}
```

dbt/models/qc/qc.vw_report_iasworld_test_all.sql
jeancochrane: **[Suggestion, optional]** Yup, good call. We name this column `source_table` in the current version of the workbooks, and it should refer to the iasWorld table name, e.g. `"pardat"` for this particular view (`qc.vw_report_iasworld_test_pardat`).

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -0,0 +1,97 @@
...
+    "cityname", "statecode", "zip1", "zip2", "user27"
+] -%}
+
+{%- set whitespace_parts = [] -%}
+{%- for col in whitespace_cols -%}
+    {%- do whitespace_parts.append(
+        "(" ~ col ~ " LIKE '% ' OR " ~ col ~ " LIKE ' %')"
+    ) -%}
+{%- endfor -%}
+
```
jeancochrane: **[Thought, non-blocking]** If we end up using this logic for testing whitespace in other tables, it might make sense to factor it out into a macro so we can reuse it. No need to do that now, however; we can do it later if we run into more examples of this type of whitespace test.

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
jeancochrane: **[Suggestion, blocking]** In theory yes, though I think the existing test that ensures `owndat` PINs are present in `pardat` is complicated by the fact that the iasWorld data model now treats `pardat` as the authoritative source for PINs, since other tables can include deactivated PINs that are being kept around for the purpose of collecting tax bills. I think it probably makes sense for us to skip migrating any tests that check to make sure PINs are present in `pardat`, and instead use `pardat` as the basis for all of these tables, e.g. inner join `owndat` to `pardat` in this query. Does that seem right to you @wrridgeway?

jeancochrane (changes requested):
Nice work! Just a couple suggested changes here, none of them serious. I'll take one final look before we merge.

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -19,12 +19,6 @@
         "condition": "cur IN ('Y', 'D')",
         "additional_select_columns": ["cur"]
     },
-    {
-        "name": "iasworld_owndat_parid_in_pardat_parid",
-        "description": "parid should be in pardat",
```
Damonamajor: This test should be removed, right?

jeancochrane: **[Praise]** Yup, that's right!

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -0,0 +1,93 @@
...
+        "name": "iasworld_owndat_cur_in_accepted_values",
+        "description": 'cur should be "Y" or "D"',
+        "category": "incorrect_values",
+        "condition": "cur IN ('Y', 'D')",
+        "additional_select_columns": ["cur"]
+    },
+    {
+        "name": "iasworld_owndat_parid_not_null",
+        "description": "parid should not be null",
+        "category": "incorrect_values",
```
jeancochrane: **[Nitpick, optional]** Oops, I totally forgot that the `run_iasworld_data_tests.py` script defines a default for some types of tests so that they don't need to specify a category. Those defaults indicate that "not null" tests should map to `"missing_values"`:

https://github.com/ccao-data/data-architecture/blob/9bd81870e6b58c791b693a4e3068500fd58d16d7/dbt/scripts/run_iasworld_data_tests.py#L59-L69

```suggestion
        "category": "missing_values",
```

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -0,0 +1,93 @@
...
+        "condition": "seq = prev_seq + 1",
+        "additional_select_columns": ["seq", "prev_seq"]
+    },
+    {
+        "name": "iasworld_owndat_address_columns_no_extra_whitespace",
+        "description": "address columns should not have whitespace",
+        "category": "column_values",
+        "condition": "NOT (" ~ whitespace_parts | join(" OR ") ~ ")",
+        "additional_select_columns": whitespace_cols
+    },
```
jeancochrane: **[Suggestion, optional]** I just noticed that we are excluding this test from the workbook due to the high volume of errors, so we should probably skip migrating this test to the view:

https://github.com/ccao-data/data-architecture/blob/9bd81870e6b58c791b693a4e3068500fd58d16d7/dbt/models/iasworld/schema/iasworld.owndat.yml#L281-L283

If we remove the test, I think we should also remove all the supporting columns and Jinja template variables, for the sake of cleanliness.

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -0,0 +1,93 @@
...
+    {
+        "name": "iasworld_owndat_unique_by_parid_taxyr",
+        "description": "owndat should be unique by parid and taxyr",
+        "category": "duplicate_rows",
```
jeancochrane: **[Nitpick, required]** My bad for not noticing sooner, but this category should be `"duplicate_records"`, per the existing map:

https://github.com/ccao-data/data-architecture/blob/9bd81870e6b58c791b693a4e3068500fd58d16d7/dbt/scripts/run_iasworld_data_tests.py#L65

```suggestion
        "category": "duplicate_records",
```

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -0,0 +1,93 @@
...
+            OVER (PARTITION BY owndat.parid, owndat.taxyr ORDER BY owndat.seq)
+            AS prev_seq,
+        {% for col in whitespace_cols %}
+        CASE WHEN owndat.{{ col }} LIKE '% ' OR owndat.{{ col }} LIKE ' %'
+            THEN owndat.{{ col }} END AS {{ col }},
+        {% endfor %}
+        COUNT(*)
+            OVER (PARTITION BY owndat.parid, owndat.taxyr)
+            AS num_duplicates
+    FROM iasworld.owndat AS owndat
```
jeancochrane: **[Nitpick, required]** I totally forgot that we should be using [the dbt `source()` function](https://docs.getdbt.com/reference/dbt-jinja-functions/source?version=1.11) to refer to iasWorld tables, so that dbt can infer the relationships between these new views and the underlying iasWorld tables that they pull from:
 
```suggestion
    FROM {{ source('iasworld', 'owndat') }} AS owndat
```

We'll also want to make a similar change to the rest of the joins in these two new views.

dbt/models/qc/qc.vw_report_iasworld_test_owndat.sql
```diff
@@ -0,0 +1,93 @@
...
+        AND legdat.cur = 'Y'
+        AND legdat.deactivat IS NULL
+    INNER JOIN (
+        SELECT DISTINCT parid, taxyr
+        FROM iasworld.pardat
+        WHERE cur = 'Y'
+            AND deactivat IS NULL
+    ) AS pardat_parids
+        ON owndat.parid = pardat_parids.parid
+        AND owndat.taxyr = pardat_parids.taxyr
```
jeancochrane: **[Suggestion, optional]** No need for the `DISTINCT` subquery here I don't think, so we can keep it simple:

```suggestion
    INNER JOIN {{ source('iasworld', 'pardat') }} AS pardat
        ON owndat.parid = pardat.parid
        AND owndat.taxyr = pardat.taxyr
        AND pardat.cur = 'Y'
        AND pardat.deactivat IS NULL
```

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,92 @@
...
+        "condition": "length(adrno) <= 5",
+        "additional_select_columns": ["adrno"]
+    },
+    {
+        "name": "iasworld_pardat_class_equals_luc",
+        "description": "class should be the same as luc",
+        "category": "class_mismatch_or_issue",
+        "condition": "class = luc",
+        "additional_select_columns": ["luc"]
+    },
```
jeancochrane, line 15: **[Nitpick, required]** Not sure how this got lost (sorry if we discussed it and we forgot), but it seems like we missed the `iasworld_pardat_class_in_ccao_class_dict` test during this migration:

https://github.com/ccao-data/data-architecture/blob/9bd81870e6b58c791b693a4e3068500fd58d16d7/dbt/models/iasworld/schema/iasworld.pardat.yml#L95-L113

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,92 @@
...
+        "name": "iasworld_pardat_seq_all_sequential_exist",
+        "description": "seq should be sequential",
+        "category": "incorrect_values",
+        "condition": "seq = prev_seq + 1",
+        "additional_select_columns": ["seq", "prev_seq"]
+    },
+    {
+        "name": "iasworld_pardat_unique_by_parid_taxyr",
+        "description": "pardat should be unique by parid and taxyr",
+        "category": "duplicate_rows",
```
jeancochrane: **[Nitpick, optional]** Same here as above:


```suggestion
        "category": "duplicate_records",
```

jeancochrane (changes requested):
Just a couple more changes!

Damonamajor:
Way to see one of each failing test

```
SELECT *
FROM (
  SELECT *,
         ROW_NUMBER() OVER (PARTITION BY test_name ORDER BY test_name) AS rn
  FROM z_ci_1088_create_tableau_dashboard_based_on_iasworld_tests_qc.vw_report_iasworld_test_all
) t
WHERE rn = 1
```

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,105 @@
...
+        AND legdat.deactivat IS NULL
+    LEFT JOIN (
+        SELECT DISTINCT town_nbhd
+        FROM {{ source('spatial', 'neighborhood') }}
+    ) AS nbhd
+        ON pardat.nbhd = nbhd.town_nbhd
+    LEFT JOIN (
+        SELECT DISTINCT class_code
+        FROM {{ ref('ccao.class_dict') }}
+    ) AS class_dict
```
jeancochrane: **[Suggestion, optional]** `ccao.class_dict` is already unique by class code ([docs](https://ccao-data.github.io/data-architecture/#!/seed/seed.ccao_data_athena.ccao.class_dict#description)), so the `DISTINCT` subquery here is superfluous:

```suggestion
    LEFT JOIN {{ ref('ccao.class_dict') }} AS class_dict
```

dbt/models/qc/qc.vw_report_iasworld_test_pardat.sql
```diff
@@ -0,0 +1,105 @@
...
+        "description": "class should be the same as luc",
+        "category": "class_mismatch_or_issue",
+        "condition": "class = luc",
+        "additional_select_columns": ["luc"]
+    },
+    {
+        "name": "iasworld_pardat_class_in_ccao_class_dict",
+        "description": "class_code should be valid",
+        "category": "class_mismatch_or_issue",
+        "condition": "class_dict_class IS NOT NULL",
```
jeancochrane: **[Nitpick, required]** We're missing a couple of conditions on the old version of the test that excluded certain classes:

https://github.com/ccao-data/data-architecture/blob/ee49d815424379a23f345464fd7f9f1e5d2ceded/dbt/models/iasworld/schema/iasworld.pardat.yml#L105-L110

```suggestion
        "condition": "class_dict_class IS NOT NULL OR class IN ('EX', 'RR') OR REGEXP_LIKE(class, '[0-9]{3}[A|B]')",
```

Based on my testing, this is the only test in the view that is not matching the YAML definition of the test.

jeancochrane (commented):
This is very close! According to my testing, just one test in the view that needs to be adjusted to match the existing YAML test. Here are the commands I ran for testing, mostly just saving these here so I can run them again before approving:

```bash
dbt test --select iasworld_pardat_adrno_length_lte_5 iasworld_pardat_class_equals_luc iasworld_pardat_class_in_ccao_class_dict iasworld_pardat_cur_in_accepted_values iasworld_pardat_nbhd_matches_legdat_township iasworld_pardat_nbhd_matches_spatial_town_nbhd iasworld_pardat_seq_all_sequential_exist iasworld_pardat_unique_by_parid_taxyr iasworld_owndat_cur_in_accepted_values iasworld_owndat_parid_not_null iasworld_owndat_seq_all_sequential_exist iasworld_owndat_unique_by_parid_taxyr --store-failures
```

```sql
WITH view_rows AS (
    FROM z_ci_1088_create_tableau_dashboard_based_on_iasworld_tests_qc.vw_report_iasworld_test_all
),
test_failure_rows AS (
    SELECT 'iasworld_pardat_adrno_length_lte_5' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_adrno_length_lte_5
    UNION ALL
    SELECT 'iasworld_pardat_class_equals_luc' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_class_equals_luc
    UNION ALL
    SELECT 'iasworld_pardat_class_in_ccao_class_dict' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_class_in_ccao_class_dict
    UNION ALL
    SELECT 'iasworld_pardat_cur_in_accepted_values' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_cur_in_accepted_values
    UNION ALL
    SELECT 'iasworld_pardat_nbhd_matches_legdat_township' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_nbhd_matches_legdat_township
    UNION ALL
    SELECT 'iasworld_pardat_nbhd_matches_spatial_town_nbhd' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_nbhd_matches_spatial_town_nbhd
    UNION ALL
    SELECT 'iasworld_pardat_seq_all_sequential_exist' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_seq_all_sequential_exist
    UNION ALL
    SELECT 'iasworld_pardat_unique_by_parid_taxyr' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_pardat_unique_by_parid_taxyr
    UNION ALL
    SELECT 'iasworld_owndat_cur_in_accepted_values' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_owndat_cur_in_accepted_values
    UNION ALL
    SELECT 'iasworld_owndat_parid_not_null' as test_name, null as parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_owndat_parid_not_null
    UNION ALL
    SELECT 'iasworld_owndat_seq_all_sequential_exist' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_owndat_seq_all_sequential_exist
    UNION ALL
    SELECT 'iasworld_owndat_unique_by_parid_taxyr' as test_name, parid, taxyr
    FROM z_dev_jecochr_test_failure.iasworld_owndat_unique_by_parid_taxyr
)
SELECT
    COALESCE(v.parid, f.parid) AS parid,
    COALESCE(v.taxyr, f.taxyr) AS taxyr,
    CASE
        WHEN v.parid IS NULL THEN 'only in old failures'
        WHEN f.parid IS NULL THEN 'only in new failures'
    END AS source
FROM view_rows v
    ON v.test_name = f.test_name
    AND v.parid = f.parid
    AND v.taxyr = f.taxyr
WHERE v.parid IS NULL OR f.parid IS NULL
ORDER BY parid, test_name
```

jeancochrane (approved):
Looks good to me! I'm excited for more tests in the new view.

PR #1094, opened 2026-08-19
Update `test_open_data_assets.py` script to handle historical mismatches

### LINKED ISSUE #1027: Update `test_open_data_assets.py` script to handle historical mismatches
The `test-open-data-assets` workflow fails basically every time due to a quirk in our data upload schedule: We only update historical (prior year) values once a year, and we only update the most recent year each month, but we test all years every month. This means that historical years will continue to fail throughout the year as our staff make small edits to historical years ([example](https://github.com/ccao-data/data-architecture/actions/runs/25244720517/job/74026878053)).

Two options we have for handling this:

1. Change historical tests to a percentage threshold (e.g. within 5% different instead of exactly matching)
2. Only test historical values once per year, when we update them

Or maybe a combination of these two approaches.

wrridgeway:
This PR adds an additional buffer level for prior years (earlier than the current and last) that can be adjusted independently of the current year buffer. It also formats our `agate` tables before output to avoid the comma in the year column and number columns with values of only 0/1 being coerced into booleans.

These changes were motivated by drift between row counts across less-frequently updated, older years of data on the open data portal and the athena assets that feed them. We should expect and thus allow larger discrepancies for these years since they're only updated once a year.

[Test run here.](https://github.com/ccao-data/data-architecture/actions/runs/32288770397/job/96184485930)

This will still surface errors for 1 vs 0 rows, which is a pretty small _absolute_ difference. I'm happy to try to address that case if we think it should also be allowed to pass through the buffer.

I also updated our sales asset after seeing the diff from the test run.

### DIFF
--- a/.github/scripts/test_open_data_assets.py
+++ b/.github/scripts/test_open_data_assets.py
@@ -53,7 +53,7 @@ REQUESTS.mount(
 # (which is updated bi-weekly or monthly). We expect differences to arise for
 # prior years due to our slower open data update cadence for those years (once
 # a year) despite their relatively static nature in Athena.
-BUFFERS = {"current": 0.02, "prior": 0.2}
+BUFFERS = {"current": 0.02, "prior": 0.1}
 
 
 def main() -> None:
@@ -159,8 +159,11 @@ def main() -> None:
             "by year:"
         )
 
+        # Declaring column types for agate helps make sure number columns with
+        # values of only 0/1 aren't coerced into booleans and makes printing
+        # our year column cleaner.
         diff_column_types = {
-            DEFAULT_YEAR_FIELD: agate.Text(),
+            "year": agate.Text(),
             source_model_name: agate.Number(),
             asset_name: agate.Number(),
         }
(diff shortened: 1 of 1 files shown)


.github/scripts/test_open_data_assets.py
```diff
@@ -47,11 +47,13 @@
...
-# change in the Athena data (which is updated daily) compared to the Open Data
-# portal (which is updated bi-weekly or monthly).
-BUFFER = 0.02
+# Data is permitted to be slightly different, according to these buffer values.
+# This is because we expect some level of change for the most current two years
+# in the Athena data (which is updated daily) compared to the Open Data portal
+# (which is updated bi-weekly or monthly). We expect differences to arise for
+# prior years due to our slower open data update cadence for those years (once
+# a year) despite their relatively static nature in Athena.
+BUFFERS = {"current": 0.02, "prior": 0.2}
```
wrridgeway, line 56: 0.2 for prior years is arbitrary. I tried a number of different values and this was the lowest I could go without triggering any errors. If we go down to 0.1 we catch differences between 8 and 9 rows.

jeancochrane: **[Thought, non-blocking]** Sounds good! Note that we might end up needing to bump the `prior` threshold as the year goes on, since I expect the number of errors will continue increasing. That seems like a reasonable tradeoff for now, though.

jeancochrane (approved):
Thanks for this!

PR #1098, opened 2026-08-24
Create final model views for assessment and training data

wrridgeway:
We join final model values into a few different reporting views and regularly output them for EI issues. It's a common join that can be a little treacherous to remember and I think these views could make working with final model values more ergonomic.

I've created three new views for `assessment_pin`, `assessment_card`, and `training_data` that limit them to final model runs, and for assessment data the triad that was being reassessed. I've added tests to make sure this is happening as expected.

I think these views could be particularly useful if we ever decide to start running the model on a township-by-township cadence.

Edit: the `training_data` view was removed from this PR.


dbt/models/model/docs.md
```diff
@@ -258,3 +278,14 @@ View to compile PIN-level model inputs shared between the residential
 
 **Primary Key**: `year`, `meta_pin`
 {% enddocs %}
+
+# vw_training_data_final
+
+{% docs view_vw_training_data_final %}
+
+Version of `training_data` limited to final model runs. Includes additional
```
jeancochrane, line 286: **[Suggestion, optional]** `model.training_data` is actually already limited to final model runs -- we chose to design it that way because we were nervous about the amount of data the table would contain if it included all model runs, especially since training data rarely changes between model runs in a given year. Here's the relevant code block:

https://github.com/ccao-data/data-architecture/blob/9bd81870e6b58c791b693a4e3068500fd58d16d7/dbt/models/model/model.training_data.py#L14-L27

All that being said, I think it's become clear that `training_data` is a confusing table name that many of us have misinterpreted. What if instead of creating a new view, we just rename the existing table to `model.training_data_final`? If we do this, we'll probably also want to do a search around our repos to find all of the references we need to change, so I'm fine punting that to a follow-up PR too.

jeancochrane (approved):
Looks good to me, other than one note about training data!

dbt/models/model/docs.md
wrridgeway: I think it's probably just fine to leave it as is, we don't need to re-architect because I'm too lazy to read code and docs. We can if you think it's best, but i don't think it's urgent.

jeancochrane: I think it's a good idea, though I agree that it's not urgent! I opened up https://github.com/ccao-data/data-architecture/issues/1099 and popped it in next sprint, we can discuss at sprint planning.

jeancochrane:
One thing I realized while thinking more about this PR: We also have the `model.parameter_final` table, which uses the suffix `_final` to refer to the fact that the parameter is the end result of a CV loop, not to refer to the fact that the table contains results for final models. That's slightly different from this PR, which future maintainers might find confusing. Not a reason to stop this PR, but something for us to be aware of, and perhaps something that merits a note in the docs for `model.parameter_final`.

PR #1102, opened 2026-09-01
Add iasworld.sale to qc view

Damonamajor:
Follow-up to https://github.com/ccao-data/data-architecture/pull/1093

https://github.com/ccao-data/data-architecture/blob/master/dbt/models/iasworld/schema/iasworld.sales.yml

<details><summary>Test returns 0 mismatches:</summary>
<p>

WITH view_rows AS (
    FROM z_dev_damajor_qc.vw_report_iasworld_test_sales
),
test_failure_rows AS (
    SELECT 'iasworld_sales_cur_in_accepted_values' AS test_name, parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_sales_cur_in_accepted_values
    UNION ALL
    SELECT 'iasworld_sales_instrtyp_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_sales_instrtyp_in_accepted_values
    UNION ALL
    SELECT 'iasworld_sales_parid_in_pardat_parid', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_sales_parid_in_pardat_parid
    UNION ALL
    SELECT 'iasworld_sales_parid_not_null', CAST(NULL AS VARCHAR) AS parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_sales_parid_not_null
    UNION ALL
    SELECT 'iasworld_sales_price_between_0_and_1b', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_sales_price_between_0_and_1b
    UNION ALL
    SELECT 'iasworld_sales_saledt_lte_now', parid, SUBSTR(saledt, 1, 4) AS taxyr
    FROM z_dev_damajor_test_failure.iasworld_sales_saledt_lte_now
    UNION ALL
    SELECT 'iasworld_sales_unique_by_parid_instruno', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_sales_unique_by_parid_instruno
),
mismatches AS (
    SELECT
        CASE
            WHEN v.taxyr IS NULL THEN 'only in test failure'
            WHEN f.taxyr IS NULL THEN 'only in qc view'
        END AS source
    FROM view_rows v
        ON v.test_name = f.test_name
        AND v.parid = f.parid
        AND v.taxyr = f.taxyr
    WHERE v.taxyr IS NULL OR f.taxyr IS NULL
)
SELECT
    test_name,
    source,
    COUNT(*) AS n
FROM mismatches
GROUP BY test_name, source
ORDER BY test_name, source

</p>
</details>

### DIFF
--- a/dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
@@ -21,7 +21,7 @@
     },
     {
         "name": "iasworld_sales_parid_in_pardat_parid",
-        "description": "parid should be in pardat",
+        "description": "parid should be in pardat for year of sale",
         "category": "relationships",
         "condition": "pardat_parid IS NOT NULL"
     },
@@ -52,7 +52,9 @@
     SELECT
         -- Identifying columns
         sales.parid,
-        SUBSTR(sales.saledt, 1, 4) AS taxyr,
+        MAX(SUBSTR(sales.saledt, 1, 4))
+            OVER (PARTITION BY sales.parid, sales.instruno)
+            AS taxyr,
         CAST(NULL AS INTEGER) AS card,
         CAST(NULL AS INTEGER) AS lline,
         legdat.user1 AS township_code,
@@ -75,13 +77,18 @@
         AND SUBSTR(sales.saledt, 1, 4) = legdat.taxyr
         AND legdat.cur = 'Y'
         AND legdat.deactivat IS NULL
-    LEFT JOIN {{ source('iasworld', 'pardat') }} AS pardat
+    LEFT JOIN (
+        SELECT parid, class,
+            ROW_NUMBER() OVER (PARTITION BY parid ORDER BY taxyr DESC) AS rn
+        FROM {{ source('iasworld', 'pardat') }}
+        WHERE cur = 'Y'
+            AND deactivat IS NULL
+    ) AS pardat
         ON sales.parid = pardat.parid
-        AND SUBSTR(sales.saledt, 1, 4) = pardat.taxyr
-        AND pardat.cur = 'Y'
-        AND pardat.deactivat IS NULL
+        AND pardat.rn = 1
     WHERE sales.cur = 'Y'
         AND sales.deactivat IS NULL
+        AND SUBSTR(sales.saledt, 1, 4) >= '2011'
 {% endset %}
 
-{{ generate_iasworld_qc_test_view(base_query, tests) }}
+{{ generate_iasworld_qc_test_view(base_query, tests, start_year='2011') }}
--- a/dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
@@ -88,7 +88,6 @@
         AND pardat.rn = 1
     WHERE sales.cur = 'Y'
         AND sales.deactivat IS NULL
-        AND SUBSTR(sales.saledt, 1, 4) >= '2011'
 {% endset %}
 
 {{ generate_iasworld_qc_test_view(base_query, tests, start_year='2011') }}
--- a/dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
@@ -1,0 +1,87 @@
+{%- set tests = [
+    {
+        "name": "iasworld_sales_cur_in_accepted_values",
+        "description": 'cur should be "Y" or "D"',
+        "category": "incorrect_values",
+        "condition": "cur IN ('Y', 'D')",
+        "additional_select_columns": ["cur"]
+    },
+    {
+        "name": "iasworld_sales_instrtyp_in_accepted_values",
+        "description": "instrtyp should be '01', '02', '03', '04', '05', '06', or 'B'",
+        "category": "incorrect_values",
+        "condition": "instrtyp IN ('01', '02', '03', '04', '05', '06', 'B')",
+        "additional_select_columns": ["instrtyp"]
+    },
+    {
+        "name": "iasworld_sales_parid_not_null",
+        "description": "parid should not be null",
+        "category": "missing_values",
+        "condition": "parid IS NOT NULL"
+    },
+    {
+        "name": "iasworld_sales_parid_in_pardat_parid",
+        "description": "parid should be in pardat",
+        "category": "relationships",
+        "condition": "pardat_parid IS NOT NULL"
+    },
+    {
+        "name": "iasworld_sales_price_between_0_and_1b",
+        "description": "price should be between 0 and 1,000,000,000",
+        "category": "incorrect_values",
+        "condition": "price IS NULL OR (price >= 0 AND price <= 1000000000)",
+        "additional_select_columns": ["price"]
+    },
+    {
+        "name": "iasworld_sales_saledt_lte_now",
+        "description": "saledt should be before the current date",
+        "category": "incorrect_values",
+        "condition": "saledt IS NULL OR DATE_PARSE(SUBSTR(saledt, 1, 10), '%Y-%m-%d') <= current_date",
+        "additional_select_columns": ["saledt"]
+    },
+    {
+        "name": "iasworld_sales_unique_by_parid_instruno",
+        "description": "sales should be unique by parid and instruno",
+        "category": "duplicate_records",
+        "condition": "num_duplicates = 1",
+        "additional_select_columns": ["num_duplicates", "instruno"]
+    }
+] -%}
+
+{%- set base_query %}
+    SELECT
+        -- Identifying columns
+        sales.parid,
+        SUBSTR(sales.saledt, 1, 4) AS taxyr,
+        CAST(NULL AS INTEGER) AS card,
+        CAST(NULL AS INTEGER) AS lline,
+        legdat.user1 AS township_code,
+        pardat.class,
+        sales.who,
+        sales.wen,
+        -- Columns to test
+        sales.cur,
+        sales.instrtyp,
+        sales.price,
+        sales.saledt,
+        sales.instruno,
+        pardat.parid AS pardat_parid,
+        COUNT(*)
+            OVER (PARTITION BY sales.parid, sales.instruno)
+            AS num_duplicates
+    FROM {{ source('iasworld', 'sales') }} AS sales
+    LEFT JOIN {{ source('iasworld', 'legdat') }} AS legdat
+        ON sales.parid = legdat.parid
+        AND SUBSTR(sales.saledt, 1, 4) = legdat.taxyr
+        AND legdat.cur = 'Y'
+        AND legdat.deactivat IS NULL
+    LEFT JOIN {{ source('iasworld', 'pardat') }} AS pardat
+        ON sales.parid = pardat.parid
+        AND SUBSTR(sales.saledt, 1, 4) = pardat.taxyr
+        AND pardat.cur = 'Y'
+        AND pardat.deactivat IS NULL
+    WHERE sales.cur = 'Y'
+        AND sales.deactivat IS NULL
+{% endset %}
+
+{{ generate_iasworld_qc_test_view(base_query, tests) }}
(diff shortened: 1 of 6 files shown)


Damonamajor:
@jeancochrane 

This is a first attempt to add iasworld.sale to the qc view. For the most part, we want to look at only the current year of data, but for this view we want to look at historical sales. This makes me believe that we will want to modify this, but I may be wrong.

https://github.com/ccao-data/data-architecture/blob/master/dbt/dbt_project.yml#L32

Let me know what you think, it returns this in the full query, which I'm not sure is what we want.

```
            from base
            where
                taxyr
                between '2026'
                and '2030'
```

jeancochrane:
> For the most part, we want to look at only the current year of data, but for this view we want to look at historical sales. This makes me believe that we will want to modify [the built-in date filter in `generate_iasworld_qc_test_view`], but I may be wrong.

Yup, good point! My suggestion would be to modify the `generate_iasworld_qc_test_view()` macro so that it accepts two new optional arguments, `start_year` and `end_year`, and have it default to the `data_test_iasworld_year_*` variables for the corresponding argument whenever the argument is missing. That way we could configure the `qc.vw_report_iasworld_test_sales` view with `start_year = '2011'` with minimal disruption to other views that rely on `generate_iasworld_qc_test_view()`.

What do you think? I'm open to other designs if you have a different proposal.

Damonamajor:
> > For the most part, we want to look at only the current year of data, but for this view we want to look at historical sales. This makes me believe that we will want to modify [the built-in date filter in `generate_iasworld_qc_test_view`], but I may be wrong.
> 
> Yup, good point! My suggestion would be to modify the `generate_iasworld_qc_test_view()` macro so that it accepts two new optional arguments, `start_year` and `end_year`, and have it default to the `data_test_iasworld_year_*` variables for the corresponding argument whenever the argument is missing. That way we could configure the `qc.vw_report_iasworld_test_sales` view with `start_year = '2011'` with minimal disruption to other views that rely on `generate_iasworld_qc_test_view()`.
> 
> What do you think? I'm open to other designs if you have a different proposal.

Yup, that was more or les what I was going to do.

dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
```diff
@@ -0,0 +1,94 @@
...
+        COUNT(*)
+            OVER (PARTITION BY sales.parid, sales.instruno)
+            AS num_duplicates
+    FROM {{ source('iasworld', 'sales') }} AS sales
+    LEFT JOIN {{ source('iasworld', 'legdat') }} AS legdat
+        ON sales.parid = legdat.parid
+        AND SUBSTR(sales.saledt, 1, 4) = legdat.taxyr
+        AND legdat.cur = 'Y'
+        AND legdat.deactivat IS NULL
+    LEFT JOIN (
```
Damonamajor, line 78: We talked briefly about this join. This is how we return 0 mismatches in the PR header so I"ll leave it like this for now. We could also join it based on taxyr so that the PIN needs to be present in the year of the sale. I'll leave that up to you though.

dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
```diff
@@ -0,0 +1,94 @@
...
+            ROW_NUMBER() OVER (PARTITION BY parid ORDER BY taxyr DESC) AS rn
+        FROM {{ source('iasworld', 'pardat') }}
+        WHERE cur = 'Y'
+            AND deactivat IS NULL
+    ) AS pardat
+        ON sales.parid = pardat.parid
+        AND pardat.rn = 1
+    WHERE sales.cur = 'Y'
+        AND sales.deactivat IS NULL
+        AND SUBSTR(sales.saledt, 1, 4) >= '2011'
```
jeancochrane: **[Nitpick, required]** I believe this line is duplicative with setting `start_year='2011'` in the `generate_iasworld_qc_test_view()` call below. If that's right, we should remove it so that we don't set ourselves up for a mistake if we ever try to update `start_year` without changing this line too.


```suggestion
```

dbt/models/qc/schema.yml
```diff
@@ -8,6 +8,12 @@ models:
   - name: qc.vw_report_iasworld_test_pardat
     description: '{{ doc("view_vw_report_iasworld_test_pardat") }}'
 
+  - name: qc.vw_report_iasworld_test_sales
+    description: '{{ doc("view_vw_report_iasworld_test_sales") }}'
+    config:
+      tags:
+        - data_test_iasworld
```
jeancochrane: **[Nitpick, required]** We use this tag to mark iasWorld tests for our dbt selector as part of the legacy iasWorld test process. The new view is separate from that legacy process, so we don't need the tag.

```suggestion
```

dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
jeancochrane: **[Suggestion, optional]** Hmm, yeah, this is tricky. I don't feel confident enough in my assumption that `pardat.taxyr` should match `sales.saledt`, and the fact that there are so many failures when you join by `taxyr` (100k+) makes me think that it can be normal for there to be no corresponding `pardat` record for a sale due to e.g. divisions. So I think it's reasonable not to join by `taxyr`.

Some further confirmation of this approach: The current version of the `iasworld_sales_parid_in_pardat_parid` test uses the `relationships` generic test, which basically just checks to make sure the `parid` exists in `pardat` at all, regardless of the year:

https://github.com/ccao-data/data-architecture/blob/a68185f3065100a67c5505d6d63f0963a7a1212d/dbt/models/iasworld/schema/iasworld.sales.yml#L116-L127

https://github.com/ccao-data/data-architecture/blob/a68185f3065100a67c5505d6d63f0963a7a1212d/dbt/tests/generic/test_relationships.sql#L3-L24

So I think it's reasonable to exclude `taxyr` when we migrate the test to the new view.

One suggestion, however: I don't think any of the legacy versions of these tests actually select `pardat.class`. The `ROW_NUMBER()` window function that is currently allowing this view to select the most recent `pardat` class is much slower than a simple `DISTINCT`, so in the interest of keeping things as snappy as possible, I think we should drop the class and implement a faster `pardat` join:

```sql
LEFT JOIN (
    SELECT DISTINCT parid
    FROM {{ source('iasworld', 'pardat') }}
    WHERE cur = 'Y'
        AND deactivat IS NULL
) AS pardat
````

File: dbt/models/qc/qc.vw_report_iasworld_test_sales.sql [hunk 4]
```diff
@@ -0,0 +1,94 @@
...
+    }
+] -%}
+
+{%- set base_query %}
+    SELECT
+        -- Identifying columns
+        sales.parid,
+        MAX(SUBSTR(sales.saledt, 1, 4))
+            OVER (PARTITION BY sales.parid, sales.instruno)
+            AS taxyr,
```
jeancochrane: **[Question, non-blocking]** Why choose the most recent sale date for this doc number, instead of just the sale date for the individual record? It seems like this might make it hard to track down the individual sale that is failing a specific test, especially since we're not including doc number in the output.

jeancochrane (commented):
Getting close! A few more nits and suggestions below.

dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
Damonamajor: I think we used this here:
https://github.com/ccao-data/data-architecture/blob/master/dbt/models/iasworld/schema/iasworld.sales.yml#L226C5-L228C32

dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
Damonamajor: I think the 100K + observations are duplicates. I got maybe 10,000 more when deduplicating before the join.

dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
Damonamajor: yup, it results in two mismatches, which I'm not exactly able to trace down the reasoning of.

dbt/models/qc/qc.vw_report_iasworld_test_sales.sql
Damonamajor: And the doc_no is in the additional_select column.

Damonamajor: This is the results when not using max

image

image

jeancochrane: Our use of `max()` in the `additional_select_columns` for the uniqueness test you linked above is actually slightly different than the use here. The `max()` aggregation function is required for that use of `additional_select_columns` because the `unique_combination_of_columns` test generic groups its underlying query by the specified columns, so we can't select additional columns without aggregating them:

https://github.com/ccao-data/data-architecture/blob/a68185f3065100a67c5505d6d63f0963a7a1212d/dbt/tests/generic/test_unique_combination_of_columns.sql#L31

Since we're using `max()` via a window function rather than a `group by` clause here, the output will not be grouped at all, meaning it should contain the same number of rows regardless of whether we select `taxyr` using a `max()` window function or just select it directly. We can confirm this with a simplified query:

```sql
select 'has_max' as type, count(*) as num_rows
from (
    select max(substr(saledt, 1, 4)) over (partition by parid, instruno) as taxyr
    from iasworld.sales
    where cur = 'Y' and deactivat is null
) as has_max
union all
select 'no_max' as type, count(*) as num_rows
from (
    select substr(saledt, 1, 4) as taxyr
    from iasworld.sales
    where cur = 'Y' and deactivat is null
) as no_max
```

| type | num_rows |
| --- | --- | 
| no_max | 3179623 |
| has_max | 3179623 |

Ultimately I think it would be simpler to just select `substr(saledt, 1, 4) as taxyr` here, but it's not a big deal either way, so I leave it up to you. I doubt that there are many cases where the sale year differs when grouped by PIN and doc number anyway.

jeancochrane (approved):
Looks great! Comparison to the existing tests only shows a mismatch in the uniqueness test:

```sql
with view_counts as (
    select test_name, count(*) as cnt
    FROM z_ci_add_iasworldsale_to_qc_view_qc.vw_report_iasworld_test_sales
    group by 1
),

test_failure_rows as (
	SELECT 'iasworld_sales_cur_in_accepted_values' as test_name, parid, taxyr
	FROM z_dev_jecochr_test_failure.iasworld_sales_cur_in_accepted_values
	UNION ALL
	SELECT 'iasworld_sales_instrtyp_in_accepted_values' as test_name, parid, taxyr
	FROM z_dev_jecochr_test_failure.iasworld_sales_instrtyp_in_accepted_values
	UNION ALL
	SELECT 'iasworld_sales_parid_not_null' as test_name, null as parid, taxyr
	FROM z_dev_jecochr_test_failure.iasworld_sales_parid_not_null
	UNION ALL
	SELECT 'iasworld_sales_parid_in_pardat_parid' as test_name, parid, taxyr
	FROM z_dev_jecochr_test_failure.iasworld_sales_parid_in_pardat_parid
	UNION ALL
	SELECT 'iasworld_sales_price_between_0_and_1b' as test_name, parid, taxyr
	FROM z_dev_jecochr_test_failure.iasworld_sales_price_between_0_and_1b
	UNION ALL
	SELECT 'iasworld_sales_saledt_lte_now' as test_name, parid, substr(saledt, 1, 4) as taxyr
	FROM z_dev_jecochr_test_failure.iasworld_sales_saledt_lte_now
	UNION ALL
	SELECT 'iasworld_sales_unique_by_parid_instruno' as test_name, parid, taxyr
	FROM z_dev_jecochr_test_failure.iasworld_sales_unique_by_parid_instruno
),

test_failure_counts as (
    select test_name, count(*) as cnt
    from test_failure_rows
    group by 1
)

select coalesce(v.test_name, t.test_name) as test_name, t.cnt as test_failure_count, v.cnt as view_count
from view_counts v
full outer join test_failure_counts t
    on v.test_name = t.test_name
```

| test_name | test_failure_count | view_count |
| --- | --- | --: |
| iasworld_sales_instrtyp_in_accepted_values | 10 | 10 |
| iasworld_sales_unique_by_parid_instruno | 17883 | 35849 |
| iasworld_sales_parid_in_pardat_parid | 2484 | 2484 |

However, as you point out, the mismatch is only due to the fact that the existing test is able to group by PIN and doc number, while this view cannot. That's probably OK; if our stakeholders don't like it, we can just create another view specific to the uniqueness test.

PR #1105, opened 2026-09-11
Resolve dbt deprecation warnings

jeancochrane:
## Background

_Apologies in advance for this absolutely massive PR. There are very few _types_ of changes contained in this PR, but a huge number of files in which we need to make those changes. I've tried to keep the diff as small as possible but unfortunately it's still quite big._

https://github.com/ccao-data/data-architecture/pull/1104 upgrades our version of dbt to the latest version in the 1.x line, but that version is raising a number of deprecation warnings. Those warnings will be escalated to errors in dbt v2. This PR cleans up our dbt code to resolve those warnings and follow the latest dbt practices.

## Deprecation reference

These are the new deprecation warnings that we aim to resolve in this PR:

- [`ArgumentsPropertyInGenericTestDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#argumentspropertyingenerictestdeprecation): Arguments to generic tests now need to be nested under an `arguments` key.
- [`PropertyMovedToConfigDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#propertymovedtoconfigdeprecation): Certain properties like `meta`, `freshness`, and `tags` now need to be nested under a `config` key.
- [`DuplicateYAMLKeysDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#duplicateyamlkeysdeprecation): Duplicate YAML keys will soon raise an error, rather than falling back to the most recent declaration in the group.

## Testing

The [dbt docs](https://docs.getdbt.com/reference/deprecations?version=1#dbt-cli) recommend running the command `dbt parse` with the `--show-all-deprecations` and `--no-partial-parse` flags enabled in order to show deprecation warnings.

Running this command on the branch behind #1104 shows 1k+ warnings:

```bash
$ git checkout jeancochrane/resolve-dependabot-alerts
$ dbt parse --show-all-deprecations --no-partial-parse
...
Summary of encountered deprecations:
- DuplicateYAMLKeysDeprecation: 160 occurrences
- MissingArgumentsPropertyInGenericTestDeprecation: 632 occurrences
- PropertyMovedToConfigDeprecation: 495 occurrences
```

Running that same command on this branch shows no warning output:

```bash
$ git checkout jeancochrane/resolve-dbt-deprecation-warnings
$ dbt parse --show-all-deprecations --no-partial-parse
20:16:11  Running with dbt=1.12.4
20:16:11  Registered adapter: athena=1.10.2
20:16:24  Performance info: ./target/perf_info.json
```

See also the workflow logs for the `build-and-test-dbt` job, which display no warnings during the build and test steps: https://github.com/ccao-data/data-architecture/actions/runs/34903880867/job/104175935151


dbt/models/default/schema/default.vw_pin_value.yml
```diff
@@ -238,7 +242,7 @@ unit_tests:
         # `class` is the important column here, and all other columns are only
         # provided to ensure proper joins when creating the dummy table
         rows:
-          - {parid: "123", taxyr: "2024", class: "200", procname: "CCAOVALUE", rolltype: "RP", valasm3: 10, class: "2.1-1)A"}
+          - {parid: "123", taxyr: "2024", procname: "CCAOVALUE", rolltype: "RP", valasm3: 10, class: "2.1-1)A"}
```
jeancochrane, line 245: This is an example of a fix for [`DuplicateYAMLKeysDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#duplicateyamlkeysdeprecation).

dbt/models/ccao/schema.yml
```diff
@@ -4,13 +4,15 @@ models:
 
 sources:
   - name: ccao
-    tags:
-      - load_manual
+    config:
+      tags:
+        - load_manual
```
jeancochrane, line 9: This is an example of a fix for [`PropertyMovedToConfigDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#propertymovedtoconfigdeprecation).

dbt/models/ccao/schema.yml
```diff
@@ -81,47 +83,55 @@ sources:
         data_tests:
           - unique_combination_of_columns:
               name: ccao_hidename_unique_by_pin
-              combination_of_columns:
-                - pin
+              arguments:
+                combination_of_columns:
+                  - pin
```
jeancochrane, line 88: This is an example of a fix for [`ArgumentsPropertyInGenericTestDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#argumentspropertyingenerictestdeprecation).

dbt/models/external/schema/external.vw_cmap.yml
```diff
@@ -1,139 +1,140 @@
...
-                AND dweldat.cur = 'Y'
-            )
-      - unique_combination_of_columns:
-          name: external_vw_cmap_unique_by_pin_card_and_year
-          combination_of_columns:
-            - pin
-            - year
-            - card
-
+models:
```
jeancochrane, line 1: This file appears as if every line was changed because I changed the line endings from Windows to Linux, to match the rest of the files in the repo. I'll call out the functional changes to this file in comments below, but you can also hide whitespace while reviewing in order to filter for just those changes.

dbt/models/external/schema/external.vw_cmap.yml
```diff
@@ -1,139 +1,140 @@
...
+        description: '{{ doc("shared_column_char_ext_wall") }}'
+      - name: year
+        description: '{{ doc("shared_column_year") }}'
+      - name: year_built
+        description: '{{ doc("shared_column_char_yrblt") }}'
+
+    data_tests:
+      - row_count:
+          name: external_vw_cmap_row_count
+          arguments:
```
jeancochrane, line 116: This is the first of two functional changes to this file: Nesting the `equals` argument under the newly-required `arguments` key to resolve [`ArgumentsPropertyInGenericTestDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#argumentspropertyingenerictestdeprecation).

dbt/models/external/schema/external.vw_cmap.yml
```diff
@@ -1,139 +1,140 @@
...
+                ) AS asmt_all
+                LEFT JOIN {{ source("iasworld", "dweldat") }} AS dweldat
+                  ON asmt_all.parid = dweldat.parid
+                  AND asmt_all.taxyr = dweldat.taxyr
+                  AND dweldat.deactivat IS NULL
+                  AND dweldat.cur = 'Y'
+              )
+      - unique_combination_of_columns:
+          name: external_vw_cmap_unique_by_pin_card_and_year
+          arguments:
```
jeancochrane, line 136: This is the second of two functional changes to this file: Nesting the `combination_of_columns` argument under the `arguments` key to resolve [`ArgumentsPropertyInGenericTestDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#argumentspropertyingenerictestdeprecation).

dbt/models/iasworld/schema/iasworld.addn.yml
```diff
@@ -1,40 +1,47 @@
+anchors:
+  - &unique-conditions
+    where: |
+      CAST(taxyr AS int) BETWEEN {{ var('data_test_iasworld_year_start') }} AND {{ var('data_test_iasworld_year_end') }}
+      AND cur = 'Y'
+      AND deactivat IS NULL
```
jeancochrane, line 5: dbt 1.10 introduced a new [top-level `anchors` key](https://docs.getdbt.com/reference/resource-properties/anchors?version=2) that we can use to define anchors that are not valid YAML on their own. Though I believe this `where` configuration is indeed valid YAML, I like the idea of moving it to a top-level anchor, because previously we would often need to move the anchor definition whenever we introduced a test that preceded the test that used to define the anchor.

We have to update all of our anchor definitions for unique conditions anyway, since the fact that we previously assigned `&unique-conditions` to the top-level `config` object means that we can't nest `meta` keys under `config` without triggering [`DuplicateYAMLKeysDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#duplicateyamlkeysdeprecation). Since we have to update all of these anchor definitions anyway, I decided to go ahead and move them to the top-level `anchors` key. I did the same for all of the `select-columns-*` anchors too.

dbt/models/default/schema/default.vw_pin_universe.yml
```diff
@@ -280,7 +285,7 @@ unit_tests:
         rows:
           # `class` is the important column here, and all other columns are
           # only set to ensure proper joins when creating the dummy tables
-          - {parid: "00000000000000", taxyr: "2024", class: "200", cur: "Y", deactivat: null, class: "2.1-1)A"}
+          - {parid: "00000000000000", taxyr: "2024", cur: "Y", deactivat: null, class: "2.1-1)A"}
```
jeancochrane: This is an example of a fix for [`DuplicateYAMLKeysDeprecation`](https://docs.getdbt.com/reference/deprecations?version=2#duplicateyamlkeysdeprecation).

dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -194,84 +222,91 @@ sources:
...
-                    description: >
-                      external_occpct (Occupancy %) should not be 0
+                  arguments:
+                    values: [0]
+                    quote: false
+                    additional_select_columns: *select-columns
+                  config:
+                    where: *unique-conditions
+                    meta:
+                      description: external_occpct (Occupancy %) should not be 0
```
jeancochrane: Some of these descriptions were short enough that they didn't actually need to be broken onto multiple lines in order to stick to our rough 80-character line limit. The [`dbt-autofix` tool](https://github.com/dbt-labs/dbt-autofix) that I used to help resolve warnings pulled these descriptions back onto one line, and I decided to leave the diff since I also prefer a single line when possible.

dbt/models/iasworld/schema/iasworld.comdat.yml
```diff
@@ -194,84 +222,91 @@ sources:
...
+                      - taxyr
+                      - parid
+                      - card
+                      - who
+                      - wen
+                      - chgrsn
+                      - mktadj
                   config:
-                    where:
+                    where: |
```
jeancochrane: Some of our multiline strings were not properly marked as such using YAML [block scalar headers](https://yaml.org/spec/1.2.2/#811-block-scalar-headers). This doesn't seem to cause incorrect dbt behavior, but it did cause `dbt-autofix` to try to collapse the multiline strings onto one line, so I added block headers where appropriate in order to preserve our formatting for these long multiline strings.

dbt/models/qc/schema.yml
```diff
@@ -1,3 +1,32 @@
+anchors:
```
jeancochrane: This file needed more refactoring than the rest, since it made heavy use of merging aliased YAML blocks in a way that raised duplicate key warnings. In general, I decomposed those block anchors in such a way that they were more specific, e.g. splitting out test arguments from `config.where` filters so that we can specify them separately.

dbt/models/sale/schema.yml
```diff
@@ -171,18 +183,11 @@ sources:
...
         description: '{{ doc("metadata") }}'
-        tags:
-          - load_auto
-        data_tests:
-          - unique_combination_of_columns:
-              name: sale_metadata_unique_by_run_id
-              combination_of_columns:
-                - run_id
-              meta:
-                description: metadata is unique by run_id
```
jeancochrane: As I mentioned in the PR body, this was a bug in the prior version of the code: By declaring two `data_tests` blocks on this source, we ended up overwriting this block, causing the `sale_metadata_unique_by_run_id` test to never run.

PR #1107, opened 2026-09-16
add iasworld.dweldat to tests

Damonamajor:
The biggest discrepency seems to be due to not aggregating from the card to PIN level as in the .yml file. That is in the `iasworld_dweldat_class_matches_pardat_class` test. I'll leave it unaggregated for now, but we may want to think about if we want to aggregate these tests moving forward.

Full query for all mismatches:

<details><summary>Details</summary>
<p>

WITH view_rows AS (
    FROM z_dev_damajor_qc.vw_report_iasworld_test_dweldat
),
test_failure_rows AS (
    SELECT 'iasworld_dweldat_attic_in_accepted_values' AS test_name, parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_attic_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_attic_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_attic_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_bsmt_and_user12_match_234_class', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_bsmt_and_user12_match_234_class
    UNION ALL
    SELECT 'iasworld_dweldat_bsmt_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_bsmt_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_bsmt_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_bsmt_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_calc_meth_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_calc_meth_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_card_gte_1', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_card_gte_1
    UNION ALL
    SELECT 'iasworld_dweldat_card_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_card_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_card_proration_rate_between_0_and_100', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_card_proration_rate_between_0_and_100
    UNION ALL
    SELECT 'iasworld_dweldat_char_air_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_air_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_air_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_air_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_apts_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_apts_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_apts_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_apts_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_attic_fnsh_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_attic_fnsh_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_attic_fnsh_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_attic_fnsh_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_bsmt_fin_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_bsmt_fin_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_bsmt_fin_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_bsmt_fin_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_frpl_between_0_and_6', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_frpl_between_0_and_6
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_area_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_area_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_area_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_area_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_att_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_att_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_att_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_att_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_cnst_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_cnst_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_cnst_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_cnst_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_cnst_null_if_gar_size_is_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_cnst_null_if_gar_size_is_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_size_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_size_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_size_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_size_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_ncu_between_0_and_5', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_between_0_and_5
    UNION ALL
    SELECT 'iasworld_dweldat_char_ncu_not_0_when_class_is_212', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_not_0_when_class_is_212
    UNION ALL
    SELECT 'iasworld_dweldat_char_ncu_not_null_when_class_is_212', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_not_null_when_class_is_212
    UNION ALL
    SELECT 'iasworld_dweldat_char_ncu_null_when_class_is_not_212_and_char_ncu_not_0', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_null_when_class_is_not_212_and_char_ncu_not_0
    UNION ALL
    SELECT 'iasworld_dweldat_char_porch_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_porch_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_renovation_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_renovation_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_roof_cnst_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_roof_cnst_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_roof_cnst_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_roof_cnst_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_tp_plan_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_tp_plan_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_tp_plan_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_tp_plan_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_use_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_use_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_char_use_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_use_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_class_in_ccao_class_dict', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_class_in_ccao_class_dict
    UNION ALL
    SELECT 'iasworld_dweldat_class_matches_pardat_class', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_class_matches_pardat_class
    UNION ALL
    SELECT 'iasworld_dweldat_cur_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_cur_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_exempt_classes_match_pardat_class', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_exempt_classes_match_pardat_class
    UNION ALL
    SELECT 'iasworld_dweldat_external_calc_rcnld_not_0', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_external_calc_rcnld_not_0
    UNION ALL
    SELECT 'iasworld_dweldat_external_occpct_not_0', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_external_occpct_not_0
    UNION ALL
    SELECT 'iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null
    UNION ALL
    SELECT 'iasworld_dweldat_external_propct_not_0', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_external_propct_not_0
    UNION ALL
    SELECT 'iasworld_dweldat_external_rcnld_not_0', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_external_rcnld_not_0
    UNION ALL
    SELECT 'iasworld_dweldat_extwall_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_extwall_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_fixbath_matches_number_of_units', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_fixbath_matches_number_of_units
    UNION ALL
    SELECT 'iasworld_dweldat_fixbath_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_fixbath_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_fixhalf_gte_0', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_fixhalf_gte_0
    UNION ALL
    SELECT 'iasworld_dweldat_fixhalf_matches_number_of_units', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_fixhalf_matches_number_of_units
    UNION ALL
    SELECT 'iasworld_dweldat_heat_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_heat_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_heat_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_heat_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null
    UNION ALL
    SELECT 'iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_parid_in_pardat_parid', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_parid_in_pardat_parid
    UNION ALL
    SELECT 'iasworld_dweldat_parid_not_null', CAST(NULL AS VARCHAR) AS parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_parid_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_rmbed_lte_rmtot', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_rmbed_lte_rmtot
    UNION ALL
    SELECT 'iasworld_dweldat_rmbed_matches_number_of_units', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_rmbed_matches_number_of_units
    UNION ALL
    SELECT 'iasworld_dweldat_rmbed_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_rmbed_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_rmtot_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_rmtot_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_rmtot_sf_between_1_and_40', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_rmtot_sf_between_1_and_40
    UNION ALL
    SELECT 'iasworld_dweldat_rmtot_sf_between_1_and_50', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_rmtot_sf_between_1_and_50
    UNION ALL
    SELECT 'iasworld_dweldat_seq_all_sequential_exist', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_seq_all_sequential_exist
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1000_and_1800_for_class_203', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1000_and_1800_for_class_203
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1801_and_25000_for_class_204', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1801_and_25000_for_class_204
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_10000_for_class_210', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_10000_for_class_210
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_10000_for_class_234', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_10000_for_class_234
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_20000_for_class_212', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_20000_for_class_212
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_20000_for_class_295', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_20000_for_class_295
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_2000_for_class_207', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_2000_for_class_207
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_2200_for_class_205', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_2200_for_class_205
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_40000_for_class_211', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_40000_for_class_211
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_1_and_999_for_class_202', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_999_for_class_202
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_2001_and_3800_for_class_278', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_2001_and_3800_for_class_278
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_2201_and_4999_for_class_206', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_2201_and_4999_for_class_206
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_3801_and_4999_for_class_208', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_3801_and_4999_for_class_208
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_between_5000_and_50000_for_class_209', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_5000_and_50000_for_class_209
    UNION ALL
    SELECT 'iasworld_dweldat_sfla_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_stories_in_accepted_values', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_stories_in_accepted_values
    UNION ALL
    SELECT 'iasworld_dweldat_stories_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_stories_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_taxyr_not_null', parid, CAST(NULL AS VARCHAR) AS taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_taxyr_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_unique_by_parid_card_taxyr', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_unique_by_parid_card_taxyr
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_205_more_than_62_years_old', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_205_more_than_62_years_old
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_206_more_than_62_years_old', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_206_more_than_62_years_old
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_207_less_than_62_years_old', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_207_less_than_62_years_old
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_208_less_than_62_years_old', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_208_less_than_62_years_old
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_210_more_than_62_years_old', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_210_more_than_62_years_old
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_278_less_than_62_years_old', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_278_less_than_62_years_old
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_295_less_than_62_years_old', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_295_less_than_62_years_old
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_between_1850_and_now', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_between_1850_and_now
    UNION ALL
    SELECT 'iasworld_dweldat_yrblt_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_not_null
),
mismatches AS (
    SELECT
        CASE
            WHEN v.taxyr IS NULL THEN 'only in test failure'
            WHEN f.taxyr IS NULL THEN 'only in qc view'
        END AS source
    FROM view_rows v
        ON v.test_name = f.test_name
        AND v.parid = f.parid
        AND v.taxyr = f.taxyr
    WHERE v.taxyr IS NULL OR f.taxyr IS NULL
)
SELECT
    test_name,
    source,
    COUNT(*) AS n
FROM mismatches
GROUP BY test_name, source
ORDER BY test_name, source

</p>
</details> 

Mismatches of counts:

<details><summary>Details</summary>
<p>

with view_counts as (
    select test_name, count(*) as cnt
    FROM z_dev_damajor_qc.vw_report_iasworld_test_dweldat
    group by 1
),

test_failure_rows as (
	SELECT 'iasworld_dweldat_attic_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_attic_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_attic_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_attic_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_bsmt_and_user12_match_234_class' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_bsmt_and_user12_match_234_class
	UNION ALL
	SELECT 'iasworld_dweldat_bsmt_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_bsmt_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_bsmt_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_bsmt_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_calc_meth_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_calc_meth_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_card_gte_1' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_card_gte_1
	UNION ALL
	SELECT 'iasworld_dweldat_card_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_card_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_card_proration_rate_between_0_and_100' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_card_proration_rate_between_0_and_100
	UNION ALL
	SELECT 'iasworld_dweldat_char_air_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_air_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_air_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_air_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_apts_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_apts_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_apts_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_apts_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_attic_fnsh_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_attic_fnsh_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_attic_fnsh_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_attic_fnsh_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_bsmt_fin_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_bsmt_fin_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_bsmt_fin_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_bsmt_fin_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_frpl_between_0_and_6' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_frpl_between_0_and_6
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_area_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_area_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_area_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_area_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_att_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_att_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_att_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_att_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_cnst_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_cnst_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_cnst_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_cnst_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_cnst_null_if_gar_size_is_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_cnst_null_if_gar_size_is_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_size_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_size_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_gar_size_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_size_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_ncu_between_0_and_5' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_between_0_and_5
	UNION ALL
	SELECT 'iasworld_dweldat_char_ncu_not_0_when_class_is_212' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_not_0_when_class_is_212
	UNION ALL
	SELECT 'iasworld_dweldat_char_ncu_not_null_when_class_is_212' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_not_null_when_class_is_212
	UNION ALL
	SELECT 'iasworld_dweldat_char_ncu_null_when_class_is_not_212_and_char_ncu_not_0' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_ncu_null_when_class_is_not_212_and_char_ncu_not_0
	UNION ALL
	SELECT 'iasworld_dweldat_char_porch_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_porch_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_renovation_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_renovation_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_roof_cnst_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_roof_cnst_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_roof_cnst_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_roof_cnst_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_tp_plan_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_tp_plan_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_tp_plan_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_tp_plan_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_char_use_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_use_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_char_use_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_char_use_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_class_in_ccao_class_dict' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_class_in_ccao_class_dict
	UNION ALL
	SELECT 'iasworld_dweldat_class_matches_pardat_class' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_class_matches_pardat_class
	UNION ALL
	SELECT 'iasworld_dweldat_cur_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_cur_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_exempt_classes_match_pardat_class' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_exempt_classes_match_pardat_class
	UNION ALL
	SELECT 'iasworld_dweldat_external_calc_rcnld_not_0' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_external_calc_rcnld_not_0
	UNION ALL
	SELECT 'iasworld_dweldat_external_occpct_not_0' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_external_occpct_not_0
	UNION ALL
	SELECT 'iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null
	UNION ALL
	SELECT 'iasworld_dweldat_external_propct_not_0' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_external_propct_not_0
	UNION ALL
	SELECT 'iasworld_dweldat_external_rcnld_not_0' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_external_rcnld_not_0
	UNION ALL
	SELECT 'iasworld_dweldat_extwall_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_extwall_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_extwall_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_extwall_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_fixbath_matches_number_of_units' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_fixbath_matches_number_of_units
	UNION ALL
	SELECT 'iasworld_dweldat_fixbath_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_fixbath_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_fixhalf_gte_0' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_fixhalf_gte_0
	UNION ALL
	SELECT 'iasworld_dweldat_fixhalf_matches_number_of_units' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_fixhalf_matches_number_of_units
	UNION ALL
	SELECT 'iasworld_dweldat_heat_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_heat_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_heat_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_heat_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null
	UNION ALL
	SELECT 'iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_parid_in_pardat_parid' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_parid_in_pardat_parid
	UNION ALL
	SELECT 'iasworld_dweldat_parid_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_parid_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_rmbed_lte_rmtot' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_rmbed_lte_rmtot
	UNION ALL
	SELECT 'iasworld_dweldat_rmbed_matches_number_of_units' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_rmbed_matches_number_of_units
	UNION ALL
	SELECT 'iasworld_dweldat_rmbed_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_rmbed_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_rmtot_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_rmtot_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_rmtot_sf_between_1_and_40' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_rmtot_sf_between_1_and_40
	UNION ALL
	SELECT 'iasworld_dweldat_rmtot_sf_between_1_and_50' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_rmtot_sf_between_1_and_50
	UNION ALL
	SELECT 'iasworld_dweldat_seq_all_sequential_exist' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_seq_all_sequential_exist
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1000_and_1800_for_class_203' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1000_and_1800_for_class_203
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1801_and_25000_for_class_204' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1801_and_25000_for_class_204
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_10000_for_class_210' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_10000_for_class_210
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_10000_for_class_234' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_10000_for_class_234
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_20000_for_class_212' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_20000_for_class_212
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_20000_for_class_295' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_20000_for_class_295
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_2000_for_class_207' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_2000_for_class_207
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_2200_for_class_205' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_2200_for_class_205
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_40000_for_class_211' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_40000_for_class_211
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_1_and_999_for_class_202' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_1_and_999_for_class_202
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_2001_and_3800_for_class_278' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_2001_and_3800_for_class_278
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_2201_and_4999_for_class_206' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_2201_and_4999_for_class_206
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_3801_and_4999_for_class_208' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_3801_and_4999_for_class_208
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_between_5000_and_50000_for_class_209' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_between_5000_and_50000_for_class_209
	UNION ALL
	SELECT 'iasworld_dweldat_sfla_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_stories_in_accepted_values' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_stories_in_accepted_values
	UNION ALL
	SELECT 'iasworld_dweldat_stories_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_stories_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_taxyr_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_taxyr_not_null
	UNION ALL
	SELECT 'iasworld_dweldat_unique_by_parid_card_taxyr' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_unique_by_parid_card_taxyr
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_205_more_than_62_years_old' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_205_more_than_62_years_old
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_206_more_than_62_years_old' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_206_more_than_62_years_old
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_207_less_than_62_years_old' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_207_less_than_62_years_old
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_208_less_than_62_years_old' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_208_less_than_62_years_old
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_210_more_than_62_years_old' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_210_more_than_62_years_old
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_278_less_than_62_years_old' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_278_less_than_62_years_old
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_295_less_than_62_years_old' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_295_less_than_62_years_old
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_between_1850_and_now' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_between_1850_and_now
	UNION ALL
	SELECT 'iasworld_dweldat_yrblt_not_null' AS test_name
	FROM z_dev_damajor_test_failure.iasworld_dweldat_yrblt_not_null
),

test_failure_counts as (
    select test_name, count(*) as cnt
    from test_failure_rows
    group by 1
)

select coalesce(v.test_name, t.test_name) as test_name, t.cnt as test_failure_count, v.cnt as view_count
from view_counts v
full outer join test_failure_counts t
    on v.test_name = t.test_name
    where coalesce(t.cnt, 0) != coalesce(v.cnt, 0)

</p>
</details> 

| test_name | old_test_failure_count | new_view_count |
| --- | --- | --- |
| iasworld_dweldat_heat_not_null | 133 | 132 |
| iasworld_dweldat_char_gar_area_not_null | 72 | 71 |
| iasworld_dweldat_class_matches_pardat_class | 40 | 4824 |
| iasworld_dweldat_sfla_not_null | 8 | |

It seems like some of these discrpencies are due to non-active PINs

```
WITH targets AS (
    SELECT 'iasworld_dweldat_sfla_not_null' AS test_name, parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_sfla_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_heat_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_heat_not_null
    UNION ALL
    SELECT 'iasworld_dweldat_char_gar_area_not_null', parid, taxyr
    FROM z_dev_damajor_test_failure.iasworld_dweldat_char_gar_area_not_null
),
active_pardat AS (
    SELECT parid, taxyr
    FROM iasworld.pardat
    WHERE cur = 'Y' AND deactivat IS NULL
)
SELECT
    t.test_name,
    CASE WHEN p.parid IS NULL THEN 'no active pardat match' ELSE 'has active pardat match' END AS pardat_status,
    COUNT(*) AS n
FROM targets t
LEFT JOIN active_pardat p ON t.parid = p.parid AND t.taxyr = p.taxyr
GROUP BY 1, 2
ORDER BY 1, 2

```

And it seems that there are a small number of `iasworld_dweldat_rmtot_sf_between_1_and_40` which uses a larger universe of classes. It's likely that these are all multi-card parcels.

```
SELECT
    f.*,
    d.class
FROM z_dev_damajor_test_failure.iasworld_dweldat_rmtot_sf_between_1_and_40 AS f
INNER JOIN iasworld.dweldat AS d
    ON f.parid = d.parid
    AND f.taxyr = d.taxyr
    AND f.card = d.card
    AND d.cur = 'Y'
    AND d.deactivat IS NULL
ORDER BY f.parid, f.taxyr, f.card;

 ```
It seems like this can be remedied by doing something like this, but I don't know if it's worth it.

```
{#-
    Most tests should exclude non-regression classes. A handful of tests 
    intentionally have other exclussion criteria. Those are listed here so
    they're skipped when the exclusion is applied below.
-#}
{%- set non_regression_classes = [
    '201', '213', '218', '219', '220', '221', '224', '225',
    '236', '240', '241', '290', '294', '297'
] -%}
{%- set tests_without_class_exclusion = [
    'iasworld_dweldat_class_in_ccao_class_dict',
    'iasworld_dweldat_rmtot_sf_between_1_and_40',
    'iasworld_dweldat_exempt_classes_match_pardat_class',
    'iasworld_dweldat_cur_in_accepted_values',
    'iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null',
    'iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null',
    'iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null'
] -%}
{%- set class_exclusion_escape = "class IN ('" ~ non_regression_classes | join("', '") ~ "') OR " -%}
{%- for test in tests if test.name not in tests_without_class_exclusion %}
    {%- do test.update(
        {"condition": class_exclusion_escape ~ "(" ~ test.condition ~ ")"}
    ) %}
{%- endfor %}
```

### DIFF
--- a/dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
+++ b/dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
@@ -1,0 +1,736 @@
+{%- set tests = [
+    {
+        "name": "iasworld_dweldat_attic_in_accepted_values",
+        "description": "attic (Attic Type) should be an integer between 1 and 5",
+        "category": "incorrect_values",
+        "condition": "attic IN ('1', '2', '3', '4', '5')",
+        "additional_select_columns": ["attic"]
+    },
+    {
+        "name": "iasworld_dweldat_attic_not_null",
+        "description": "attic (Attic Type) should not be null",
+        "category": "missing_values",
+        "condition": "attic IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_bsmt_in_accepted_values",
+        "description": "bsmt (Basement Type) should be an integer between 1 and 7",
+        "category": "incorrect_values",
+        "condition": "bsmt IN ('1', '2', '3', '4', '5', '6', '7')",
+        "additional_select_columns": ["bsmt"]
+    },
+    {
+        "name": "iasworld_dweldat_bsmt_not_null",
+        "description": "bsmt (Basement Type) should not be null",
+        "category": "missing_values",
+        "condition": "bsmt IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_bsmt_and_user12_match_234_class",
+        "description": "bsmt (Basement Type) must be 3 - PARTIAL and user12 (Basement Finished) must be 1 - FAMILY ROOM when class is 234 (split level)",
+        "category": "incorrect_values",
+        "condition": "class != '234' OR (bsmt = '3' AND user12 = '1')",
+        "additional_select_columns": ["bsmt", "user12"]
+    },
+    {
+        "name": "iasworld_dweldat_calc_meth_accepted_values",
+        "description": 'calc_meth (Calculation Method) should be null or "E"',
+        "category": "incorrect_values",
+        "condition": "calc_meth IS NULL OR calc_meth = 'E'",
+        "additional_select_columns": ["calc_meth"]
+    },
+    {
+        "name": "iasworld_dweldat_card_gte_1",
+        "description": "card should be >= 1",
+        "category": "incorrect_values",
+        "condition": "card IS NULL OR card >= 1"
+    },
+    {
+        "name": "iasworld_dweldat_card_not_null",
+        "description": "card should not be null",
+        "category": "missing_values",
+        "condition": "card IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_card_proration_rate_between_0_and_100",
+        "description": "user24 (Proration %) should be between 0 and 100",
+        "category": "incorrect_values",
+        "condition": "user24 IS NULL OR CAST(user24 AS decimal) BETWEEN 0.00 AND 100.00",
+        "additional_select_columns": ["user24"]
+    },
+    {
+        "name": "iasworld_dweldat_class_in_ccao_class_dict",
+        "description": "class code must be valid",
+        "category": "class_mismatch_or_issue",
+        "condition": "class IN ('EX', 'RR') OR class_dict_class IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_class_matches_pardat_class",
+        "description": "at least one class should match pardat class",
+        "category": "class_mismatch_or_issue",
+        "condition": "comdat_parid IS NOT NULL OR pardat_class IS NULL OR class = pardat_class",
+        "additional_select_columns": ["pardat_class"]
+    },
+    {
+        "name": "iasworld_dweldat_exempt_classes_match_pardat_class",
+        "description": "at least one class should be exempt or omitted if pardat is exempt",
+        "category": "class_mismatch_or_issue",
+        "condition": "class LIKE 'OA%' OR pardat_class IS NULL OR pardat_class != 'EX' OR class = 'EX'",
+        "additional_select_columns": ["pardat_class"]
+    },
+    {
+        "name": "iasworld_dweldat_cur_in_accepted_values",
+        "description": 'cur should be "Y" or "D"',
+        "category": "incorrect_values",
+        "condition": "cur IN ('Y', 'D')",
+        "additional_select_columns": ["cur"]
+    },
+    {
+        "name": "iasworld_dweldat_external_calc_rcnld_not_0",
+        "description": "external_calc_rcnld (Net Market Value) should not be 0",
+        "category": "incorrect_values",
+        "condition": "external_calc_rcnld IS NULL OR external_calc_rcnld != 0",
+        "additional_select_columns": ["external_calc_rcnld"]
+    },
+    {
+        "name": "iasworld_dweldat_external_occpct_not_0",
+        "description": "external_occpct (Occupancy %) should not be 0",
+        "category": "incorrect_values",
+        "condition": "external_occpct IS NULL OR external_occpct != 0",
+        "additional_select_columns": ["external_occpct"]
+    },
+    {
+        "name": "iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null",
+        "description": "external_occpct (Occupancy % [current]) should not be null if mktrsn is 5 or 5B and mktadj is null",
+        "category": "missing_values",
+        "condition": "mktrsn NOT IN ('5', '5B') OR mktadj IS NOT NULL OR external_occpct IS NOT NULL",
+        "additional_select_columns": ["external_occpct", "mktrsn", "mktadj"]
+    },
+    {
+        "name": "iasworld_dweldat_external_propct_not_0",
+        "description": "external_propct (Proration %) should not be 0",
+        "category": "incorrect_values",
+        "condition": "external_propct IS NULL OR external_propct != 0",
+        "additional_select_columns": ["external_propct"]
+    },
+    {
+        "name": "iasworld_dweldat_external_rcnld_not_0",
+        "description": "external_rcnld (Full Market Value) should not be 0",
+        "category": "incorrect_values",
+        "condition": "external_rcnld IS NULL OR external_rcnld != 0",
+        "additional_select_columns": ["external_rcnld"]
+    },
+    {
+        "name": "iasworld_dweldat_extwall_in_accepted_values",
+        "description": "extwall (Exterior Construction) should be 1, 2, 3, 4, 6, 7, 8, or 9",
+        "category": "incorrect_values",
+        "condition": "extwall IN ('1', '2', '3', '4', '6', '7', '8', '9')",
+        "additional_select_columns": ["extwall"]
+    },
+    {
+        "name": "iasworld_dweldat_extwall_not_null",
+        "description": "extwall (Exterior Construction) should not be null",
+        "category": "missing_values",
+        "condition": "extwall IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_fixbath_matches_number_of_units",
+        "description": "fixbath (Number of Full Baths) should be <= 7 times the number of units (user14)",
+        "category": "incorrect_values",
+        "condition": "fixbath IS NULL OR (fixbath >= 1 AND fixbath <= CASE WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 7 WHEN user14 = '1' THEN 14 WHEN user14 = '2' THEN 21 WHEN user14 = '3' THEN 28 WHEN user14 = '4' THEN 35 WHEN user14 = '5' THEN 42 ELSE 7 END)",
+        "additional_select_columns": ["fixbath", "user14"]
+    },
+    {
+        "name": "iasworld_dweldat_fixbath_not_null",
+        "description": "fixbath (Number of Full Baths) should not be null",
+        "category": "missing_values",
+        "condition": "fixbath IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_fixhalf_gte_0",
+        "description": "fixhalf (Number of Half Baths) should be >= 0",
+        "category": "incorrect_values",
+        "condition": "fixhalf IS NULL OR fixhalf >= 0",
+        "additional_select_columns": ["fixhalf"]
+    },
+    {
+        "name": "iasworld_dweldat_fixhalf_matches_number_of_units",
+        "description": "fixhalf (Number of Half Baths) should be <= 5 times the number of units (user14)",
+        "category": "incorrect_values",
+        "condition": "fixhalf IS NULL OR (fixhalf >= 0 AND fixhalf <= CASE WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 5 WHEN user14 = '1' THEN 10 WHEN user14 = '2' THEN 15 WHEN user14 = '3' THEN 20 WHEN user14 = '4' THEN 25 WHEN user14 = '5' THEN 30 ELSE 5 END)",
+        "additional_select_columns": ["fixhalf", "user14"]
+    },
+    {
+        "name": "iasworld_dweldat_heat_in_accepted_values",
+        "description": "heat (Heating) should be an integer between 1 and 4",
+        "category": "incorrect_values",
+        "condition": "heat IN ('1', '2', '3', '4')",
+        "additional_select_columns": ["heat"]
+    },
+    {
+        "name": "iasworld_dweldat_heat_not_null",
+        "description": "heat (Heating) should not be null",
+        "category": "missing_values",
+        "condition": "heat IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null",
+        "description": "mktadj (Occupancy % [deprecated]) should not be null if mktrsn is 5 or 5B and external_occpct is null",
+        "category": "missing_values",
+        "condition": "mktrsn NOT IN ('5', '5B') OR external_occpct IS NOT NULL OR mktadj IS NOT NULL",
+        "additional_select_columns": ["mktadj", "mktrsn", "external_occpct"]
+    },
+    {
+        "name": "iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null",
+        "description": "mktrsn (Reason for Override) should be 5 or 5B if external_occpct or mktadj is not null",
+        "category": "incorrect_values",
+        "condition": "(external_occpct IS NULL AND mktadj IS NULL) OR mktrsn IN ('5', '5B')",
+        "additional_select_columns": ["mktrsn", "external_occpct", "mktadj"]
+    },
+    {
+        "name": "iasworld_dweldat_parid_not_null",
+        "description": "parid should not be null",
+        "category": "missing_values",
+        "condition": "parid IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_parid_in_pardat_parid",
+        "description": "parid should be in pardat",
+        "category": "relationships",
+        "condition": "pardat_parid IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_rmbed_lte_rmtot",
+        "description": "rmbed (Number of Bedrooms) should be <= rmtot (Number of Rooms)",
+        "category": "relationships",
+        "condition": "rmbed IS NULL OR rmtot IS NULL OR rmbed <= rmtot",
+        "additional_select_columns": ["rmbed", "rmtot"]
+    },
+    {
+        "name": "iasworld_dweldat_rmbed_matches_number_of_units",
+        "description": "rmbed (Number of Bedrooms) should be <= 8 times the number of units (user14)",
+        "category": "incorrect_values",
+        "condition": "rmbed IS NULL OR (rmbed >= 1 AND rmbed <= CASE WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 8 WHEN user14 = '1' THEN 16 WHEN user14 = '2' THEN 24 WHEN user14 = '3' THEN 32 WHEN user14 = '4' THEN 40 WHEN user14 = '5' THEN 48 ELSE 8 END)",
+        "additional_select_columns": ["rmbed", "user14"]
+    },
+    {
+        "name": "iasworld_dweldat_rmbed_not_null",
+        "description": "rmbed (Number of Bedrooms) should not be null",
+        "category": "missing_values",
+        "condition": "rmbed IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_rmtot_not_null",
+        "description": "rmtot (Number of Rooms) should not be null",
+        "category": "missing_values",
+        "condition": "rmtot IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_rmtot_sf_between_1_and_40",
+        "description": "rmtot (Number of Rooms) should be between 1 and 40",
+        "category": "incorrect_values",
+        "condition": "class IN ('211', '212') OR rmtot IS NULL OR (rmtot >= 1 AND rmtot <= 40)",
+        "additional_select_columns": ["rmtot"]
+    },
+    {
+        "name": "iasworld_dweldat_rmtot_sf_between_1_and_50",
+        "description": "rmtot (Number of Rooms) should be between 1 and 50",
+        "category": "incorrect_values",
+        "condition": "class NOT IN ('211', '212') OR rmtot IS NULL OR (rmtot >= 1 AND rmtot <= 50)",
+        "additional_select_columns": ["rmtot"]
+    },
+    {
+        "name": "iasworld_dweldat_seq_all_sequential_exist",
+        "description": "seq should be sequential",
+        "category": "incorrect_values",
+        "condition": "prev_seq IS NULL OR seq = prev_seq + 1",
+        "additional_select_columns": ["seq", "prev_seq"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_999_for_class_202",
+        "description": "sfla (Building Square Footage) should be between 1 and 999 for class 202 cards",
+        "category": "incorrect_values",
+        "condition": "class != '202' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 999)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_2000_for_class_207",
+        "description": "sfla (Building Square Footage) should be between 1 and 2000 for class 207 cards",
+        "category": "incorrect_values",
+        "condition": "class != '207' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 2000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_2200_for_class_205",
+        "description": "sfla (Building Square Footage) should be between 1 and 2200 for class 205 cards",
+        "category": "incorrect_values",
+        "condition": "class != '205' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 2200)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_10000_for_class_210",
+        "description": "sfla (Building Square Footage) should be between 1 and 10000 for class 210 cards",
+        "category": "incorrect_values",
+        "condition": "class != '210' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 10000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_10000_for_class_234",
+        "description": "sfla (Building Square Footage) should be between 1 and 10000 for class 234 cards",
+        "category": "incorrect_values",
+        "condition": "class != '234' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 10000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_20000_for_class_212",
+        "description": "sfla (Building Square Footage) should be between 1 and 20000 for class 212 cards",
+        "category": "incorrect_values",
+        "condition": "class != '212' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 20000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_20000_for_class_295",
+        "description": "sfla (Building Square Footage) should be between 1 and 20000 for class 295 cards",
+        "category": "incorrect_values",
+        "condition": "class != '295' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 20000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1_and_40000_for_class_211",
+        "description": "sfla (Building Square Footage) should be between 1 and 40000 for class 211 cards",
+        "category": "incorrect_values",
+        "condition": "class != '211' OR sfla IS NULL OR (sfla >= 1 AND sfla <= 40000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1000_and_1800_for_class_203",
+        "description": "sfla (Building Square Footage) should be between 1000 and 1800 for class 203 cards",
+        "category": "incorrect_values",
+        "condition": "class != '203' OR sfla IS NULL OR (sfla >= 1000 AND sfla <= 1800)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_1801_and_25000_for_class_204",
+        "description": "sfla (Building Square Footage) should be between 1801 and 25000 for class 204 cards",
+        "category": "incorrect_values",
+        "condition": "class != '204' OR sfla IS NULL OR (sfla >= 1801 AND sfla <= 25000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_2001_and_3800_for_class_278",
+        "description": "sfla (Building Square Footage) should be between 2001 and 3800 for class 278 cards",
+        "category": "incorrect_values",
+        "condition": "class != '278' OR sfla IS NULL OR (sfla >= 2001 AND sfla <= 3800)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_2201_and_4999_for_class_206",
+        "description": "sfla (Building Square Footage) should be between 2201 and 4999 for class 206 cards",
+        "category": "incorrect_values",
+        "condition": "class != '206' OR sfla IS NULL OR (sfla >= 2201 AND sfla <= 4999)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_3801_and_4999_for_class_208",
+        "description": "sfla (Building Square Footage) should be between 3801 and 4999 for class 208 cards",
+        "category": "incorrect_values",
+        "condition": "class != '208' OR sfla IS NULL OR (sfla >= 3801 AND sfla <= 4999)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_between_5000_and_50000_for_class_209",
+        "description": "sfla (Building Square Footage) should be between 5000 and 50000 for class 209 cards",
+        "category": "incorrect_values",
+        "condition": "class != '209' OR sfla IS NULL OR (sfla >= 5000 AND sfla <= 50000)",
+        "additional_select_columns": ["sfla"]
+    },
+    {
+        "name": "iasworld_dweldat_sfla_not_null",
+        "description": "sfla (Building Square Footage) should not be null",
+        "category": "missing_values",
+        "condition": "sfla IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_stories_in_accepted_values",
+        "description": "stories (Type of Residence) should be one of 1.00, 2.00, 3.00, 4.00, 5.00, 6.00, 7.00, 8.00, 9.00, or 9.90",
+        "category": "incorrect_values",
+        "condition": "stories IN (1.00, 2.00, 3.00, 4.00, 5.00, 6.00, 7.00, 8.00, 9.00, 9.90)",
+        "additional_select_columns": ["stories"]
+    },
+    {
+        "name": "iasworld_dweldat_stories_not_null",
+        "description": "stories (Type of Residence) should not be null",
+        "category": "missing_values",
+        "condition": "stories IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_taxyr_not_null",
+        "description": "taxyr should not be null",
+        "category": "missing_values",
+        "condition": "taxyr IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_renovation_in_accepted_values",
+        "description": 'user3 (Renovated) should be "0" or "1"',
+        "category": "incorrect_values",
+        "condition": "user3 IN ('0', '1')",
+        "additional_select_columns": ["user3"]
+    },
+    {
+        "name": "iasworld_dweldat_char_tp_plan_in_accepted_values",
+        "description": 'user5 (Plan of Design) should be "0", "1", or "2"',
+        "category": "incorrect_values",
+        "condition": "user5 IN ('0', '1', '2')",
+        "additional_select_columns": ["user5"]
+    },
+    {
+        "name": "iasworld_dweldat_char_tp_plan_not_null",
+        "description": "user5 (Plan of Design) should not be null",
+        "category": "missing_values",
+        "condition": "user5 IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_attic_fnsh_in_accepted_values",
+        "description": 'user6 (Attic Finish) should be "1", "2", or "3"',
+        "category": "incorrect_values",
+        "condition": "user6 IN ('1', '2', '3')",
+        "additional_select_columns": ["user6"]
+    },
+    {
+        "name": "iasworld_dweldat_char_attic_fnsh_not_null",
+        "description": "user6 (Attic Finish) should not be null unless attic is 3 (None)",
+        "category": "missing_values",
+        "condition": "attic = '3' OR user6 IS NOT NULL",
+        "additional_select_columns": ["user6", "attic"]
+    },
+    {
+        "name": "iasworld_dweldat_char_air_accepted_values",
+        "description": 'user7 (Central Air Conditioning) should be "1" or "2"',
+        "category": "incorrect_values",
+        "condition": "user7 IN ('1', '2')",
+        "additional_select_columns": ["user7"]
+    },
+    {
+        "name": "iasworld_dweldat_char_air_not_null",
+        "description": "user7 (Central Air Conditioning) should not be null",
+        "category": "missing_values",
+        "condition": "user7 IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_bsmt_fin_accepted_values",
+        "description": "user12 (Basement Finished) should be an integer between 1 and 6",
+        "category": "incorrect_values",
+        "condition": "user12 IN ('1', '2', '3', '4', '5', '6')",
+        "additional_select_columns": ["user12"]
+    },
+    {
+        "name": "iasworld_dweldat_char_bsmt_fin_not_null",
+        "description": "user12 (Basement Finished) should not be null unless bsmt is 2 (Slab)",
+        "category": "missing_values",
+        "condition": "bsmt = '2' OR user12 IS NOT NULL",
+        "additional_select_columns": ["user12", "bsmt"]
+    },
+    {
+        "name": "iasworld_dweldat_char_roof_cnst_accepted_values",
+        "description": "user13 (Roof Construction) should be an integer between 1 and 6",
+        "category": "incorrect_values",
+        "condition": "user13 IN ('1', '2', '3', '4', '5', '6')",
+        "additional_select_columns": ["user13"]
+    },
+    {
+        "name": "iasworld_dweldat_char_roof_cnst_not_null",
+        "description": "user13 (Roof Construction) should not be null",
+        "category": "missing_values",
+        "condition": "user13 IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_apts_accepted_values",
+        "description": "user14 (Total Number of Units) should be an integer between 1 and 6 for class 211 and 212",
+        "category": "incorrect_values",
+        "condition": "class NOT IN ('211', '212') OR user14 IN ('1', '2', '3', '4', '5', '6')",
+        "additional_select_columns": ["user14"]
+    },
+    {
+        "name": "iasworld_dweldat_char_apts_not_null",
+        "description": "user14 (Total Number of Units) should not be null for class 211 and 212",
+        "category": "missing_values",
+        "condition": "class NOT IN ('211', '212') OR user14 IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_use_accepted_values",
+        "description": 'user15 (Use) should be "1" or "2"',
+        "category": "incorrect_values",
+        "condition": "user15 IN ('1', '2')",
+        "additional_select_columns": ["user15"]
+    },
+    {
+        "name": "iasworld_dweldat_char_use_not_null",
+        "description": "user15 (Use) should not be null",
+        "category": "missing_values",
+        "condition": "user15 IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_ncu_between_0_and_5",
+        "description": "user20 (No. of Commercial Units) should be an integer between 0 and 5",
+        "category": "incorrect_values",
+        "condition": "user20 IN ('0', '1', '2', '3', '4', '5')",
+        "additional_select_columns": ["user20"]
+    },
+    {
+        "name": "iasworld_dweldat_char_ncu_not_0_when_class_is_212",
+        "description": "user20 (No. of Commercial Units) should not be 0 when class is 212",
+        "category": "incorrect_values",
+        "condition": "class != '212' OR user20 != '0'",
+        "additional_select_columns": ["user20"]
+    },
+    {
+        "name": "iasworld_dweldat_char_ncu_not_null_when_class_is_212",
+        "description": "user20 (No. of Commercial Units) should not be null when class is 212",
+        "category": "missing_values",
+        "condition": "class != '212' OR user20 IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_ncu_null_when_class_is_not_212_and_char_ncu_not_0",
+        "description": "user20 (No. of Commercial Units) should be null or 0 when class is not 212",
+        "category": "incorrect_values",
+        "condition": "class IN ('EX', 'RR') OR class = '212' OR user20 = '0' OR user20 IS NULL",
+        "additional_select_columns": ["user20"]
+    },
+    {
+        "name": "iasworld_dweldat_char_porch_accepted_values",
+        "description": 'user30 (Porch) should be "0", "1", or "2"',
+        "category": "incorrect_values",
+        "condition": "user30 IN ('0', '1', '2')",
+        "additional_select_columns": ["user30"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_att_accepted_values",
+        "description": 'user31 (Garage Attached) should be "1" or "2"',
+        "category": "incorrect_values",
+        "condition": "user31 IN ('1', '2')",
+        "additional_select_columns": ["user31"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_att_not_null",
+        "description": "user31 (Garage Attached) should not be null if user33 (Garage Size) is not 7 (NONE)",
+        "category": "missing_values",
+        "condition": "user33 = '7' OR user31 IS NOT NULL",
+        "additional_select_columns": ["user31", "user33"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_area_accepted_values",
+        "description": 'user32 (Garage in Area) should be "1" or "2"',
+        "category": "incorrect_values",
+        "condition": "user32 IN ('1', '2')",
+        "additional_select_columns": ["user32"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_area_not_null",
+        "description": "user32 (Garage in Area) should not be null if user33 (Garage Size) is not 7 (NONE)",
+        "category": "missing_values",
+        "condition": "user33 = '7' OR user32 IS NOT NULL",
+        "additional_select_columns": ["user32", "user33"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_size_accepted_values",
+        "description": "user33 (Garage Size) should be an integer between 1 and 10",
+        "category": "incorrect_values",
+        "condition": "user33 IN ('1', '2', '3', '4', '5', '6', '7', '8', '9', '10')",
+        "additional_select_columns": ["user33"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_size_not_null",
+        "description": "user33 (Garage Size) should not be null",
+        "category": "missing_values",
+        "condition": "user33 IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_cnst_accepted_values",
+        "description": "user34 (Garage Construction) should be an integer between 1 and 4",
+        "category": "incorrect_values",
+        "condition": "user34 IN ('1', '2', '3', '4')",
+        "additional_select_columns": ["user34"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_cnst_not_null",
+        "description": "user34 (Garage Construction) should not be null if user33 (Garage Size) is not 7 (NONE)",
+        "category": "missing_values",
+        "condition": "user33 = '7' OR user34 IS NOT NULL",
+        "additional_select_columns": ["user34", "user33"]
+    },
+    {
+        "name": "iasworld_dweldat_char_gar_cnst_null_if_gar_size_is_null",
+        "description": "user34 (Garage Construction) should be null or 0 if user33 (Garage Size) is 7 (NONE)",
+        "category": "incorrect_values",
+        "condition": "user33 != '7' OR user34 IS NULL OR user34 = '0'",
+        "additional_select_columns": ["user34", "user33"]
+    },
+    {
+        "name": "iasworld_dweldat_char_frpl_between_0_and_6",
+        "description": "wbfp_o (Number of Fireplaces) should be between 0 and 6",
+        "category": "incorrect_values",
+        "condition": "wbfp_o IS NULL OR (wbfp_o >= 0 AND wbfp_o <= 6)",
+        "additional_select_columns": ["wbfp_o"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_205_more_than_62_years_old",
+        "description": "yrblt should be > 62 years old when class is 205",
+        "category": "incorrect_values",
+        "condition": "class != '205' OR yrblt IS NULL OR yrblt <= year(current_date) - 63",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_206_more_than_62_years_old",
+        "description": "yrblt should be > 62 years old when class is 206",
+        "category": "incorrect_values",
+        "condition": "class != '206' OR yrblt IS NULL OR yrblt <= year(current_date) - 63",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_207_less_than_62_years_old",
+        "description": "yrblt should be <= 62 years old when class is 207",
+        "category": "incorrect_values",
+        "condition": "class != '207' OR yrblt IS NULL OR yrblt >= year(current_date) - 62",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_208_less_than_62_years_old",
+        "description": "yrblt should be <= 62 years old when class is 208",
+        "category": "incorrect_values",
+        "condition": "class != '208' OR yrblt IS NULL OR yrblt >= year(current_date) - 62",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_210_more_than_62_years_old",
+        "description": "yrblt should be > 62 years old when class is 210",
+        "category": "incorrect_values",
+        "condition": "class != '210' OR yrblt IS NULL OR yrblt <= year(current_date) - 63",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_278_less_than_62_years_old",
+        "description": "yrblt should be <= 62 years old when class is 278",
+        "category": "incorrect_values",
+        "condition": "class != '278' OR yrblt IS NULL OR yrblt >= year(current_date) - 62",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_295_less_than_62_years_old",
+        "description": "yrblt should be <= 62 years old when class is 295",
+        "category": "incorrect_values",
+        "condition": "class != '295' OR yrblt IS NULL OR yrblt >= year(current_date) - 62",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_between_1850_and_now",
+        "description": "yrblt should be between 1850 and now",
+        "category": "incorrect_values",
+        "condition": "yrblt IS NULL OR (yrblt >= 1850 AND yrblt <= year(current_date))",
+        "additional_select_columns": ["yrblt"]
+    },
+    {
+        "name": "iasworld_dweldat_yrblt_not_null",
+        "description": "yrblt should not be null",
+        "category": "missing_values",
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql [hunk 1]
```diff
@@ -0,0 +1,729 @@
...
+        SELECT DISTINCT parid, taxyr
+        FROM {{ source('iasworld', 'comdat') }}
+        WHERE cur = 'Y'
+            AND deactivat IS NULL
+    ) AS comdat
+        ON dweldat.parid = comdat.parid
+        AND dweldat.taxyr = comdat.taxyr
+    LEFT JOIN {{ ref('ccao.class_dict') }} AS class_dict
+        ON dweldat.class = class_dict.class_code
+    WHERE dweldat.cur = 'Y'
```
Damonamajor, line 721: The main question I have is that we use unique conditions for most but not all tests in the .yml file.
https://github.com/ccao-data/data-architecture/blob/master/dbt/models/iasworld/schema/iasworld.dweldat.yml#L62-L67

Is there an issue with using just regression classes for this query, or do we want to seperate it out individually as we did in the .sql file.

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+        "description": 'calc_meth (Calculation Method) should be null or "E"',
+        "category": "incorrect_values",
+        "condition": "calc_meth IS NULL OR calc_meth = 'E'",
+        "additional_select_columns": ["calc_meth"]
+    },
+    {
+        "name": "iasworld_dweldat_card_gte_1",
+        "description": "card should be >= 1",
+        "category": "incorrect_values",
+        "condition": "card IS NULL OR card >= 1"
```
jeancochrane, line 46: **[Nitpick, optional]** We actually don't need to explicitly test for `card IS NULL` here, since Athena will handle nulls properly in this case. If `card` is null, then the bare condition `card >= 1` will also evaluate to null (rather than `true` or `false`), and the `not (card >= 1`) condition won't return the row, which is what we want in the case of nulls. Evidence of this behavior in [the Trino docs](https://trino.io/docs/current/functions/comparison.html#is-distinct-from-and-is-not-distinct-from) (Trino is the underlying query engine for Athena):

> In SQL a `NULL` value signifies an unknown value, so any comparison involving a `NULL` produces `NULL`.

Here's a quick SQL gut check to confirm that `not null` will never evaluate to true. This query returns no rows:

```sql
select parid, card
from iasworld.dweldat
where cur = 'Y'
    and deactivat is null
    and taxyr = '2026'
    and not null
```

Compare that behavior to `not false`, which returns all rows:

```sql
select parid, card
from iasworld.dweldat
where cur = 'Y'
    and deactivat is null
    and taxyr = '2026'
    and not false
```

It doesn't make a huge difference either way, since the condition is also valid with `card IS NOT NULL` included, so I leave it up to you to decide you want to keep it in there or not. My gut feeling is that we should only include nulls in the condition if we're explicitly testing for it (like in `iasworld_dweldat_card_not_null`) or if the test description mentions it as an accepted value (like in `iasworld_dweldat_calc_meth_accepted_values`), since that seems like the clearest pattern. But again, it's technically fine either way.


```suggestion
        "condition": "card >= 1"
```

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+        "name": "iasworld_dweldat_card_not_null",
+        "description": "card should not be null",
+        "category": "missing_values",
+        "condition": "card IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_card_proration_rate_between_0_and_100",
+        "description": "user24 (Proration %) should be between 0 and 100",
+        "category": "incorrect_values",
+        "condition": "user24 IS NULL OR CAST(user24 AS decimal) BETWEEN 0.00 AND 100.00",
```
jeancochrane: **[Nitpick, optional]** Same here as above: `NOT (CAST(user24 AS decimal) BETWEEN 0.00 AND 100.00)` will return null if `user24` is null, so we don't actually need the `user24` is null condition here.


```suggestion
        "condition": "CAST(user24 AS decimal) BETWEEN 0.00 AND 100.00",
```

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+        "name": "iasworld_dweldat_extwall_not_null",
+        "description": "extwall (Exterior Construction) should not be null",
+        "category": "missing_values",
+        "condition": "extwall IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_fixbath_matches_number_of_units",
+        "description": "fixbath (Number of Full Baths) should be <= 7 times the number of units (user14)",
+        "category": "incorrect_values",
+        "condition": "fixbath IS NULL OR (fixbath >= 1 AND fixbath <= CASE WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 7 WHEN user14 = '1' THEN 14 WHEN user14 = '2' THEN 21 WHEN user14 = '3' THEN 28 WHEN user14 = '4' THEN 35 WHEN user14 = '5' THEN 42 ELSE 7 END)",
```
jeancochrane, line 140: **[Nitpick, optional]** Let's break super long lines like this one for the sake of readability (also removing the unnecessary `fixbath IS NULL` condition):

```suggestion
        "condition": (
            "fixbath BETWEEN 1 AND CASE "
                "WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 7 "
                "WHEN user14 = '1' THEN 14 "
                "WHEN user14 = '2' THEN 21 "
                "WHEN user14 = '3' THEN 28 "
                "WHEN user14 = '4'THEN 35 "
                "WHEN user14 = '5' THEN 42 "
                "ELSE 7 "
            "END"
        ),
```

There are a few more tests below where we could also break long lines like this one.

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+        "description": "fixhalf (Number of Half Baths) should be >= 0",
+        "category": "incorrect_values",
+        "condition": "fixhalf IS NULL OR fixhalf >= 0",
+        "additional_select_columns": ["fixhalf"]
+    },
+    {
+        "name": "iasworld_dweldat_fixhalf_matches_number_of_units",
+        "description": "fixhalf (Number of Half Baths) should be <= 5 times the number of units (user14)",
+        "category": "incorrect_values",
+        "condition": "fixhalf IS NULL OR (fixhalf >= 0 AND fixhalf <= CASE WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 5 WHEN user14 = '1' THEN 10 WHEN user14 = '2' THEN 15 WHEN user14 = '3' THEN 20 WHEN user14 = '4' THEN 25 WHEN user14 = '5' THEN 30 ELSE 5 END)",
```
jeancochrane, line 160: **[Nitpick, required]** We already have a test to confirm that `fixhalf >= 0`, so by adding that same condition to this test, we run the risk of duplicate test failures for rows where `fixhalf` is negative. We could easily resolve that concern by removing the `fixhalf >= 0` condition here.

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+        "description": "mktadj (Occupancy % [deprecated]) should not be null if mktrsn is 5 or 5B and external_occpct is null",
+        "category": "missing_values",
+        "condition": "mktrsn NOT IN ('5', '5B') OR external_occpct IS NOT NULL OR mktadj IS NOT NULL",
+        "additional_select_columns": ["mktadj", "mktrsn", "external_occpct"]
+    },
+    {
+        "name": "iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null",
+        "description": "mktrsn (Reason for Override) should be 5 or 5B if external_occpct or mktadj is not null",
+        "category": "incorrect_values",
+        "condition": "(external_occpct IS NULL AND mktadj IS NULL) OR mktrsn IN ('5', '5B')",
```
jeancochrane, line 187: **[Nitpick, required]** This is actually a case where we _do_ need to check to make sure `mktrsn` is not null. The logic is a little complicated, but basically: If one of `external_occpct` or `mkatdj` is not null, and `mktrsn` is null, then we want the row to fail the test, because `mktrsn` needs to be `'5'` or `'5B'` in that case. But if `mktrsn` is null, then this full condition will evaluate to null, as will the `not ({{ condition }})` filter in the `generate_iasworld_qc_test_view()` macro, causing the row not to be returned by the `where {{ test.name }}` filter in that same macro.

This general rule of thumb might help for reasoning about nulls in `accepted_values` tests like this one:

- If null is a passing value for the test, then we don't need to worry about handling it, because Athena will generally filter it out from the results.
- If null is a failing value for the test, then we either need a dedicated test to check for nulls (like how we handle `attic`), or we need to make sure the test fails rows with null values.

Does that make sense? The docs for Trino's [effect of NULL on logical operators](https://trino.io/docs/current/functions/logical.html#effect-of-null-on-logical-operators) might also help for reasoning about cases like this one that involve multiple boolean clauses.


```suggestion
        "condition": "(external_occpct IS NULL AND mktadj IS NULL) OR (mktrsn IS NOT NULL AND mktrsn IN ('5', '5B'))",
```

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+        "condition": "comdat_parid IS NOT NULL OR class = pardat_class",
+        "additional_select_columns": ["pardat_class"]
+    },
+    {
+        "name": "iasworld_dweldat_exempt_classes_match_pardat_class",
+        "description": "at least one class should be exempt or omitted if pardat is exempt",
+        "category": "class_mismatch_or_issue",
+        "condition": "class LIKE 'OA%' OR pardat_class != 'EX' OR class = 'EX'",
+        "additional_select_columns": ["pardat_class"]
+    },
```
jeancochrane, line 80: **[Suggestion, optional]** As you mention in the PR body, these pardat class tests pose a problem, since they require that we aggregate dweldat in a way that isn't currently possible. I see two simple workarounds:

1. Create a new file with a new query dedicated to these tests, something like `qc.vw_report_iasworld_test_dweldat_agg.sql`
    - Pro: Maximum flexibility, since we can define a whole separate query
    - Con: Somewhat confusing and duplicative to have a second file
2. Use a window function in this base query to aggregate classes into a new column like `all_classes`
    - Pro: Simple organizational structure, since we stick with one file per base table
    - Con: We'll end up with duplicate rows in the event of a multicard failure, one per card

I lean towards option 1 since its output would be more correct, though I'm curious what you think! I'm also open to other solutions.

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+    {
+        "name": "iasworld_dweldat_rmtot_not_null",
+        "description": "rmtot (Number of Rooms) should not be null",
+        "category": "missing_values",
+        "condition": "rmtot IS NOT NULL"
+    },
+    {
+        "name": "iasworld_dweldat_rmtot_sf_between_1_and_40",
+        "description": "rmtot (Number of Rooms) should be between 1 and 40",
```
jeancochrane: **[Nitpick, optional]** I don't understand why we didn't do this in the original test description (it was probably a mistake) but we should distinguish this test from the 50 SF version, in case they both appear in the same report:

```suggestion
        "description": "rmtot (Number of Rooms) should be between 1 and 40 when class is 211 or 212",
```

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,729 @@
...
+        "category": "incorrect_values",
+        "condition": "class IN ('211', '212') OR rmtot IS NULL OR (rmtot >= 1 AND rmtot <= 40)",
+        "additional_select_columns": ["rmtot"]
+    },
+    {
+        "name": "iasworld_dweldat_rmtot_sf_between_1_and_50",
+        "description": "rmtot (Number of Rooms) should be between 1 and 50",
```
jeancochrane, line 231: **[Nitpick, optional]** I don't understand why we didn't do this in the original test description (it was probably a mistake) but we should distinguish this test from the 40 SF version, in case they both appear in the same report:

```suggestion
        "description": "rmtot (Number of Rooms) should be between 1 and 50 when class is 211 or 212",
```

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
jeancochrane: **[Suggestion, optional]** Hmm, good question. I made a list of all of the tests that currently include classes outside the regression class list:

| Test name | Notes |
| --- | --- |
| `iasworld_dweldat_class_in_ccao_class_dict` | Seems genuinely incompatible with the regression class filter, since by definition, regression class cards will be in `ccao.class_dict` |
| `iasworld_dweldat_class_matches_pardat_class` | I think it would probably be good to keep all classes in here, since our stakeholders will want to fix non-regression classes. But if we move this test to a separate query (per my note above about aggregation), we can solve the problem automatically |
| `iasworld_dweldat_exempt_classes_match_pardat_class` | Same as `iasworld_dweldat_class_matches_pardat_class`, we could resolve this by moving it to a separate query |
| `iasworld_dweldat_external_occpct_not_null_when_mktrsn_eq_5_or_5b_and_mktadj_is_null` | Probably fine to switch to regression classes only, seems like a mistake that we forgot |
| `iasworld_dweldat_external_rcnld_not_0` | Probably fine to switch to regression classes only |
| `iasworld_dweldat_mktadj_not_null_when_mktrsn_eq_5_or_5b_and_external_occpct_is_null` | Probably fine to switch to regression classes only |
| `iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null` | Probably fine to switch to regression classes only |
| `iasworld_dweldat_rmtot_sf_between_1_and_40` | Probably fine to switch to regression classes only |

The only test that seems tricky to fit into this framework is `iasworld_dweldat_class_in_ccao_class_dict`. I think we can resolve the issue for the rest of the tests easily by either switching to testing regression classes only, or moving them to a separate query for aggregate tests, per the Notes column in the table above.

If we switch to a separate file for tests that require aggregation, maybe there's a way we could move `iasworld_dweldat_class_in_ccao_class_dict` into that file? I'm curious if that sounds plausible to you. We could also just remove this test, but I'd be sad to lose it, since it's the type of easy fix that it'd be nice for us to flag for our stakeholders.

The nuclear option, of course, is to move this filter into a new column named something like `is_regression_class` and then use that in the `condition` for all tests that need it. But since there are so many tests in this file, and the vast majority of them implement this filter, I think that approach would introduce an overwhelming amount of boilerplate. It'd be nice to find a more elegant solution if possible.

jeancochrane (changes requested):
Awesome work! Most of my comments below are optional nitpicks, related to suggestions about how we could clean up a bunch of these tests by removing unnecessary null checks and breaking long lines. The biggest-picture thoughts are related to the aggregate tests and the regression class filter, which seem like the trickiest unresolved questions. Let me know if you want to discuss either out loud, otherwise I'll take a second look at this once we've decided on a path forward and implemented it.

Damonamajor:
image

`iasworld_dweldat_heat_not_null` - pardat mismatch

`iasworld_dweldat_sfla_not_null` - pardat mismatch

`iasworld_dweldat_char_gar_area_not_null` - pardat mismatch

`iasworld_dweldat_rmtot_sf_between_1_and_40` relates to more classes being allowed

`iasworld_dweldat_mktrsn_eq_5_or_5b_when_external_occpct_or_mktadj_not_null` - updated null mismatching

`iasworld_dweldat_class_matches_pardat_class` - combination of pardat and what appears to be looking at class(1,3) vs all

`iasworld_dweldat_fixhalf_matches_number_of_units` - unexplained

dbt/models/qc/qc.vw_report_iasworld_test_dweldat_agg.sql
```diff
@@ -0,0 +1,82 @@
...
+        "condition": "pardat_class != 'EX' OR any_class_is_exempt_or_omitted",
+        "additional_select_columns": ["pardat_class", "classes"]
+    },
+    {
+        "name": "iasworld_dweldat_class_in_ccao_class_dict",
+        "description": "class code must be valid for every card",
+        "category": "class_mismatch_or_issue",
+        "condition": "NOT any_class_invalid",
+        "additional_select_columns": ["classes"]
+    }
```
jeancochrane: **[Nitpick, optional]** Does this test need to be defined here on the aggregate view? It seems to me that it would make more sense defined on the non-aggregate view (`qc.vw_report_iasworld_test_dweldat`) since it should fail for any card with a mismatching class (in contrast to `iasworld_dweldat_class_matches_pardat_class` and `iasworld_dweldat_exempt_classes_match_pardat_class`, which could pass if another card on the PIN has a match).

dbt/models/qc/qc.vw_report_iasworld_test_dweldat_agg.sql
```diff
@@ -0,0 +1,82 @@
...
+            AND deactivat IS NULL
+    ) AS comdat
+        ON dweldat.parid = comdat.parid
+        AND dweldat.taxyr = comdat.taxyr
+    WHERE dweldat.cur = 'Y'
+        AND dweldat.deactivat IS NULL
+        AND dweldat.class NOT IN (
+            '201', '213', '218', '219', '220', '221', '224', '225',
+            '236', '240', '241', '290', '294', '297'
+        )
```
jeancochrane: **[Suggestion, required]** I'm pretty sure we don't want to apply this class filter to the whole view, since `iasworld_dweldat_exempt_classes_match_pardat_class` relies on being able to search dweldat for `EX` and `OA` classes, which this filter would exclude. What do you think about moving this `NOT IN` clause to a new column that we `SELECT` and alias to a name like `is_not_residential`, so that we can make the `iasworld_dweldat_class_matches_pardat_class` condition  something like `is_not_residential OR any_class_matches_pardat_class`?

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -150,14 +164,22 @@
...
+        "description": (
+            "fixhalf (Number of Half Baths) should be <= 5 times the number of "
+            "units (user14)"
+        ),
+        "category": "incorrect_values",
+        "condition": (
+            "fixhalf <= CASE WHEN user14 IS NULL OR "
+            "user14 = '0' OR user14 = '6' THEN 5 WHEN user14 = '1' THEN 10 "
+            "WHEN user14 = '2' THEN 15 WHEN user14 = '3' THEN 20 WHEN "
+            "user14 = '4' THEN 25 WHEN user14 = '5' THEN 30 ELSE 5 END"
```
jeancochrane: **[Nitpick, optional]** I find these conditions to be easier to read when we give each `WHEN` clause its own line:

```suggestion
            "fixhalf <= CASE "
                "WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 5 "
                "WHEN user14 = '1' THEN 10 "
                "WHEN user14 = '2' THEN 15 "
                "WHEN user14 = '3' THEN 20 "
                "WHEN user14 = '4' THEN 25 "
                "WHEN user14 = '5' THEN 30 "
                "ELSE 5 "
            "END"
```

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -195,16 +235,26 @@
...
+        "description": (
+            "rmbed (Number of Bedrooms) should be <= 8 times the number of "
+            "units (user14)"
+        ),
+        "category": "incorrect_values",
+        "condition": (
+            "(rmbed >= 1 AND rmbed <= CASE WHEN user14 IS "
+            "NULL OR user14 = '0' OR user14 = '6' THEN 8 WHEN user14 = '1' "
+            "THEN 16 WHEN user14 = '2' THEN 24 WHEN user14 = '3' THEN 32 WHEN "
+            "user14 = '4' THEN 40 WHEN user14 = '5' THEN 48 ELSE 8 END)"
```
jeancochrane: **[Nitpick, optional]** Same here as above:

```suggestion
            "rmbed >= 1 AND rmbed <= CASE "
            "    WHEN user14 IS NULL OR user14 = '0' OR user14 = '6' THEN 8 "
            "    WHEN user14 = '1' THEN 16 "
            "    WHEN user14 = '2' THEN 24 "
            "    WHEN user14 = '3' THEN 32 "
            "    WHEN user14 = '4' THEN 40 "
            "    WHEN user14 = '5' THEN 48 "
            "    ELSE 8 "
            "END"
```

dbt/models/qc/qc.vw_report_iasworld_test_dweldat.sql
```diff
@@ -0,0 +1,895 @@
...
+        "condition": "CAST(user24 AS decimal) BETWEEN 0.00 AND 100.00",
+        "additional_select_columns": ["user24"]
+    },
+    {
+        "name": "iasworld_dweldat_cur_in_accepted_values",
+        "description": 'cur should be "Y" or "D"',
+        "category": "incorrect_values",
+        "condition": "cur IN ('Y', 'D')",
+        "additional_select_columns": ["cur"]
+    },
```
jeancochrane: **[Suggestion, optional]** I just realized that this `cur` test is going to be a no-op because the base query filters for `cur = 'Y'` 🤦🏻‍♀️ This same issue will affect all of the `iasworld_*_cur_in_accepted_values` tests that we've defined in our test views so far.

Between this issue and the issues with our class tests, I'm starting to sense a common friction in the design of our new test framework: **It's hard to implement per-test filter overrides**. Currently, we only offer two ways of overriding test filters, and both of them have serious downsides:

- **Option 1**: Extract any tests that need filter overrides to separate view files, where we can define separate base queries with different `WHERE` clause filters
    - **Example**:  `iasworld_dweldat_class_in_ccao_class_dict`, which we moved to `qc.vw_report_aisworld_test_dweldat_agg` so that it wouldn't be subject to the regression class filtering that we implement in the `WHERE` clause of this view's base query
    - **Pro**: Maximally flexible filter logic, since all filters are just `WHERE` clauses defined on base queries
    - **Con**: Can lead to large amounts of boilerplate code due to duplicative query logic, and scatters related tests across disparate files
- **Option 2**: Move the filter out of the `WHERE` clause of the base query and into the `condition` definition for any test that needs that filter
    - **Example**:  The `class IN ('211', '212')` clause that forms part of the condition for the `iasworld_dweldat_rmtot_sf_between_1_and_40` test
    - **Pro**: Allows as many related tests as possible to be co-located in the same view, since each test is responsible for its own `condition` filtering
    - **Con**: Can lead to large amounts of boilerplate code if a filter is common across many tests

Since this issue has recurred across a few different types of problems, I'm starting to think that we need to tweak the design of our test framework to make per-test filters easier to configure. Here's a proposal that would be similar to the way our dbt tests currently work:

1. Add a new `default_where` parameter to the `generate_iasworld_qc_test_view()` macro that we can use to set a default `WHERE` clause for a base query[^1]:

```diff
 {% macro generate_iasworld_qc_test_view(
-    base_query, tests, start_year=none, end_year=none
+    base_query, tests, start_year=none, end_year=none, default_where=none
 ) %}
```

[^1]: Though the `default_where` parameter might seem duplicative with the existing `start_year` and `end_year` parameters, I think it's probably useful to keep both of them, since implementing year filtering in the `WHERE` clause is helpful for improving the speed and cost of our queries by making use of our common `taxyr` partitioning.

2. Adjust the `generate_iasworld_qc_test_view()` macro definition to implement an additional layer of filtering for each test based on the new `default_where` parameter and a new optional `where` parameter that the user can define on individual tests:

```diff
     {% for test in tests %}
+        {%- set test_where = test.where if test.where is defined else default_where -%}
         select
             parid,
             taxyr,
             ...
         from test_result
-        where {{ test.name }} {{ "UNION ALL" if not loop.last }}
+        where {{ test.name }}
+        {%- if test_where %}
+            AND ({{ test_where }})
+        {%- endif %}
+        {{ "UNION ALL" if not loop.last }}
     {% endfor %}
 {% endmacro %}
```

3. Modify each call site for `generate_iasworld_qc_test_view()` so that we explicitly set `default_where` to the basic filters for each view (this will generally correspond to the current definition of the `unique-conditions` anchor that we use for dbt tests defined on each iasWorld source table). Here's an example of how we'd modify the call for dweldat tests:

```diff
-{{ generate_iasworld_qc_test_view(base_query, tests) }}
+{{ generate_iasworld_qc_test_view(
+    base_query,
+    tests,
+    default_where=(
+        "cur = 'Y' "
+        "AND deactivat IS NULL "
+        "AND class NOT IN ("
+            "'201', '213', '218', '219', '220', '221', '224', '225', "
+            "'236', '240', '241', '290', '294', '297'"
+        ")"
+    )
+) }}
```

4. Add a `where` key to each test that needs to override the conditions set by the `default_where` parameter. For tests that simply need to disable the `default_where` filter, like this one, we can set `where` to an empty string. Here's an example of how we'd modify this particular test:

```diff
    {
        "name": "iasworld_dweldat_cur_in_accepted_values",
        "description": 'cur should be "Y" or "D"',
        "category": "incorrect_values",
        "condition": "cur IN ('Y', 'D')",
+       "where": "",
        "additional_select_columns": ["cur"]
    },
```

Here's another example for `iasworld_dweldat_class_in_ccao_class_dict`, which would require us to set `where` to a non-empty string in order to match the current definition:

```diff
    {
        "name": "iasworld_dweldat_class_in_ccao_class_dict",
        "description": "class code must be valid for every card",
        "category": "class_mismatch_or_issue",
-        "condition": "NOT any_class_invalid",
+        "condition": "class_dict.class_code IS NULL",
+        "where": (
+            "cur = 'Y' "
+            "AND deactivat IS NULL "
+            "AND class NOT IN ('EX', 'RR')"
+        )
        "additional_select_columns": ["classes"]
    }
```

What do you think about all this? Since it's a complex issue that affects all of our test views, and since it may require some custom code to resolve, I'm fine punting a resolution to a follow-up PR if you don't feel like dealing with it right now.

jeancochrane (commented):
This is looking very close to done! I'd like to do some final testing tomorrow morning to make sure I fully understand all the mismatches. In the meantime, I had a couple tiny nitpicks and one high-level thought about the design of our filter override system that I wanted to give you extra time to think about.

